PR #553, opened 2018-02-27
CPU nan/inf tensor validation

Krovatkin:


Krovatkin:
@rkimballn1 removed a WIP tag. When this passes CI can I merge?

PR #617, opened 2018-03-09
Zero Dimension Tensor Elimination

Krovatkin:


src/ngraph/pass/zero_dim_tensor_elimination.cpp
```diff
@@ -0,0 +1,176 @@
...
+static std::unordered_map<std::type_index,
+                          std::function<std::shared_ptr<Node>(std::shared_ptr<Node>)>>
+    ops_to_const_values = initialize_const_values_to_ops();
+
+static std::shared_ptr<Node> get_const_value_for_op(std::shared_ptr<Node> n)
+{
+    const Node& node = *n;
+    auto entry = ops_to_const_values.find(TI(node));
+    return entry == ops_to_const_values.end() ? nullptr : entry->second(n);
+}
```
diyessi: Why not use a virtual `Node::get_const_value_for_op()` whose default returns `nullptr`?

PR #645, opened 2018-03-14
Dot op that can handle more than 2D on GPU

fengleitian:
Dot that can handle more than 2D on GPU.

limitation:
1. Still only support float point.


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -263,23 +265,60 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                        {
+                            throw std::runtime_error(
+                                "input and output shape does not match for dot;");
+                        }
+                        else
+                        {
+                            m *= arg0_shape[i];
+                        }
+                    }
+                    for (size_t i = 0, idxa = arg1_shape.size() - 1, idxb = out_shape.size() - 1;
```
csullivan: The use of these multi-index for loops make this difficult to read. Since these loops are helpers (not performance heavy), I think a loss in efficiency for the sake of readability would be better.

fengleitian: good point.

rkimballn1: With modern compilers there is no loss of efficiency. Always shoot for "normal" and "readable".

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -263,23 +265,60 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                         i < arg1_shape.size() - reduction_axes;
+                         i++, idxa--, idxb--)
+                    {
+                        if (arg1_shape[idxa] != out_shape[idxb])
+                        {
+                            throw std::runtime_error(
```
rkimballn1: std::invalid_argument

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -241,6 +242,7 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
                     writer.indent--;
                     writer << "}\n";
                 }
+                //matrix vector
```
rkimballn1: Is is a standard to not put a space after `//` in a comment? I see it a lot lately. It looks better with a space there?

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -263,23 +265,60 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
-                         (dot->get_reduction_axes_count() == 1))
+                //dot can be treat as matrix multiply
+                else
                 {
-                    // GEMM Call
-                    if (arg0_shape[0] != out[0].get_shape()[0] || // m
-                        arg1_shape[1] != out[0].get_shape()[1] || // n
-                        arg0_shape[1] != arg1_shape[0])           // k
+                    size_t reduction_axes = dot->get_reduction_axes_count();
+                    size_t m = 1, n = 1, k = 1;
```
rkimballn1: One decl per line please. No commas.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -263,23 +265,60 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                    //check if input and output size correct
+                    for (size_t i = 0, idxa = arg0_shape.size() - 1; i < reduction_axes;
```
rkimballn1: No commas here either.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -263,23 +265,60 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                         i++, idxa--)
                     {
-                        throw std::runtime_error("input and output shape does not match for dot;");
+                        if (arg0_shape[idxa] != arg1_shape[i])
+                        {
+                            throw std::runtime_error(
```
rkimballn1: std::invalid_argument, remove the semicolon from the string there and everywhere else.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: Thanks, fixed.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: fixed.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: fixed.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: Thank you for point this out. changed.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -263,23 +265,64 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
-                else if ((arg0_shape.size() == 2) && (arg1_shape.size() == 2) &&
-                         (dot->get_reduction_axes_count() == 1))
+                // cases that can be treat as matrix multiply
+                else
                 {
-                    // GEMM Call
-                    if (arg0_shape[0] != out[0].get_shape()[0] || // m
-                        arg1_shape[1] != out[0].get_shape()[1] || // n
-                        arg0_shape[1] != arg1_shape[0])           // k
+                    // treat as out[m,n] = arg0[m,k] * arg1[k,n]
```
csullivan: Thanks, it reads better. I've been working through the mathematics of NumPy/Tensorflow tensordot and I realized that op::Dot is less general, is that correct? There isn't any mention of contraction in ngraph op::Dot so I assume so. The matrix multiplication impl. looks good.

fengleitian: There was some spec on generalized dot, 
https://github.com/NervanaSystems/ngraph/blob/2246d9ba05ace2368494bb098da0b4964437d76a/src/ngraph/ops/dot.hpp
Let me check with @aprocter

aprocter:
The docstring in the older revision linked to by @fengleitian remains accurate. Not sure if all the detail made it into the new Sphinx docs.

src/ngraph/runtime/gpu/gpu_emitter.cpp
aprocter: The intended semantics is the same as https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.tensordot.html, in the case where `axes` is an "integer_like". In other words, you cannot arbitrarily pair reduction axes in nGraph, unlike NumPy and TF.

fengleitian: Thanks @aprocter

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -227,8 +228,8 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
                     {
                         if (arg0_shape[i] != arg1_shape[i])
                         {
-                            throw std::runtime_error(
-                                "input1 and input2 shape does not match for dot;");
+                            throw std::invalid_argument(
+                                "arg0 and arg0 shape does not match for dot.");
```
adstraw: arg0 and 'arg1'

fengleitian: fixed. thank you for point this out.

PR #657, opened 2018-03-15
[v0.1.0] Multi-output fprop_cache tentative fix

Krovatkin:


src/ngraph/ops/batch_norm.cpp
```diff
@@ -148,8 +148,17 @@ void ngraph::op::BatchNorm::generate_adjoints(autodiff::Adjoints& adjoints,
     auto gamma = get_input_op(0);
     auto beta = get_input_op(1);
     auto input = get_input_op(2);
-    auto mean = std::make_shared<op::GetOutputElement>(shared_from_this(), 1);
-    auto var = std::make_shared<op::GetOutputElement>(shared_from_this(), 2);
+
```
diyessi: This needs comments about what it is doing. Is it snapping out the `GetOutputElement`?

src/ngraph/autodiff/adjoints.cpp
```diff
@@ -125,7 +125,7 @@ std::shared_ptr<Node> autodiff::Adjoints::get(const std::shared_ptr<Node>& x)
 void autodiff::Adjoints::add_delta(const std::shared_ptr<Node>& x,
                                    const std::shared_ptr<Node>& delta)
 {
-    if (!x->has_same_type(delta))
+    if (!x->has_same_type(delta) && delta->get_shape() != x->get_outputs().at(0).get_shape())
```
mbrookhart: This is a bit of technical debt that doesn't matter for batchnorm but will matter for other multi-output ops. I've created NGMX-422 to track it.

src/ngraph/ops/batch_norm.cpp
Krovatkin: will do

Krovatkin: @diyessi added the comment

Krovatkin: @diyessi added another comment since I wasn't sure if you were referring to `BatchNorm::generate_adjoints` or `GetOutputElement::generate_adjoints`

PR #667, opened 2018-03-16
[Python] Rename ngraph APIs

postrational:
Renames top-level modules:
* `ngraph` -> `ngraph_bind`
* `ngraph_api` -> `ngraph`


silee2:
I've put up #674 which is a similar attempt but instead of two folders, it uses one.
-ngraph/impl holds the low level wrappers
-ngraph holds the pythonic wrappers

diyessi:
Replaced by #674

PR #674, opened 2018-03-17
[v0.1.0] Silee2/python binding

silee2:
This is a proposal for what the python binding code and package structure should be like.
The PR builds and functions correctly.
The main ideas is to 
- put all c++ pybind code under python/pyngraph
- put all the low level direct wrappers under python/ngraph/impl
- put the high level pythoninc API under python/ngraph

Most users can just add
``
import ngraph as ng
``
And enjoy the pythonic API

This is an alternative to #667 

The PR looks big but most of the changes are just for updating docstrings

By the way, I slipped in a patch to make ngraph build again on Mac. (the first commit)
Some SDL-3 Linker options are only applicable to ELF based systems. And MacOS it not one of them.
The patch enables those linker flags only if the system is not a Mac.
Should I split the PR?


silee2:
Matching changes for ngraph-neon are here:
https://github.com/NervanaSystems/ngraph-neon/pull/85

PR #679, opened 2018-03-18
Add `--visualize` to nbench

Krovatkin:


src/tools/nbench/nbench.cpp
```diff
@@ -98,6 +106,7 @@ OPTIONS
         -b|--backend       Backend to use (default: CPU)
         -i|--iterations    Iterations (default: 10)
         -s|--statistics    Display op stastics
+        -v|--visualize     Visualize a model (WARNING: requires dot installed)
```
rkimballn1: It doesn't require dot to be installed, but it will only generate a dot file if you do not have graphviz installed.
The package is call graphviz, not dot.

src/tools/nbench/nbench.cpp
```diff
@@ -106,6 +115,22 @@ OPTIONS
     const string json_string = file_util::read_file_to_string(model);
     stringstream ss(json_string);
     shared_ptr<Function> f = deserialize(ss);
+
+    if (visualize)
+    {
+        const char* format = std::getenv("NGRAPH_VISUALIZE_TREE_OUTPUT_FORMAT");
```
rkimballn1: We should probably handle the cases where the format is either "png" or ".png"

src/tools/nbench/nbench.cpp
Krovatkin: @rkimballn1 I might be misreading stn but it does seem that we do require `GraphViz` to do both:

```cpp
void pass::VisualizeTree::render() const
{
#ifdef GRAPHVIZ_FOUND
    auto tmp_file = m_name + ".tmp";
    ofstream out(tmp_file);
    if (out)
    {
        out << "digraph ngraph\n{\n";
        out << m_ss.str();
        out << "}\n";
        out.close();

        stringstream ss;



        ss << "dot -T" << get_file_ext() << " " << tmp_file << " -o " << m_name;
        auto cmd = ss.str();
        auto stream = popen(cmd.c_str(), "r");
        pclose(stream);

        remove(tmp_file.c_str());
    }
#endif
}
```

rkimballn1: With the current code you are correct. We could change the code to emit the dot file but only call graphviz if it is installed. Meh, whatever.

Krovatkin: @rkimballn1  A workflow where only a dot file is generated but then it needs to be moved to a machine where `GraphViz` is installed *seems* to be rather cumbersome one.  I don't know if there's a lot of value in supporting it.

Krovatkin:
TESTNOW

Krovatkin:
ping... @rkimballn1

Krovatkin:
@rkimballn1 , could you please approve ?   .. I updated the comment!

PR #680, opened 2018-03-18
add python dir to check-code-format and re-run apply-code-format on it

Krovatkin:
Otherwise folks will get annoying changes in `python` due to `check-code-format` didn't make sure those files were formatted properly before a check-in


maint/check-code-format.sh
```diff
@@ -47,9 +47,10 @@ pushd "${THIS_SCRIPT_DIR}/.."
 
 declare ARGON_SRC_DIR="build/third-party/argon_transformer/src/ext_argon_transformer/src"
 declare ARGON_TEST_DIR="build/third-party/argon_transformer/src/ext_argon_transformer/test"
+declare PYBIND_WRAPPER="python/pyngraph"
 
 declare ROOT_SUBDIR
-for ROOT_SUBDIR in src test ${ARGON_SRC_DIR} ${ARGON_TEST_DIR}; do
+for ROOT_SUBDIR in src test ${ARGON_SRC_DIR} ${ARGON_TEST_DIR} ${PYBIND_WRAPPER}; do
```
diyessi: Aren't these wrappers generated code? We shouldn't be formatting generated code.

silee2:
I guess this is not urgent but the idea was to make the python wrappers follow the same c++ style guideline. The wrappers are not auto generated. @nbpatel ran the script manually for open sourcing int would make sense to integrate into the script in the long run.

Krovatkin:
@silee2 @diyessi unless we also update `check` any1 who runs `apply` will get surprising updates to `python` dir because `check` didn't gate the previous check-ins.

Krovatkin:
@diyessi Could you please let me know if I could merge this when it passes CI?

PR #684, opened 2018-03-19
Doc validation2

indie:


jennifermyers:
before merging, let's make sure the image appears in the ngraph-unittest jenkins job build of the docs.

diyessi:
There are some blog comments on changes for the figures; so we might want to update it if Tunde agrees.

PR #686, opened 2018-03-19
 Update GraphRewrite API

Krovatkin:
Finally updating GraphRewrite API to take advantage of the existence of `op::Result` No more weird contracts between callbacks and `GraphRewrite`. `GraphRewrite` should be able to replace as many nodes as needed.


src/ngraph/pattern/core_fusion.cpp
```diff
@@ -61,19 +61,19 @@ void pass::CoreFusion::construct_relu_pattern()
     pattern::gr_callback_fn callback = [val, zero](pattern::Matcher& m) {
         NGRAPH_DEBUG << "In a callback for construct_relu_pattern against "
                      << m.match_root()->get_name();
-        auto pattern_map = m.get_pattern_map();
-        shared_ptr<Node> nn;
 
+        auto pattern_map = m.get_pattern_map();
```
sasadep: @Krovatkin looks like you have made the necessary changes required for `CoreFusion`. This is the only fusion that is currently used for NNP backend. So, the  conclusion is if this works then its should not effect us much.

PR #689, opened 2018-03-19
ConvolutionRelu

Krovatkin:


Krovatkin:
Interestingly enough, it fails the new test I added when compiled w/ GCC. I'll investigate tomorrow.

Krovatkin:
@jbobba could you please take a look when you have a moment? I think it's semi-ready and should be passing CI

adstraw:
please fix merge conflict

Krovatkin:
@jbobba 

> Dont we need more checks to see if the Convolution is supported by MKLDNN before fusing it to ConvolutionRelu?

We talked about it a few days ago. Basically, I was asking if I need to do any checks in `CPUFusion` and you seem to have indicated that we can always fuse :-)

@adstraw will do shortly

Krovatkin:
@jbobba I added the checks from `CPUAssignment` pass. Please let me know if those are good enough? 
@adstraw alright, resolved the merge conflicts.

adstraw:
"Can you also add a comment to cpu fusion as to why we are checking for these specific attributes or pull that code out into a utility function" <- this needs to get done pre-merge, yes?

Krovatkin:
@jbobba @adstraw done.

Krovatkin:
> LGTM, this include the forward part, is the backprop part in a separate PR?

My understanding was we weren't doing backprop atm? @jbobba ?

jbobba:
Right, backprop should be a different PR.

PR #691, opened 2018-03-19
cleanup neon integration documentation

indie:
Thanks @nbpatel for helping me get this straightened out.  Here's how it turned out


doc/sphinx/source/framework-integration-guides.rst
```diff
@@ -46,55 +46,31 @@ system that already has an ``ngraph_dist`` installed.
    .. code-block:: console
 
       $ sudo apt-get install python3-pip python3-venv
+      $ python3 -m venv .venv
+      (.venv)$ . bin/activate
```
nbpatel: type here? you need to do . .venv/bin/activate and then the (.venv) prompt will show up from the next steps.

nbpatel: Rest of it LGTM.

doc/sphinx/source/framework-integration-guides.rst
```diff
@@ -46,55 +46,31 @@ system that already has an ``ngraph_dist`` installed.
...
 
-         $ sudo mkdir -p /opt/frameworks   
-         $ sudo chown -R username:username /opt/frameworks
-         $ cd /opt/frameworks
+      (.venv)$ cd ngraph/python
+      (.venv)$ git clone --recursive -b allow-nonconstructible-holders https://github.com/jagerman/pybind11.git
+      (.venv)$ pip install -U . 
 
-   * This section is required:       
+#. Finally are we ready to install the `neon` integration:
```
nbpatel: typo...we are* instead of are we

doc/sphinx/source/framework-integration-guides.rst
```diff
@@ -46,55 +46,31 @@ system that already has an ``ngraph_dist`` installed.
...
+   .. code-block:: console
 
-         $ git clone git@github.com:NervanaSystems/ngraph-neon
-         $ cd ngraph-neon
+      (.venv)$ git clone git@github.com:NervanaSystems/ngraph-neon
+      (.venv)$ cd ngraph-neon
+      (.venv)$ make install
 
-#. *Optional* Activate a virtualenv if you prefer working with virtualenvs:
+#. To test a training example, you can run the following from ``ngraph-neon/examples/cifar10``
```
nbpatel: instead of words here ..how about giving the command with the entire path in the next step...
python examples/cifar10/cifar10_conv.py

doc/sphinx/source/framework-integration-guides.rst
diyessi: I prefer giving it a visible name, like `pydoc` since macos makes directories starting with `.` invisible to file finders, like PyCharm, which means you can't use them as an environment within the IDE.

doc/sphinx/source/framework-integration-guides.rst
indie: This is actually a language thing; I did intend it to be that way.  Varying sentence structure so reading instructions is not boring.  Because, seriously.  By the time people get this far through the instructions, they are going to be thinking *finally!*

doc/sphinx/source/framework-integration-guides.rst
indie: @postrational had commented that it's not clear which directory holds the correct "example" to run whether it's the ngraph-python repo or the ngraph-neon we cloned.  Change is not needed.

doc/sphinx/source/framework-integration-guides.rst
indie: What makes the most sense is to create a venv named `frameworks`, I think.  This is kind of the line I was going down originally with my previous instructions for system install.  People will be able to have a single `venv` that represents nGraph-enabled frameworks.  Tested it; it works.  Awesome.

doc/sphinx/source/framework-integration-guides.rst
diyessi: Compromise: `Finally our neon integration is ready to install.`

doc/sphinx/source/framework-integration-guides.rst
nbpatel: ok. i will get rid of examples from ngraph/python just to avoid confusion. They were added there just for testing the wrapper code.

postrational: I think that we can be more clear in the docs that an example comes from this or that repo. We may want to keep `examples` in the ngraph repo.

PR #698, opened 2018-03-20
Remove examples from pybind wrapper

nbpatel:


postrational:
Is there a reason to remove these? They are kind of our only documentation of the Python API at the moment. 

I would suggest leaving ` python/examples/basic.py` until we have more documentation.

nbpatel:
@postrational this is to avoid confusion for the users as to which examples to run. We can keep basic.py but maybe we can have it in a dir called something else and not examples just to avoid the confusion. We can remove mnist as it was added just to test some basic wrapper functionalities.

nbpatel:
@postrational added back examples/basic.py

PR #702, opened 2018-03-20
CallFrame order

yxlao:
Making it consistent s.t.
```
Function(outputs, inputs)
CallFrame.call(outputs, inputs)
```
Tagging component owners for review since this changes API.


avijit-nervana:
Adding Adam who would be able to get the review done from TF side much faster that I can.

postrational:
This won't require much work in the ONNX bridge either.

PR #704, opened 2018-03-21
Remove XLA compatibility fig leaves

aprocter:
Marking WIP because we haven't yet adapted `ngraph-tensorflow-1.3`.


Krovatkin:
@diyessi @aprocter 
I was wondering if we might want to repurpose an `XLATuple` into a `Tuple` which we could use for multi-output ops in autodiff? Unless you guys already decided how we are going to support `generate_adjoints` for multi-output ops?

diyessi:
No needed for that.

aprocter:
Removing WIP because we have a PR ready to deal with this in ngtf-1.3 now.

adstraw:
@diyessi can you review today?

PR #705, opened 2018-03-21
Update nnp git URL and branch

yxlao:
- Update nnp-transformer repo address.
- Switched to https. It seems there are more people using https (rather than ssh) internally for easier proxy setup. Please let me know if we should keep the ssh address instead.


cmake/external_nnp_transformer.cmake
```diff
@@ -33,7 +33,7 @@ if (NGRAPH_NNP_ENABLE)
     if (DEFINED CUSTOM_NNP_TRANSFORMER_GIT_REPOSITORY)
         set(NNP_TRANSFORMER_GIT_REPOSITORY ${CUSTOM_NNP_TRANSFORMER_GIT_REPOSITORY})
     else()
-        set(NNP_TRANSFORMER_GIT_REPOSITORY git@github.com:NervanaSystems/argon-transformer.git)
+        set(NNP_TRANSFORMER_GIT_REPOSITORY https://github.com/NervanaSystems/nnp-transformer.git)
```
sasadep: do you want to leave it as `git@github.com:NervanaSystems/nnp-transformer.git` thought you had that for CI help

yxlao: I remember it's https, @DawnStone could you verify this?
In both cases, we could set `cmake -DCUSTOM_NNP_TRANSFORMER_GIT_REPOSITORY`

DawnStone: For Jenkins, we needed to checkout the argon-transformer repo before running cmake, and reference the directory using `-DCUSTOM_NNP_TRANSFORMER_REPOSITORY`, although I am currently setting this as `-DCUSTOM_NNP_TRANSFORMER_DIR`.  I'll need to test this out to find out what it has really been doing during the build process.

sasadep: @DawnStone I guess you will need to check if you have  prefer  `https://github.com` over `git@github.com:` or vice versa

DawnStone: @yxlao: I have verified that the Jenkins CI process does not use CUSTOM_NNP_TRANSFORMER_GIT_REPOSITORY, but successfully builds with the nnp-transformer repo using the CUSTOM_NNP_TRANSFORMER_DIR as the SOURCE_DIR.

adstraw:
@varun-intel can you review?

ransford2011:
@adstraw .  Varun is PTO today.  If this is gating, then go ahead and merge.  It's not a major change on our end.

adstraw:
It is not gating me so if it isn't gating you we can leave it as-is.

yxlao:
@adstraw we could merge this now since it's a minor change

PR #707, opened 2018-03-21
Update bezel version

lwencel:
TensorFlow 1.7 will not compile with bezel 0.5.4.
The minimum version of bazel is 0.10.0, but I recommend using version 0.11.1.


avijit-nervana:
We should make this change as the version of TF we have open sourced is 1.3 based which won't work with 0.11.

lwencel:
@avijit-nervana ok, I was not sure that you are using TF 1.3, because in documentation you recommand to use `git clone git@github.com:NervanaSystems/ngraph-tensorflow.git`. This will clone private repository with TensorFlow 1.7 which requires bazel 0.10.0 (point 5 in http://ngraph.nervanasys.com/docs/latest/framework-integration-guides.html#tensorflow-intg)
Should I use `https://github.com/NervanaSystems/ngraph-tensorflow-1.3` instead of `git@github.com:NervanaSystems/ngraph-tensorflow.git`?

avijit-nervana:
@lwencel Yes - please clone the 1.3 branch. @indie Could you please fix the documentation? Thanks.

adstraw:
is this PR dead?  can we close it>  @avijit-nervana @lwencel

avijit-nervana:
We need to update the documentation - so once that is done we should close this.

adstraw:
Closing this PR per @avijit-nervana, fixed in #726

PR #708, opened 2018-03-21
[Py] Add UT for avg_pooling_3d

tsocha:
-Cover uncontigous data input from python to ngraph
-Add UT for average pooling op


python/test/ngraph/test_avg_pooling.py
```diff
@@ -0,0 +1,46 @@
...
+
+import ngraph as ng
+
+
+def test_avg_pooling_3d():
+    manager_name = pytest.config.getoption('backend', default='CPU')
+    rt = ng.runtime(manager_name=manager_name)
+
+    input = np.array(range(16), dtype=np.float32)
+    input = input + 11
```
postrational: These 2 lines can be simplified to `np.arange(11, 27, dtype=np.float32)`

python/test/ngraph/test_avg_pooling.py
```diff
@@ -0,0 +1,46 @@
...
+import pytest
+
+import ngraph as ng
+
+
+def test_avg_pooling_3d():
+    manager_name = pytest.config.getoption('backend', default='CPU')
+    rt = ng.runtime(manager_name=manager_name)
+
+    input = np.array(range(16), dtype=np.float32)
```
postrational: Name input shadows built-in function name. Shouldn't use it.

python/test/ngraph/test_avg_pooling.py
```diff
@@ -0,0 +1,46 @@
...
+    input = input + 11
+    input = input.reshape((1, 1, 4, 4))
+    input = np.broadcast_to(input, (1, 1, 4, 4, 4))
+
+    param = ng.parameter(input.shape)
+
+    avgpool = ng.avg_pool(param,
+                          [2, 2, 2],
+                          [2, 2, 2],
+                          zero_pad=False)
```
postrational: Why does `zero_pad` default to True? What would happen if it wasn't set to `False` here?

arogowie-intel: `zero_pad` decides whether or not to include filled with zeros artificially extended area of image into average computations. Great example [here](http://ngraph.nervanasys.com/docs/cpp/ops/avg_pool.html).

tsocha: I think it can be deleted because we have 0 padding here. Thanks for mention it.

python/test/ngraph/test_avg_pooling.py
tsocha: Thanks and done.

python/test/ngraph/test_avg_pooling.py
tsocha: changed input -> data

tsocha:
Branch rebased on master

adstraw:
waiting for @arogowie-intel to approve

PR #710, opened 2018-03-21
[WIP] [Py] Update source directory for cloned pybind11 repository.

arogowie-intel:
Fix path to be consistent with what is stated inside python/setup.py. 
line 182:
```
include_dirs = [  # Path to pybind11 headers
    'pybind11/include',
```
Before this change pybind11 was cloned into `python/build` directory during building.


silee2:
@nbpatel Does the installation doc need to be updated?

nbpatel:
@silee2 No, i think this PR is a follow up of the change we made in the install doc on where pybind needs to be cloned.

silee2:
Putting external code outside of the build directory may have some weird side effect. Currently the build script (the .sh file) removes existing build directory and git clone all external dependency. So pybind11 folder is not removed by the build script and then the script will attempt to clone into a existing folder. I think that will cause a build failure. Please check.

nbpatel:
Currently the way the things are, pybind sits under ngraph/python and things work fine if you follow the installation doc. The installation doc is based on the assumption that the user has already installed 
ngraph. I have not tested the process with .sh script.

postrational:
I'm not convinced yet that this is the way we should go.

It would be preferable to have only one file `setup.py` and not 2 (`setup.py` and `setup.py.in`).

I think it should be possible to make `setup.py` smart enough to work as part of the `cmake` build process as well as outside of it.

We should be checking if `pybind` and a built `ngraph` are available and complain if they are not.

Their locations could be passed in via `ENV` flags.

We should verify the chosen solution works in:
1) `cmake` and `build_python2&3_wheel.sh` scripts
2) Independent of `cmake` when calling `python setup.py *`
3) as part of `tox`
4) as part of our CI process

arogowie-intel:
Yes, this quick change was initiated by changes where `pybind11` needs to be cloned, since tests run with `tox` failed to compile `pyngraph`. Currently this is WIP as we search for clearer solution working in all scenarios Michal mentioned.

nbpatel:
@postrational The second setup.py was introduced as a last min solution to fix tox

postrational:
I would suggest going in this direction:
https://github.com/NervanaSystems/ngraph/pull/761

postrational:
I think this can be closed after #761

PR #711, opened 2018-03-21
Change the call frame order in pybind wrapper tests

nbpatel:


silee2:
Update test cases after changes made by #702

adstraw:
Scott is in meetings all day.  You might add @yxlao who made the original change instead.

adstraw:
Do we need to wait on @baojun-nervana ?

nbpatel:
He was the one who wrote the tests hence i added him. But I think we should be ok with two approvals

adstraw:
I would prefer to see all reviewers +1 but given these tests are broken I plan to merge this when it passes CI.

PR #714, opened 2018-03-22
make sure deserializer doesn't add op::Result twice

Krovatkin:


src/ngraph/serializer.cpp
```diff
@@ -851,11 +851,29 @@ static shared_ptr<ngraph::Function>
...
+        {
+            result.push_back(res);
+        }
+        else
+        {
+            result.push_back(std::make_shared<op::Result>(fr));
+        }
     }
+
+    if (results != 0 && results != func_result.size())
```
jbobba: Where is results getting updated?

Krovatkin: grrr 😢  somebody took it! fixed in the next commit.

src/ngraph/serializer.cpp
```diff
@@ -851,11 +851,31 @@ static shared_ptr<ngraph::Function>
...
+    {
+        throw ngraph_error(
+            " Graph serialization is incosistent. Some op::Results appear to be missing");
```
rkimballn1: inconsistent is misspelled.

PR #716, opened 2018-03-22
gpu reshape n-dimension (n>2)

fengleitian:


src/ngraph/runtime/gpu/gpu_cuda_kernel_emitters.cpp
```diff
@@ -92,3 +92,39 @@ void runtime::gpu::emit_onehot(std::string name,
                                   0));  // arguments
     CUDA_SAFE_CALL(cuCtxSynchronize()); // Retrieve and print output.
 }
+
+void runtime::gpu::emit_reshape(std::string name,
```
rkimballn1: const string& for arg. Same comment for other instances of the same thing.

src/ngraph/runtime/gpu/gpu_cuda_kernel_emitters.cpp
```diff
@@ -92,3 +92,39 @@ void runtime::gpu::emit_onehot(std::string name,
...
+                                CUdeviceptr in,
+                                CUdeviceptr out,
+                                std::array<std::string, 2> data_types,
```
rkimballn1: can we do const array& or does it make calling more difficult?

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -569,9 +569,65 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
                 // Other cases (reordering of axes for tensors with rank>2) are not handled yet.
                 else
                 {
-                    throw runtime_error(
-                        "Axis permutation in reshape is not implemented yet for tensors with "
-                        "rank>2");
+                    size_t* input_strides = new size_t[arg_rank];
```
rkimballn1: Use a vector<size_t> here. Slightly less efficient but it does bounds checking.

fengleitian: Yes, right, changed.

src/ngraph/runtime/gpu/gpu_cuda_kernel_emitters.cpp
fengleitian: Updated.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -571,9 +571,64 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                    }
+                    writer << "};\n";
+
+                    writer << "std::vector<size_t> trans_strides_h = {" << trans_strides[0] << "UL";
+                    for (int i = 1; i < arg_rank; i++)
+                    {
+                        writer << ", " << trans_strides[i] << "UL";
+                    }
+                    writer << "};\n";
+
```
csullivan: Because we are runtime compiling these kernels, ideally we should not have to perform this memcpy. The only reason it is necessary here is because the cuda is emitted from the LLVM compiled function instead of the transformer. If the cuda kernels were emitted at this level, it could be adjusted from within the gpu_emitter to include these vectors/arrays, and then be compiled before the LLVM call is made. 

Last weekend I started a retooling to decouple the cuda kernel emission and execution to enable this kind of functionality -- I haven't followed up on it during the week yet, instead focusing on ops. For now these copies are OK but we should keep a mental note about coming back to this when the compilation order is changed.

fengleitian: Absolutely true. This is a hack for now.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -571,9 +571,64 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
                 // Other cases (reordering of axes for tensors with rank>2) are not handled yet.
```
csullivan: Change this comment. 

Also, I noticed an extra indent that should be removed on line (556) - https://github.com/NervanaSystems/ngraph/blob/61623fb00f93b4d95d259337cf03ec6b2d3a286b/src/ngraph/runtime/gpu/gpu_emitter.cpp#L556

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -571,9 +571,64 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                    for (int i = arg_rank - 1; i >= 0; i--)
+                    {
+                        output_strides[i] = stride;
+                        stride *= arg_shape[input_order[i]];
+                    }
+                    for (int i = 0; i < arg_rank; i++)
+                    {
+                        trans_strides[input_order[i]] = output_strides[i];
+                    }
+                    writer << "{   // " << node->get_name() << "\n";
```
csullivan: use writer.block_begin()/end()

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: Thanks, updated.

fengleitian:
TESTNOW

fengleitian:
TESTNOW

PR #719, opened 2018-03-22
[Py] Python wrappers for nGraph operations

arogowie-intel:
add:
- ng.absolute
- ng.acos
- ng.asin
- ng.atan

### DIFF
--- a/python/test/ngraph/test_ops_unary.py
+++ b/python/test/ngraph/test_ops_unary.py
@@ -34,20 +34,9 @@ def _run_unary_op_node(input_data, unary_op):
 
 def _run_unary_op_numeric_data(input_data, unary_op):
     runtime = _get_runtime()
-    parameter_a = ng.parameter(input_data.shape, name='A', dtype=np.float32)
     node = unary_op(input_data)
-    computation = runtime.computation(node, parameter_a)
-    return computation(input_data)
-
-
-@pytest.mark.parametrize('ng_api_fn, numpy_fn, input_data', [
-    (ng.absolute, np.abs, -1 + np.random.rand(2, 3, 4) * 2),
-    (ng.acos, np.arccos, -1 + np.random.rand(2, 3, 4) * 2),
-])
-def test_unary_op(ng_api_fn, numpy_fn, input_data):
-    result = _run_unary_op_node(input_data, ng_api_fn)
-    expected = numpy_fn(input_data)
-    assert np.allclose(result, expected)
+    computation = runtime.computation(node)
+    return computation()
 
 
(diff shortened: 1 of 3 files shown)


python/test/ngraph/test_ops_unary.py
```diff
@@ -0,0 +1,58 @@
...
+
+import ngraph as ng
+
+
+def _get_runtime():
+    manager_name = pytest.config.getoption('backend', default='CPU')
+    return ng.runtime(manager_name=manager_name)
+
+
+def _run_unary_op_node(input_data, unary_op, as_node=True):
```
postrational: If you use `as_node` you don't need to create or pass the parameter.
Let's split this into 2 functions, one to run the operation on a node, the other on numeric data.
You can then run both in each test, so no need to pass the extra `as_node` param.

arogowie-intel: updated

python/test/ngraph/test_ops_unary.py
```diff
@@ -0,0 +1,62 @@
...
+    runtime = _get_runtime()
+    parameter_a = ng.parameter(input_data.shape, name='A', dtype=np.float32)
+    node = unary_op(parameter_a)
+    computation = runtime.computation(node, parameter_a)
+    return computation(input_data)
+
+
+def _run_unary_op_numeric_data(input_data, unary_op):
+    runtime = _get_runtime()
+    parameter_a = ng.parameter(input_data.shape, name='A', dtype=np.float32)
```
postrational: `parameter` is redundant in this case

python/test/ngraph/test_ops_unary.py
```diff
@@ -0,0 +1,62 @@
...
+    assert np.allclose(result, expected)
+
+
+@pytest.mark.parametrize('ng_api_fn, numpy_fn, input_data', [
+    (ng.absolute, np.abs, -1 + np.random.rand(2, 3, 4) * 2),
+    (ng.absolute, np.abs, np.float32(-3)),
+    (ng.acos, np.arccos, -1 + np.random.rand(2, 3, 4) * 2),
+    (ng.acos, np.arccos, np.float32(-0.5)),
+])
+def test_unary_op_with_numeric_data(ng_api_fn, numpy_fn, input_data):
```
postrational: We can combine these 2 tests into one.

python/test/ngraph/test_ops_unary.py
arogowie-intel: Fixed

python/test/ngraph/test_ops_unary.py
arogowie-intel: Fixed.

PR #720, opened 2018-03-22
Normalize op implementations

rkimballn1:
This is a PR along the path of making copy_with_new_args support copying things such as metadata and placement.
Each header now has a corresponding source file. Many implementations are moved from the header to the source file. That should be all that is changed in this PR. Rather than bury the "real" changes I wanted to make in this large PR I thought I would make two PRs instead.


rkimballn1:
This thing got messed up in a merge. Will recreate

PR #721, opened 2018-03-22
[WIP] Add support for prebuilt LLVM on Mac.

silee2:
WIP: Two issues to resolve.
1. limit the prebuilt LLVM to macOS 10.13
2. solve unit failures related to c++ exception handling.


cmake/external_llvm_prebuilt.cmake
```diff
@@ -61,83 +73,163 @@ if (NGRAPH_CPU_ENABLE)
...
-        LLVMCore
-        LLVMBinaryFormat
-        LLVMSupport
-        LLVMDemangle
-        tinfo
-        z
-        m
-        PARENT_SCOPE)
-
+    if(APPLE)
```
diyessi: Any way we could have a list of things common to everything augmented by the platform-specific stuff?

silee2: Sure.
Want to do the same for  cmake/external_llvm.cmake ? I copied the code from that file.

silee2: There is only one line difference across platforms so the updated code will look much cleaner.

aprocter:
This is awesome! Last time I tried this I ran into some unit test failures (some kind of ABI mismatch wrt `std::exception`, presumably) that do not show up if you build LLVM from source. Were you able to fix those? (Or have you not seen them?)

silee2:
wasn't aware of the issue. I just ran the unit tests and there are some issues.
`
[==========] 1216 tests from 34 test cases ran. (209447 ms total)
[  PASSED  ] 1210 tests.
[  FAILED  ] 6 tests, listed below:
[  FAILED  ] memory_manager.bad_free
[  FAILED  ] CPU.one_hot_scalar_fp_nonint_in_3
[  FAILED  ] CPU.one_hot_scalar_oob_in_3
[  FAILED  ] CPU.one_hot_vector_1_barely_oob
[  FAILED  ] CPU.one_hot_vector_1_far_oob
[  FAILED  ] CPU.one_hot_vector_1_fp_nonint

 6 FAILED TESTS
  YOU HAVE 15 DISABLED TESTS

make[3]: *** [test/CMakeFiles/unit-test-check] Error 1
make[2]: *** [test/CMakeFiles/unit-test-check.dir/all] Error 2
make[1]: *** [test/CMakeFiles/unit-test-check.dir/rule] Error 2
make: *** [unit-test-check] Error 2
`

silee2:
[ RUN      ] memory_manager.bad_free
/Users/silee2/Projects/ngraph/test/pass_memory_layout.cpp:181: Failure
Expected: mm.free(10) throws an exception of type std::runtime_error.
  Actual: it throws a different type.
[  FAILED  ] memory_manager.bad_free (0 ms)

diyessi:
I wonder if some stuct changed size between what's in the headers we're using and what's in the binary llvm.

aprocter:
When I chatted with @jmenon about this his theory is that we are linking against LLVM's libc++ instead of the system libc++. This would make sense, because when we build LLVM from source we do not include libc++ so it would be forced to default to the system library.

silee2:
By the way, why would LLVM matter for this case?
[ FAILED ] memory_manager.bad_free

aprocter:
@silee2 It's the same underlying error. That test is supposed to catch a bad free in our memory manager, which throws `std::invalid_argument` or something like that.

Of note is that the many tests which check for `ngraph_error` are not failing. So it's seemingly only standard library exceptions where this happens.

postrational:
Unfortunately does not work on older versions of MacOS:
```
[ 76%] Linking CXX shared library libngraph.so
ld: warning: -pie being ignored. It is only used when linking a main executable
ld: warning: object file (/.../ngraph/build/third-party/ext_llvm-prefix/src/ext_llvm/lib/libLLVMX86AsmParser.a(X86AsmParser.cpp.o)) was built for newer OSX version (10.13) than being linked (10.12)```

silee2:
Changed the status to WIP

silee2:
Not too much value at this point.

PR #723, opened 2018-03-22
Catch performance issues due to op::Result early on

Krovatkin:
Note, this will start passing as soon as #714 gets merged.


Krovatkin:
TESTNOW

adstraw:
@diyessi is unlikely to be able to look at this today due to all-day meetings.  can someone else review?

PR #724, opened 2018-03-22
ngmx-372 Fuse RNN matrix operations

louisfeng:
Added a pass to fuse multi time step RNN matrix operations.


src/ngraph/runtime/cpu/pass/cpu_rnn_mat_fusion.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    // remove ops with less than segment_count number of segments
+    for (auto op_it = op_seg_map.cbegin(); op_it != op_seg_map.cend();)
+    {
+        // remove ops with less than expected segements
+        bool valid = true;
+        for (auto& seg : op_it->second)
+        {
+            if (seg.empty())
```
Krovatkin: I think this needs a comment. This seems to mean a segment of a particular type is missing (kinda `nullptr`) so that slot is initialized w/ a default segment.

src/ngraph/runtime/cpu/pass/cpu_rnn_mat_fusion.cpp
```diff
@@ -0,0 +1,288 @@
...
+    std::vector<NodeSegment> segment_bundle;
+    for (auto& node : param_nodes)
+    {
+        NodeSegment segment;
+        FindValidSegments(node, segment, segment_bundle, valid_sequences, 0);
+    }
+
+    // combined all segments by last operator
+    std::map<NodePtr, std::vector<NodeSegment>> op_seg_map;
+    for (const auto& segment : segment_bundle)
```
Krovatkin: heh, why `bundle` why not just `segments`?

src/ngraph/runtime/cpu/pass/cpu_rnn_mat_fusion.cpp
```diff
@@ -0,0 +1,288 @@
...
+        {
+            it = param_list.erase(it);
+        }
+        else
+        {
+            ++it;
+        }
+    }
+
+    // Expecting in put data shape D=[x, y, z], weights W=[u, v], bias B = [w]
```
Krovatkin: a typo `input`

src/ngraph/runtime/cpu/pass/cpu_rnn_mat_fusion.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+struct OrderedParams
+{
+public:
+    OrderedParams()
+        : m_params{{nullptr, nullptr, nullptr}}
+    {
+    }
+
+    bool exist(const NodeSegment::Type type) { return m_params[type] != nullptr; }
```
Krovatkin: is `exist` used anywhere?

src/ngraph/runtime/cpu/pass/cpu_rnn_mat_fusion.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    void set(const NodeSegment::Type type, const NodePtr& node) { m_params[type] = node; }
+    NodePtr get(const NodeSegment::Type type) const { return m_params.at(type); }
+    friend bool operator<(const OrderedParams& a, const OrderedParams& b);
+
+private:
+    // order based on NodeSegment::Type
+    // <data, weights, bias>
+    std::array<NodePtr, 3> m_params;
```
Krovatkin: Should we replace `3` w/ some constant `NUM_MMB_ARGS`?

test/cpu_fusion.cpp
```diff
@@ -1012,3 +1014,102 @@ TEST(cpu_fusion, batchnorm_fprop_inference_b2c2h2w1)
...
+    Shape bias_shape{6};
+
+    size_t data_size =
+        std::accumulate(begin(data_shape), end(data_shape), 1, std::multiplies<size_t>());
+    size_t weights_size =
+        std::accumulate(begin(weights_shape), end(weights_shape), 1, std::multiplies<size_t>());
+    size_t bias_size =
+        std::accumulate(begin(bias_shape), end(bias_shape), 1, std::multiplies<size_t>());
+    std::mt19937_64 gen(0);
+    std::uniform_real_distribution<> real_random(0.0, 1.0);
```
Krovatkin: Can we use our `rng` which is used for both `backend` and `backwards` tests?

test/cpu_fusion.cpp
```diff
@@ -1012,3 +1014,102 @@ TEST(cpu_fusion, batchnorm_fprop_inference_b2c2h2w1)
...
+
+TEST(cpu_fusion, rnn_matrix_fusion_eval_pass)
+{
+    const size_t time_steps = 4;
+    Shape data_shape{3, time_steps, 5};
+    Shape weights_shape{6, data_shape[2]};
+    Shape bias_shape{6};
+
+    size_t data_size =
+        std::accumulate(begin(data_shape), end(data_shape), 1, std::multiplies<size_t>());
```
Krovatkin: We have a helper for this `shape_size`

src/ngraph/runtime/cpu/pass/cpu_rnn_mat_fusion.cpp
```diff
@@ -0,0 +1,288 @@
...
+{
+    bool modified = false;
+
+    const NodeTypeSequenceList valid_sequences{
+        {NodeSegment::DATA,
+         {TI(op::Parameter), TI(op::Slice), TI(op::Reshape), TI(op::Dot), TI(op::Add)}},
+        {NodeSegment::WEIGHTS, {TI(op::Parameter), TI(op::Reshape), TI(op::Dot), TI(op::Add)}},
+        {NodeSegment::BIAS, {TI(op::Parameter), TI(op::Broadcast), TI(op::Add)}}};
+    // this is the expected sequence count for fusing
+    const size_t segment_count = 3;
```
Krovatkin: I guess we won't be handling a case where we only have a dot and no bias?

src/ngraph/runtime/cpu/pass/cpu_rnn_mat_fusion.cpp
```diff
@@ -0,0 +1,288 @@
...
+};
+typedef std::pair<NodeSegment::Type, std::vector<std::type_index>> NodeTypeSequence;
+typedef std::list<NodeTypeSequence> NodeTypeSequenceList;
+
+// Preorder traversal to collect all valid segments in the graph
+// precondition: all valid sequences must be unique
+// [a, b, c] and [a, c, b] are different, for valid sequences like [a, b, c] and [a, b], the
+// longest sequence will be matched.
+void FindValidSegments(const NodePtr& node,
+                       NodeSegment segment,
```
Krovatkin: Heh, this is literally a matcher in reverse 😄 I'm kinda wondering if it might be easier to simply go over the entire graph and give a matcher your sequences (in the reverse order) if it returns a match -- excellent, you got a new sequence of a particular type and a node it starts from. if @jbobba doesn't mind we could spend some time to try to simplify this. As a bonus, the labels would map to parameters and we could be constructing some of the maps @louis constructs later on, in parallel w/ a graph walk.

test/cpu_fusion.cpp
```diff
@@ -1012,3 +1014,102 @@ TEST(cpu_fusion, batchnorm_fprop_inference_b2c2h2w1)
     ASSERT_TRUE(
         ngraph::test::all_close(expected_result, read_vector<float>(bn_output), 1e-3f, 1e-4f));
 }
+
+std::vector<shared_ptr<runtime::TensorView>>
+    rnn_matrix_fusion_eval(const size_t time_steps,
```
Krovatkin: awesome tests!

src/ngraph/runtime/cpu/pass/cpu_rnn_mat_fusion.cpp
Krovatkin: I'd replace a `preorder` traversal with `users' chain` or stn like dat.

louisfeng:
TEST NOW

mbrookhart:
Would we see a similar performance gain if we did the same thing on bprop? i.e., a bunch of dots going into a concat transformed into a concat going into a bunch of dots?

louisfeng:
@mbrookhart Something I will look into next. If there's any chance to combine these ops, we should probably do it.

Krovatkin:
> We need to generalize this to handle RNN fusion deeper in the graph and not just at the top of it. Also will be nice if we can figure out a way to use the pattern matcher.

@jbobba @louisfeng @mbrookhart 

I'd like to clarify that I'm partially to blame for the fact @louisfeng didn't use PM for his rnn fusion pass since I suggested that we might need to walk user chains to collect the info the pass needed. After looking at @louisfeng implementation I realized that there might be another way (a *simpler* one, IMO) to skin a proverbial cat.

Issue #725, opened 2018-03-22
Getting a 404 for links to ngraph-tensorflow in http://ngraph.nervanasys.com/docs/latest/framework-integration-guides.html


Ajay191191:
The [doc](http://ngraph.nervanasys.com/docs/latest/framework-integration-guides.html#run-mnist-mlp-through-the-tensorflow-xla-plugin-to-ngraph) points to this tensorflow repo: https://github.com/NervanaSystems/ngraph-tensorflow which is not accessible.

yxlao:
Thanks @Ajay191191!

Currently, the doc in http://ngraph.nervanasys.com/docs/latest/framework-integration-guides.html#tensorflow represents our efforts integrating with TF1.3. The repository with TF1.3 integration has been moved to https://github.com/NervanaSystems/ngraph-tensorflow-1.3. We'll update this doc accordingly. 

We're also actively working on a "dynamic-plugin" approach for newer versions of TF.  Please stay tuned.

avijit-nervana:
Yes - we are fixing this typo in the documentation.

PR #726, opened 2018-03-22
Fixed the bug in the TensorFlow git URL.

avijit-nervana:
Issue #725


adstraw:
does this replace #707 ?

PR #732, opened 2018-03-23
[WIP] Change CPU emitter function signature.

rkimballn1:
Make CPU emitters callable from NNP backend without the need for a CPU External Function.

For the hybrid NNP backend we want to use some of the CPU functions for things not yet implemented on the NNP. This change makes it easier.


jbobba:
@rkimballn1 is trying an alternate approach using non-static emitter objects.

PR #735, opened 2018-03-23
tensor (sequence) mask

adstraw:


src/ngraph/builder/tensor_mask.cpp
```diff
@@ -0,0 +1,66 @@
...
+    auto broadcast_sequence_lengths =
+        std::make_shared<op::Broadcast>(sequence_lengths, mask_shape, broadcast_axes);
+
+    // create sequence data [0, ..., max_sequence_length]
+    auto max_sequence_length = mask_shape[sequence_axis];
+    std::vector<uint32_t> sequence_data(max_sequence_length);
+    std::iota(sequence_data.begin(), sequence_data.end(), 0);
+
+    // create sequence constant
+    // TODO: how to handle sequence type / enforce sequence_lengths type?
```
aprocter: I can see a couple of things you could do here. One is to insist on a particular element type for sequence_lengths; the other is query the element type of sequence_lengths and force `sequence` to have the same element type, either through a `Convert` or by explicit construction.

I think a `Convert` actually makes reasonable sense here, because there's really no reason we shouldn't have an optimization pass that collapses a cast of a constant into a constant in the target type. If we don't use a `Convert` here we run into the usual bureaucratic nightmare.

PR #739, opened 2018-03-24
New backend/transformer API

rkimballn1:
Ignore the backend implementation. Just look at the runtime::Backend header. All of the stuff below the comment `New backend` is the backend API. No CallFrame, ExternalFunction or Manager exist with the new API.
You use it like this:
```
    auto backend = runtime::Backend::create("CPU");
    // two different ways to create tensors
    auto a = backend->create_tensor<float>(shape);
    auto b = backend->create_tensor(element::f32, shape);
    auto result = ...
    backend->call(f, {result}, {a, b});
```
I'm not sure about the `create_device/host_tensor()` part. We may just want a `create_tensor()`

API CHANGE:
Backend::compile now takes a `const Function&` instead of `shared_ptr<Function>`


rkimballn1:
I implemented the INT backend just to make sure things work. It passes the single unit test that I converted to the new API. This PR also leaves the old API untouched so it is backwards compatible. We can remove the old API after things have been converted.

mbrookhart:
I like the API change, it's a lot cleaner, but I have one main concern. 

We currently use the same backend for multiple functions.

Last I heard, we couldn't initialize more than 1 backend for Argon, so the MXNet bridge is setup to store backends centrally. We'd need to change Argon to allow multiple backends, and then refactor the bridge quite a bit.

diyessi:
Can you compile the same function on more than one backend?

rkimballn1:
In order to compile the same function on multiple backends we need to make a copy of the function. I think I should change the signature of compile to be `compile(const Function&)` to force the backend to make a copy.

rkimballn1:
I don't see any reason we can't create multiple instances of the same backend. For things like GPU we can do something like `backend::create("GPU:0")` and pass everything after the colon to the backend. That way we can create one backend per GPU.

diyessi:
I think CUDA has a way to tie multiple backends together, but the string could include that information for the backend.

yxlao:
Argon transformer could be made compatible by calling HW initialize and the first Argon backend constructor and HW finalize at the last Argon backend destructor. This would allow us to have multiple Argon backend instances.

src/ngraph/runtime/backend.hpp
```diff
@@ -44,20 +40,58 @@ namespace ngraph
...
+
+            template <typename T>
+            std::shared_ptr<ngraph::runtime::TensorView> create_tensor(const Shape& shape)
+            {
+                return create_tensor(element::from<T>(), shape);
+            }
+
+            virtual bool compile(const ngraph::Function& fun) = 0;
+
+            virtual bool is_callable() const = 0;
```
aprocter: What does this function mean?

src/ngraph/op/function_call.cpp
```diff
@@ -51,7 +52,9 @@ op::FunctionCall::FunctionCall(shared_ptr<Function> function, const NodeVector&
 
 shared_ptr<Node> op::FunctionCall::copy_with_new_args(const NodeVector& new_args) const
 {
-    return make_shared<FunctionCall>(m_function, new_args);
+    shared_ptr<FunctionCall> fc = make_shared<FunctionCall>(m_function, new_args);
+    fc->m_function = clone_function(*m_function);
```
aprocter: I don't think we should simply clone the function here, in case a single function has multiple call sites. (Of course, it's hard to say how big a deal this actually is, since currently only the TF bridge is using embedded functions, and then only as a last resort.) I guess we would have to maintain a `FunctionMap` akin to the `NodeMap` over in `clone_function`, and pass `new_functions` in to `copy_with_new_args`.

rkimballn1:
1) Tensor reuse across functions is certainly possible as the management of these tensors is handled by the caller/framework. A single copy of a tensor could be passed to multiple functions. We should probably make sure that input tensors are const so we know they wont be changed.
2) If you want to avoid the copy of the Function then we could change the parameter for `compile()` to take a unique_ptr instead. That way the compile function has ownership of the function and can mutate it at will without fear of changing the Function for other users.

src/ngraph/op/function_call.cpp
rkimballn1: When we `compile()` a graph we mutate it. Running the same function through a CPU backend and then an NNP backend would cause mayhem, and probably fail, if we don't actually copy stuff.
copy_with_new_args should actually make a copy, it was not before this change as it maintained a pointer to the same function as the original op.
I don't understand the need for FunctionMap, please `splain

src/ngraph/runtime/backend.cpp
```diff
@@ -0,0 +1,40 @@
...
+using namespace std;
+using namespace ngraph;
+
+std::shared_ptr<runtime::Backend> runtime::Backend::create(const std::string& type)
+{
+    std::shared_ptr<Manager> manager = runtime::Manager::get(type);
+    return manager->allocate_backend();
+}
+
+vector<string> get_registered_devices()
```
ransford2011: Are these just placeholder functions? Currently there is no behavior being performed here

src/ngraph/runtime/backend.hpp
```diff
@@ -44,20 +40,58 @@ namespace ngraph
...
+            static std::shared_ptr<Backend> create(const std::string& type);
+
+            /// @brief Query the list of registered devices
+            /// @returns A vector of all registered devices.
+            static std::vector<std::string> get_registered_devices();
+
+            /// @brief Query the list of available subdevices of a particular device.
+            /// @param type The name of a registered backend, such as "CPU" or "GPU"
+            /// @returns A vector of available devices of the specified type.
+            static std::vector<size_t> get_subdevices(const std::string& type);
```
ransford2011: This seems a bit dangerous having a static function defined along with a member function with the same name( `get_subdevices`). Although they have different signatures, looks real easy to mess things up.

src/ngraph/runtime/backend.cpp
rkimballn1: Yea, placeholders. This PR is not fully functional. It is intended to get the API designed, but I should fix this one.

src/ngraph/runtime/backend.hpp
rkimballn1: One of them is private so not in the API

rkimballn1:
Ok, the new, new, new API is up. The big change is that `call(outputs, inputs)` is now `call(function, outputs, inputs)`. This makes things more understandable. There can be multiple functions on a backend. The first call to this method will compile `function` and following calls will used this cached compile.
I left the old call function in for backwards compatibility but it is deprecated and will be remove in a PR that will quickly follow this one.

src/ngraph/runtime/backend.cpp
rkimballn1: I added an implementation for this

src/ngraph/runtime/backend.hpp
rkimballn1: I'm not sure so I removed it.

src/ngraph/runtime/gpu/gpu_backend.cpp
```diff
@@ -34,3 +34,20 @@ std::shared_ptr<ngraph::runtime::TensorView>
     auto rc = make_shared<runtime::gpu::GPU_TensorView>(element_type, shape);
     return dynamic_pointer_cast<runtime::TensorView>(rc);
 }
+
+bool runtime::interpreter::GPU_Backend::compile(const std::shared_ptr<ngraph::Function>& fun)
+{
+    return false;
+}
+
+bool runtime::interpreter::GPU_Backend::is_callable() const
```
ransford2011: Shouldn't this be removed?

src/ngraph/runtime/gpu/gpu_backend.hpp
```diff
@@ -36,6 +36,13 @@ namespace ngraph
                 std::shared_ptr<ngraph::runtime::TensorView>
                     make_primary_tensor_view(const ngraph::element::Type& element_type,
                                              const Shape& shape) override;
+
+                bool compile(const std::shared_ptr<ngraph::Function>& fun) override;
+
+                bool is_callable() const override;
```
ransford2011: I'm surprised this didn't fail at compile time since you removed the inherited function call.

rkimballn1:
@ransford2011 The public API has a single definition of `get_subdevices(const string& backend)` so the user only sees one function.
The second function is internal implementation details. For now I have move the second non-static version to the runtime::Manager class where is will probably stay.

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -32,6 +35,55 @@ std::shared_ptr<ngraph::runtime::TensorView>
...
+                                     const vector<shared_ptr<runtime::TensorView>>& outputs,
+                                     const vector<shared_ptr<runtime::TensorView>>& inputs)
+{
+    bool rc = true;
+    try
+    {
+        compile(fun);
+        call(outputs, inputs);
+    }
+    catch (...)
```
aprocter: This will change the expectations of a couple of unit tests (specifically the one_hot tests, and a couple others I can't remember) which are expecting to catch certain exceptions thrown inside the generated code. But in retrospect perhaps that's not something we should be doing in the first place.

src/ngraph/op/function_call.cpp
aprocter: I still think we should clone a function only once per graph-copy, i.e., if a function is used by more than one op in the original graph, then we should preserve that sharing in the new graph, albeit with a new copy of the function. I understand that it doesn't change anything semantically, but if someone is generating a graph that calls the same function in many places, this may be unexpected and undesired behavior.

That said, AFAIK we don't actually use multiple call sites in practice (in AFAIK TF is the only backend using embedded functions at all, and even there we try to avoid them whenever we can), so if we just document that functions with multiple call sites may be duplicated, I don't suppose it's that big a deal.

rkimballn1:
The python code that the user sees such as this example program does not need to change in any way. It works just as it is. There will be some changes inside of the python wrapper, but this PR does not require those changes as I left the old API intact. The old API will be removed in a separate PR.

diyessi:
I think some of the name choices in python might want to be more compatible. The python-ngraph names had to start out in a world that co-existed with the pre-ngraph version of neon, which had its own backend, and that is why "runtime" was used. We're no longer in that world.
For the most part, the customers of this API are the frameworks and transformers, all still within Intel. Now is the time to straighten this out (and the doc examples will need updating) before it's outside of AIPG.

crlishka:
TESTNOW

src/ngraph/runtime/backend.hpp
avijit-nervana: Is it possible to remove the DEPRICATED functions?

src/ngraph/runtime/manager.hpp
```diff
@@ -36,13 +37,17 @@ namespace ngraph
...
+            friend class runtime::Backend;
+
         public:
             virtual ~Manager() {}
             /// @brief Allocate a backend for this transformer.
             ///
             /// Specific transformers may provide addtional methods for allocating customized backends.
             virtual std::shared_ptr<Backend> allocate_backend() = 0;
 
+            virtual std::vector<size_t> get_subdevices() const = 0;
```
avijit-nervana: Could you please add some comments with example usage for tis function?

src/ngraph/runtime/manager.hpp
```diff
@@ -36,13 +37,17 @@ namespace ngraph
...
         class Manager
         {
+            friend class runtime::Backend;
+
         public:
             virtual ~Manager() {}
             /// @brief Allocate a backend for this transformer.
             ///
             /// Specific transformers may provide addtional methods for allocating customized backends.
             virtual std::shared_ptr<Backend> allocate_backend() = 0;
```
avijit-nervana: Please describe the return value i.e., if allocate fails what is the return (possibly nullptr?)
Also, how does the caller know the reason for the failure. Can this function either a failure object or an actual value - may be a good idea to adopt something similar here?
(This is applicable to error reporting in general - so may be we could address it outside of this PR)

rkimballn1:
TESTNOW

PR #743, opened 2018-03-25
lighten weight of ngraph_theme by not customizing webtheme w font files so large 

indie:


Krovatkin:
@rkimballn1 I think this is also about documentation reading experience for our users. If they don't have to load 19MBs to read our documentation, that could be a win!

indie:
@rkimballn1  I don't quite understand what you want me to do here in these "requesting changes" ...  Can you please clarify?  Close the PR w/no changes?  Something else? Thx.

PR #744, opened 2018-03-25
switching to using get_users()

Krovatkin:
After this change, the only uses of `users()` are the ones that keep I/O and node views in sync. In other words, we should be able to remove `m_arguments` and `m_users`


Krovatkin:
TEST NOW

PR #745, opened 2018-03-25
Forbid replacing multi-output nodes directly until we have a clear valid use case in mind

Krovatkin:


Krovatkin:
TEST NOW

DawnStone:
TESTNOW

src/ngraph/graph_util.cpp
```diff
@@ -123,6 +123,11 @@ void ngraph::replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> re
         throw ngraph_error("Result nodes cannot be replaced.");
     }
 
+    if (target->get_outputs().size() > 1)
+    {
+        throw ngraph_error("Multi-output nodes cannot be replaced");
```
diyessi: "Multi-output node replacement is not yet implemented"

Krovatkin: @diyessi I'm not sure we want to implement it 😄 https://github.com/NervanaSystems/ngraph/pull/757 is a very good example where we replace a multi-output op by replacing all its GOEs. It might be a bit verbose, but it's very explicit and simple.

diyessi: Then the error should give some hint about how to achieve what the user is trying to do, rather than say it is impossible.

Krovatkin: that's a very good point. I'll reword the message.

PR #746, opened 2018-03-25
BatchNorm training tests

Krovatkin:


Krovatkin:
@aprocter I still owe you inference tests. I'll do them next. I noticed if I try to do all the tests at once, I end up not checking in anything at ll :-(

PR #748, opened 2018-03-26
Split cpu_fusion.cpp into cpu_fusion.cpp and cpu_test.cpp; clean up headers

Krovatkin:
I was meaning to split `cpu_fusion.cpp` into two files for a long time since it's been harbouring quite a few tests that aren't fusion specific and run only on CPU. @rkimballn1 is this appropriate? I could try moving them to `backend_test.in.cpp` and disabling all the backends but it seems rather awkward?


DawnStone:
TESTNOW

DawnStone:
TESTNOW

PR #749, opened 2018-03-26
[Py] Enable Softmax op

tsocha:


python/ngraph/ops.py
```diff
@@ -421,3 +421,16 @@ def concat(nodes, axis):  # type: (List[Node], int) -> Node
     :return: Return new node that is a concatenation of input nodes.
     """
     return Concat(NodeVector(nodes), axis)
+
+
+@nameable_op
+def softmax(node, axes):  # type: (Node, Iterable[int]) -> Node
+    """Concatenate input nodes into single new node along specified axis.
```
postrational: What?

python/ngraph/ops.py
```diff
@@ -421,3 +421,16 @@ def concat(nodes, axis):  # type: (List[Node], int) -> Node
...
+
+    :param nodes: The nodes we want concatenate into single new node.
+    :param axis: The axis along which we want to concatenate input nodes.
+    :return: Return new node that is a concatenation of input nodes.
+    """
+    if axes is not set:
```
postrational: And this is not checking type.

PR #753, opened 2018-03-26
Fix a Relu(Conv) test by making sure Relu(Conv) won't be fused

Krovatkin:


Krovatkin:
TESTNOW

test/cpu_fusion.cpp
```diff
@@ -1003,46 +1003,90 @@ TEST(cpu_fusion, fuse_conv_relu)
...
+    return result_vectors;
+}
+
+TEST(cpu_fusion, conv_relu_n2c1h2w2_2)
+{
+    Shape shape_a{2, 1, 6, 6};
+    Shape shape_weights{1, 1, 2, 2};
 
-    cf->call({_conv_relu, _relu}, {_a, _weights});
-    EXPECT_TRUE(test::all_close(read_vector<float>(_conv_relu), read_vector<float>(_relu)));
```
pruthviIntel: may be better to use full name here, make_interpreter_function. ? :)

test/cpu_fusion.cpp
```diff
@@ -1003,46 +1003,90 @@ TEST(cpu_fusion, fuse_conv_relu)
...
+    auto make_int_function = [shape_a, shape_weights]() {
```
pruthviIntel: may be better to use full name here, make_interpreter_function. ? :)

Krovatkin: @pruthviIntel , actually, it looks so aligned everywhere else 😄 
```cpp
    auto int_results = execute(int_f, args, "INTERPRETER");
    auto cpu_results = execute(cpu_f, args, "CPU");
```

how strongly do you feel about this?

pruthviIntel: hmm, true!. its okay, then.

PR #756, opened 2018-03-26
add exception handing in deserializer

rkimballn1:
add exception handing in deserializer to tell which op actually caused the exception. The exception now contains the name of the op that caused the exception. I know the PR looks like a lot changed, but that is just the indenting change. it is really just a try block around all of the deseriailization code.


rkimballn1:
TESTNOW

PR #757, opened 2018-03-26
Relu(BatchNorm) Fusion

Krovatkin:


src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -681,6 +682,65 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias()
...
+            m_bn->get_eps_value(), pattern_map[gamma], pattern_map[beta], pattern_map[input]);
+
+        auto bn_relu_output = std::make_shared<op::GetOutputElement>(bn_relu, 0);
+        auto bn_relu_mean = std::make_shared<op::GetOutputElement>(bn_relu, 1);
+        auto bn_relu_var = std::make_shared<op::GetOutputElement>(bn_relu, 2);
+
+        std::shared_ptr<Node> new_nodes[] = {bn_relu_output, bn_relu_mean, bn_relu_var};
+
+        std::vector<std::shared_ptr<Node>> mgoes(m_bn->get_outputs().size());
+        for (auto _input : m_bn->get_output_inputs(0))
```
pruthviIntel: Nick, can you explain whats happening here? R we inserting GetOutputElement to bnRelu output's?

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -578,21 +578,26 @@ size_t MKLDNNEmitter::build_batchnorm_forward(const mkldnn::memory::desc& input_
...
     size_t mean_index = build_memory_primitive(mean_desc);
     size_t variance_index = build_memory_primitive(variance_desc);
 
+    mkldnn::primitive_attr conv_attr;
+    conv_attr.set_post_ops(pops);
+
     if (bn_training_flag)
     {
         size_t batchnorm_index = insert_primitive(new mkldnn::batch_normalization_forward(
             {{mkldnn::prop_kind::forward_training,
```
pruthviIntel: Can you point me to the constructor which takes post_ops, I dint find the relevant one in the bn constructor.

pruthviIntel: also the workspace requirement is only needed for bprop ? just wanted to clarify.

Krovatkin: Yes, we are replacing the old GOEs with the new ones. This is the easiest. Otherwise, we have to deal with I/O views.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
Krovatkin: yes, we are replacing the old GOEs with the new ones.

test/cpu_fusion.cpp
```diff
@@ -986,6 +989,79 @@ TEST(cpu_fusion, sigmoid_bprop_n1c1h4)
...
+    auto f = make_shared<Function>(
+        NodeVector{output_relu, mean_rt, variance_rt, output_rt_bnr, mean_rt_bnr, variance_rt_bnr},
+        op::ParameterVector{input, gamma, beta});
+    auto manager = runtime::Manager::get("CPU");
+    auto external = manager->compile(f);
+    auto backend = manager->allocate_backend();
+    auto cf = backend->make_call_frame(external);
+
+    // Create some tensors for input/output
+    auto _input = backend->make_primary_tensor_view(element::f32, Shape{1, 2, 2, 2});
```
diyessi: Style and C++ violation.

Krovatkin: @diyessi will fix!

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -473,6 +474,80 @@ namespace ngraph
...
+                }
+
+                const float ops_scale = 1.f;
+                const float ops_alpha = -0.f; // relu negative slope
+                const float ops_beta = 0.f;
+
+                mkldnn::post_ops ops;
+                ops.append_eltwise(ops_scale, mkldnn::algorithm::eltwise_relu, ops_alpha, ops_beta);
+
+                writer.indent++;
```
jbobba: Use block_begin() and block_end()

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -473,6 +474,80 @@ namespace ngraph
...
+                       << out[1].get_name() << ");\n";
+                writer << "cpu::mkldnn_utils::set_memory_ptr(ctx, " << to_string(deps[4]) << ", "
+                       << out[2].get_name() << ");\n";
+
+                writer << "cpu::mkldnn_utils::mkldnn_invoke_primitive(ctx, "
+                       << to_string(batchnorm_index) << ");\n";
+
+                writer.indent--;
+                writer << "}\n";
+            }
```
jbobba: Check for case where BatchNormRelu is not assigned to MKLDNN and throw error

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -578,21 +578,26 @@ size_t MKLDNNEmitter::build_batchnorm_forward(const mkldnn::memory::desc& input_
...
+                                              bool bn_training_flag,
+                                              const mkldnn::post_ops& pops)
 {
     size_t input_index = build_memory_primitive(input_desc);
     size_t weights_index = build_memory_primitive(weights_desc);
     size_t result_index = build_memory_primitive(result_desc);
     size_t mean_index = build_memory_primitive(mean_desc);
     size_t variance_index = build_memory_primitive(variance_desc);
 
+    mkldnn::primitive_attr conv_attr;
```
jbobba: why conv_attr?

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
Krovatkin: @pruthviIntel it's on `primitive_desc`. This one is for Convolution, but batch normalization has the same one : 
https://github.com/intel/mkl-dnn/blob/master/examples/simple_net_int8.cpp#L114

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
Krovatkin: Will rename

src/ngraph/runtime/cpu/cpu_emitter.cpp
Krovatkin: Could you please elaborate?

src/ngraph/runtime/cpu/cpu_emitter.cpp
Krovatkin: Will do

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
Krovatkin: > also the workspace requirement is only needed for bprop ? just wanted to clarify. 


I think so. @jbobba  ?

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -477,6 +477,11 @@ namespace ngraph
             template <>
             void CPU_Emitter::EMITTER_DECL(ngraph::op::BatchNormRelu)
```
diyessi: Wouldn't it be better to check for this condition in the node constructor since you'd catch the error closer to its origin?

Krovatkin: @diyessi If I understand @jbobba 's master plan, we will have multiple libraries (e.g. `libxsmm`, `mkldnn`) that might handle different cases and one or the other might be able to handle a case where batchnorm's input rank isn't necessarily 4. The checks in `BatchNormRelu::BatchNormRelu` and `CPUFusion` would then be the superset of the checks for individual libraries.

diyessi: But you still wouldn't want to create the special op if the library it needed wasn't available; so the test in the constructor would check that the required lib existed. Presumably the optimizer that introduces the fused op doesn't happen if the lib isn't available.

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -477,6 +477,11 @@ namespace ngraph
...
             {
+                if (!mkldnn_utils::use_mkldnn_kernel(node))
```
jbobba: This needs to be similar to the conv+bias emitter for consistency.

src/ngraph/runtime/cpu/cpu_emitter.cpp
jbobba: CPUfusion should have a check to not generate a Batchnorm+Relu unless its of shape supported by the backend at that moment.

jbobba: cpu_emitter check is more of a paranoia check.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -681,6 +682,65 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias()
...
+
+        auto pattern_map = m.get_pattern_map();
+        auto m_bn = std::dynamic_pointer_cast<op::BatchNorm>(
+            m.match_root()->get_input_op(0)->get_inputs().at(0).get_output().get_node());
+
+        if (!m_bn->get_training_flag())
+        {
+            NGRAPH_DEBUG << " This is an inference batchnorm, so skipping fusion";
+            return false;
+        }
```
jbobba: Add a check to ensure that Relu is the only consumer of Batchnorm data output. Otherwise, there is no point in fusing.

Krovatkin:
@jbobba @diyessi Guys could you please take another look? I believe I addressed all yr comments.

Krovatkin:
@pruthviIntel if the PR looks good to you, could you please approve it?

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
pruthviIntel: @Krovatkin , where the above condition as @jbobba pointed out is taken care?. if the fused Op output going to multiple op's then essentially we are computing bn+relu multiple times.  I guess we need to check this as well Relu->get_output_size() != 1 don't fuse

pruthviIntel:
@Krovatkin added some comments. check it out. rest looks okay for me

jbobba:
@Krovatkin please add support for the new op in the layout pass. Just realized that it was missing.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -112,6 +113,19 @@ namespace ngraph
                     convolution->set_op_annotations(op_annotations);
                 }
 
+                template <>
+                void CPUAssignment::ASSIGN_DECL(ngraph::op::BatchNormRelu)
+                {
+                    if (node->get_input_op(2 /*input data*/)->get_shape().size() == 4)
+                    {
+                        auto convolution = static_cast<op::BatchNormRelu*>(node);
```
jbobba: change variable name.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
jbobba: If we have something like BN{A}->Relu{B}, BN{A}->Conv{C}, will this fuse nodes A and B? If so, we end up with two BN operations, which is going to be less efficient.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -681,6 +682,80 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias()
...
+        //and it requires input data's rank to be equal to 4
+        if (pattern_map[input]->get_shape().size() != 4)
+        {
+            NGRAPH_DEBUG << " Input data's rank isn't equal to 4. Shape = "
+                         << pattern_map[input]->get_shape().size();
+            return false;
+        }
+
+        std::vector<std::shared_ptr<Node>> mgoes(m_bn->get_outputs().size());
+        for (auto _input : m_bn->get_output_inputs(0))
```
diyessi: _input

PR #759, opened 2018-03-26
Build MKL-DNN without tests

chaseadams509:
Set ExternalProject_Add MKL-DNN to only build the mkl.so without any tests.


aprocter:
TESTNOW

PR #762, opened 2018-03-27
make BatchNorm backwards compatible

rkimballn1:


src/ngraph/serializer.cpp
```diff
@@ -411,7 +411,8 @@ static shared_ptr<ngraph::Function>
         else if (node_op == "BatchNorm")
         {
             auto epsilon = node_js.at("eps").get<double>();
-            if (node_js.at("training"))
+            bool training = get_or_default<bool>(node_js, "training", true);
+            if (training)
             {
```
Krovatkin: Maybe we could add a check that if this flag is set then a batchnorm *must* have 3 arguments and 5 otherwise? @rkimballn1  @jbobba

jbobba: Agree that a check would be good. The other option is to not rely on this attribute and only look at the number of arguments to infer batchnorm type.

PR #761, opened 2018-03-27
[WIP] Unify setup.py

postrational:
Work in progress. Not yet fully tested. **Do not merge**.

This change is meant to address issues we are currently facing due to the existence of 2 `setup.py` files (`setup.py` and `setup.py.in`).

One is used by cmake, the other is used by Python (e.g. `python setup.py develop`, `tox`).

It would make sense to have just one file which can be used in both cases.

This should be tested in these scenarios:

1. `cmake` and `build_python2&3_wheel.sh` scripts
2. Independent of `cmake` when calling `python setup.py *`
3. as part of `tox`
4. as part of our CI process

PR #764, opened 2018-03-27
Update README to make it consistent with ngraph-neon

nbpatel:
Making the installation doc consistent with http://ngraph.nervanasys.com/docs/latest/framework-integration-guides.html#neon-trade and having the .sh script as an alternative installation procedure.


postrational:
You should also mention:
```
export DYLD_LIBRARY_PATH=$HOME/ngraph_dist/lib  # On MacOS
```
Only needed on MacOS, but I think it's required.

silee2:
How is pybind11 downloaded in this case?

DawnStone:
TESTNOW

raramer01:
The python_ci  check is failing due to the following compilation error when building the wrappers:

[branch_cpu_py2] /root/ng/python/pyngraph/serializer.cpp: In function 'void regclass_pyngraph_Serializer(pybind11::module)':
[branch_cpu_py2] /root/ng/python/pyngraph/serializer.cpp:33:87: error: address of overloaded function with no contextual type information
[branch_cpu_py2]            (std::string(*)(std::shared_ptr<ngraph::Function>, size_t, bool)) & ngraph::serialize,
[branch_cpu_py2]                  

not sure why this is happening...has there been a change to the version of pybind11 that we are using?                                                                      ^

nbpatel:
@raramer01 i created #780. That should fix it. The master has changed and hence the failure in the wrapper.

raramer01:
thanks!  Yep, Sang Ik and Baojun found the api change as well...hopefully we can get past this issue soon!

postrational:
Since we merged #761 we can mention the env var `PYBIND_HEADERS_PATH` which will be honored by `setup.py` and can be used to point to the location of the PyBind11 download.

python/README.md
```diff
@@ -1,38 +1,27 @@
...
-To build python2 bdist wheel type
 ```
-./build_python2_wheel.sh
-```
-To build python3 bdist wheel type
-```
-./build_python3_wheel.sh
+export NGRAPH_CPP_BUILD_PATH=$HOME/ngraph_dist/
+export LD_LIBRARY_PATH=$HOME/ngraph_dist/lib/
+export DYLD_LIBRARY_PATH=$HOME/ngraph_dist/lib (Only needed on MacOS)
```
postrational: Would be better written as:
`export DYLD_LIBRARY_PATH=$HOME/ngraph_dist/lib # (Only needed on MacOS) `
The `#` prevents a problem if someone copy-pastes this whole block.

python/README.md
```diff
@@ -1,38 +1,27 @@
...
+Set the environment variables
 
-To build python2 bdist wheel type
 ```
-./build_python2_wheel.sh
-```
-To build python3 bdist wheel type
-```
-./build_python3_wheel.sh
+export NGRAPH_CPP_BUILD_PATH=$HOME/ngraph_dist/
```
postrational: I don't think we need the trailing `/` in any of these.

python/README.md
```diff
@@ -1,38 +1,27 @@
...
+export LD_LIBRARY_PATH=$HOME/ngraph_dist/lib/
+export DYLD_LIBRARY_PATH=$HOME/ngraph_dist/lib (Only needed on MacOS)
 ```
```
postrational: You should also add `export PYBIND_HEADERS_PATH=pybind11` here. Perhaps move the entire block, so that it's after cloning of PYBIND.

python/README.md
```diff
@@ -1,38 +1,27 @@
...
-pip install -U <full path to the bdist wheel>
-```
-
-For example, On MacOS you would run a command like,
-
-```
-pip install -U build/dist/ngraph-0.0.1-cp35-cp35m-linux_x86_64.whl
+cd ngraph/python
+git clone --recursive -b allow-nonconstructible-holders https://github.com/jagerman/pybind11.git
+pip install -U .
```
postrational: I think `python setup.py install` works better (is more verbose) and faster then `pip install .`

postrational:
You will also need this for `tox` to work with non-standard pybind header locations: https://github.com/NervanaSystems/ngraph/pull/794

nbpatel:
@postrational made the changes. Do take a look when you have time.

python/README.md
```diff
@@ -42,6 +37,7 @@ pip install -r test_requirements.txt
 ```
 
 Then run a test.
+
 ```
 pytest test/test_ops.py
 pytest test/ngraph/*
```
postrational: I just noticed this, the `*` will break this command. Specify the path to directory, not a filesystem glob

python/README.md
```diff
@@ -79,3 +73,38 @@ rm -rf .tox
...
+Activate your virtual environment and install the bdist wheel
+
+```
+pip install -U <full path to the bdist wheel>
+```
+
+For example, On MacOS you would run a command like,
+
+```
+pip install -U build/dist/ngraph-0.0.1-cp35-cp35m-linux_x86_64.whl
```
postrational: The directory changed from `build/dist/` to simply `dist`

python/README.md
nbpatel: what directory? there are test files under this directory....i am giving the command for test_basic just as an example.

postrational: The "test/" directory, instead of the "test/*" filename pattern. Just removing the * would be fine.

nbpatel: Done.

PR #766, opened 2018-03-27
Convolution Bias bug fix

louisfeng:
This fixed two bugs in conv bias. I also refactored convolution's `infer_convolution_output_shape()` to be used by other convolution* classes.


src/ngraph/op/convolution.cpp
```diff
@@ -26,20 +26,20 @@
 using namespace std;
 using namespace ngraph;
 
-static Shape infer_convolution_output_shape(const Shape& data_batch_shape,
```
aprocter: I guess don't have a super-strong objection to this, but the reason this wasn't placed inside `op::Convolution` is that it's used by more classes than just `Convolution`. Any particular reason for the change?

aprocter: I do see that it can't be `static` anymore since you're using it from a separate file. Still think it would be better to have it outside the class, though.

louisfeng: I agree, fixed.

mbrookhart: It's still static in the header. Just moved to the header instead of being a local function.

Probably could (should?) be a utility in convolution.hpp instead of a static method.

Krovatkin: @aprocter @louisfeng I was wondering if we could move it into its own namespace so it doesn't leak into `ngaph::op` ? We could also consider moving it to `graph_util`?

louisfeng: Just looked at `graph_util.hpp`, it doesn't feel like belong in that file because these utils seemed more general for generic nodes. I wonder if it's more natural to place this free function under a new namespace `ngraph::op::util` and keep it in `convolution.hpp`. That way other op level utility free functions can also share that namespace but stay close to their own ops.

Krovatkin: frankly, i have a very minor gripe w/ this function being in `ngraph::op` namespace if you guys are fine with this, I'm also more than fine 😄

src/ngraph/op/convolution.cpp
```diff
@@ -26,20 +26,20 @@
...
-                                            const CoordinateDiff& padding_above,
-                                            const Strides& data_dilation_strides,
-                                            size_t batch_axis_data,
-                                            size_t input_channel_axis_data,
-                                            size_t input_channel_axis_filters,
-                                            size_t output_channel_axis_filters,
-                                            size_t batch_axis_result,
-                                            size_t output_channel_axis_result,
-                                            string error_prefix)
+Shape op::infer_convolution_output_shape(const Shape& data_batch_shape,
```
diyessi: This should be in `op::util`

src/ngraph/op/convolution.hpp
```diff
@@ -340,5 +340,20 @@ namespace ngraph
             CoordinateDiff m_padding_above_backward;
             Strides m_data_dilation_strides_backward;
         };
+
+        Shape infer_convolution_output_shape(const Shape& data_batch_shape,
```
diyessi: move into `util`

PR #775, opened 2018-03-28
[WIP]Silee2/neon cmake

silee2:
Add options for controlling gpu build and prebuilt llvm when building python wrapper.
ngraph-neon needs these options for controlling builds.
Would be great if the PR makes it to v0.2.0
(update) since v0.2.0 is for TF tagging. does not need to be part of v0.2.0
(update) Need to adjust some parts after #761


python/CMakeLists.txt
```diff
@@ -41,31 +41,29 @@ if (DEFINED NGRAPH_INSTALL_PREFIX)
...
         ngraph
         DOWNLOAD_COMMAND ""
         SOURCE_DIR ${NGRAPH_SOURCE_DIR}
         CMAKE_ARGS ${NGRAPH_CMAKE_ARGS}
         )
 endif()
+
+set(SETUP_PY_IN "${CMAKE_SOURCE_DIR}/setup.py.in")
+set(SETUP_PY    "${CMAKE_BINARY_DIR}/setup.py")
+configure_file(${SETUP_PY_IN} ${SETUP_PY})
```
postrational: I guess these lines may not be needed any more after #761

silee2:
I don't think anyone is using this Cmake file anymore since it is not the official supported way of building python wrappers.. I''m closing the PR.  May open a new PR that removes the Cmake file.

PR #777, opened 2018-03-28
CPU: Rename PREFER_EIGEN to something clearer

jmenon:


src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -106,7 +106,8 @@
 using namespace std;
 using namespace ngraph;
 
-#define PREFER_EIGEN 0
+// Enables old unoptimized Eigen code paths
+#define PREFER_OLD_INLINE_EIGEN 0
```
jbobba: "USE_EIGEN_CORE" or "USE_INLINE_EIGEN"? Dont like "OLD"

DawnStone:
TESTNOW

DawnStone:
TESTNOW

src/ngraph/runtime/cpu/cpu_emitter.cpp
jmenon: How about both? USE_EIGEN_CORE_INLINE

jbobba: ok.

PR #780, opened 2018-03-28
Change serialize routine

nbpatel:


baojun-nervana:
do we need change the api test as well?
https://github.com/NervanaSystems/ngraph/blob/master/python/ngraph/runtime.py#L96

nbpatel:
@baojun-nervana I assume so, i'll add @arogowie-intel as reviewer. The wrapper tests are failing due to some other change as well. I'll look into it and create a PR

python/pyngraph/serializer.cpp
```diff
@@ -30,8 +30,7 @@ void regclass_pyngraph_Serializer(py::module m)
                           bool binary_constant_data = false);
```
silee2: Can you update the comment as well? "bool binary_constant_data = false" is not an argument anymore.

nbpatel:
@silee2 Done.

raramer01:
python_ci stilling failing due to this compilation error:

[branch_cpu_py2] /root/ng/python/pyngraph/serializer.cpp: In function 'void regclass_pyngraph_Serializer(pybind11::module)':
[branch_cpu_py2] /root/ng/python/pyngraph/serializer.cpp:33:87: error: address of overloaded function with no contextual type information

silee2:
Looks like ci is pulling the old version of code.
/root/ng/python/pyngraph/serializer.cpp:33:87: 
There is no column 87 on line 33 in the PR.

nbpatel:
Yes, it compiles locally for me.

raramer01:
TESTNOW

raramer01:
TESTNOW

raramer01:
after modifying the test setup and more importantly fixing a Jenkins job configuration issue, this PR is almost there, the only remaining item is fixing one of the unit test that calls the serializer, see: 

http://aipg-jenkins.fm.intel.com/algo/job/testing-ngraph-python-merge/678/testReport/

nbpatel:
@raramer01 I pushed a fix for that as well. Test should pass now. @arogowie-intel please take a look.

python/ngraph/runtime.py
```diff
@@ -93,14 +93,13 @@ def __call__(self, *input_values):  # type: (*NumericData) -> NumericData
         result_arr = result_arr.reshape(result_shape)
         return result_arr
 
-    def serialize(self, indent=0, bin_const_data=False):  # type: (int, bool) -> str
+    def serialize(self, indent=0):  # type: (int, bool) -> str
```
arogowie-intel: Remove `bool` type annotation from function type signature.

postrational:
Let's merge this to unblock Python CI.

python/pyngraph/serializer.cpp
rkimballn1: binary_constant_data still exists on a private version of this function. The user can't ever set binary_constant_data to true so I removed the parameter. This parameter is used with one of the other serializer functions that stores the serialized data in a cpio file (kind of like a tar file). It was never intended to be in the public API and so it was removed.

PR #781, opened 2018-03-28
Fix LLVM commit hashes to work around recent version bump incompatibl…

aprocter:
…e with our hard-coded "5.0.1" paths


mbrookhart:
We'll need to tag v0.2.1 for the TF release?

aprocter:
> We'll need to tag v0.2.1 for the TF release?

Yup. I will take care of that.

PR #782, opened 2018-03-28
[WIP] Support for Recurring Patterns

Krovatkin:


Krovatkin:
@jbobba @pruthviIntel this is a matching part to support recurring patterns. We will need to figure out how to integrate it into `GraphRewrite` or if it even makes sense to do that.

test/pattern.cpp
```diff
@@ -530,3 +530,92 @@ TEST(pattern, variance)
...
+    ASSERT_EQ(recurrent_matches.at(2), b);
+
+    //Multiple labels in a reccuring pattern
+    auto iconst1 = construct_constant_node(1);
+    auto iconst_label = std::make_shared<pattern::op::Label>(iconst1, nullptr, NodeVector{iconst1});
+    auto add2_2 = iconst1 + add1;
+    auto add3_2 = iconst0 + add2_2;
+    auto padd2 = iconst_label + rpattern;
+    matches.clear();
+    Matcher::match_recurring_pattern(add3_2, padd2, rpattern, matches, empty_correlated_matches);
```
nirhasabnis: @Krovatkin not sure I understand the argument here.. `rpattern` is a label for recurring pattern? But then `iconst_label` is also a label for recurring pattern, right? Not sure why we need to pass label for recurring pattern explicitly? It is already part of the pattern right?

nhasabni: @Krovatkin not sure I understand the argument here.. rpattern is a label for recurring pattern? But then iconst_label is also a label for recurring pattern, right? Not sure why we need to pass label for recurring pattern explicitly? It is already part of the pattern right?

src/ngraph/pattern/matcher.cpp
```diff
@@ -38,6 +38,48 @@ namespace ngraph
...
+
+            NGRAPH_DEBUG << "matching graph to " << graph->get_name() << std::endl;
+            while (m.match(graph, previous_matches))
+            {
+                no_match = true;
+                //for the next round
+                graph = m.m_pattern_map[rpattern];
+                NGRAPH_DEBUG << "setting graph to " << graph->get_name() << std::endl;
+                for (auto me : m.m_pattern_map)
+                {
```
diyessi: `auto&` ? I'm not sure what's in a pattern map. Is `me` something like a `map entry`? That would be a better variable name then.

src/ngraph/pattern/matcher.cpp
```diff
@@ -38,6 +38,48 @@ namespace ngraph
...
+                    patterns[me.first].push_back(m.m_pattern_map[me.first]);
+                }
+
+                for (auto l : correlated_patterns)
+                {
```
diyessi: `pattern` instead of `l`.

src/ngraph/pattern/matcher.cpp
```diff
@@ -243,5 +285,31 @@ namespace ngraph
...
+            //clear our state
+            m_match_root.reset();
+            m_pattern_map.clear();
+
+            //insert previous matches
+            m_pattern_map.insert(previous_matches.cbegin(), previous_matches.cend());
+
+            if (!m_pattern_node || !graph_node)
+            {
+                throw "m_pattern_node or graph_node are not set!";
```
diyessi: Someone is asking to be frog for a month.

Krovatkin: Loooooool :-) Will fix :-)

src/ngraph/pattern/matcher.cpp
Krovatkin: yes, that's pretty much what it stands for.  I can call it *local_match*
if matched a *cell* we go over all labels and add bound nodes to a global map.

src/ngraph/pattern/matcher.cpp
Krovatkin: will fix

test/pattern.cpp
Krovatkin: @nhasabni Since a pattern graph for individual cells can include any number of `Label`s, we need to give a matcher a hint as to which label is bound to the output of a previous cell, so in the next iteration the matcher moves from the current cell to the previous one.

src/ngraph/pattern/matcher.cpp
Krovatkin: @diyessi actually in my defense, I copied this from the original `match` which was written before we started using `ngraph_error` . Fixed both.

Krovatkin:
TESTNOW

PR #783, opened 2018-03-28
Pruthvi/bn to support globalstats

pruthviIntel:
- Support for bn training Op to compute with global stats
- this is needed to pass all the MxNet test cases


src/ngraph/op/batch_norm.cpp
```diff
@@ -69,6 +69,69 @@ ngraph::op::BatchNorm::BatchNorm(double eps,
...
+        {
+            auto err_msg = std::string("The element type of ") + input_names[i] +
+                           " isn't equal to input data's type";
+            throw ngraph_error(err_msg.c_str());
+        }
+    }
+
+    for (size_t index = 0; index < get_input_size(); index++)
+    {
+        if (index != 2 && get_input_op(index)->get_shape().size() != 1)
```
cconvey: Would this method be a little easier to read through if you used symbolic names for the inputs?  For example:
``` c++
enum { kInputGamma = 0, kInputBeta, kInputData, kInputMean, kInputVariance };
...
if (index != kInputData && ...) {
...
```

src/ngraph/op/batch_norm.hpp
```diff
@@ -36,6 +36,15 @@ namespace ngraph
                       std::shared_ptr<Node> beta,
                       std::shared_ptr<Node> input);
 
+            // BatchNorm Training
```
cconvey: IMO it would be helpful to increase the header comments regarding each of these constructors.

For example, it would be helpful if each constructor's comments indicated what the output type of this op is when using that particular c'tor.

src/ngraph/op/batch_norm.hpp
```diff
@@ -36,6 +36,15 @@ namespace ngraph
...
+            BatchNorm(bool training,
```
cconvey: IMO it's a bit confusing to have a comment indicating that this is a "training" version of the c'tor, but then to also parameter whose name suggests that "training" is optional.

src/ngraph/op/batch_norm.cpp
cconvey: Heck, I'd be a fan of having these in the class's public interface.  Something like:
``` c++
class BatchnormTrainingOp ... {
   public:
      enum { kInputGamma = 0, kInputBeta, kInputData, kInputMean, kInputVariance };
      enum { kOutputData = 0, kBatchMean, kBatchVariance };
      enum { kNumInputs = kInputVariance, kNumOutputs = kBatchVariance };
   ...
};
```

src/ngraph/op/batch_norm.cpp
```diff
@@ -247,16 +325,23 @@ void ngraph::op::BatchNorm::generate_adjoints(autodiff::Adjoints& adjoints,
     //and get_n() is used to sort the inputs in the same order as Batchnorm's outputs
     //Next, Mean and Variance (`at(1)` and `at(2)`) are extracted
     //Please see `add_output` in `BatchNorm::BatchNorm` for more details
-    std::vector<std::shared_ptr<Node>> goes(get_outputs().size());
 
-    for (auto _input : get_output_inputs(0))
+    std::vector<std::shared_ptr<Node>> goes(get_outputs().size());
+    if (this->get_training_flag() && get_input_size() == 3)
+    {
+        for (auto _input : get_output_inputs(0))
```
aprocter: We can't use variable names beginning with underscores.

src/ngraph/op/batch_norm.hpp
aprocter: I agree. I think it'd be better to extend the existing constructor with a boolean parameter in the last position, whose default value is `false`. The behavior of this constructor is identical to the existing five-argument inference constructor.

src/ngraph/op/batch_norm.cpp
```diff
@@ -69,6 +69,69 @@ ngraph::op::BatchNorm::BatchNorm(double eps,
...
+                                 std::shared_ptr<ngraph::Node> beta,
+                                 std::shared_ptr<ngraph::Node> input,
+                                 std::shared_ptr<ngraph::Node> mean,
+                                 std::shared_ptr<ngraph::Node> variance)
+    : RequiresTensorViewArgs("BatchNorm", {gamma, beta, input, mean, variance})
+    , m_bn_input_shape(input->get_shape())
+    , m_bn_variance_shape(variance->get_shape())
+    , m_bn_mean_shape(mean->get_shape())
+    , m_epsilon(eps)
+    , m_training(training)
```
Krovatkin: @pruthviIntel we will need to check how this interacts with `Relu(BatchNorm)` fusion all the way down to MKLDNN : 
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/pass/cpu_fusion.cpp#L709

src/ngraph/op/batch_norm.cpp
Krovatkin: I second that. Though, I'm not sure we need a full-blown enum. A simple constant w/ a name might suffice.
```
const size_t INPUT_INDEX = 2;
```

src/ngraph/op/batch_norm.cpp
```diff
@@ -69,6 +69,69 @@ ngraph::op::BatchNorm::BatchNorm(double eps,
...
+        if (index != 2 && get_input_op(index)->get_shape()[0] != m_bn_input_shape[1])
+        {
+            auto err_msg = std::string(input_names[index]) +
+                           " shape should match the input channel size (" +
+                           std::to_string(m_bn_input_shape[1]) + ",)";
+            throw ngraph_error(err_msg.c_str());
+        }
+    }
+
+    if (variance->get_shape()[0] != mean->get_shape()[0])
```
Krovatkin: isn't this subsumed by a check at 
https://github.com/NervanaSystems/ngraph/pull/783/files#diff-0540104e61a41e6ff57c2cee5830d723R118 ?

src/ngraph/op/batch_norm.cpp
```diff
@@ -136,12 +199,25 @@ std::shared_ptr<ngraph::Node>
 {
     if (this->m_training)
     {
-        if (new_args.size() != 3)
+        if (new_args.size() == 3)
```
Krovatkin: Maybe a switch statement might be cleaner ? I dunno?

src/ngraph/op/batch_norm.cpp
```diff
@@ -247,16 +325,23 @@ void ngraph::op::BatchNorm::generate_adjoints(autodiff::Adjoints& adjoints,
     //and get_n() is used to sort the inputs in the same order as Batchnorm's outputs
     //Next, Mean and Variance (`at(1)` and `at(2)`) are extracted
     //Please see `add_output` in `BatchNorm::BatchNorm` for more details
-    std::vector<std::shared_ptr<Node>> goes(get_outputs().size());
 
-    for (auto _input : get_output_inputs(0))
+    std::vector<std::shared_ptr<Node>> goes(get_outputs().size());
+    if (this->get_training_flag() && get_input_size() == 3)
```
Krovatkin: I think we are using this pattern to order outputs in multiple spots. We should probably eventually refactor it into a helper.

src/ngraph/op/batch_norm.hpp
Krovatkin: I second the suggestion to have separate classes. It would simplify constructors, generate_adjoints and even emitter handlers. This is getting really hard to follow.

src/ngraph/op/batch_norm.cpp
pruthviIntel: agreed, will make necessary changes

pruthviIntel:
@aprocter @Krovatkin @cconvey addressed the comments.

src/ngraph/serializer.cpp
```diff
@@ -427,7 +427,7 @@ static shared_ptr<ngraph::Function>
                 else
                 {
                     node = make_shared<op::BatchNorm>(
-                        epsilon, args[0], args[1], args[2], args[3], args[4]);
```
aprocter: We'll need to update the serializer to handle the five-arg training variant as well.

src/ngraph/op/batch_norm.hpp
```diff
@@ -30,19 +30,22 @@ namespace ngraph
...
+            // bool training(true) -> BatcNorm Training using global stats
+            // bool training(false) ->  BatcNorm Inference using global stats)
             BatchNorm(double eps,
                       std::shared_ptr<ngraph::Node> gamma,
                       std::shared_ptr<ngraph::Node> beta,
                       std::shared_ptr<ngraph::Node> input,
                       std::shared_ptr<ngraph::Node> mean,
-                      std::shared_ptr<ngraph::Node> variance);
+                      std::shared_ptr<ngraph::Node> variance,
+                      bool training);
```
aprocter: Can we default this to `false` for backwards compatibility?

src/ngraph/serializer.cpp
aprocter: (On a re-read, my comment was not very clear.) What I mean is that this won't correctly deserialize the five-argument version of the op when `training` is `true`.

src/ngraph/op/batch_norm.hpp
```diff
@@ -30,19 +30,22 @@ namespace ngraph
         class BatchNorm : public util::RequiresTensorViewArgs
         {
         public:
-            // BatchNorm Training
+            // BatchNorm Training with scale shift. mean and variance
+            // are computed internal to MKLDNN
```
jbobba: Update comment to say 'batch mean and variance are computed in the kernel'

pruthviIntel:
@jbobba , can you take a look?

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -578,6 +578,7 @@ size_t MKLDNNEmitter::build_batchnorm_forward(const mkldnn::memory::desc& input_
                                               const mkldnn::memory::desc& mean_desc,
                                               const mkldnn::memory::desc& variance_desc,
                                               const double eps,
+                                              size_t bn_total_inputs,
```
jbobba: instead of bn_total_inputs, make this into something more meaningful like 'use_computed_stats'

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -609,6 +610,25 @@ size_t MKLDNNEmitter::build_batchnorm_forward(const mkldnn::memory::desc& input_
             input_index, weights_index, result_index, mean_index, variance_index};
         return batchnorm_index;
     }
+    else if (bn_training_flag && bn_total_inputs == 5)
+    {
+        size_t batchnorm_index = insert_primitive(new mkldnn::batch_normalization_forward(
+            {{mkldnn::prop_kind::forward_training,
```
jbobba: Might be more performant to use "forward_inference"

src/ngraph/op/batch_norm.hpp
```diff
@@ -30,19 +30,22 @@ namespace ngraph
...
             BatchNorm(double eps,
                       std::shared_ptr<Node> gamma,
                       std::shared_ptr<Node> beta,
                       std::shared_ptr<Node> input);
 
-            //BatchNorm Inference
+            // bool training(true) -> BatcNorm Training using global stats
```
jbobba: spelling.

src/ngraph/op/batch_norm.hpp
```diff
@@ -30,19 +30,22 @@ namespace ngraph
...
+            // bool training(false) ->  BatcNorm Inference using global stats)
```
jbobba: spelling 'BatcNorm' -> 'BatchNorm'

DawnStone:
TESTNOW

PR #790, opened 2018-03-30
Add GPURuntimeContext to GPU transformer

csullivan:
This PR adds a GPURuntimeContext object which mirrors that in the IA transformer. It is composed of various pointers to memory used during graph execution as well as the CudaFunctionPool class.

* GPU_ExternalFunction has a unique GPURuntimeContext (ctx).
* The context object ctx is accessible from the gpu_emitters(.hh) and so the CudaFunctionPool can be constructed prior to graph compilation. In this PR they are still constructed at graph execution.
* GPU_CallFrame passes (ctx) through to the compiled function for access at graph execution.


src/ngraph/runtime/gpu/gpu_call_frame.cpp
```diff
@@ -32,9 +32,6 @@ runtime::gpu::GPU_CallFrame::GPU_CallFrame(std::shared_ptr<GPU_ExternalFunction>
     : m_external_function(external_function)
     , m_compiled_function(compiled_function)
 {
-    //Create context use driver API and make it current, the runtime call will pickup the context
-    //http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#interoperability-between-runtime-and-driver-apis
-    ngraph::runtime::gpu::CudaContextManager::instance();
     cublasStatus_t cublasStatus = cublasCreate(&m_cublas_handle);
```
fengleitian: we can also move `cublasCreate` and `cudnnCreate` to `gpu_runtime_context.hpp`, so we don't need the `link_runtime_context`;

src/ngraph/runtime/gpu/gpu_cuda_function_pool.cpp
```diff
@@ -28,21 +28,18 @@ static const std::string s_output_dir = "gpu_codegen";
...
     const char* opts[] = {"--gpu-architecture=compute_35", "--relocatable-device-code=true"};
     std::string filename =
         file_util::path_join(s_output_dir, "cuda_kernel_" + name + "_codegen.cu");
     std::ofstream out(filename);
     out << kernel;
     out.close();
-    m_function_map.insert({name, CudaFunctionBuilder::get("cuda_" + name, kernel, 2, opts)});
+    auto cu_compiled_function = CudaFunctionBuilder::get("cuda_" + name, kernel, 2, opts);
+    m_function_map.insert({name, cu_compiled_function});
+    return cu_compiled_function;
```
fengleitian: nice.

src/ngraph/runtime/gpu/gpu_cuda_kernel_emitters.cpp
```diff
@@ -24,27 +24,29 @@ using namespace ngraph;
...
-                                  const std::array<std::string, 2>& data_types,
                                   size_t repeat_size,
                                   size_t repeat_times,
                                   size_t count)
 {
     std::string name_signature = name + "_" + data_types[0] + "_" + data_types[1];
     std::replace(name_signature.begin(), name_signature.end(), ' ', '_');
     // Create an instance of nvrtcProgram with the code string.
-    if (CudaFunctionPool::instance().get(name_signature) == nullptr)
+    auto compiled_kernel = ctx->nvrtc_cache->get(name_signature);
```
fengleitian: maybe call it `compiled_kernel_pool` instead of `nvrtc_cache`? easier to understand, not everyone know what nvrtc is.

src/ngraph/runtime/gpu/gpu_cuda_kernel_emitters.cpp
```diff
@@ -94,28 +98,29 @@ void runtime::gpu::emit_onehot(const std::string& name,
...
         CudaKernelBuilder::add_pod_typedefs(writer);
         CudaKernelBuilder::get_reshape_op(writer, name_signature, data_types);
         std::string kernel = writer.get_code();
-        CudaFunctionPool::instance().set(name_signature, kernel);
+        compiled_kernel = ctx->nvrtc_cache->set(name_signature, kernel);
     }
 
     void* args_list[] = {&in, &out, &input_strides, &trans_strides, &rank, &count};
-    CUDA_SAFE_CALL(cuLaunchKernel(*CudaFunctionPool::instance().get(name_signature).get(),
+    CUDA_SAFE_CALL(cuLaunchKernel(*compiled_kernel.get(),
```
fengleitian: nice.

src/ngraph/runtime/gpu/gpu_cuda_kernel_emitters.hpp
```diff
@@ -88,25 +83,34 @@ namespace ngraph
...
+                                              1,
+                                              1, // block dim
+                                              0,
+                                              NULL, // shared mem and stream
+                                              args_list,
+                                              0));  // arguments
                 CUDA_SAFE_CALL(cuCtxSynchronize()); // Retrieve and print output.
             }
+
+            void emit_reshape(const std::string& name,
```
fengleitian: since we are changing the order anyway, does it make sense to put it in alphabet order?

src/ngraph/runtime/gpu/gpu_call_frame.cpp
```diff
@@ -77,7 +76,7 @@ void runtime::gpu::GPU_CallFrame::tensor_call(
         outputs.push_back(tv->m_allocated_buffer_pool);
     }
 
-    m_compiled_function(inputs.data(), outputs.data(), m_cublas_handle, m_cudnn_handle);
+    m_compiled_function(inputs.data(), outputs.data(), m_external_function->m_ctx.get());
```
fengleitian: what if we also pass here as a reference to unique_ptr

src/ngraph/runtime/gpu/gpu_runtime_context.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include "ngraph/runtime/gpu/gpu_cuda_function_pool.hpp"
+#include "ngraph/runtime/gpu/gpu_util.hpp"
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace gpu
+        {
+            extern "C" {
```
rkimballn1: extern "C" here seems wrong on many levels. namespaces are C++ and the ctor/dtor in the struct are C++. I don't understand this.

csullivan: @rkimballn1, this was a mistake in overall design of the GPURuntimeContext, which I have corrected in the most recent commit. 

For the GPURuntimeContext struct it is quite important that it be extern "C" so that we are only using the C API when operating on ngraph data in the JIT compiled function. This point was raised to me by @jmenon and @jbobba, where the local clang implementation may target a different GLIBC ABI than the native compiler that was used to build nGraph. In these cases, calling nGraph C++ from within the clang could cause runtime errors. By forcing the interface to be C-only, this problem is avoided.

src/ngraph/runtime/gpu/gpu_call_frame.cpp
csullivan: See my comment below regarding interoperability between the clang and native compiled code. For this reason we should use only raw pointers.

csullivan:
@fengleitian 

> we can also move cublasCreate and cudnnCreate to gpu_runtime_context.hpp, so we don't need the link_runtime_context;

See my comments above about using the C-only api. To be safer, I think it probably better to leave them in the ngraph side and to pass them via a pointer. If the interface for cudnn and cublas is C then we could do as you suggest safely. For now I have left them to be setup in the gpu call frame.

src/ngraph/runtime/gpu/gpu_call_frame.cpp
fengleitian: OK. I see.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -241,7 +242,18 @@ runtime::gpu::GPU_ExternalFunction::GPU_ExternalFunction(
     : ngraph::runtime::ExternalFunction(function, release_function)
     , m_compiled_function(nullptr)
     , m_emit_timing(std::getenv("NGRAPH_GPU_EMIT_TIMING") != nullptr)
+    , m_ctx(new GPURuntimeContext)
 {
+    // Create context use driver API and make it current, the runtime call will pickup the context
```
fengleitian: why not put in `setup_runtime_context`? I'm think it's better to put cuda/cublas/cudnn initializations together?

csullivan: The problem is that the external_function::compile() is called prior to the consturction of the call_frame, and the gpu_emitters require the GPURuntimeContext. I'm open to suggestions on a cleaner place for this to be instantiated. We could possibly move the setup_runtime_context to GPUExternalFunction.

csullivan:
Closing in favor of superseding PR #837 as it is easier to review the changes along with those from the other branch. Your comments from this review have been worked into that PR.

PR #792, opened 2018-03-30
Add CUDNNEmitter for cudnn emission via closures (e.g. op::Sum rework)

csullivan:
**Notice:** #790 should be reviewed/merged first as this branch is based off of it - for now this PR is pointing to cs/gpu_runtime_context for easier review. It should be switched to point to master prior to merging.
-
This PR adds CUDNNEmitter, an analog of the MKLDNNEmitter in the IA transformer. It is a thin wrapper on cudnn function calls that utilizes lambda closures to capture cudnn operations that can then be passed and executed by the compiled graph. 

The GPU emitter for op::Sum was reworked to use the CUDNNEmitter as an example. In principle this functionality does not affect performance, but is a quality of life improvement for two reasons:
* cudnn code can be written directly without need for wrapping it as strings that are emitted.
* compiling of cudnn now occurs at compile time instead of just-in-time. This makes catching compiler errors easier.
* cudnn code becomes easier to debug as gdb has some trouble with the JIT code.

Note, these changes are non-API breaking and so the old method of emitting cudnn strings can persist until there is time to refactor or if there is an advantage to do so.


csullivan:
TESTNOW

csullivan:
GitHub appears to be showing some overlap of the PRs, however the core parts of this PR is different from #790. This PR (#792) principally adds the CUDNNEmitter. #790 adds the GPURuntimeContext object which is used during emission and passed into the graph execution runtime. I'll see if I can fix the diff.

csullivan:
@rkimballn1 I fixed the diff so that the changes shown are now distinct from #790. Please review, thank you.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -0,0 +1,177 @@
...
+                          const std::vector<void*>& result)
+{
+    m_cudnn_primitives[primitive_index](args, result);
+}
+
+size_t CUDNNEmitter::register_primitive(
+    const std::function<void(std::vector<void*>, std::vector<void*>)>& f)
+{
+    // try emplace
+    m_cudnn_primitives.push_back(f);
```
fengleitian: Is It possible two reduce share the same function?

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -781,91 +776,24 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
                             kernel::emit_memset(writer, out[0], 0);
                         }
                         // no change in dimensions, reduction not necessary
-                        else if (input_shape.size() == out[0].get_shape().size())
+                        else if (args[0].get_shape().size() == out[0].get_shape().size())
                         {
                             kernel::emit_memcpyDtD(writer, out[0], args[0]);
                         }
                         // descriptors for tensors  with <= 4 dimensions
```
fengleitian: comments should change any dimension or just remove.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: I think it's fine for now. but we need to revisit it in the future for optimization. We can do something similar to pool. If we can find in pool, we don't need to add another one. save some space and overhead.

csullivan:
Closing in favor of superseding PR #837 as it is easier to review the changes along with those from the other branch. Your comments from this review have been worked into that PR.

PR #795, opened 2018-03-30
enable TensorView to use pre-allocated mem

ashokei:
Enable users to pass aligned pre-allocated mem, for use by TensorView.

### DIFF
--- a/src/ngraph/runtime/backend.hpp
+++ b/src/ngraph/runtime/backend.hpp
@@ -52,6 +52,12 @@ namespace ngraph
                 make_primary_tensor_view(const ngraph::element::Type& element_type,
                                          const Shape& shape) = 0;
 
+            /// @brief Return a handle for a tensor for given mem on backend device
+            virtual std::shared_ptr<ngraph::runtime::TensorView>
+                make_primary_tensor_view(const ngraph::element::Type& element_type,
+                                         const Shape& shape,
+                                         void* mem_handle) = 0;
+
             template <typename T>
             std::shared_ptr<ngraph::runtime::TensorView>
                 make_primary_tensor_view(const Shape& shape)
--- a/src/ngraph/runtime/backend.hpp
+++ b/src/ngraph/runtime/backend.hpp
@@ -56,7 +56,7 @@ namespace ngraph
             virtual std::shared_ptr<ngraph::runtime::TensorView>
                 make_primary_tensor_view(const ngraph::element::Type& element_type,
                                          const Shape& shape,
-                                         void* mem_handle) = 0;
+                                         void* memory_pointer) = 0;
 
             template <typename T>
             std::shared_ptr<ngraph::runtime::TensorView>
--- a/src/ngraph/runtime/cpu/cpu_tensor_view.cpp
+++ b/src/ngraph/runtime/cpu/cpu_tensor_view.cpp
@@ -51,7 +51,7 @@ runtime::cpu::CPUTensorView::CPUTensorView(const ngraph::element::Type& element_
 
     buffer_size = shape_size(shape) * element_type.size();
 
-    if (mem_handle)
+    if (mem_handle != nullptr)
     {
         aligned_buffer = static_cast<char*>(mem_handle);
         return;
--- a/src/ngraph/runtime/cpu/cpu_tensor_view.cpp
+++ b/src/ngraph/runtime/cpu/cpu_tensor_view.cpp
@@ -36,6 +36,7 @@ const size_t runtime::cpu::CPUTensorView::BufferAlignment = 64;
 
 runtime::cpu::CPUTensorView::CPUTensorView(const ngraph::element::Type& element_type,
                                            const Shape& shape,
+                                           void* mem_handle,
                                            const string& name)
     : runtime::TensorView(std::make_shared<ngraph::descriptor::PrimaryTensorView>(
           std::make_shared<ngraph::TensorViewType>(element_type, shape), name, true, true, false))
@@ -49,6 +50,13 @@ runtime::cpu::CPUTensorView::CPUTensorView(const ngraph::element::Type& element_
         *m_descriptor, runtime::cpu::LayoutDescriptor::create_native_axis_order(shape.size())));
 
     buffer_size = shape_size(shape) * element_type.size();
+
+    if (mem_handle)
+    {
+        aligned_buffer = static_cast<char*>(mem_handle);
+        return;
+    }
+
     if (buffer_size)
     {
         size_t allocation_size = buffer_size + BufferAlignment;
@@ -72,9 +80,17 @@ runtime::cpu::CPUTensorView::CPUTensorView(const ngraph::element::Type& element_
     }
 }
 
+runtime::cpu::CPUTensorView::CPUTensorView(const ngraph::element::Type& element_type,
+                                           const Shape& shape,
+                                           const string& name)
+    : CPUTensorView(element_type, shape, nullptr, name)
+{
+}
+
 runtime::cpu::CPUTensorView::~CPUTensorView()
 {
-    free(buffer);
+    if (buffer != nullptr)
+        free(buffer);
 }
 
 char* runtime::cpu::CPUTensorView::get_data_ptr()
--- a/src/ngraph/runtime/cpu/cpu_tensor_view.cpp
+++ b/src/ngraph/runtime/cpu/cpu_tensor_view.cpp
@@ -36,7 +36,7 @@ const size_t runtime::cpu::CPUTensorView::BufferAlignment = 64;
 
 runtime::cpu::CPUTensorView::CPUTensorView(const ngraph::element::Type& element_type,
                                            const Shape& shape,
-                                           void* mem_handle,
+                                           void* memory_pointer,
                                            const string& name)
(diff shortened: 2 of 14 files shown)


src/ngraph/runtime/cpu/cpu_tensor_view.cpp
```diff
@@ -72,9 +80,17 @@ runtime::cpu::CPUTensorView::CPUTensorView(const ngraph::element::Type& element_
...
+                                           const string& name)
+    : CPUTensorView(element_type, shape, nullptr, name)
+{
+}
+
 runtime::cpu::CPUTensorView::~CPUTensorView()
 {
-    free(buffer);
+    if (buffer != nullptr)
+        free(buffer);
```
rkimballn1: all code blocks require curly braces, even single lines

src/ngraph/runtime/backend.hpp
```diff
@@ -52,6 +52,12 @@ namespace ngraph
                 make_primary_tensor_view(const ngraph::element::Type& element_type,
                                          const Shape& shape) = 0;
 
+            /// @brief Return a handle for a tensor for given mem on backend device
+            virtual std::shared_ptr<ngraph::runtime::TensorView>
+                make_primary_tensor_view(const ngraph::element::Type& element_type,
+                                         const Shape& shape,
+                                         void* mem_handle) = 0;
```
rkimballn1: use full words. rename to memory_handle, or better yet, memory_pointer as it is a pointer

src/ngraph/runtime/cpu/cpu_tensor_view.hpp
```diff
@@ -33,6 +33,10 @@ namespace ngraph
                 CPUTensorView(const ngraph::element::Type& element_type,
                               const Shape& shape,
                               const std::string& name = "external");
+                CPUTensorView(const ngraph::element::Type& element_type,
+                              const Shape& shape,
+                              void* mem_handle,
```
rkimballn1: This needs a default value, memory_handle = nullptr

src/ngraph/runtime/host_tensor_view.cpp
```diff
@@ -37,6 +38,13 @@ runtime::HostTensorView::HostTensorView(const ngraph::element::Type& element_typ
         std::make_shared<ngraph::descriptor::layout::DenseTensorViewLayout>(*m_descriptor));
 
     m_buffer_size = m_descriptor->get_tensor_view_layout()->get_size() * element_type.size();
+
+    if (mem_handle != nullptr)
+    {
+        m_aligned_buffer_pool = static_cast<char*>(mem_handle);
+        return;
```
rkimballn1: You could use an else statement rather than return here. You don't need to change anything, just putting up as an alternative.

src/ngraph/runtime/cpu/cpu_tensor_view.cpp
```diff
@@ -36,6 +36,7 @@ const size_t runtime::cpu::CPUTensorView::BufferAlignment = 64;
 
 runtime::cpu::CPUTensorView::CPUTensorView(const ngraph::element::Type& element_type,
                                            const Shape& shape,
+                                           void* mem_handle,
```
rkimballn1: would it be better to add new parameters to the end of the parameter list? This parameter should have a default value so all existing calls are unchanged. Only calls making use of memory handle would need the extra argument.

ashokei: @rkimballn1 existing calls continue to work, we just added new api. I was thinking of modifying existing api with default param, but default parameters can add additional api user awareness, and also in this case we want to explicitly force users to specify a memory_pointer (no defaults used). existing code continued to work. i am ok changing, let me know will modify as needed.

src/ngraph/runtime/backend.hpp
```diff
@@ -52,6 +52,12 @@ namespace ngraph
                 make_primary_tensor_view(const ngraph::element::Type& element_type,
                                          const Shape& shape) = 0;
 
+            /// @brief Return a handle for a tensor for given mem on backend device
```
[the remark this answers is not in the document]
ashokei: do we want to allow users to use preallocated cuda mem for GPU, and similar for NNP.  For this PR, i can make it specific to cpu.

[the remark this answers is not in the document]
rkimballn1: I added @fengleitian and @yxlao to the PR to get their feedback on GPU and NNP backends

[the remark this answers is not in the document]
yxlao: Yeah, this wouldn't generalize to NNP. If we were to implement similar thing, it would be a `ArTensorId` instead of `void*`.

Another way to use it would be: instead of making `void* memory_pointer` as "use a pre-allocated NNP buffer", we treat it as "create a NNP tensor and copy the data from host `memory_pointer` to NNP". But we already have another APIs to do that.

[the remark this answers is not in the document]
fengleitian: CUDA side should be able to use `void*`, but need to allocate with `cudaMalloc` and free with `cudaFree`.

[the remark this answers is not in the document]
ashokei: this PR is to enable that use-case in general. 

So we have 3 backends that can use this, and 4th one (NNP) cannot use this directly. For consistency, can we still "create a NNP tensor and copy the data from host memory_pointer to NNP" .  Or just proxy it to existing NNP APIs that do it.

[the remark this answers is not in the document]
ashokei: latest commit has GPU added, this allows user to specify cuda mem that they may manage themselves.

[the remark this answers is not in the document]
fengleitian: I mean it will work on GPU, but I have some dumb questions:
1. Is the tensorviews itself unique, why we need memory to uniquely tag tensorviews? 
2. if user need to create memory, they first need to get information from tensorviews? like size of the memory? why not just get a pointer to the data?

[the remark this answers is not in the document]
ashokei: 1. yes, tensorview itself is expected to be unique. This PR makes this more transparent.
2. some frameworks already have their own memory management, this API allows to just reuse existing pre-allocated mem as needed. Also, shape/type (size) is already known before creating tensorview.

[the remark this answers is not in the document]
fengleitian: OK, that make sense. Thanks @ashokei

src/ngraph/runtime/cpu/cpu_tensor_view.cpp
```diff
@@ -72,9 +78,19 @@ runtime::cpu::CPUTensorView::CPUTensorView(const ngraph::element::Type& element_
...
+                                           const Shape& shape,
+                                           const string& name)
+    : CPUTensorView(element_type, shape, nullptr, name)
+{
+}
+
 runtime::cpu::CPUTensorView::~CPUTensorView()
 {
-    free(buffer);
+    if (buffer != nullptr)
```
jmenon: This check should be removed

src/ngraph/runtime/cpu/cpu_tensor_view.hpp
```diff
@@ -33,7 +33,13 @@ namespace ngraph
...
                               const Shape& shape,
                               const std::string& name = "external");
-                virtual ~CPUTensorView() override;
+
+                CPUTensorView(const ngraph::element::Type& element_type,
+                              const Shape& shape,
+                              void* memory_pointer,
+                              const std::string& name = "external");
+                
+                virtual ~CPUTensorView();
```
jmenon: Was the override specifier removal intentional?

src/ngraph/runtime/cpu/cpu_tensor_view.cpp
ashokei: will do; just curious, is it because it is unnecessary?  this check is used in interpreter before this PR, and it was missing here.

src/ngraph/runtime/cpu/cpu_tensor_view.hpp
ashokei: it is a side-effect of merging, will revert back.

src/ngraph/runtime/cpu/cpu_tensor_view.cpp
jmenon: It wasn't here because it's unnecessary, yes

src/ngraph/runtime/cpu/cpu_tensor_view.hpp
jmenon: Thanks. Also, we should probably add a note somewhere in the bridge that alignments are tied to what the nGraph backend (ex. IA) assumes. Currently it's the maximal vector word length but that's not really ideal for all tensors. I'm not sure we'll be changing that assumption anytime soon though.

ashokei: yes, it is the responsibility of API users to use properly aligned mem for optimal performance. We already do this for DO in mxnet (64 byte aligned), we will add comments as needed.

rkimballn1:
This can be made to work on NNP so it should be generalized to everything

PR #796, opened 2018-03-30
Support multi-output ops in Adjoints

Krovatkin:


src/ngraph/autodiff/adjoints.cpp
```diff
@@ -50,8 +50,37 @@ std::shared_ptr<Node> make_zero(const element::Type& element_type, const Shape&
...
+    if (ys.size() != cs.size())
+    {
+        throw ngraph_error("ys and cs must be equal size");
+    }
+
+    for (size_t i = 0; i < ys.size(); i++)
+    {
+        if (ys.at(i)->get_outputs().size() > 1 || cs.at(i)->get_outputs().size() > 1)
+        {
+            throw ngraph_error("Adjoins for multi-output ops aren't supported directly");
```
diyessi: Adjoints

diyessi: Maybe say to use the goe instead.

src/ngraph/autodiff/adjoints.cpp
```diff
@@ -106,48 +140,60 @@ autodiff::Adjoints::Adjoints(const std::shared_ptr<Node>& y, const std::shared_p
                 nodes_to_check.push_front(arg);
             }
         }
-        node->generate_adjoints(*this, get(node));
+        auto deltas = get(node);
+        if (node->get_outputs().size() > 1)
+        {
+            node->generate_adjoints(*this, deltas);
+        }
+        else
```
diyessi: Better make sure there is at least one output before doing this. We don't have any 0-output nodes, but who knows what might happen in the future.

src/ngraph/autodiff/adjoints.cpp
```diff
@@ -50,8 +50,39 @@ std::shared_ptr<Node> make_zero(const element::Type& element_type, const Shape&
...
+    {
+        throw ngraph_error("ys and cs must be equal size");
+    }
+
+    for (size_t i = 0; i < ys.size(); i++)
+    {
+        if (ys.at(i)->get_outputs().size() > 1 || cs.at(i)->get_outputs().size() > 1)
+        {
+            throw ngraph_error(
+                "Adjoitns for multi-output ops aren't supported directly.\nProvide deltas for "
```
mbrookhart: `Adjoitns` typo.

src/ngraph/node.hpp
```diff
@@ -90,11 +90,8 @@ namespace ngraph
                 input.get_output().remove_input(&input);
             }
         }
-        virtual void generate_adjoints(autodiff::Adjoints& adjoints,
-                                       const std::shared_ptr<Node>& delta)
-        {
-        }
 
+        virtual void generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas) {}
```
mbrookhart: It seems you only ever use the first value in these vectors?

Krovatkin: so far, yes. Even `BatchNorm` only uses `deltas.at(0)` since the first delta corresponds to `dinput`. It could have easily been any `deltas.at(1)` or `deltas.at(2)` depending an order of the outputs of `BatchNorm`. We can now handle all these scenarios.

https://github.com/NervanaSystems/ngraph/pull/796/files#diff-0540104e61a41e6ff57c2cee5830d723L260

Also, note the implementation of `GetOutputElement::generate_adjoints`

https://github.com/NervanaSystems/ngraph/pull/796/files#diff-0065dc21e94515d9e5b9dd2f63a1021dR54

It pushes its update to the corresponding output.

src/ngraph/autodiff/adjoints.cpp
Krovatkin: Wlil fxi!

Krovatkin:
@mbrookhart They don't use our autodiff do they?

mbrookhart:
Good point

src/ngraph/op/add.cpp
```diff
@@ -33,8 +33,10 @@ shared_ptr<Node> op::Add::copy_with_new_args(const NodeVector& new_args) const
     return make_shared<Add>(new_args.at(0), new_args.at(1));
 }
```
diyessi: What if there were an overloading of `genearate_adjoints` that took a `Node` as its argument, and the vector'd `generate_adjoints` default implementation called it with the `deltas.at(0)`. Then you wouldn't need to change all these.

PR #797, opened 2018-03-30
simplify contrib/docker Makefile and make targets

DawnStone:


contrib/docker/Makefile
```diff
@@ -25,7 +25,7 @@
 # Override with make PARALLEL="-j <num_parallel_processes>" where
 #    <num_parallel_processes> = the number of make processes to run in parallel
 # Turn off with make PARALLEL=
-PARALLEL=-j 22 
+PARALLEL=22
```
crlishka: One negative I see with changing PARALLEL to be only the number of threads is that you can never run _make_ without the -j option.  Also, you can not easily run _make_ with the default number of threads for -j, since the build-and-test script will set PARALLEL=22 if the PARALLEL option is empty (but set).

contrib/docker/Makefile
```diff
@@ -25,7 +25,7 @@
 # Override with make PARALLEL="-j <num_parallel_processes>" where
 #    <num_parallel_processes> = the number of make processes to run in parallel
 # Turn off with make PARALLEL=
```
crlishka: As you have PARALLEL implemented now, this comment is incorrect.  If you specify "PARALLEL=" with no number, then you will either get "-j 22" (because PARALLEL is empty and the build-and-test script will set PARALLEL to the default 22) or you will get "-j" (if the build-and-test script does not detect PARALLEL as being empty ... which I don't think will happen, but bash variable expansion sometimes fools me).  So the comment is wrong.  To use single-threaded parallel (i.e. "-j 1"), you need to use "PARALLEL=1".

contrib/docker/Makefile
DawnStone: Agreed.  I restored the original setting methodology.

DawnStone:
@crlishka: FYI - I also found that I needed to add the creation of the ngraph_dist_gcc.tgz tarball to maintain compatibility with the current `contrib/docker/Makefile` behavior for the install_gcc target.

contrib/docker/Makefile
crlishka: This change looks good

contrib/docker/build-ngraph-and-test.sh
```diff
@@ -120,23 +118,35 @@ echo "    NGRAPH_REPO=${NGRAPH_REPO}"
...
-    env VERBOSE=1 make ${CMD_TO_RUN} 2>&1 | tee ${OUTPUT_DIR}/make_${CMD_TO_RUN}_${TEST_SUITE}.log
+    echo "Running make ${MAKE_CMD_TO_RUN}"
+    env VERBOSE=1 make ${MAKE_CMD_TO_RUN} 2>&1 | tee ${OUTPUT_DIR}/make_${CMD_TO_RUN}.log
+
+    # create the ngraph_dist_gcc.tgz file to align with existing make install behavior
+    if [ "${MAKE_CMD_TO_RUN}" == "install" ] ; then
+        echo "Building ngraph_dist_${COMPILER}.tgz"
+        tar czf ngraph_dist_${COMPILER}.tgz ngraph_dist 2>&1 | tee make_tarball_${COMPILER}.log
+        ls -l ngraph_dist_*.tgz
+    fi
```
crlishka: The install_* make targets are my own creations, so that I could provide a tarball which could be used in automation. 

If a final decision is made on packaging, we can modify the results of install_* targets to match the packaging requirements and modify the automation to use the new packaging.

DawnStone:
Fixed hidden dependencies after testing with CI integration scripts.

DawnStone:
TESTNOW

DawnStone:
TESTNOW

PR #799, opened 2018-03-31
add random init for all input types, not just float

rkimballn1:


ashokei:
LGTM, as well. thanks!

PR #800, opened 2018-04-01
Cache the order ops during codegen for a nice compile speedup

rkimballn1:
With this change the compile time for mxnet/LSTM_backward.json went from 9.6 to 8.7 seconds.
The stopwatch changes we done while I was hunting down the issue and I liked them, so I kept them.


DawnStone:
TESTNOW

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -518,14 +524,15 @@ using namespace ngraph::runtime;
 
     for (shared_ptr<Function> current_function : pass_manager.get_state().get_functions())
     {
+        auto ordered_ops = function_ordered_ops.at(current_function);
```
Krovatkin: does this need to be a copy?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -459,7 +465,7 @@ using namespace ngraph::runtime;
     unordered_map<Node*, string> match_functions;
     for (shared_ptr<Function> current_function : pass_manager.get_state().get_functions())
     {
-        const list<shared_ptr<Node>>& tmp = current_function->get_ordered_ops();
+        list<shared_ptr<Node>> tmp = function_ordered_ops.at(current_function);
```
Krovatkin: does this need to be a copy?

Krovatkin:
@rkimballn1 we should reconsider (in future) reviving caching in `Function` in some form or another. It might look simpler

aprocter:
TESTNOW

PR #806, opened 2018-04-03
Enable Travis CI

postrational:
This is a proposal to enable Travis for public-facing CI of ngraph.

### DIFF
--- a/.ci/travis/ubuntu/Dockerfile
+++ b/.ci/travis/ubuntu/Dockerfile
@@ -1,0 +1,45 @@
+FROM ubuntu:16.04
+
+# nGraph dependencies
+RUN apt-get update && apt-get install -y \
+  build-essential \
+  cmake \
+  clang-3.9 \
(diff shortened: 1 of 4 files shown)


jennifermyers:
LGTM

diyessi:
Are we duplicating docker setups we have other places?

postrational:
The current setup actually runs tests on both Python 2 and 3.

Python 2: https://travis-ci.org/NervanaSystems/ngraph/jobs/362110086#L6406-L6423
Python 3: https://travis-ci.org/NervanaSystems/ngraph/jobs/362110086#L7413-L7430

postrational:
I did notice a different potential problem however.
Our build and testing takes about 41 minutes. 
This is dangerously close to the 50 minute timeout set for a job. 

We may need to split the build into multiple jobs to avoid timeouts.

postrational:
Added a way to run tests in parallel, this should avoid the 50 minute timeout.

PR #814, opened 2018-04-04
add GPU backend support for contrib/docker make process

DawnStone:
This PR adds dockerfiles and support for the GPU backend in the `contrib/docker/Makefile`.

Merge after PR #797.


DawnStone:
TODOs:
-Check for GPU support on the host system and fall back to CPU if not present
-Update README.md

contrib/docker/Makefile
```diff
@@ -65,13 +76,30 @@ endif
 
 ifeq ("$(shell echo ${OS} | grep centos)","centos74")
     RUN_AS_USER_SCRIPT=${DOCKUSER_HOME}/ngraph-test/contrib/docker/run_as_centos_user.sh
-    DOCKERFILE=Dockerfile.ngraph.centos74
+    CPU_DOCKERFILE=Dockerfile.ngraph.centos74
```
crlishka: This looks inconsistent with the "else" case below.  In the "else" case, you use conditional assigment (?=), but here you are using absolute assignment.  Should both the then and else cases use conditional, or both use absolute?

crlishka: I am also a little confused as to why the Dockerfile name does not have the "_gpu" at the end, since you have created _gpu Dockerfiles for this commit?

raramer01: crlishka: possibly because the generically named Dockerfile.ngraph.centos74 is used for cpu and is assigned to CPU_Dockerfile, while if the Nvidia-smi command is available, then the GPU_Dockerfile variable is set to the 'generically' named docker file with the appended "_gpu"...see line 89 below...

crlishka: @raramer01 Yes you are correct.  Looking further, I see that GPU_DOCKERFILE is derived from CPU_DOCKERFILE a little later.  That clears my question up.

PR #817, opened 2018-04-04
Add op::MaxPool and op::MaxPoolBackprop to GPU transformer

csullivan:


src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -27,3 +31,148 @@ CUDAEmitter::CUDAEmitter(GPUPrimitiveEmitter* emitter)
...
+                              const ngraph::Shape& output_shape,
+                              const ngraph::Shape& padding_below,
+                              const ngraph::Shape& padding_above,
+                              const ngraph::Shape& padding_interior)
+{
+    if (padding_interior.size())
+    {
+        throw std::runtime_error("Interior padding is not yet supported in the GPU transformer.");
+    }
+    std::string hash = "pad_i" + join(input_shape) + "_pb" + join(padding_below) + "_pa" +
```
rkimballn1: join has an optional second parameter which is the separator. You could use calls like `join(input_shape, "_")` and then skip the replace stuff below.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -13,12 +13,16 @@
...
 #include <vector>
 
 #include "ngraph/runtime/gpu/cudnn_emitter.hpp"
 #include "ngraph/runtime/gpu/gpu_primitive_emitter.hpp"
 #include "ngraph/runtime/gpu/gpu_runtime_context.hpp"
+#include "ngraph/runtime/gpu/gpu_util.hpp"
+#include "ngraph/util.hpp"
 
 using namespace ngraph;
 using namespace ngraph::runtime::gpu;
```
rkimballn1: I know it is not part of your code, but we should not use `using namespace` beyond the first namespace, so `using namespace ngraph;` is ok but `using namespace ngraph::runtime::gpu;` is not

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -28,10 +32,10 @@ CUDNNEmitter::CUDNNEmitter(GPUPrimitiveEmitter* emitter)
 {
 }
 
-size_t CUDNNEmitter::build_reduce_forward(GPURuntimeContext* ctx,
-                                          const Shape& input_shape,
-                                          const AxisSet& reduction_axes,
-                                          const cudnnReduceTensorOp_t& reduce_op)
+size_t CUDNNEmitter::build_reduce_forward(const GPURuntimeContext* ctx,
+                                          const cudnnReduceTensorOp_t& reduce_op,
+                                          const ngraph::Shape& input_shape,
```
rkimballn1: If we are using namespace ngraph then we should not need `ngraph::`. You might need it because of the using ngraph::runtime::gpu if there is a Shape in that namespace, but just remove the multi level usings

src/ngraph/runtime/gpu/gpu_cuda_kernel_emitters.cpp
```diff
@@ -133,3 +133,38 @@ void runtime::gpu::emit_reshape(const std::string& name,
                                   0));  // arguments
     CUDA_SAFE_CALL(cuCtxSynchronize()); // Retrieve and print output.
 }
+
+// void runtime::gpu::emit_1d_max_pool(GPURuntimeContext* ctx,
```
rkimballn1: Why is this comment out code here?

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -800,6 +800,245 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
                 writer.block_end();
                 return;
             }
+
+            template <>
+            void GPU_Emitter::EMITTER_DECL(ngraph::op::Pad)
+            {
+                auto pad = static_cast<const ngraph::op::Pad*>(node);
+                writer.block_begin("  // " + node->get_name());
```
rkimballn1: We have removed all of the trailing comments for block_begin() because we have put a comment description of the function prior to emitting the op. Look at CPU emitted code to see what it looks like. Comments in emitted function makes Common Function Elimination more difficult.

src/ngraph/runtime/gpu/gpu_util.hpp
```diff
@@ -52,6 +52,16 @@
         }                                                                                          \
     } while (0)
 
+#define CUDNN_SAFE_CALL(func)                                                                      \
```
rkimballn1: Why is this a macro and not a function?

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -166,3 +170,148 @@ std::vector<int> cudnn_util::compute_strides(const std::vector<int>& dim)
...
+                                   const ngraph::Shape& output_shape,
+                                   const ngraph::Strides& window_strides,
+                                   const ngraph::Shape& window_shape,
+                                   const ngraph::Shape& padding_below,
+                                   const ngraph::Shape& padding_above)
+{
+    // construct hash to determine if kernel needs to be emitted
+    // or if it already exists in the primitive list
+    std::stringstream ss;
+    ss << "pool_op" << pool_op << "_dir" << static_cast<int>(direction) << "_i" << join(input_shape)
```
fengleitian: My thought is if the parameter is not hard coded into kernel, it does not need to go into hash, since they can share same primitive.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -166,3 +170,148 @@ std::vector<int> cudnn_util::compute_strides(const std::vector<int>& dim)
...
+                                    pool_op,
+                                    CUDNN_NOT_PROPAGATE_NAN,
+                                    window_shape[0],
+                                    window_shape[1],
+                                    0,
+                                    0,
+                                    window_strides[0],
+                                    window_strides[1]);
+    }
+    else if (input_shape.size() == 5)
```
fengleitian: Did I miss a branch with `input_shape.size() > 5`

test/autodiff.in.cpp
```diff
@@ -42,7 +41,7 @@ TEST(${BACKEND_NAME}, backwards_maxpool_n4_c1_hw4_2x2_max)
 
     auto A = make_shared<op::Parameter>(element::i32, shape_a);
     auto reshape = make_shared<op::Reshape>(
-        A, AxisVector{0, 3, 1, 2}, Shape{1, 4, 4, 4}); //convert CHWN to CNHW
+        A, AxisVector{3, 0, 1, 2}, Shape{4, 1, 4, 4}); //convert CHWN to NCHW
```
fengleitian: will this affect other backend. Or it was some bug?

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: good to know. :)

src/ngraph/runtime/gpu/gpu_util.hpp
fengleitian: My understanding is Macro is easier to take any function as an input. It will be much harder to do with a function. I agree, at some point we need to have our own error handle.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
csullivan: I will change these, but if it is ok, I would prefer to make a separate PR to remove them because we do this in many places and it might obfuscate this PR. I also notice this in the the IA transformer, e.g. mkldnn_emitter.cpp

src/ngraph/runtime/gpu/gpu_emitter.cpp
csullivan: Really? The comments screw the compiler up when optimizing common expressions? That almost seems fragile! I like how the CPU emitted code looks. In some cases we still need the {} for scope. I will look into how this is implemented on the CPU side.

src/ngraph/runtime/gpu/gpu_cuda_kernel_emitters.cpp
csullivan: Removed.

src/ngraph/runtime/gpu/gpu_util.hpp
csullivan: Also, a lot of frameworks implement these as defines (e.g. https://github.com/NervanaSystems/ngraph-mxnet/blob/3857caef796fc9d4c3df4ea5e9f83f8af3f089ae/src/common/cuda_utils.h#L175) so that they can be wrapped up with ifndef NDEBUG so that the asserted code is only checked for debug builds. Something we should add.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -27,3 +31,148 @@ CUDAEmitter::CUDAEmitter(GPUPrimitiveEmitter* emitter)
...
+
+    // if the kernel has not been compiled, build it
+    auto compiled_kernel = ctx->compiled_kernel_pool->get(hash);
+    auto nthreads = shape_size(input_shape);
+    if (compiled_kernel == nullptr)
+    {
+        // normalize pad dimensions to shape dimensions
+        Shape pad_below(input_shape.size(), 0);
+        Shape pad_above(input_shape.size(), 0);
+        if (padding_below.size() != input_shape.size())
```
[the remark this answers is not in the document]
csullivan: There are two different convention for padding in nGraph, one for op::Pad and one for op::MaxPool/op::Convolution.

For op::MaxPool/Convolution, the padding rank and the tensor rank will not be equal. In that case padding_rank == tensor_rank-2, or the number of spatial dimensions. That is what this conditional is checking for. For op::Pad, you are right, the ranks will be equal.

This relates to the discussion in #824. I found this to be a little confusing, I'm not sure if the difference in convention is necessary or not. I'm guessing it is there to prevent people from accidentally padding the N and C dimensions for pooling and convolution.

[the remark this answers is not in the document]
fengleitian: Yes, got it.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -166,3 +170,148 @@ std::vector<int> cudnn_util::compute_strides(const std::vector<int>& dim)
 
     return strides;
 }
+
+cudnnTensorDescriptor_t cudnn_util::tensor_descriptor_4d_from_shape(const Shape& shape)
```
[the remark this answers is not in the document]
csullivan: Really great point, I will do that and rework the kernels to use the single function.

test/autodiff.in.cpp
csullivan: After discussing with @Krovatkin, we determined that the indices in the test were wrong (bug), but because the size of one of the dimensions was 1, the in-memory layout was equivalent. Since CNHW is not a common format, I believe the change I've made to the test was it's original intent (to NCHW). It shouldn't affect other backends.

fengleitian: 👍

PR #821, opened 2018-04-05
Python nGraph operations wrappers.

arogowie-intel:
- OneHot, Power, Reduce,  Sqrt, Relu, Sign, Sin, Sinh, Tan, Subtract, Select, Tanh, Sum, Reduce,
  Softmax, ReplaceSlice, Reverse


python/ngraph/utils/types.py
```diff
@@ -32,6 +32,7 @@
 TensorShape = List[int]
 NumericData = Union[int, float, np.ndarray]
 NumericType = Union[type, np.dtype]
+ScalarType = Union[int, float]
```
postrational: This should be named `ScalarData`, by analogy to `NumericData` above.

python/ngraph/ops.py
```diff
@@ -521,9 +628,31 @@ def prod(node, reduction_axes=None, name=None):
     :param node: The tensor we want to product-reduce.
     :param reduction_axes: The axes to eliminate through product operation.
     :param name: Optional name for output node.
+    :return: The new node performing product-reduction operation.
+    """
+    return Product(node, AxisSet(get_reduction_axes(node, reduction_axes)))
+
+
+@nameable_op
+def reduce(node, init_val, reduce_fun, reduce_axes, name=None):
```
postrational: Does this function accept a Python callable?
Let's check and add a unit test for this one.

arogowie-intel: Yup, it accepts Python callable. It seems to me that it should be a constructor of an object deriving from BinaryElementwiseArithmetic.

python/ngraph/utils/types.py
arogowie-intel: Fixed.

PR #824, opened 2018-04-05
gpu convolution support nd(n<4)

fengleitian:
Currently 2D(4D tensor) and 3D(5D tensor) convolution works.

because of cudnn, currently there are limitation:
1. does not support asymmetric padding, which means only supply when padding_below = padding_above
2. does not support data dilation
3. 4D (>=6D tensor) convolution has issue on cuDNN so far, will continue debug it.

future work:
1. add a separated  padding for asymmetric padding.
2. add data dilation (deconvolution) support.


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -171,6 +171,247 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                Strides window_dilation_strides = convolution->get_window_dilation_strides();
+                Strides window_movement_strides = convolution->get_window_movement_strides();
+                Strides data_dilation_strides = convolution->get_data_dilation_strides();
+                CoordinateDiff padding = convolution->get_padding_below();
+                CoordinateDiff padding_above = convolution->get_padding_above();
+
+                if (padding.size() > 3)
+                {
+                    throw std::runtime_error(node->get_name() +
+                                             "with more than 3D still not working.");
```
rkimballn1: Say `not implemented` not `not working`

fengleitian: Got it.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -171,6 +171,247 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                auto convolution = static_cast<const ngraph::op::ConvolutionBackpropData*>(node);
+                Strides window_dilation_strides =
+                    convolution->get_window_dilation_strides_forward();
+                Strides window_movement_strides =
+                    convolution->get_window_movement_strides_forward();
+                Strides data_dilation_strides = convolution->get_data_dilation_strides_forward();
+                CoordinateDiff padding = convolution->get_padding_below_forward();
+                CoordinateDiff padding_above = convolution->get_padding_above_forward();
+
+                if (padding.size() > 3)
```
csullivan: Same comment as above.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -171,6 +171,247 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                       << "&alpha, " << args0 << ", " << args[0].get_name() << ", " << args1 << ", "
+                       << args[1].get_name() << ", " << conv_descriptor << ", " << conv_algo << ", "
+                       << "workspace, " << workSapceSizeInBytes << ", "
+                       << "&beta, " << out0 << ", " << out[0].get_name() << ");\n";
+
+                writer << "runtime::gpu::free_gpu_buffer(workspace);\n";
+                writer.block_end();
+            }
+
+            template <>
```
csullivan: I noticed that these three ops (Conv,BPData,BPFilters) all have essentially the same boiler-plate/prologue code for emitting them. This could be refactored into a single function for maintainability, but is maybe not a high priority for now.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -171,6 +171,247 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                const std::string conv_algo = "CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM";
+                const size_t workSapceSizeInBytes = 1024;
+                auto convolution = static_cast<const ngraph::op::Convolution*>(node);
+                Strides window_dilation_strides = convolution->get_window_dilation_strides();
+                Strides window_movement_strides = convolution->get_window_movement_strides();
+                Strides data_dilation_strides = convolution->get_data_dilation_strides();
+                CoordinateDiff padding = convolution->get_padding_below();
+                CoordinateDiff padding_above = convolution->get_padding_above();
+
+                if (padding.size() > 3)
```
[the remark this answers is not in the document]
fengleitian: for convolution, it does not pad NC.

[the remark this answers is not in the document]
csullivan: Right, but pad = {0,0,1,1}, also does not pad NC, only HW. If the padding array is passed in as this, instead of {1,1}, it will it your assertion even though it is only 2 dimensional data.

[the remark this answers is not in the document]
csullivan: Maybe try this with one of the tests and see where it is caught?

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -171,6 +171,247 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+
+                const std::string args0 = "x_descriptor";
+                const std::string args1 = "w_descriptor";
+                const std::string out0 = "y_descriptor";
+                const std::string conv_descriptor = "conv_descriptor";
+                const std::string data_type = "CUDNN_DATA_FLOAT";
+                const std::string tensor_format = "CUDNN_TENSOR_NCHW";
+                const std::string mode = "CUDNN_CROSS_CORRELATION";
+                const std::string conv_algo = "CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM";
+                const size_t workSapceSizeInBytes = 1024;
```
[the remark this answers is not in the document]
fengleitian: There is a function to decide the space needed, the algorithm I chose here request 0 space.

src/ngraph/runtime/gpu/gpu_emitter.cpp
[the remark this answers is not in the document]
aprocter: Looks like the vector is coming from nGraph, so it will not include the N, C dimensions.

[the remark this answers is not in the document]
csullivan: @aprocter In that case, I would suggest that the above mentioned test name is a bit misleading for op::Pad. In that test, it is an exterior pad of a 4d tensor, and the pad is provided as {0,0,1,1} which includes N and C. 

If I am supposed to interpret this not as a NCHW tensor, but rather just an NDArray, then it is only a partial exterior padding and then the name is wrong.

Please correct me if I am missing a third alternative interpretation, that's just where I got with this test :)

src/ngraph/runtime/gpu/gpu_emitter.cpp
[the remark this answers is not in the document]
fengleitian: Let me add the function.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: Right, most of the code are very similar. there are slight different with each other.
Since those are not completed yet(missing asymmetric padding and data dilation), let's do it later.

src/ngraph/runtime/gpu/gpu_emitter.cpp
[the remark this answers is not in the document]
fengleitian: will op::Pad actually pad NC?

src/ngraph/runtime/gpu/gpu_kernel_emitters.cpp
```diff
@@ -45,6 +45,129 @@ void runtime::gpu::kernel::emit_memcpyDtD(codegen::CodeWriter& writer,
...
+    }
+}
+
+void runtime::gpu::kernel::emit_cudnnFilterDescriptor(codegen::CodeWriter& writer,
+                                                      const std::string& name,
+                                                      const std::string& format,
+                                                      const std::string& data_type,
+                                                      const Shape& shape)
+{
+    std::vector<size_t> dimensions;
```
Krovatkin: nitpick: this logic seems to be duplicated across `emit_cudnnFilterDescriptor` and `emit_cudnnTensorDescriptor`. Can we refactor it into a small helper?

src/ngraph/runtime/gpu/gpu_util.cpp
```diff
@@ -32,14 +32,15 @@ using namespace std;
 
 void runtime::gpu::print_gpu_f32_tensor(void* p, size_t element_count, size_t element_size)
 {
-    float* local;
+    float* local = new float[element_count];
```
Krovatkin: nitpick: we have a helper `vector_to_string` I'd just recommend allocating the vector of the right length getting `&v[0]` and using `vector_to_string` to print out tensors' values.

src/ngraph/runtime/gpu/gpu_kernel_emitters.cpp
```diff
@@ -45,6 +45,129 @@ void runtime::gpu::kernel::emit_memcpyDtD(codegen::CodeWriter& writer,
...
+        writer << "                 /*format=*/" << format << ",\n";
+        writer << "                 /*dataType=*/" << data_type << ",\n";
+        writer << "                 /*dimension_size*/" << dimensions[0] << ",\n";
+        writer << "                 /*dimension_size*/" << dimensions[1] << ",\n";
+        writer << "                 /*dimension_size*/" << dimensions[2] << ",\n";
+        writer << "                 /*dimension_size*/" << dimensions[3] << ");\n";
+    }
+    else
+    {
+        auto compute_strides = [](const std::vector<size_t>& dim) {
```
Krovatkin: nitpick: I wonder if @aprocter might have already written a helper for `compute_strides`?

src/ngraph/runtime/gpu/gpu_kernel_emitters.cpp
```diff
@@ -45,6 +45,129 @@ void runtime::gpu::kernel::emit_memcpyDtD(codegen::CodeWriter& writer,
...
+    }
+}
+
+void runtime::gpu::kernel::emit_cudnnTensorDescriptor(codegen::CodeWriter& writer,
+                                                      const std::string& name,
+                                                      const std::string& format,
+                                                      const std::string& data_type,
+                                                      const Shape& shape)
+{
+    std::vector<size_t> dimensions;
```
Krovatkin: nitpick: this just pads `dimensions` to at least 4 (if less) with 1s? I wonder if we could add a short comment?

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -171,6 +171,259 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                Strides data_dilation_strides = convolution->get_data_dilation_strides();
+                CoordinateDiff padding = convolution->get_padding_below();
+                CoordinateDiff padding_above = convolution->get_padding_above();
+
+                if (padding.size() > 3)
+                {
+                    throw std::runtime_error(node->get_name() +
+                                             "with more than 3D is not implemented.");
+                }
+                for (auto a : data_dilation_strides)
```
Krovatkin: nitpick: checks look the same for convolution forward and backward. I wonder if we could maybe refactor them into a helper?

src/ngraph/runtime/gpu/gpu_kernel_emitters.cpp
rkimballn1: Or just initialize dimensions with {1, 1, 1, 1}

src/ngraph/runtime/gpu/gpu_kernel_emitters.cpp
rkimballn1: The function is `row_major_strides` in shape.hpp
I have been seeing this computation in other PRs. Can we clean this up?

src/ngraph/runtime/gpu/gpu_util.cpp
rkimballn1: Or use the `join` function in util.hpp

src/ngraph/runtime/gpu/gpu_kernel_emitters.cpp
fengleitian: yes, let me change it. Thanks.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: good point, we already had a discussion about this, will do it later.

PR #825, opened 2018-04-05
Support concat with mkldnn and add a test case

nbpatel:


src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -678,3 +678,36 @@ size_t MKLDNNEmitter::build_batchnorm_backward(const mkldnn::memory::desc& weigh
                                          dweights_index};
     return batchnorm_index;
 }
+
+size_t MKLDNNEmitter::build_concat(const std::vector<mkldnn::memory::desc>& inputs_data_desc,
+                                   const mkldnn::memory::desc& result_desc,
+                                   int concat_dims,
```
pruthviIntel: can u make at as const, `const int dims`

src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -184,6 +184,11 @@ namespace ngraph
                                                 const mkldnn::memory::desc& dweights_desc,
                                                 const double eps);
 
+                size_t build_concat(const std::vector<mkldnn::memory::desc>& inputs_data_desc,
+                                    const mkldnn::memory::desc& result_desc,
+                                    int concat_dims,
```
pruthviIntel: same as above, make it const.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -76,6 +77,31 @@ namespace ngraph
...
+                    bool flag = true;
+
+                    for (size_t i = 0; i < num_inputs; i++)
+                    {
+                        if (node->get_input_element_type(i) == element::f32 &&
+                            (node->get_input_shape(i)).size() == 4)
+                            continue;
+                        else
+                            flag = false;
+                    }
```
pruthviIntel: I think, we don't need to have this `flag` here. ngraph is strongly typed. so it ensures all the arguments have same rank during compilation time. so you can simplify here. Also I think you can allow it as mkldnn op,even if the rank size is 2.

nbpatel:
TESTNOW

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -851,25 +851,76 @@ namespace ngraph
...
+                    for (size_t i = 0; i < args.size(); i++)
+                    {
+                        inputs_data_desc.push_back(
+                            mkldnn_emitter->build_memory_descriptor(args[i], inputs_format[i]));
+                    }
+
+                    auto result_desc =
+                        mkldnn_emitter->build_memory_descriptor(out[0], result_format);
+                    for (size_t i = 0; i < args.size(); i++)
+                    {
```
jbobba: Create this inside build_concat in mkldnn_emitter.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
jbobba: Agree. We might need to special case later for performance reasons but API seems broad enough for now.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -1168,6 +1169,33 @@ namespace ngraph
...
+                            runtime::cpu::mkldnn_utils::get_input_mkldnn_format(node.get(), 0);
+                        size_t num_inputs = node->get_input_size();
+
+                        vector<memory::format> prim_input_formats;
+                        vector<memory::format> prim_output_formats;
+                        for (size_t i = 0; i < num_inputs; i++)
+                        {
+                            prim_input_formats.push_back(input0_layout);
+                        }
+                        prim_output_formats.push_back(input0_layout);
```
jbobba: can you check with MKLDNN folks on what the output format should be? http://intel.github.io/mkl-dnn/group__c__api__concat.html Their API says we can either set output format/desc to NULL or create the primitive with mkldnn::any and query for the right output format.

test/backend_test.in.cpp
```diff
@@ -483,6 +483,35 @@ TEST(${BACKEND_NAME}, concat_vector)
     EXPECT_EQ((vector<float>{2, 4, 8, 16, 1, 2, 4, 8, 16, 32, 18, 19}), read_vector<float>(result));
 }
 
+TEST(${BACKEND_NAME}, concat_mkldnn)
```
jbobba: Why do we need a special test case for mkldnn?

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
nbpatel: ok so you guys are saying API should be good enough to handle any dimensioned input and also we dont need to check for the element type of the inputs?

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
nbpatel: sure.

test/backend_test.in.cpp
nbpatel: The other concat test cases did not have 4D inputs and i only enable concat for mkldnn for 4D inputs. Since you are specifying the API should be able to handle any dimension inputs, i guess we dont need this anymore.

jbobba: you can add it as concat_4d_tensor instead of calling it concat_mkldnn

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
jbobba: we will still need element type check. @pruthviIntel was referring to this: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/concat.cpp#L49-L57

You only need to check element_type and rank for the first input tensor. Rest will be the same.

nbpatel: @jbobba ok makes sense now. Also, should be just have the input limited to rank 4 only for mkldnn? or as @pruthviIntel mentioned above rank 2 should also be ok

jmenon: I'd suggest doing a quick timing experiment to make sure MKLDNN still performs well on smaller tensors

DawnStone:
TESTNOW

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
nbpatel: @jbobba evarist said we can do either. Have any and query the output format or have the user define the output format. So what do you think is our way forward? Also, i've added a condition to allow only 2D and 4D inputs to use mkldnn concat, since its optimized for that. I have also made the other changes you and @pruthviIntel requested. Please take a look

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -851,25 +851,70 @@ namespace ngraph
...
+                    writer << "cpu::mkldnn_utils::set_memory_ptr(ctx, " << to_string(deps[i])
+                           << ", " << out[0].get_name() << ");\n";
+
+                    writer << "cpu::mkldnn_utils::mkldnn_invoke_primitive(ctx, "
+                           << to_string(concat_index) << ");\n";
                 }
+                else
+                {
+                    auto axis =
+                        (dynamic_cast<const ngraph::op::Concat*>(node))->get_concatenation_axis();
```
Krovatkin: I think we use static casts everywhere for this specific case?

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -851,25 +851,70 @@ namespace ngraph
...
+                    {
+                        inputs_data_desc.push_back(
+                            mkldnn_emitter->build_memory_descriptor(args[i], inputs_format[i]));
+                    }
+
+                    auto result_desc =
+                        mkldnn_emitter->build_memory_descriptor(out[0], result_format);
+
+                    size_t concat_index = 0;
+                    size_t concat_dims =
```
Krovatkin: nitpick: `concant_dim` or `axis`? it reads like it's plural...

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
Krovatkin: nitpick: I don't know if matters that much but should we make an effort and set the input layout to one of the biggest concat chunk(s) to minimize possible `ConvertLayout`s?

src/ngraph/runtime/cpu/cpu_emitter.cpp
nbpatel: Its not added by me, its just moved a bit further down after the if condition which is added by me. But, if you'd like i can change it.

src/ngraph/runtime/cpu/cpu_emitter.cpp
nbpatel: my bad, i'll change it to concat_dim, which is what mkldnn api likes it to call

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
jbobba: evarist wanted us to query for output layouts. We need to create a primitive with 'any' as output layout and query that primitive for the actual layout.

test/backend_test.in.cpp
```diff
@@ -483,6 +483,64 @@ TEST(${BACKEND_NAME}, concat_vector)
     EXPECT_EQ((vector<float>{2, 4, 8, 16, 1, 2, 4, 8, 16, 32, 18, 19}), read_vector<float>(result));
 }
 
+TEST(${BACKEND_NAME}, concat_4d_tensor)
+{
+    SKIP_TEST_FOR("GPU", "${BACKEND_NAME}");
```
jbobba: Why skip 'GPU'?

test/backend_test.in.cpp
```diff
@@ -483,6 +483,64 @@ TEST(${BACKEND_NAME}, concat_vector)
...
+    copy_data(c, vector<float>{3});
+    auto result = backend->make_primary_tensor_view(element::f32, shape_r);
+
+    cf->call({result}, {a, b, c});
+    EXPECT_EQ((vector<float>{1, 2, 3}), read_vector<float>(result));
+}
+
+TEST(${BACKEND_NAME}, concat_2d_tensor)
+{
+    SKIP_TEST_FOR("GPU", "${BACKEND_NAME}");
```
jbobba: why skip 'GPU'?

nbpatel:
TESTNOW

nbpatel:
TESTNOW

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -1168,6 +1169,70 @@ namespace ngraph
...
+                        }
+                        for (size_t i = 0; i < num_inputs; i++)
+                        {
+                            inputs_format.push_back(
+                                runtime::cpu::mkldnn_utils::get_input_mkldnn_format(concat, i));
+                        }
+                        for (size_t i = 0; i < num_inputs; i++)
+                        {
+                            inputs_data_desc.push_back(
+                                mkldnn_emitter->build_memory_descriptor(in[i], inputs_format[i]));
```
jbobba: Create mkldnn memory descriptors directly instead of going through mkldnn_emitter

raramer01:
TESTNOW

nbpatel:
TESTNOW

test/backend_test.in.cpp
```diff
@@ -417,6 +417,56 @@ TEST(${BACKEND_NAME}, concat_vector)
     EXPECT_EQ((vector<float>{2, 4, 8, 16, 1, 2, 4, 8, 16, 32, 18, 19}), read_vector<float>(result));
 }
 
+TEST(${BACKEND_NAME}, concat_4d_tensor)
```
csullivan: In the future please add appropriate backend skips for new tests if other tests of the same op also skip those backends. Thanks!

PR #827, opened 2018-04-06
[Py] Add python wrapper for nGraph Reduce operation.

arogowie-intel:
- Add UT.

### DIFF
--- a/python/ngraph/ops.py
+++ b/python/ngraph/ops.py
@@ -17,13 +17,13 @@
 """Factory functions for all ngraph ops."""
 import numpy as np
 
-from ngraph.impl import AxisSet, AxisVector, Coordinate, CoordinateDiff, Node, NodeVector, \
-    Shape, Strides
+from ngraph.impl import AxisSet, AxisVector, Coordinate, CoordinateDiff, Function, Node, \
+    NodeVector, Shape, Strides
 
 from ngraph.impl.op import Abs, Acos, Add, Asin, Atan, AvgPool, Broadcast, Ceiling, Concat, \
     Constant, Convert, Convolution, Cos, Cosh, Divide, Dot, Equal, Exp, Floor, Greater, GreaterEq,\
     Less, LessEq, Log, Max, Maximum, MaxPool, Min, Minimum, Multiply, Negative, Not, NotEqual, \
-    Pad, Parameter, Product, Reshape, Slice, Softmax, Sqrt, Subtract, Sum, Tanh
+    Pad, Parameter, Product, Reduce, Reshape, Slice, Softmax, Sqrt, Subtract, Sum, Tanh
 
 from typing import Iterable, List
 
@@ -32,7 +32,7 @@ from ngraph.utils.decorators import nameable_op, binary_op, unary_op
 from ngraph.utils.input_validation import assert_list_of_ints
 from ngraph.utils.reduction import get_reduction_axes
 from ngraph.utils.types import NumericType, NumericData, TensorShape, make_constant_node, \
-    NodeInput
+    NodeInput, ScalarData
 from ngraph.utils.types import get_element_type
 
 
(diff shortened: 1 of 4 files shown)


python/ngraph/ops.py
```diff
@@ -633,6 +633,36 @@ def prod(node, reduction_axes=None, name=None):
     return Product(node, AxisSet(get_reduction_axes(node, reduction_axes)))
 
 
+@nameable_op
+def reduce(node,                 # type: Node
+           initial_value,        # type: ScalarData
+           reduction_node,       # type: Node
```
aprocter: This will need to be an nGraph `Function`. The idea is that you have some function on scalars `f(x,y) = ???`, and you're using that to reduce ("fold" in a functional programming sense) some of the axes. For example, if `f(x,y) = x+y`, and `n = {{1,2,3},{4,5,6}}` then (note: the following notation is pseudocode but it might be pretty close to the actual Python notation that falls out):

`reduce(n,0,f,reduction_axes=[1]) --> {0+1+2+3,0+4+5+6} --> {6,15}`

`reduce(n,0,f,reduction_axes=[0]) --> {0+1+4,0+2+5,0+3+6} --> {5,7,9}`

`reduce(n,0,f,reduction_axes=[0,1]) --> (0+1+2+3)+(0+4+5+6) --> 21`

_Except_ that I'm implying a higher degree of determinism here than actually exists: it's acceptable, and the backend may choose, to sum the columns first.

This can be used to implement more general reductions. For example, if `initial_value = 1` and `f(x,y) = x*y` then you can use it to take the product (of each row, or each column, or of an entire matrix, or of analogous things in higher dimensions). If `initial_value = -inf` and `f(x,y) = max(x,y)` you can take the maximum, and so on.

The reason it has to be a `Function` instead of a `Node` is that a `Node` itself can't be free-standing if it takes inputs---it has to be hooked up to some other node(|s) that generate(s|) its input(|s). So even if we are doing a simple `Add`, that `Add` still has to be wrapped up in a nested `Function`, whose inputs are represented as `Parameter` nodes. Think of it like the `lambda` that shows up in this example: http://book.pythontips.com/en/latest/map_filter.html#reduce .

Just as an aside, note that this op made its way into nGraph originally because we needed it in order to implement this op from XLA:

https://www.tensorflow.org/performance/xla/operation_semantics#reduce

In practice, we tend to avoid its use (favoring instead ops like `Sum` or `Max`) because it is hard to compile it efficiently in the fully general case. (There are ways it can be done, but for the moment implementing them is a low priority.)

python/test/ngraph/test_reduction.py
```diff
@@ -42,3 +42,38 @@ def test_reduction_ops(ng_api_helper, numpy_function, reduction_axes):
...
+    init_val = np.float32(0.)
+    input_data = np.random.randn(3, 4, 5).astype(np.float32)
+    expected = np.sum(input_data, axis=reduction_axes)
+    reduction_function_args = [init_val, ng.impl.op.Add, list(reduction_axes)]
+    result = run_op_node([input_data], ng.reduce, *reduction_function_args)
+    assert np.allclose(result, expected)
+
+    reduction_axes = (0, )
+    input_data = np.random.randn(100).astype(np.float32)
+    expected = reduce(lambda x, y: x - y, input_data, np.float32(0.))
```
aprocter: Having just written the wall of text above, I now see that you are doing the wrapping here. :) (Though I think the order of the `node` and `reduction_node` is flipped, and the docstrings above are incorrect---what you have here isn't really a `Node`, but a function on two `Node`s that produces a `Node`.) That should work just fine, and if it feels more "Pythonic" (I'm not a Python expert by the remotest stretch of the imagination) I think it makes sense. Would probably be good to also have a variant that takes an nGraph `Function` object in place of the Python-level `lambda`.

python/ngraph/ops.py
```diff
@@ -633,6 +633,36 @@ def prod(node, reduction_axes=None, name=None):
...
+    :param name: The new name for output node.
+    :return: The node performing reduction operation with provided reduction node.
+    """
+    if reduction_axes is None:
+        reduction_axes = list(range(len(node.shape)))
+    init_val_node = constant(initial_value)
+    # wrap reduction node into Function object
+    param1 = Parameter(node.get_element_type(), Shape([]))
+    param2 = Parameter(node.get_element_type(), Shape([]))
+    reduction_operation = Function(NodeVector([reduction_node(param1, param2)]), [param1, param2],
```
postrational: Let's add an option to accept a `Function` instance directly.

The current code, which takes a Python callable should be one of 2 possible inputs.

python/ngraph/ops.py
```diff
@@ -633,6 +633,36 @@ def prod(node, reduction_axes=None, name=None):
...
+                           produces a scalar result.
+    :param reduction_axes: The list of axes indices to be reduced. Default to reduce all axes.
+    :param name: The new name for output node.
+    :return: The node performing reduction operation with provided reduction node.
+    """
+    if reduction_axes is None:
+        reduction_axes = list(range(len(node.shape)))
+    init_val_node = constant(initial_value)
+    # wrap reduction node into Function object
+    param1 = Parameter(node.get_element_type(), Shape([]))
```
postrational: @aprocter Does it make sense here to assume that `param`s will have an empty shape (scalar values)?

aprocter: Yup, it’ll always be scalar.

PR #829, opened 2018-04-06
Use less complex pass base where possible

rkimballn1:


Krovatkin:
@diyessi I was wondering if we could just remove `NodePass` and `CallGraphPass` and leave only `FunctionPass` so a pass uses whichever traversal order it needs by calling an appropriate `Function`'s getter.

jbobba:
@Krovatkin I dont see any changes to CPULayout. Am i missing something here?

PR #831, opened 2018-04-06
Remove the no longer supported alternative installation method for

silee2:
python binding.


silee2:
@diyessi This PR is removing unsupported installation method. #755 was an attempt to keep supporting it but since there is no user. Just removing it.

PR #836, opened 2018-04-09
Optimize 4D Reshape

jmenon:


mbrookhart:
What would it take to make these Eigen kernels more generic than specific element types/shapes? Would we expect a more general speedup by doing that?

jmenon:
The original plan was to instantiate these during runtime compilation (similar to the core eigen paths) but that'll have an impact on the compilation time (which works for me but might not be acceptable generally for all bridges). There is some amount of redesign that can be done but given the narrow ranks/shape space that gets used in models currently, I don't think there's any immediate performance improvements to be had by doing all that work.

jbobba:
2D can be a separate PR but was curious why you created a 2D specialization for Eigen as well?

jmenon:
I think that was for unit tests but I didn't keep that change in the emitter. Replaced it with a 3d instance.

PR #837, opened 2018-04-09
Add GPURuntimeContext and GPUPrimitiveEmitter to the gpu transformer 

csullivan:
This is a backward compatible restructuring of the GPU transformer that enables several capabilities. It is a superset of PRs #790 and #792 which are now closed. Review comments from those PRs have been applied to this branch.

Features:
* extern C structure (GPURuntimeContext) passed from ngraph runtime to JIT runtime. 
* GPUPrimitiveEmitter - CUDA and CUDNN primitives can be registered with this class so that they can be safely invoked from the JIT runtime.
* The GPUPrimitiveEmitter has two main components the CUDNNEmitter and CUDAEmitter. The CUDNNEmitter enables CUDNN to be compiled with the ngraph core, rather than JIT. The CUDAEmitter allows CUDA kernels to be compiled from ngraph instead of from the JIT code.
* Both CUDA and CUDNN emitters can cache their primitives according to a hash defined at the primitive definition to avoid recompiling primitives that operate identically (an example of this is provided in the upcoming PR for MaxPool and op::Pad)

Limitation:
While this PR is backwards compatible with the previous implementation of cuda kernels and emitted cudnn code, it should be noted that any calls from the JIT into ngraph are unsafe and should eventually be refactored. If the native compiler and the clang compiler link against different GLIBCXX, runtime errors will occur.


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -772,92 +775,25 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
-                                                           args[0],
-                                                           out[0],
-                                                           "CUDNN_REDUCE_TENSOR_ADD",
-                                                           tensor_type,
-                                                           "CUDNN_NOT_PROPAGATE_NAN",
-                                                           input_desc,
-                                                           output_desc,
-                                                           1.0,
-                                                           0.0);
+                            auto& cudnn_emitter =
```
fengleitian: This is good, some of my thought:
```
int instance_idx = OpFactory.getOp().build_instance(sth);
writer << "OpFactory,getOp(instance_idx).run(input, output);
```
Basically we can create an instance with specialized parameters, and the emitter just run it.

This can be shared for all Ops, whether use cudnn/cuda can be managed in the Ops. 
Let me know how you think about it.

csullivan: I actually have it setup to do this already with the CUDAEmitter for Pad, see this for example: 
https://github.com/NervanaSystems/ngraph/blob/42911c8f7dd7aa398c0a5707235ca75b8b76a7f8/src/ngraph/runtime/gpu/cuda_emitter.cpp#L35.

The emitted kernel is specific to the tensor input and the requested padding and this is how the emitted kernel looks for test GPU.max_pool_2d_1channel_1image_padded:

```
more gpu_codegen/cuda_kernel_pad_i1__1__5__5_pb1__0_pa1__2_pi_codegen.cu
extern "C" __global__ void cuda_pad_i1__1__5__5_pb1__0_pa1__2_pi(float* in, float* out)
{
    size_t tid = blockIdx.x * blockDim.x + threadIdx.x;
    if (tid < 25)
    {
        size_t idx = 7 + tid % 5 + ((tid / 5) % 5) * 7 + ((tid / 25) % 5) * 49;
        out[idx] = in[tid];
    }
}
```

Notice how no variables are passed in other than input and output as you suggest, since they are all emitted into the writer for compilation. 

It is then called in the emitter via (https://github.com/NervanaSystems/ngraph/blob/42911c8f7dd7aa398c0a5707235ca75b8b76a7f8/src/ngraph/runtime/gpu/gpu_emitter.cpp#L819)

```
                        writer << "gpu::invoke_primitive(ctx, " << pad_index << ", ";
                        writer << "std::vector<void*>{" << args[0].get_name() << "}.data(), ";
                        writer << "std::vector<void*>{pad_buffer}.data()";
                        writer << ");\n";
                    }
```

The importance of the gpu::invoke_primitive function is that it is compiled in nGraph as `extern "C"` so that it provides a pure C bridge between the JIT runtime and the ngraph runtime. I believe this function is providing the interface that you suggest with the proposed ``OpFactory``.

csullivan: These code snippets are from PR #817 which depends on this PR.

fengleitian: I noticed you already have the builder. I'm looking for the possibility to wrap all those into a more unified interface.

Since we already have ctx during the build, should we pass another one during the run?

My understanding is we just need to make the interface `extern "C"`, and it will work, correct me if I'm wrong.

csullivan: Ah, I _think_ I understand now, please do correct me if I am wrong:

Instead of 
```gpu_primitive_emitter::get_cuda_emitter()``` and ```gpu_primitive_emitter::get_cudnn_emitter()``` 

your suggestion is that we could possibly just have ```idx = gpu_primitive_emitter/OpFactory.getOp(idx).build(…)``` 

and for invocation instead of ```gpu::invoke_primitive(idx)``` 

we have ```writer<<OpFactory.getOp(idx).run(input,output)```.

This would be nice, to have the building and invocation using the same interface, however, I'm not sure how to do it safely. You are correct, the interface needs to be ```extern "C"```, but that precludes this kind of an interface,

```writer<<"OpFactory.getOp(idx).run(input,output)"```

because this is a C++ call. However, if what you are meaning is:

```
writer <<"invoke(cts,idx,input,output)";
...
extern "C" invoke(ctx,idx,input,output) 
{
   ctx->OpFactory.getOp(idx).run(input,output);
}
```

That would be doable, and is pretty close to the current implementation. The only reason to avoid doing so is that by having a C++ pointer exist in the runtime context, it seems to tell the user that it is safe to be called within the JIT code when it is actually not. This is why the std::vector<gpu::primitives*> is stored in the runtime context as a gpu::primitive**, so that the JIT code has no idea that it is actually coming from a C++ object. Only numbers are being passed. We could save it in the runtime context as a void* to avoid this problem, however. I see merits both ways.

**Comment by csullivan** [comment]
@fengleitian's comment above reminded me of one other thing.

In MKLDNN, there are callable primitives and *memory primitives*. This PR brings in the support for callable primitives (cuda or cudnn) on the GPU side, but not memory primitives. I could see bringing in memory primitives at some point too, as a way of managing memory on the device and for passing/binding memory-primimitive-inputs that are lazily computed in the JIT code and so cannot be "baked" in to the kernel/primitives themselves from ngraph during emission.

Doing something like this would allow us to have a class that handles the memory allocations in the GPU and can reuse space that is needed for transient/temporary allocations (like what we use for padding, or for the workspaces needed for op::Sum). 

In addition, we could use it to cache results that will be needed by backprop, and then they can be bound to the backprop op primitive so that it knows where to look at a later time. An example of this use case would be for (MaxPool and MaxPoolBackprop, where the backprop needs the indices of the Maxpool. If they were cached in the forward pass, then an additional computation could be avoided).

File: src/ngraph/runtime/gpu/cudnn_emitter.cpp [hunk 2]
```diff
@@ -0,0 +1,168 @@
...
+                              &beta,
+                              output_desc,
+                              outputs[0]);
+            free_gpu_buffer(workspace_ptr);
+        }};
+
+    return this->m_primitive_emitter->insert(reduce);
+}
+
+std::vector<int> cudnn_util::compute_strides(const std::vector<int>& dim)
```
rkimballn1: This must exist somewhere else, but I haven't looked. Ask Scott

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -241,7 +242,18 @@ runtime::gpu::GPU_ExternalFunction::GPU_ExternalFunction(
     : ngraph::runtime::ExternalFunction(function, release_function)
     , m_compiled_function(nullptr)
     , m_emit_timing(std::getenv("NGRAPH_GPU_EMIT_TIMING") != nullptr)
+    , m_ctx(new GPURuntimeContext)
 {
+    // Create context use driver API and make it current, the runtime call will pickup the context
+    // http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html
+    // #interoperability-between-runtime-and-driver-apis
+    ngraph::runtime::gpu::CudaContextManager::instance();
```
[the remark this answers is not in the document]
csullivan: The cudaInit function needs to be called before the gpu_emitter functions are called, since some of the cuda kernels are now compiled there. Unfortunately, the call_frame does not exist until after the gpu_emitters are run, and so this setup call has to happen earlier. That said because of @rkimballn1's backend change I suspect we will refactor this at some point.

.gitignore
```diff
@@ -100,3 +100,8 @@ python-wheels/
 
 # python wheels
 python/share/*
+
+# git merge
+*.orig
+\#*
+\.#*
```
csullivan: @fengleitian You made a comment about the whitespace here, is there an issue?

fengleitian: I think clang build with warning as error might fail due to this.

fengleitian: I was wrong, for this .gitignore file, here is the reason:
https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file

src/ngraph/runtime/gpu/gpu_external_function.cpp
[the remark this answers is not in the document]
fengleitian: OK, let do it later.

.gitignore
diyessi: For a long time, many UNIX programs did not see the last line of a file if it did not have a new line at the end of it, which is why this check exists.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
csullivan: Thanks, I have added a cudnn_util functions to use the ngraph::row_major_strides function. But actual in this case, I need to convert the shape from td::vector<size_t> to std::vector<int> and then do the stride calculation, as both the shape and strides are required to be arrays of ints when passed to cudnn.

PR #841, opened 2018-04-10
add reference OS to the docker image name for the contrib/docker/Makefile

DawnStone:
added the reference OS marker to the image name defined in the contrib/docker/Makefile
fixed variable settings in contrib/docker/make-dimage.sh script

If a docker image exists for CentOS 7.4, it has the same name as the docker image that would be built with Ubuntu 16.04 for the same branch.

Although this should not be an issue for an end-user building a docker image and running make targets from the `contrib/docker/Makefile` from the command line, I had observed an issue attempting to use the scripts to automate simultaneous make targets with different reference OS settings.


DawnStone:
TESTNOW

DawnStone:
TESTNOW

PR #843, opened 2018-04-10
gpu slice

fengleitian:
slice that working for general case without performance considering yet.

limitation:
1. only support float point


fengleitian:
@csullivan 
Exactly.

fengleitian:
:TESTNOW

PR #851, opened 2018-04-11
[WIP] switch std::deque in node.cpp to std::vector

Krovatkin:


Krovatkin:
@diyessi @rkimballn1 To rid the descriptor classes from `std::deque`s we need to enable move constructors as they are required by `std::vector` methods. I'd love to hear yr opinions on this PR!

Krovatkin:
@diyessi what if we turn them into copyable values? Then, users will always have a valid copy albeit it could be a stale one.

diyessi:
I was thinking something like an `InputRef` that's a node + input offset and is copyable and has accessors for the `Input` things. Same for `OutputRef` and `Output`. But most things could just do the numbered info access via `Node` like we already have.

aprocter:
> When looking a this a while back, I was thinking of referencing inputs/outputs from each other by node+offset.

:+1: for this. It ensures the node doesn't disappear, without requiring some weird owning linkage from each of node N's outputs back to node N.

diyessi:
BTW, it'd be good if we didn't need `get_inputs()` and `get_outputs()` since they're exposing too much implementation -- removing an output from that vector would be bad,

Krovatkin:
@diyessi @aprocter I also like this `(node, index)` idea. However, given the fact that this seems to be a non-trivial change, I'm actually wondering if it's even worthwhile to do it? It seems to be somewhat orthogonal to @diyessi 's other point : 

> BTW, it'd be good if we didn't need get_inputs() and get_outputs() since they're exposing too much implementation -- removing an output from that vector would be bad,

Also, these `deque`s are typically pretty small, so it's unlikely there's any huge performance difference between  using `deque`s and `vector`s.  So it sounds mostly like a consistency change?

diyessi:
Okay, here's a vastly more radical idea where we also get rid of `GetOutputElement`. It combines with a change I played with yesterday for factories instead of `make_shared`.
We make something called `NodeValue` which is a shared pointer to a node `n` and an offset `i`; it's handle to output `i` of the node `n`. Inputs to nodes are `NodeValue`s.
When we create nodes, like `ngraph::make<Add>(...)` we get back a `NodeValue` rather than a `Node`, and we want args to be `NodeValue` rather than `Node`. Patterns might need a little adjustment, or maybe not.

Krovatkin:
Yup, I'm closing this one for now.

PR #856, opened 2018-04-12
BatchNorm documentation

diyessi:


diyessi:
There are a lot of things I don't like about this BatchNorm op. Now there are more.

mbrookhart:
This is an example of what we're currently doing in the bridge to work around this: https://github.com/NervanaSystems/ngraph-mxnet/pull/207

DawnStone:
TESTNOW

diyessi:
I've changed this to what I think we want BatchNorm to be.

PR #858, opened 2018-04-13
add min, max reduction op

fengleitian:
limitation:
only support float point.


csullivan:
Oh, do we want to add a check to the type so that it throws if it is not float? Would be more easily caught on the ngraph side than the JIT side.

fengleitian:
@csullivan we need to add float point check to almost all Ops for now. We can do this in one PR.

PR #860, opened 2018-04-13
GPU Padding - add support for custom pad value and interior padding

csullivan:
This updates GPU op::Pad support to include a padding value specified by the user or for use in max pooling where the lowest c-type value is needed for the pad value. It also extends op::Pad to handle interior padding. All op::Pad tests now pass.

Additional features:
Add TypeInfo class that allows dispatching of c-type info based on ngraph element::Type.
Refactor max_pool_1d CUDA kernel to use CUDAEmitter interface.


src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -92,33 +101,165 @@ size_t runtime::gpu::CUDAEmitter::build_pad(const runtime::gpu::GPURuntimeContex
...
+        {
+            writer << dtypes[0] << "* val, ";
+        }
+        writer << dtypes[0] << "* in, " << dtypes[1] << "* out)\n";
         writer.block_begin();
         {
             writer << "size_t tid = blockIdx.x * blockDim.x + threadIdx.x; \n";
+
+            // fill kernel
             writer << "if (tid < " << nthreads << ")\n";
```
rkimballn1: What happens if this conditional or the one on line 132 fails? Does everything do the right thing? Do we have unit tests for this?

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1168,10 +1170,10 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
                     auto& result_shape = out[0].get_shape();
                     auto padding_below = max_pool->get_padding_below();
                     auto padding_above = max_pool->get_padding_above();
-                    if (padding_below.size() != padding_above.size())
+                    if (input_shape.size() < 3)
                     {
                         throw std::runtime_error(
-                            "Padding below and above are of different dimension.");
+                            "Invalid tensor type, tensors should be of at least 3 dimensions.");
```
rkimballn1: What is the value of the invalid type? More info is always better.

test/backend_test.in.cpp
```diff
@@ -4232,15 +4231,16 @@ TEST(${BACKEND_NAME}, max_pool_2d_1channel_1image_overpadded)
...
-    EXPECT_EQ((test::NDArray<float, 4>({{{{min, min, min, min, min},
-                                          {1, 2, 2, 2, 1},
-                                          {3, 3, 2, 2, 1},
-                                          {3, 3, 2, 1, 1},
-                                          {2, 1, 2, 2, 2},
-                                          {2, 2, 2, 2, 2},
-                                          {2, 2, 1, 0, 0}}}})
-                   .get_vector()),
-              read_vector<float>(result));
+    //EXPECT_TRUE(test::all_close(vector<float>{1, 1, 729, 125}, read_vector<float>(result)));
```
rkimballn1: remove this line

src/ngraph/runtime/gpu/cuda_emitter.cpp
csullivan: These two conditionals allow only the first however many threads to perform the kernel. Since there are almost always more threads than needed for kernel execution (because the gpu schedules threads in warps of 32), these conditionals are needed and are present in many CUDA kernels. Threads that fall outside this range may still execute these operations (due to the SIMT arch.) but are guaranteed not change the program state.

fengleitian: @csullivan is right, this is used for padding threads to multiply of warps(i.e. 32).

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1308,40 +1314,40 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
-                               << input_shape.back() << ", " << result_shape.back() << ");\n";
-                    }
-                    // 2d and 3d max pool (NCHW)
-                    else if (input_shape.size() == 4 || input_shape.size() == 5)
+                    if (input_shape.size() <= 5)
                     {
-                        auto& cudnn_emitter =
-                            external_function->get_primitive_emitter()->get_cudnn_emitter();
+                        size_t max_pool_index = 0;
+                        // 1d max pool (NCW)
```
fengleitian: BTW, is 1D a special case for 2D? with H=1?

src/ngraph/runtime/gpu/type_info.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "ngraph/type/element_type.hpp"
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace gpu
+        {
+            class TypeInfo
```
fengleitian: COOL.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1276,8 +1280,9 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
                             shape_size(shape_to_pool) * args[0].get_element_type().size();
                         writer << "void* pad_buffer = "
                                << "runtime::gpu::create_gpu_buffer(" << temp_size << ");\n";
-                        writer << "runtime::gpu::cuda_memset(pad_buffer, 0, " << temp_size
-                               << ");\n";
+
+                        std::stringstream ss;
+                        ss << TypeInfo::Get(args[0].get_element_type())->lowest();
```
fengleitian: Already have a `to_string()`?

PR #861, opened 2018-04-13
A reference implementation of batchnorm fprop and tests.

varun-intel:
-> Implements batchnorm fprop in the interpreter backends and adds a couple tests to verify that it is correct.
-> We need the interpreter implementation to run our nnp tests through the NNP_TESTER class, which compares against the interpreter.
-> Bprop to be added in another pr.


varun-intel:
TESTNOW

varun-intel:
TESTNOW

aprocter:
TESTNOW

aprocter:
TESTNOW

PR #865, opened 2018-04-15
move HostTensorView to interpreter as that is the only backend using it.

rkimballn1:


yxlao:
Will update NNP side to be compatible

rkimballn1:
There are now multiple interpreters. We will leave these files where they are for now.

PR #871, opened 2018-04-16
Remove usage of CMAKE_MAKE_PROGRAM as it slows down parallel build

silee2:
Current build process for external mkl-dnn tries to speed up compilation by invoking just part of the Make rules bt doing `BUILD_COMMAND ${CMAKE_MAKE_PROGRAM} mkldnn` But that effectively disables parallel build. This PR remove the explicit BUILD_COMMAND and enables parallel build. Now more work is done but in parallel and the build time is faster.

### DIFF
--- a/cmake/external_mkldnn.cmake
+++ b/cmake/external_mkldnn.cmake
@@ -39,7 +39,6 @@ if(NGRAPH_CPU_ENABLE)
                 -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
                 -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
                 -DCMAKE_INSTALL_PREFIX=${EXTERNAL_PROJECTS_ROOT}/mkldnn
-            BUILD_COMMAND ${CMAKE_MAKE_PROGRAM} mkldnn
             TMP_DIR "${EXTERNAL_PROJECTS_ROOT}/mkldnn/tmp"
             STAMP_DIR "${EXTERNAL_PROJECTS_ROOT}/mkldnn/stamp"
             DOWNLOAD_DIR "${EXTERNAL_PROJECTS_ROOT}/mkldnn/download"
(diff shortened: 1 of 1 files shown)


jmenon:
To be clear, this has nothing really do with CMAKE_MAKE_PROGRAM because what you really want to do is pass arguments along. Also, if I recall correctly, Chase's original change didn't help with build time because the 'install' rule will build everything anyway.

silee2:
@jmenon It does. On Mac it does not have any impact but on BDW or SKX Ubuntu 16.04 
﻿you get something like "make[6]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule."

cmake/external_mkldnn.cmake
```diff
@@ -39,7 +39,7 @@ if(NGRAPH_CPU_ENABLE)
                 -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
                 -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
                 -DCMAKE_INSTALL_PREFIX=${EXTERNAL_PROJECTS_ROOT}/mkldnn
-            BUILD_COMMAND ${CMAKE_MAKE_PROGRAM} mkldnn
+            BUILD_COMMAND $(MAKE) mkldnn
```
jmenon: This is incorrect and will break non Makefile generators

silee2: I'm aware that it will break non Makefile generators. Are we supporting other generators? I would love to use Ninja. But why do we have an explicit BUILD_COMMAND in the first place? We don't specify and explicit INSTALL_COMMAND and the default INSTALL_COMMAND builds the skipped targets anyway. I can go back to my original proposal that removes the BUILD_COMMAND. That will not break non Makefile generators and does not cause "warning: jobserver unavailable: using -j1. Add '+' to parent make rule." Does that sound like the right solution?

jmenon: I can't speak to "official" support etc but following CMake best practices typically results in build scripts that users can run with any generator. Ninja works right now and I try to fix issues or point out changes that will break it.

Regarding the substance of the issue, I have no comment really. I didn't introduce the BUILD_COMMAND change. Removing it entirely works for me FWIW. My goal is to just make sure my personal workflow doesn't break

PR #874, opened 2018-04-17
[WIP] transient CrossEntropySoftmax for numerically stable autodiff

Krovatkin:


Krovatkin:
shelving it off till the better times.

PR #883, opened 2018-04-18
Add Inference Engine (IE) backend

adstraw:
IE backend framework = a slimmed down version of the interpreter backend.  Calls to inference engine graph API will replace calls to reference kernels over time.


adstraw:
TESTNOW

RanACohen:
sorry for the late reply.

I'll check it out on Sunday.

adstraw:
TESTNOW

adstraw:
TESTNOW

rkimballn1:
TESTNOW

PR #884, opened 2018-04-18
Enable users to request default/row-major layouts on result nodes

jbobba:


ashokei:
is it possible to add a unittest for your PR using mem_reuse API. (mem reuse API has one unittest)

PR #885, opened 2018-04-19
[Py] Python nGraph API doc

arogowie-intel:
Currently there are few layout issues:
- [x] How to remove from the sidebar elements under 'Python API' (subparagraphs from respective page)
- [x] How to enforce that when we go under Python API -> ngraph  page to be visible in breadcrumb.
- [x]  Is it possible to unroll inplace the page "Modules Index" ?


arogowie-intel:
Rebased to master.

PR #887, opened 2018-04-19
fix bug the 'codegen' dir create after the first _ops.txt store

fengleitian:
Thanks @rkimballn1 to help figure out this.


fengleitian:
@csullivan I can not reproduce the issue. the gpu_codegen is generated without problem after delete. 
Let's sync the details offline.

PR #889, opened 2018-04-19
Add CUDNN_SAFE_CALL and CUBLAS_SAFE_CALL

fengleitian:
Those two macro will catch the cudnn and cublas running error.


fengleitian:
@csullivan is it the latest version? It runs without problem here.

rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

PR #892, opened 2018-04-19
Add logical-and, logical-or ops

aprocter:
This adds logical-and and logical-or ops to nGraph core. We'll need this for TensorFlow.

Currently they just work on `element::boolean`. My understanding is that MXNet probably still cannot use these, because MXNet expects boolean-ish ops to work on other element types. My thinking here is that these ops should be kept consistent with existing logical-not and comparison operators, until such time as we come up with a general solution to, as we've said in the core sync, "cut the bool".


aprocter:
> What do we think it would take to make the logical ops work with types other than element::boolean?

No problem in core AFAIK, it was really just the TF bridge blocking it. The real sticky wicket is actually just comparison ops---XLA wants a boolean return no matter the input type. I'll try to look into it soon.

aprocter:
@rkimballn1: Will @mbrookhart's suggestion above cause problems for the interpreter? That's the only place I can think of where it might be an issue.

aprocter:
Actually now that I think about it I don't think it should be an issue for the interpreter. We would just have to change:

```
        else if (node_op == "Equal")
        {
            reference::equal<T>(reinterpret_cast<T*>(args[0]->get_data_ptr()),
                                reinterpret_cast<T*>(args[1]->get_data_ptr()),
                                reinterpret_cast<char*>(out[0]->get_data_ptr()),
                                out[0]->get_element_count());
        }
```

to:

```
        else if (node_op == "Equal")
        {
            if(/* same-type-as-inputs flag is set*/)
            {
                reference::equal<T,T>(reinterpret_cast<T*>(args[0]->get_data_ptr()),
                                    reinterpret_cast<T*>(args[1]->get_data_ptr()),
                                    reinterpret_cast<T*>(out[0]->get_data_ptr()),
                                    out[0]->get_element_count());
            }
            else
            {
                reference::equal<T,char>(reinterpret_cast<T*>(args[0]->get_data_ptr()),
                                    reinterpret_cast<T*>(args[1]->get_data_ptr()),
                                    reinterpret_cast<char*>(out[0]->get_data_ptr()),
                                    out[0]->get_element_count());
            }
        }
```

The concern I'd had is that we might have had to add a template parameter to the interpreter machinery here in addition to `T`, but it looks like that's not the case.

test/type_prop.cpp
```diff
@@ -591,12 +591,81 @@ TEST(type_prop, multiply_bad_arguments)
                 });
 }
 
-TEST(type_prop, subtract_bad_arguments)
```
aprocter: Whoops, didn't mean to delete this! Will fix.

rkimballn1:
The interpreter has separate template types for special ops where we can have inputs vary as in Select or outputs different from inputs as in Convert. We can make these ops have an output that are not the same type as their inputs, it should not be a problem. I would love to "cut the bool" and make  everything (ignoring Convert) have consistent input/output types.

rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

test/backend_test.in.cpp
```diff
@@ -7933,3 +7933,43 @@ TEST(${BACKEND_NAME}, validate_call_output_shape)
 
     EXPECT_ANY_THROW(backend->call(f, {a}, {c, b}));
 }
+
```
csullivan: When adding tests for new ops, please remember to add skips for the other backends. 

e.g. SKIP_TEST_FOR("GPU", "${BACKEND_NAME}");

Thanks!

PR #893, opened 2018-04-19
CUDNN BatchNorm (inference/forward/backward)

csullivan:
This PR adds a CUDNN implementation of  batch normalization. 

Notes:

* CUDNN supports mean and inverse variance cacheing to speed up the backward propagation branch, but this would require a change in the ngraph batch norm op, and likely pattern matching to fuse these operations at the graph level. For now Inference and Forward/Backward training all run independently.

* CUDNN can calculate the moving average and variance but this feature is also not used in this implementation and thus must be handled by the bridge (similar to IA transformer)

* According to the original batchnorm publication (https://arxiv.org/pdf/1502.03167v3.pdf), 
population (unbiased) statistics should be used for inference, and mini-batch (biased)
statistics should be used training (forward/backward). For the variance this
means utlitizing the following equations, respectively:

  (biased)   Var[X] = 1/m * Sum_i(x_i-mu)^2      :: used in training
  (unbiased) Var[X] = 1/(m-1) * Sum_i(x_i-mu)^2  :: used in inference

  s.t. x_i are elements of X and m = NHW.

Internally, MKLDNN always returns the biased variance and CUDNN always returns the unbiased variance. To convert the biased variance to the unbiased variance, we currently force this to be done at the framework level via Var_unbiased[X] = m/(m-1)*Var[X]_biased. Thus, in addition to the cudnn forward operation during the forward branch of training, a cuddnOpTensor call is made that converts the variance tensor from unbiased to biased in order to match MKLDNN. It is the frameworks responsibility to ensure the correct variance is then used for inference (unbiased) vs training (biased - implicitly chosen via ngraph IA and GPU backends).

* Also adds support for op::GetOutputElement in the GPU transformer (needed for unit tests).


csullivan:
Need to add JIRA ticket for SKIPPED interpreter ops.

csullivan:
TESTNOW

csullivan:
TESTNOW

csullivan:
Need to add CUDNN_SAFE_CALL to all CUDNN calls.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1412,6 +1412,102 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
             }
 
+            template <>
+            void GPU_Emitter::EMITTER_DECL(ngraph::op::BatchNorm)
+            {
+                const ngraph::op::BatchNorm* batchnorm =
+                    static_cast<const ngraph::op::BatchNorm*>(node);
+
+                auto& cudnn_emitter =
+                    external_function->get_primitive_emitter()->get_cudnn_emitter();
```
fengleitian: My understanding is primitive, ctx etc is not part of externel_function, should be under backend. For example, multiple function can share primitive, ctx.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -343,3 +343,144 @@ size_t runtime::gpu::CUDNNEmitter::build_pooling(const GPURuntimeContext* ctx,
...
+        throw std::runtime_error("Batch Norm epsilon is less than CUDNN_BN_MIN_EPSILON");
+    }
+
+    cudnnTensorDescriptor_t derived_param_desc;
+    cudnnCreateTensorDescriptor(&derived_param_desc);
+    auto tensor_desc = runtime::gpu::cudnn_util::tensor_descriptor_from_shape(tensor_shape);
+    cudnnDeriveBNTensorDescriptor(derived_param_desc, tensor_desc, bn_op);
+
+    float alpha = 1.0, beta = 0.0;
+    gpu::primitive* batchnorm;
```
fengleitian: Possible to use shared_ptr/unique_ptr?

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -343,3 +343,144 @@ size_t runtime::gpu::CUDNNEmitter::build_pooling(const GPURuntimeContext* ctx,
...
+        return primitive_index;
+    }
+
+    if (epsilon < CUDNN_BN_MIN_EPSILON)
+    {
+        throw std::runtime_error("Batch Norm epsilon is less than CUDNN_BN_MIN_EPSILON");
+    }
+
+    cudnnTensorDescriptor_t derived_param_desc;
+    cudnnCreateTensorDescriptor(&derived_param_desc);
```
fengleitian: CUDNN_SAFE_CALL as you already mentioned.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
rkimballn1: I'm almost OK with using naked pointers if you do a new which is followed fairly quicly by a delete. I don't even see how this is deleted. Use *_ptr for this as @fengleitian suggests

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -343,3 +343,144 @@ size_t runtime::gpu::CUDNNEmitter::build_pooling(const GPURuntimeContext* ctx,
...
+    auto tensor_desc = runtime::gpu::cudnn_util::tensor_descriptor_from_shape(tensor_shape);
+    cudnnDeriveBNTensorDescriptor(derived_param_desc, tensor_desc, bn_op);
+
+    float alpha = 1.0, beta = 0.0;
+    gpu::primitive* batchnorm;
+    switch (direction)
+    {
+    case Prop::Inference:
+    {
+        batchnorm = new gpu::primitive{[=](void** inputs, void** outputs) {
```
rkimballn1: Not sure that you want [=] as that makes copies and then you take references of those copies. Not sure if that is even kosher. Use [&] instead
and read this https://www.cprogramming.com/c++11/c++11-lambda-closures.html

src/ngraph/runtime/gpu/gpu_emitter.cpp
rkimballn1: At some point we should refactor the GPU backend to more align with the new backend API.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1412,6 +1412,102 @@ cudnnSetOpTensorDescriptor(opTensorDesc,
...
+
+                CUDNNEmitter::Prop dir;
+                if (batchnorm->get_training_flag() && args.size() == 3)
+                {
+                    // forward train
```
rkimballn1: I'm all for comments, but does this comment add anything? If `dir` were renamed `direction` then it would be even more clear. The best code has no comments in it because the intent is clear.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
csullivan: I believe [=] here is actually quite important. This lambda is capturing by value because the references in this scope will have long since expired when this closure is executed (during graph execution). Capturing by reference would result in undefined behavior as the local variables, such as the tensor_shape, no longer exists on the stack. Thus it is better to copy by value. Additionally [=] only copies those items which are needed by the closure, it is essentially an easier way of doing a std::bind, so that this function call can have consistent signature `void(void**,void**)` with all of the other arguments capture by value at emission time.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
csullivan: Good suggestions, updated in 582abbadddf9b9bed9e115ea749b3323c9f1dfda.

Previously, allocation and deletion of primitives were occuring in seperate
translation units with raw c pointers. Because of this, it was not
clear that these were being freed appropriately, nor did it indicate
ownership of the pointers.

In the above commit these raw pointers have been converted over to
std::unique_ptr so that the construction/destruction is managed
automatically. Furthermore, GPUPrimitiveEmitter::insert now only
takes an r-value reference, requiring move-semantics to indicate
that when inserting a primitive, the GPUPrimitiveEmitter takes
ownership of the pointer.

All instances of primitive creation have been modified.

src/ngraph/runtime/gpu/gpu_emitter.cpp
csullivan: Completely agreed, that is what I was going for with the enum, the comments were left over from before. I also have renamed dir->direction as it is better that way.

PR #894, opened 2018-04-20
fix convolution bug by retain primary context for device

fengleitian:


fengleitian:
Good point @csullivan , added destructor with release.

PR #898, opened 2018-04-20
add enable gpu convolution tests flag in py

fengleitian:


fengleitian:
@rkimballn1 good point, let me change it.

fengleitian:
updated with a skip list, just add the backend name that need to skip in the list. it will create a string to skip the test.

fengleitian:
@diyessi add a flag to make comments.
`"TODO: Need to implement this op for GPU later;"`

rkimballn1:
TESTNOW

PR #900, opened 2018-04-23
MaxPoolWithIndices

Krovatkin:


src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -3006,6 +3046,49 @@ namespace ngraph
                 }
             }
 
+            template <>
+            void CPU_Emitter::EMITTER_DECL(ngraph::op::MaxPoolWithIndicesBackprop)
+            {
+                auto mpb = static_cast<const ngraph::op::MaxPoolWithIndicesBackprop*>(node);
+
+                auto delta_shape = args[1].get_shape();
+                auto out_shape = out[0].get_shape();
```
jbobba: unused variables.

src/ngraph/runtime/cpu/op/max_pool_with_indices.cpp
```diff
@@ -0,0 +1,350 @@
...
+    , m_window_movement_strides(window_movement_strides)
+    , m_padding_below(padding_below)
+    , m_padding_above(padding_above)
+{
+    auto& arg_shape = get_input_shape(0);
+
+    //
+    // Make sure arg: NCDi for some Di of rank>0, N != 0, C != 0.
+    //
+    if (arg_shape.size() < 3)
```
jbobba: Can we do something similar to ConvolutionRelu. Create a Maxpool op and that would do all this checking and use that to create a MaxPoolWithIndices.

src/ngraph/runtime/cpu/op/max_pool_with_indices.cpp
```diff
@@ -0,0 +1,350 @@
...
+    //
+    // Construct result shape: NCDo.
+    //
+    Shape result_shape(1 + 1 + spatial_dimension_count);
+    result_shape[0] = batch_size;
+    result_shape[1] = channel_count;
+    copy(output_item_shape.begin(), output_item_shape.end(), result_shape.begin() + 2);
+
+    add_output(get_input_element_type(0), result_shape);
+    add_output(element::i32, result_shape);
```
jbobba: We will need a defintion/comment about the value of indices. Get the definition from mkldnn folks. Depending on that we can also define the element type.

src/ngraph/runtime/cpu/op/max_pool_with_indices.hpp
```diff
@@ -0,0 +1,113 @@
...
+        /// 2. <i>(the window movement strides, optional)</i> a vector of positive integers \f$(s_1,\dots,s_n)\f$.
+        ///
+        /// The output has the shape \f$(N,C,d'_1,\dots,d'_n)\f$, where \f$d'_n = \lceil \frac{d_i - w_i + 1}{s_i} \rceil\f$.
+        ///
+        /// Given an input data batch tensor \f$T_\textit{in}\f$, the output tensor is defined by the equation
+        ///
+        /// \f[
+        ///      T_\textit{out}[a,c,i_1,\dots,i_n] = \max_{j_1 = s_1 i_1, \dots, j_n = s_n i_n}^{j_1 = s_1 i_1 + w_1 - 1, \dots, j_n = s_n i_n + w_n - 1} (T_\textit{in}[a,c,j_1,\dots,j_n])
+        /// \f]
+        ///
```
jbobba: Remove and point to documentation in MaxPool op. Dont want to duplicate the comments.

src/ngraph/runtime/cpu/pass/cpu_fusion_bprop.cpp
```diff
@@ -0,0 +1,147 @@
+/*******************************************************************************
```
jbobba: ngraph/runtime/cpu/pass/workspace_insertion.cpp? Probably ask around if folks have better name.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -765,6 +766,79 @@ namespace ngraph
...
+                        node =
+                            insert_input_conversions(external_function, node, prim_input_formats);
+                        set_output_layouts(node, prim_output_formats);
+                    }
+                    else
+                    {
+                        set_default_layouts(external_function, node);
+                    }
+                }
+
```
jbobba: Reuse code. Look at ConvolutionLayout.

Krovatkin:
> This is also going to need a max_pool_with_indices .rst file. I will be making a PR with max_pool.rst soon.

This is a backend-specific/MKLDNN op. Are we documenting those now?

diyessi:
Then you just need something useful to an IA transformer person in the .hpp file.

jbobba:
Agree with @diyessi  My main point is to not duplicate

src/ngraph/runtime/cpu/op/max_pool_with_indices.cpp
Krovatkin: MKLDNN uses either `u8` or `s32` to encode indices , [source](https://github.com/intel/mkl-dnn/blob/2df65197f1c841473144a6fa7e3118628baac773/src/cpu/cpu_pooling_pd.hpp#L35) I'd suggest we could always use `i32` until there's some evidence that we need to optimize for space or to reduce copying.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
Krovatkin: Will do

src/ngraph/runtime/cpu/op/max_pool_with_indices.cpp
Krovatkin: This is actually a **preferred** way as per @aprocter and @diyessi . We kinda cut a few corners for `ConvolutionRelu`. I'd rather change `ConvolutionRelu` to include more checks.

src/ngraph/runtime/cpu/cpu_emitter.cpp
Krovatkin: Will fix.

src/ngraph/runtime/cpu/pass/cpu_fusion_bprop.cpp
Krovatkin: *Loud thudding sounds of Nick's head banging against a keyboard*

src/ngraph/runtime/cpu/op/max_pool_with_indices.hpp
Krovatkin: Will do

src/ngraph/runtime/cpu/op/max_pool_with_indices.cpp
jbobba: i meant this: https://github.com/intel/mkl-dnn/blob/2df65197f1c841473144a6fa7e3118628baac773/src/cpu/ref_pooling.cpp#L92 
'kh*KW + kw'

TF MaxPoolwithArgMax has a different definition of ArgMax.

PR #902, opened 2018-04-23
add cudnn include path to compiler.cpp

fengleitian:
There was issue in generated code compiling
```
[ RUN      ] GPU.abs
code.cpp:5:14: fatal error: 'cudnn.h' file not found
```
Add cudnn header path to compile.cpp to fix it.


fengleitian:
@diyessi you are right, CUDA_HOME is usually where cuda is located, usually `/usr/local/cuda/`, but for some reason, in some system, cudnn include is actually in `/usr/include` instead of `$CUDA_HOME/include`. the  `CUDNN_INCLUDE_DIRS` is searched from system from `FindCUDNN.cmake`.

diyessi:
So shouldn't `CUDA_INCLUDE_DIRS` be set from `${CUDA_HOME}/include`?

fengleitian:
@diyessi No, CUDA and CUDNN are separate packages. It's not necessary to be in the same folder(although the CUDNN installing guideline suggest to copy `cudnn.h` to  `CUDA_HOME/include`)

diyessi:
I'm referencing the lines right above the ones in this PR.

fengleitian:
Here is the reference:
https://docs.nvidia.com/deeplearning/sdk/cudnn-install/index.html#install-linux

fengleitian:
@diyessi sorry, you write `CUDA_INCLUDE_DIRS`, yes, it is. Here we are missing `CUDNN_INCLUDE_DIRS`

fengleitian:
`CUDNN_INCLUDE_DIRS` from: https://github.com/NervanaSystems/ngraph/blob/master/cmake/Modules/FindCUDNN.cmake
`CUDA_INCLUDE_DIRS` from: 
https://cmake.org/cmake/help/v3.0/module/FindCUDA.html

fengleitian:
TESTNOW

fengleitian:
TESTNOW

PR #905, opened 2018-04-24
Fix ngraph ONNX install directions

indie:
Previous instructions to install nGraph from scratch did not work because there is not a build_python3 script any more.


doc/sphinx/source/howto/import.rst
```diff
@@ -45,16 +45,28 @@ everything at once; if an `ngraph_dist` is already installed on your system,
...
-      # apt update
-      # apt install python3 python3-pip python3-dev
-      # apt install build-essential cmake curl clang-3.9 git zlib1g zlib1g-dev libtinfo-dev
+      $ apt update
+      $ apt install python3 python3-pip python3-dev
+      $ apt install build-essential cmake curl clang-3.9 git zlib1g zlib1g-dev libtinfo-dev
       $ git clone https://github.com/NervanaSystems/ngraph.git
+      $ cd ngraph && mkdir build
+      $ cd build && cmake ../ -DNGRAPH_USE_PREBUILT_LLVM=TRUE
+      $ make -j 8 # no need to do this step if you are running a VM
```
diyessi: -j 8 would kill my macbook.
I don't understand why running a vm removes the need for make.

rkimballn1: Life would really be better if we removed reference to the -j option of make. It works perfectly if you don't add it. Working slowly is better than failing quickly. Savvy users will add it themselves. If a non-savvy user add the -j option they won't know what to do when there is a problem.

indie: @diyessi  This was lifted from Michal's instructions on the `ngraph-onnx` repo. 

https://github.com/NervanaSystems/ngraph-onnx

I added the comment about "not necessary if you are using a VM" because it killed the build on my macbook (Ubuntu VM).

diyessi:
TESTNOW

PR #906, opened 2018-04-24
Added the capability of directly compiling cuda and ptx to CMake.

csullivan:
With these CMAKE changes we add support for compiling CUDA with NVCC. This compliments the current RT compilation via NVRTC.


csullivan:
TESTNOW

csullivan:
I have done so successfully with my local install of gcc. Not sure about clang. I will try it with clang and report back.

csullivan:
Good catch, clang-GPU hits the following error:
```
nvcc fatal   : The version ('30901') of the host compiler ('clang') is not supported

```

fengleitian:
I see somewhere clang is going to support it. but not sure which version.

csullivan:
PR #1009 adds CUDA 9.1 support while retaining backward compatibility with CUDA 8. Once the CI and dev nodes are upgraded, we can merge this in.

csullivan:
Finally resurrecting this one to support direct NVCC compilation.

csullivan:
Failing CI due to `-Wreserved-id-macro` conflict with `cuda_runtime.h`
```
In file included from <built-in>:1:
/usr/local/cuda/include/cuda_runtime.h:51:9: error: macro name is a reserved identifier [-Werror,-Wreserved-id-macro]
#define __CUDA_RUNTIME_H__
        ^
```

PR #908, opened 2018-04-25
Replace with Broadcast if exists in Algebraic Simplifier

Krovatkin:


Krovatkin:
@diyessi I tried my **eloquent** best.

PR #913, opened 2018-04-26
Adding support for mkldnn convolution+bias+relu kernel

jbobba:


src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -121,10 +121,48 @@ namespace ngraph
                 void CPUAssignment::ASSIGN_DECL(ngraph::op::ConvolutionRelu)
                 {
                     auto convolution = static_cast<op::ConvolutionRelu*>(node);
-                    auto op_annotations =
-                        std::make_shared<ngraph::runtime::cpu::CPUOpAnnotations>();
-                    op_annotations->set_mkldnn_op(true);
-                    convolution->set_op_annotations(op_annotations);
+
+                    auto arg0_rank = node->get_input_shape(0).size();
```
Krovatkin: the checks seem to be same for `ConvolutionRelu` and `ConvolutionBiasRelu`. Could we refactor the common logic into a helper?

src/ngraph/runtime/cpu/op/conv_relu.hpp
```diff
@@ -45,15 +54,39 @@ namespace ngraph
...
+                                const std::shared_ptr<Node>& filters,
+                                const std::shared_ptr<Node>& bias,
+                                const Strides& window_movement_strides,
+                                const Strides& window_dilation_strides,
+                                const CoordinateDiff& padding_below,
+                                const CoordinateDiff& padding_above,
+                                const Strides& data_dilation_strides);
+
+            const Strides& get_window_movement_strides() const { return m_window_movement_strides; }
+            const Strides& get_window_dilation_strides() const { return m_window_dilation_strides; }
```
Krovatkin: do we even need dilation if we ignore any convolution whose dilation strides aren't equal to 1?

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -377,6 +377,25 @@ namespace ngraph
                     }
                 }
 
+                template <>
+                void CPULayout::LAYOUT_DECL(ngraph::op::ConvolutionBiasRelu)
+                {
```
Krovatkin: I wonder if this logic is similar to `ConvolutionRelu` and could be refactored into a helper?

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -2511,6 +2511,11 @@ namespace ngraph
                     auto data_format = mkldnn_utils::get_input_mkldnn_format(node, 0);
                     auto weights_format = mkldnn_utils::get_input_mkldnn_format(node, 1);
                     auto bias_format = mkldnn_utils::get_input_mkldnn_format(node, 2);
+                    // HACK to help MKLDNN pick the right implementation
+                    if (weights_format == mkldnn::memory::format::nchw)
```
Krovatkin: should we just move this hack into `get_input_mkldnn_format` which would apply the  hack if op is a `Convolution`?

PR #914, opened 2018-04-26
Add Travis testing on CentOS

postrational:


postrational:
Compilation time is too long to make this practical.

postrational:
Compilation without the CPU backend is fast enough to build on Travis.

PR #915, opened 2018-04-26
Bob/new int

rkimballn1:


src/ngraph/runtime/interpreter/int_backend.cpp
```diff
@@ -15,97 +15,325 @@
...
+    return make_shared<runtime::HostTensorView>(type, shape, memory_pointer, "external");
 }
 
-bool runtime::interpreter::INT_Backend::compile(shared_ptr<Function> func)
+bool runtime::interpreter::INTBackend::compile(shared_ptr<Function> function)
 {
-    FunctionInstance& instance = m_function_map[func];
-    if (instance.m_external_function == nullptr)
+    FunctionInstance& instance = m_function_map[function];
+    if (!instance.m_is_compiled)
```
adstraw: I like the old check for nullptr.  let's just kill the is_compiled flag and check for nullptr here.

PR #920, opened 2018-04-27
Reverse Sequence

Krovatkin:


Krovatkin:
@diyessi @mbrookhart I'm thinking do adding the autodiff logic in a separate PR, while it seems pretty simple, I'll need to figure out how to do bprop in TF and adapt my tests.

diyessi:
The autodiff is just going to be a sequence reverse on the deltas.

src/ngraph/op/reverse_sequence.hpp
```diff
@@ -0,0 +1,54 @@
...
+        class ReverseSequence : public util::RequiresTensorViewArgs
+        {
+        public:
+            /// \brief Constructs an arcsin operation.
+            ///
+            /// \param arg Node that produces the input tensor.
+            ReverseSequence(const std::shared_ptr<Node> arg,
+                            size_t batch_axis,
+                            size_t seq_axis,
+                            const Shape& seq_lengths);
```
mbrookhart: the sequence lengths needs to be a Node, we chatted about it in person.

diyessi: And an attribute the specifies the axis that is lengthed, and the axis that indexes the node length.

Krovatkin:
@mbrookhart not yet :-) i don't hook up this kernel to interpreter yet :-)

PR #925, opened 2018-04-27
Auto generate version number and apply it to install dir and libngraph.so

rkimballn1:
Creates a version number based on tags and the hash of the current commit.
Requirement is that every build has a unique version number.
1) If the current commit is tagged with a version label (e.g. v1.2.3) then that version label is used as the release version, so the version would be `v1.2.3`.
2) If the current commit is not tagged then the version label is created from the most recent tag plus the first seven digits of the current commit hash. With this you might get a version like `v1.2.3+1234567` which tells you that it is based off of v1.2.3 but is not the release v1.2.3.

This version number is applied to the libngraph.so so you get file names like
```
libngraph.so -> libngraph.so.0.3
libngraph.so.0.3 -> libngraph.so.v0.3.0+191f4d6
libngraph.so.v0.3.0+191f4d6
```
This should allow users of the ngraph library so request specific versions of the ngraph API

The version number is also stored in a text file VERSION in the root of the install directory

The install directory now looks like this
```
-rw-r--r-- 1 rkimball rkimball   1842723 Apr 22 08:35 libiomp5.so
lrwxrwxrwx 1 rkimball rkimball        14 Apr 27 10:32 libmkldnn.so -> libmkldnn.so.0
lrwxrwxrwx 1 rkimball rkimball        19 Apr 27 10:32 libmkldnn.so.0 -> libmkldnn.so.0.13.0
-rw-r--r-- 1 rkimball rkimball   6508592 Apr 22 08:37 libmkldnn.so.0.13.0
-rw-r--r-- 1 rkimball rkimball 152275293 Apr 22 08:35 libmklml_intel.so
lrwxrwxrwx 1 rkimball rkimball        16 Apr 27 10:32 libngraph.so -> libngraph.so.0.3
lrwxrwxrwx 1 rkimball rkimball        27 Apr 27 10:32 libngraph.so.0.3 -> libngraph.so.v0.3.0+191f4d6
-rw-r--r-- 1 rkimball rkimball  84362640 Apr 27 10:31 libngraph.so.v0.3.0+191f4d6
-rw-r--r-- 1 rkimball rkimball        26 Apr 22 08:33 libtbb_debug.so
-rw-r--r-- 1 rkimball rkimball   2075016 Apr 22 08:33 libtbb_debug.so.2
-rw-r--r-- 1 rkimball rkimball        20 Apr 22 08:33 libtbb.so
-rw-r--r-- 1 rkimball rkimball   1867232 Apr 22 08:33 libtbb.so.2
```


raramer01:
To be more precise the travis ci environment clones the repo with the --depth option which causes the the routines in the new git_tags.cmake module to fail...the Jenkins ci environment, aka ngraph-unittest job uses a refspec to retrieve git refs which specifies the local ngraph repo, specifically '+refs/heads/*:refs/remotes/ngraph/*' which also causes the routines in git_tags.cmake to fail. Interestingly enough the python ci environment (the Jenkins testing-ngraph-python-merge job) which uses the refspec '+refs/pull/*:refs/remotes/origin/pr/*' to retrieve git refs does not fail...

raramer01:
TESTNOW

rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

DawnStone:
TESTNOW

rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

PR #926, opened 2018-04-27
Backprop documentation got borked with new code added.

indie:
Also the ONNX how to import file was tested and it was discovered
that the wheel was actually output to a different directory than
the one specified.  Also added some more useful info about venv and the onnx load process via command line


rkimballn1:
I do enjoy seeing  the word `borked`. Thank you.

PR #930, opened 2018-04-30
Do math in the same type as an underlying constant

Krovatkin:


src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -141,6 +141,36 @@ static size_t reduction_shape_size(const AxisSet& axes, const Shape& shape)
...
+        return multiply_by<float>(cnst->get_element_type(), multiplier, cnst);
+    }
+    else if (cnst->get_element_type() == element::f64)
+    {
+        return multiply_by<double>(cnst->get_element_type(), multiplier, cnst);
+    }
+
+    return nullptr;
+}
+
```
diyessi: It would be better to have `multiply_by` be a method on element type.

PR #933, opened 2018-04-30
Rename initialize_const_values_to_ops and const_values_to_ops to initialize_ops_to_simplifiers and ops_to_simplifiers

Krovatkin:


src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -241,8 +241,8 @@ bool ngraph::pass::AlgebraicSimplification::run_on_function(std::shared_ptr<ngra
         }
 
         const Node& node = *n;
-        auto eh = ops_to_const_values.find(TI(node));
-        if (eh == ops_to_const_values.end())
+        auto eh = ops_to_simplifiers.find(TI(node));
```
diyessi: Does this want to be `auto&`?

Krovatkin: `const Node&` -> `auto&`?

diyessi: No, `auto& eh = ...`
I'm not sure what the find is returning.

Krovatkin: @diyessi could you please approve this one when you have a moment?

PR #936, opened 2018-05-01
Doc through relu, use mathtt instead of texttt in equations.

diyessi:


doc/sphinx/source/ops/parameter.rst
```diff
@@ -0,0 +1,51 @@
+.. parameter.rst
+
+#########
+Parameter
+#########
+
+.. code-block: cpp
+
+   Paramerer // A function parameter.
```
aprocter: Typo (`Paramerer`)

doc/sphinx/source/ops/parameter.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Description
+===========
+
+Parameters are nodes that represent the arguments that will be passed to user-defined functions.  Function creation requires a sequence of parameters.  Basic graph operations do not need parameters attached to a function.
```
aprocter: > Basic graph operations do not need parameters attached to a function.

Not sure what this means. (Is this just a holdover from our existing docs?)

doc/sphinx/source/ops/parameter.rst
diyessi: Got it.

doc/sphinx/source/ops/parameter.rst
diyessi: Holdover that is no longer with us.

PR #937, opened 2018-05-01
add is_mkldnn_format for cpu layout descriptor

ashokei:
add is_mkldnn_format for cpu layout descriptor. This will allow frameworks to optimize tensor view handling.


jbobba:
@ashokei can we close this PR?

PR #938, opened 2018-05-01
Force backends to register by calling dlopen

rkimballn1:
If a user requests a backend like this `Backend::create("IE")` and that backend has not been registered yet then try opening it via dlopen which will force it to register.
We try the following name combinations for the name of the shared library:
```
libIE.so
libIE_backend.so
libie.so
libie_backend.so
```
If any of those work then the backend will register itself and all is well.

### DIFF
--- a/src/ngraph/runtime/backend.cpp
+++ b/src/ngraph/runtime/backend.cpp
@@ -53,6 +53,11 @@ void* runtime::Backend::open_shared_library(const string& type)
         handle = dlopen(name.c_str(), RTLD_NOW);
         if (handle)
         {
+            function<void()> create = reinterpret_cast<void (*)()>(dlsym(handle, "create_backend"));
+            if (create)
+            {
+                create();
+            }
             break;
         }
     }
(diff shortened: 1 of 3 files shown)


src/ngraph/runtime/backend.cpp
```diff
@@ -39,12 +40,41 @@ runtime::Backend::~Backend()
 {
 }
 
+void* runtime::Backend::open_shared_library(const string& type)
+{
+    void* handle = nullptr;
+    vector<string> names;
+    names.push_back("lib" + type + ".so");
```
diyessi: Why do we need all this DWIM for names? Can't we just define what the name needs to look like and enforce it?

rkimballn1: Sure. What should we define it to be?

rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

PR #941, opened 2018-05-02
Make temp memory pools static to avoid memory allocation overheads

jbobba:


rkimballn1:
TESTNOW

jbobba:
This was one of the reasons we had to do a custom MKL allocator in TF. Should have caught it earlier. Hoping it doesn't introduce a leak though.

jbobba:
TESTNOW

jbobba:
TESTNOW

PR #943, opened 2018-05-03
Clean up is_constant is_input is_output in Tensor

yxlao:
- `is_persistent` is always `false` in our current implementation, so it can be removed
- `is_constant` `is_input` and `is_output` is only consumed in `liveness.cpp`, so we could compute them there directly
- The general idea is that, a node should have info about whether its output tensor(s) is(are) persistent or not, which is related to the node type. The tensors shouldn't need to have that info


src/ngraph/pass/memory_visualize.cpp
```diff
@@ -246,25 +243,17 @@ int pass::MemoryVisualize::compute_op_weight(const shared_ptr<Node> exop)
     int mass = 0;
     // for input_decl in exop.input_decls:
```
ransford2011: This looks like Python++.  I'm not seeing the value in keeping it around.  This should be removed.

yxlao:
TESTNOW

yxlao:
TESTNOW

yxlao:
TESTNOW

DawnStone:
TESTNOW

DawnStone:
TESTNOW

DawnStone:
TESTNOW

yxlao:
@Krovatkin fixed

PR #945, opened 2018-05-03
Computation reuse

jbobba:


src/ngraph/op/parameter.cpp
```diff
@@ -21,8 +21,11 @@
 using namespace std;
 using namespace ngraph;
 
-op::Parameter::Parameter(const element::Type& element_type, const Shape& shape)
+op::Parameter::Parameter(const element::Type& element_type,
+                         const Shape& shape,
+                         const bool memoizable)
     : Op("Parameter", {})
+    , m_memoizable(memoizable)
```
diyessi: Should be on `descriptor::Output`.
For example, in drop out, the value used for a weight is scaled by `p`, so if you update the weight, you would want to scale that by `p` and it would be the result of that multiply that would be going into the dot.

jbobba: This is a cosmetic flag and the only reason I have it there is to help with nbench (see Line 225 in benchmark.cpp). We dont do any optimizations in the graph based on this information currently. I can move it to output but I dont see how that is going to  help drop out. All the control information originates from runtime tensors and the graph data flow.

diyessi: In that case, just add it to parameter.rst

jbobba: will do.

jbobba:
[WIP] until some tests show up here.

src/ngraph/runtime/cpu/cpu_runtime_context.hpp
```diff
@@ -38,6 +38,7 @@ namespace ngraph
             struct CPURuntimeContext
             {
                 int64_t* op_durations;
+                bool* p_en;
```
rkimballn1: I'm all for short variable names, but I have no idea what this variable is or does. Can you name it more descriptively.

src/ngraph/pass/memory_layout.hpp
```diff
@@ -35,11 +35,12 @@ namespace ngraph
 class ngraph::pass::MemoryLayout : public FunctionPass
 {
 public:
-    MemoryLayout(size_t alignment = 1);
+    MemoryLayout(size_t alignment = 1, bool free_temps = true);
```
rkimballn1: A description of what free_temps does would be good. How about naming it something like `disable_memory_sharing` and have it default to false. If you named it disable_memory_sharing then I would not really need a comment.

jbobba:
Sample generated code with control flags: 
p_en -> parameter enable (inputs)
t_en -> tensor enable (internal tensors)

// Convolution_131(ConvertLayout_129_0, ConvertLayout_130_0, Convolution_131_0)
105     if (Function_0_init  || t_en[3] || t_en[4]) {
106         cpu::mkldnn_utils::set_memory_ptr(ctx, 22, ((float*)(pool_base_ptr + 4096)));
107         cpu::mkldnn_utils::set_memory_ptr(ctx, 23, ((float*)(pool_base_ptr + 0)));
108         cpu::mkldnn_utils::set_memory_ptr(ctx, 24, ((float*)(pool_base_ptr + 12288)));
109         cpu::mkldnn_utils::mkldnn_invoke_primitive(ctx, 25);
110         t_en[5] = true;
111     } else {
112         t_en[5] = false;
113     }
114 
115     // AvgPool_3(Convolution_131_0, AvgPool_3_0)
116     if (Function_0_init  || t_en[5]) {
117         cpu::mkldnn_utils::set_memory_ptr(ctx, 26, ((float*)(pool_base_ptr + 12288)));
118         cpu::mkldnn_utils::set_memory_ptr(ctx, 27, ((float*)(pool_base_ptr + 16384)));
119         cpu::mkldnn_utils::mkldnn_invoke_primitive(ctx, 28);
120         t_en[6] = true;
121     } else {
122         t_en[6] = false;
123     }

PR #948, opened 2018-05-03
Slice Concat Elimination

Krovatkin:
@diyessi This should simplify @pruthviIntel 's pass to fuse multiple layers.


src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -55,6 +59,71 @@ static std::shared_ptr<pattern::op::Label>
...
+                return false;
+            }
+
+            it = it->get_argument(0);
+        }
+    }
+
+    if (!std::dynamic_pointer_cast<op::GetOutputElement>(goe))
+    {
+        NGRAPH_DEBUG << goe->get_name() << " isn't GOE ";
```
Krovatkin: @diyessi 
This isn't specific to GOE... we should be able to do it for any op but I thought it might be better if we make this opt more targeted?

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -55,6 +59,82 @@ static std::shared_ptr<pattern::op::Label>
...
+
+    auto slice =
+        std::make_shared<op::Slice>(lgoe, Coordinate{0, 0}, Coordinate{2, 1}, Strides{1, 1});
+
+    auto reshape_pred = [](std::shared_ptr<Node> r) {
+        return std::dynamic_pointer_cast<op::Reshape>(r) != nullptr;
+    };
+
+    auto skip_reshape = std::make_shared<pattern::op::Skip>(slice, reshape_pred);
+
```
pruthviIntel: Can we have this without reshape predicate?, it will cause a layout exception if we have case like this  `GOE(32, 200) -> Slice( 16, 200) -> Reshape(16, 1, 200) -> Result(16, 1, 200)` this pass, will reduce it to `GOE(32, 200) -> Result(16, 1, 200)`. which will cause layout exception.

Krovatkin:
@pruthviIntel if the fix looks good to you; please approve!

pruthviIntel:
https://aipg.slack.com/files/U29MD4D5H/FANMW9JGM/pasted_image_at_2018_05_11_02_53_pm.png

pruthviIntel:
I think their is one more case which is failing to capture, can we eliminate  this as well. 
pasted image at 2018_05_11 02_53 pm

aprocter:
:frog: report: CI is failing because of a missing `#include` for the `VisualizeTree` pass in `test/algebraic_simplification.cpp`.

PR #949, opened 2018-05-03
Algebraic Simplification for Product

Krovatkin:


src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -171,13 +184,40 @@ static std::shared_ptr<Node> get_sum_constant(std::shared_ptr<op::Constant> cnst
...
+static std::shared_ptr<Node> get_prod_constant(std::shared_ptr<op::Constant> cnst,
+                                               size_t multiplier)
+{
+    if (cnst->get_element_type() == element::i32)
+    {
+        return pow_by<int>(cnst->get_element_type(), multiplier, cnst);
+    }
+    else if (cnst->get_element_type() == element::i8)
+    {
+        return pow_by<char>(cnst->get_element_type(), multiplier, cnst);
```
diyessi: Use `signed char` since compilers `char` is signed on some compilers and unsigned on others.

Even better would be to put these in a template class so we can say `element::c_type<element::i32>::type` instead of `int`, etc. and have the correct types one place.

Krovatkin: > Even better would be to put these in a template class so we can say element::c_type<element::i32>::type instead of int, etc. and have the correct types one place.

Agreed. If you don't mind I'll do it in a separate PR since it will most likely require changes to more than just this logic.

Krovatkin:
@diyessi i've made the change you requested; please approve!!!! pretty please?

PR #951, opened 2018-05-03
CUDNN and CUDA kernels for AvgPool (forward/backward)

csullivan:
This enables all forward and backward AvgPool tests.

Features:
* CUDNN is used primarily for forward/backward pooling in 2 and 3 dimensions with symmetric padding.
* Cuda-C kernel handles pooling for 1, 2 and 3 dimensions and has support for asymmetric padding.
* Added some helper functions for fast integer division and constant memory access

Limitations:
* Feature pooling -- where values from different channels are pooled together -- is not currently supported
* alpha and beta blending factors in the Cuda-C kernel are present but not currently implemented
* Will need additional work for extension to Nd.

Notes for future:
* The performance could likely be increased here by considering a separate tensor format to increase coalescing (CDHWN) or by reducing the number of blocks and moving the shuffle reduce results of multiple warps into shared memory.


csullivan:
TESTNOW

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -22,11 +22,39 @@
...
 #include "ngraph/runtime/gpu/gpu_runtime_context.hpp"
+#include "ngraph/runtime/gpu/gpu_util.hpp"
 #include "ngraph/runtime/gpu/type_info.hpp"
 #include "ngraph/util.hpp"
 
 using namespace ngraph;
 
+inline namespace enums
+{
+    enum pooling_indices
```
rkimballn1: I think you can replace this enum and vector<int> with a class that has int members N_, C_, ...
This should end up being more clear.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -280,6 +308,257 @@ size_t runtime::gpu::CUDAEmitter::build_1d_max_pool(const GPURuntimeContext* ctx
...
+                                                 const Shape& output_shape,
+                                                 const Shape& window_shape,
+                                                 const Shape& window_stride,
+                                                 const Shape& padding_below,
+                                                 bool include_pad)
+{
+    // assumes NCDHW format
+    auto input_width = input_shape.back();
+    auto output_width = output_shape.back();
+    auto s =
```
rkimballn1: If you are calling a function `get_avgpool_params` then name the variable `avgpool_params` if for no other reason than to make it easy to search for in the code. It took me a while to find it's use because I can't search for `s`.
I think this is not used until way later in the code. Why not move this line down to be closer to where it is ultimately used.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -280,6 +308,257 @@ size_t runtime::gpu::CUDAEmitter::build_1d_max_pool(const GPURuntimeContext* ctx
...
+    int DHW = s[D_] * HW;
+    int CDHW = s[C_] * DHW;
+    int PQ = s[P_] * s[Q_];
+    int MPQ = s[M_] * PQ;
+    int KMPQ = s[K_] * MPQ;
+    int RS = s[R_] * s[S_];
+    int TRS = s[T_] * RS;
+
+    // precompute magic numbers and shifts for fast integer division
+    int magic_N, shift_N;
```
rkimballn1: The coding standard allows for only one declaration per line. Don't use commas here. or in the following code.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -314,3 +593,57 @@ void runtime::gpu::CUDAEmitter::print_tensor_from_gpu(codegen::CodeWriter& write
...
+    // div64: fast integer division via magic multiplication and shifting
+    // if value is a power of 2, magic will be 1 and only shifting
+    // is required (predicate p in div64)
+    // load: helper to load from constant memory for fast access
+    std::stringstream ss;
+    ss << R"(
+__device__ __forceinline__ int div64(int value, int magic, int shift)
+{
+    int result;
+    asm("{\n\t"
```
rkimballn1: Emitting code that emits code...mind blown

src/ngraph/runtime/gpu/cuda_emitter.cpp
csullivan: Agreed, I think it is more clear this way, I replaced the inline enum with a struct as you suggest. However, the pooling shape parameters are used in the function hash of the kernel to determine if the kernel has been used previously. This is why this line appears early on. I also updated the name as suggested.

src/ngraph/runtime/gpu/cuda_emitter.cpp
csullivan: Fixed.

src/ngraph/runtime/gpu/cuda_emitter.cpp
csullivan: Replaced with a struct and an ostream operator overload for printing it.

csullivan:
Thanks @rkimballn1, your comments have been implemented

src/ngraph/runtime/gpu/gpu_util.cpp
```diff
@@ -77,3 +77,101 @@ void runtime::gpu::cuda_memset(void* dst, int value, size_t buffer_size)
 {
     cudaMemset(dst, value, buffer_size);
 }
+
+namespace
+{
+    uint64_t _powU64(uint64_t base, uint64_t exp)
```
fengleitian: can this be shared for all platform?

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -280,6 +314,295 @@ size_t runtime::gpu::CUDAEmitter::build_1d_max_pool(const GPURuntimeContext* ctx
...
+
+                writer << "rcp_pool_size = 1.0f / (float)pool_size;\n";
+                // reduce pooling window within warp.
+                // this could be improved by calculating the
+                // pooling windows each thread can partake in to
+                // reduce loads and increase coalescing. in that case,
+                // multiple warps per block would be required and the
+                // warp reduced sums would need to be accumulated in
+                // shared memory
+                writer << "for (int i = 16; i > 0; i >>= 1)\n";
```
fengleitian: what if possible warp < 32.

src/ngraph/runtime/gpu/gpu_util.cpp
rkimballn1: any name starting with an underscore is reserved in c++ so don't start functions or variable names with an underscore.
https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: Looks like it should be fine since all cuda device now using warp 32.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1560,6 +1560,132 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
...
+            {
+                // assumes NC{d1,d2,...} format
+                auto avg_pool = static_cast<const ngraph::op::AvgPool*>(node);
+                writer.block_begin("  // " + node->get_name());
+                {
+                    auto& input_shape = args[0].get_shape();
+                    auto& result_shape = out[0].get_shape();
+                    auto padding_below = avg_pool->get_padding_below();
+                    auto padding_above = avg_pool->get_padding_above();
+                    if (input_shape.size() < 3)
```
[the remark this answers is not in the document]
csullivan: @fengleitian, you are correct, I was being overly cautious. I've removed the check as it is a duplicate of (https://github.com/NervanaSystems/ngraph/blob/cs%2Fgpu_avgpool/src/ngraph/op/avg_pool.cpp#L41)

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -280,6 +314,295 @@ size_t runtime::gpu::CUDAEmitter::build_1d_max_pool(const GPURuntimeContext* ctx
...
+    const Shape& in, const Shape& out, const Shape& window, const Shape& strides, const Shape& pad)
+{
+    pooling_op_shape shape;
+    shape.N = static_cast<int>(in[0]);
+    shape.C = static_cast<int>(in[1]);
+    shape.K = shape.C; // pooling feature maps is
+    shape.J = shape.C; // not currently supported
+    if (in.size() == 3)
+    {
+        shape.D = 1;
```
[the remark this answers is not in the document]
csullivan: You are correct, this would need some additional engineering to move to N dimensions. However, from what I can tell, the frameworks are typically using only 2 or 3 dimensions. For example, Tensorflow only supports 2/3d for average pooling 

* https://www.tensorflow.org/api_docs/python/tf/nn/avg_pool
* https://www.tensorflow.org/api_docs/python/tf/nn/avg_pool3d

src/ngraph/runtime/gpu/gpu_util.cpp
csullivan: Yeah, I originally put them in an anonymous namespace for that reason but I've removed the prefix just in case.

csullivan: I \*believe\* so. Any particular concern with it? We can add additional reviewers if it would help

src/ngraph/runtime/gpu/cuda_emitter.cpp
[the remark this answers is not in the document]
fengleitian: I see. Let's not worry it for now.

src/ngraph/runtime/gpu/gpu_util.cpp
fengleitian: Just think if this will benefit other platform.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -280,6 +314,295 @@ size_t runtime::gpu::CUDAEmitter::build_1d_max_pool(const GPURuntimeContext* ctx
...
+                // warp reduced sums would need to be accumulated in
+                // shared memory
+                writer << "for (int i = 16; i > 0; i >>= 1)\n";
+                writer.block_begin();
+                {
+                    writer << "sum += __shfl_xor(sum,i);\n";
+                }
+                writer.block_end();
+                // write result to output
+                writer << "if (tid == 0)\n";
```
fengleitian: since you only need tid 0 for result, have you try __shfl_down.

csullivan: I may be wrong on this but isn't shuffle_down just: 
```reg(tid) += register(tid+1),```
and so to accumulate all 32 threads I would need to call __shfl_down 32 times.

Since __shfl_xor uses a butterfly merge, each half of the selected threads are added to the other half, and this is done recursively, so only 4 calls to __shfl_xor are needed.

Comment (fengleitian) [reply src/ngraph/runtime/gpu/cuda_emitter.cpp]: so far as I know, shuffle down is similar to shuffle xor, but can only go in one direction, so the result will only be stored in tid = 0; something like.
reg(tid) += reg(tid+n); (n>0)
it will be less work than _shfl_xor,  `n + n/2 + n/4 +..`  vs `n + n + n + ...`

Comment (fengleitian) [reply src/ngraph/runtime/gpu/cuda_emitter.cpp]: https://devblogs.nvidia.com/faster-parallel-reductions-kepler/

Comment (fengleitian) [reply src/ngraph/runtime/gpu/cuda_emitter.cpp]: __shfl_down(reg, 16, 32);
__shfl_down(reg, 8, 16);
__shfl_down(reg, 4, 8);
...

Comment (fengleitian) [reply src/ngraph/runtime/gpu/cuda_emitter.cpp]: for _shfl_xor, it should be the case. 
for my understanding,  _shfl_down should not be. let me check.

Comment (csullivan) [reply src/ngraph/runtime/gpu/cuda_emitter.cpp]: Correction: In shuffle_down the values are not broadcast. They threads out of the range just return their local value.

Ah maybe I see. Exchanging data between part of the warp vs the whole warp should still cost the same number of instructions (down vs xor), but with down, you have branch divergence in the warp, but you also don't have to wait for the data movement from the non participating threads. Since this is within the warp, the data movement is small, so I think it would really need some testing to see if it is actually an optimization, but I see where you are coming from now.

Comment (fengleitian) [reply src/ngraph/runtime/gpu/cuda_emitter.cpp]: performance side. it may or may not be optimization depends on hardware. I'm thinking the  tid 16-31 should be sum = 1; 8-15 should have sum = 2; Looks little either my understanding is wrong or we used it in a wrong way.

Comment (fengleitian) [reply src/ngraph/runtime/gpu/cuda_emitter.cpp]: looks like __shfl_xor and __shfl_down will be deprecated in cuda 9.

Comment (csullivan) [reply src/ngraph/runtime/gpu/cuda_emitter.cpp]: Yeah, so is __ballot, they are replaced with __*_sync functions, and we can make our code backwards compatible with earlier versions via

```
#if defined(CUDA_VERSION) && CUDA_VERSION < 9000
    #define __ballot_sync(mask, predicate) __ballot(predicate)
    #define __shfl_down_sync(mask, val, delta, width) __shfl_down(val, delta, width)
    #define __shfl_xor_sync(mask, val, laneMask, width) __shfl_xor(val, laneMask, width)
#endif
```
once we move to CUDA 9.

fengleitian: yes.

PR #952, opened 2018-05-04
add gpu reverse

fengleitian:


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -971,6 +971,62 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                    kernel::emit_memcpyDtD(writer, out[0], args[0]);
+                }
+                else
+                {
+                    writer << "size_t rank = " << arg_rank << ";\n";
+                    writer << "std::vector<size_t> input_shapes_h = {" << join(arg_shape, "UL,")
+                           << "UL};\n";
+                    writer << "std::vector<size_t> reverse_axes_h = {"
+                           << join(reverse_axes_flag, "UL,") << "UL};\n";
+
```
csullivan: Why not use the new CUDAEmitter instead of the old codegen API? Technically the old API is unsafe as it makes calls directly into nGraph from clang code. 

Additionally, I feel as though the codegen is getting in the way, because, as seen below, it forces us to do expensive cuda memcpys into the device from the JIT code rather than just passing them as arguments to the kernel from nGraph. I was planning to refactor the code that uses this old interface eventually, but am hoping that we can stop using the old API for new ops. Is that OK?

To some extent we can refactor in the second pass when we focus on performance, but it also doesn't make sense for us to have to redo those ops which haven't yet been implemented.

csullivan: Since this is working and already done, I'm not saying op::Reverse should be redone right now, just that if we can I personally think it would be better to use the new API moving forward.

fengleitian: I see what you mean , good point.  Let's do it later.

PR #960, opened 2018-05-04
Aprocter/attribute maps

dkannaiy:


dkannaiy:
Collaborator:
Review for this pull request was created. [Link to review](https://sbox-codecollab.intel.com:443/ui#review:id=77537)

src/ngraph/pass/pretty_print.cpp
```diff
@@ -0,0 +1,90 @@
+/*******************************************************************************
```
rkimballn1: We already have a pretty print pass, it is called dump_sorted. We don't need two passes that do the same thing so merge them together or delete one.

aprocter:
I didn't open this PR.

PR #961, opened 2018-05-04
allow for building in a directory not under the repo

rkimballn1:


rkimballn1:
#963 is a better solution

PR #964, opened 2018-05-04
NGRAPH-1605 Sigmoid multiply fusion

louisfeng:
This is the forward part of the sigmoid (logistic and tanh) fusion with element wise multiply. Compiler reports vectorized loop, and it gives about 2.5X speed up over the unfused ops (measured using unit tests before/after on reasonably large data size).

This is what it would generate:
```
    // SigmoidMultiply_265(Parameter_0_0, Parameter_1_0, SigmoidMultiply_265_0)
    {
        #pragma omp parallel for simd
        for (size_t i=0; i<4; i++)
        {
            float numer_0;
            float denom_0;
            {
                auto ex = exp(((float*)(inputs[0]))[i]);
                numer_0 = ex;
                denom_0 = ex+1;
            }
            float numer_1;
            float denom_1;
            {
                auto ex = exp(2.0*((float*)(inputs[1]))[i]);
                numer_1 = ex-1;
                denom_1 = ex+1;
            }
            ((float*)(outputs[0]))[i] = (numer_0 * numer_1) / (denom_0 * denom_1);
        }
    }
```


jbobba:
Looks reasonable. Can we make this more general, in terms of a sequence of unary ops leading into a binary op? Sigmoid, Multiply could be attributes inside this fused-op. Might be harder to support autodiff on this fused-op though.

src/ngraph/runtime/cpu/op/sigmoid_mul.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+shared_ptr<Node> op::SigmoidMultiply::copy_with_new_args(const NodeVector& new_args) const
+{
+    if (new_args.size() != 2)
+    {
+        throw ngraph_error("SigmoidMultiply incorrect number of new arguments");
+    }
+
+    return make_shared<SigmoidMultiply>(new_args.at(0), new_args.at(1));
```
Krovatkin: `SigmoidMultiply` c-tor seems to expect either `Tanh` or `Sigmoid` nodes, while `new_args` , I think, contains the **children** of `Tanh` and `Sigmoid`. Could you please double check if this does what you expect it to do?

src/ngraph/runtime/cpu/op/sigmoid_mul.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
```
Krovatkin: missing a new line... this will probably fail CI

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -943,3 +945,38 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_relu()
     auto m = std::make_shared<pattern::Matcher>(prelu, callback);
     this->add_matcher(m);
 }
+
+void ngraph::runtime::cpu::pass::CPUFusion::construct_sigmoid_multiply()
+{
+    //construct predicate to match sigmoid and tanh
+    auto sigmoid_pred = [](std::shared_ptr<Node> n) {
+        bool result = (std::dynamic_pointer_cast<op::Sigmoid>(n) != nullptr) ||
```
Krovatkin: in this particular case, it might be cleaner to just return the expression rather than assign it to a result variable.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -943,3 +945,38 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_relu()
...
+
+    //Define a call back that needs to called once the DFG matches the pattern
+    ngraph::pattern::graph_rewrite_callback callback = [sigmoid_1, sigmoid_2](pattern::Matcher& m) {
+        NGRAPH_DEBUG << "In a callback for construct_sigmoid_multiply pattern against "
+                     << m.get_match_root()->get_name();
+        auto pattern_map = m.get_pattern_map();
+
+        if (m.get_match_root()->get_element_type() != element::f32)
+        {
+            NGRAPH_DEBUG << "mpattern = " << m.get_match_root()->get_name()
```
Krovatkin: thank you for debug messages! they help a lot 👍

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -943,3 +945,38 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_relu()
...
+                         << " type is not float!";
+            return false;
+        }
+
+        auto sigmoid_mul_node =
+            std::make_shared<op::SigmoidMultiply>(pattern_map[sigmoid_1], pattern_map[sigmoid_2]);
+        ngraph::replace_node(m.get_match_root(), sigmoid_mul_node);
```
Krovatkin: could we add a message saying that we performing a transformation here?

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -943,3 +945,38 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_relu()
...
+        return true;
+    };
+
+    auto m = std::make_shared<ngraph::pattern::Matcher>(elem_mul, callback);
+    this->add_matcher(m);
+}
```
Krovatkin: missing a new line

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -3488,6 +3489,71 @@ namespace ngraph
...
+                                              numer_0,
+                                              denom_0);
+                std::string input_1_func_string =
+                    generate_sigmoid_mul_func(sigmoid_mul->get_input_2_func_type(),
+                                              args[1].get_name() + "[i]",
+                                              numer_1,
+                                              denom_1);
+
+                writer.block_begin();
+                writer << "#pragma omp parallel for simd\n";
```
Krovatkin: out of curiosity, does this pragma actually help?

test/cpu_fusion.cpp
```diff
@@ -1125,3 +1127,64 @@ TEST(cpu_fusion, weight_fusion)
                   new_convert_layout->get_argument(0)),
               cvt_lt_conv);
 }
+
+TEST(cpu_fusion, sigmoid_multiply_fusion_compute)
+{
+    auto backend = runtime::Backend::create("CPU");
+
+    const size_t w = 2;
```
Krovatkin: `w, h` seem a bit redundant here. `SigmoidMultiply` doesn't seem to care what those dimensions mean.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
louisfeng: Yep, it was a debugging left over. I'll fix that.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
louisfeng: Will fix.

louisfeng:
Added backprop through generate_adjoint() and fusing sigmoid * broadcast.

Krovatkin:
@jbobba 

> Looks reasonable. Can we make this more general, in terms of a sequence of unary ops leading into a binary op? Sigmoid, Multiply could be attributes inside this fused-op. Might be harder to support autodiff on this fused-op though.

I've been thinking about it for quite a while now. I believe we should be able to fuse arbitrary graphs consisting of unary and binary ops into 1-dimensional loops, but it will be a different change rather than an extension of @louisfeng 's PR. If you are okay with me spending time on this, I think we could get this done!

src/ngraph/runtime/cpu/cpu_emitter.cpp
louisfeng: yes it does.

test/cpu_fusion.cpp
louisfeng: fixed.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
louisfeng: actually clang seems to remove any extra new line I add.

louisfeng:
TEST NOW

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -943,3 +945,52 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_relu()
...
+                     << m.get_match_root()->get_name();
+        auto pattern_map = m.get_pattern_map();
+
+        if (m.get_match_root()->get_element_type() != element::f32)
+        {
+            NGRAPH_DEBUG << "mpattern = " << m.get_match_root()->get_name()
+                         << " type is not float!";
+            return false;
+        }
+
```
jbobba: If the sigmoid ops have multiple users, then skip fusion.

louisfeng: good catch, thanks.

louisfeng: @jbobba This is fixed.

louisfeng:
TEST NOW

louisfeng:
TESTNOW

PR #972, opened 2018-05-08
Add CLion files to .gitignore

postrational:


.gitignore
```diff
@@ -63,6 +63,10 @@ nervana_aeon.egg-info/
 *.swo
 tags
 
+# CLion
+.idea/
+cmake-build-debug/
```
rkimballn1: Is this cmake-build-debug directory for clion or is it a build directory. If it is a build directory then it should be put into your global gitignore file, not the local .gitignore.

rkimballn1: The more I think about this the more I think all of these changes belong in your global git ignore file.
https://help.github.com/articles/ignoring-files/

postrational: I'm not sure why we would ever want anyone to commit these files to the repo, but I can also exclude them globally on my end.

iamthebot: Keep in mind you'd also want to exclude `cmake-build-release`, etc.
In general, excluding `.idea/*` and `*build*` is a combo that's worked well in plenty of projects.

PR #979, opened 2018-05-08
Add op::Or and op::And to GPU transformer

csullivan:
Moved elementwise ops implementation into CUDAEmitter and added logical_and and logical_or ops.

### DIFF
--- a/src/ngraph/runtime/gpu/cuda_emitter.cpp
+++ b/src/ngraph/runtime/gpu/cuda_emitter.cpp
@@ -306,8 +306,8 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
 
     // check if the kernel has already been compiled. if so, create
     // a launch primitive for it based on the input tensor shape
-    // but do not recompile the kernel. if not, recompile the
-    // kernel and then create the primitive
+    // but do not recompile the kernel. otherwise, do it all:
+    // recompile the kernel and then create the primitive
     auto compiled_kernel = ctx->compiled_kernel_pool->get(kernel_name.str());
     if (compiled_kernel == nullptr)
     {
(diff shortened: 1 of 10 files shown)


src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -23,9 +23,9 @@ using namespace ngraph;
 void runtime::gpu::CudaKernelBuilder::get_elementwise_op(codegen::CodeWriter& writer,
                                                          const std::string& name,
                                                          const std::string& op,
-                                                         const std::vector<std::string>& data_types,
-                                                         const size_t& num_inputs)
+                                                         const std::vector<std::string>& data_types)
 {
+    auto num_inputs = data_types.size() - 1;
```
fengleitian: need to double check if output size will always be one.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -280,6 +281,79 @@ size_t runtime::gpu::CUDAEmitter::build_1d_max_pool(const GPURuntimeContext* ctx
     return primitive_index;
 }
 
+size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeContext* ctx,
```
fengleitian: what's the `n_to_1` mean here?

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
csullivan: This cannot handle multi-output ew ops. Is there an example of one that we need to support? I can add an assertion just in case.

src/ngraph/runtime/gpu/cuda_emitter.cpp
csullivan: That it handles unary/binary/ternary/.../n input elementwise operations where there is only a single output.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
csullivan: Added the check in GPU_Emitter (https://github.com/NervanaSystems/ngraph/pull/979/files#diff-88bf3ba31e0b08d55a3811bc6932533aR72)

PR #983, opened 2018-05-09
New manifest driven method for disabling backend unit tests

rkimballn1:
The old method for skipping tests, placing SKIP_TEST_FOR macros in the code, does not scale well. Implementors of backends should not be editing ngraph unit tests to disable them.
New method is to use a manifest file which is located in the backend's directory. From cmake you register the backend for test, pointing to the manifest file. Magic.


rkimballn1:
TESTNOW

PR #985, opened 2018-05-10
add gpu reduction

fengleitian:
Here we implemented a general reduce the can take a reduction function.

Because of the limitation from CUDNN and currently nGrpah framework. There are some limitation:

1. The reduction function can only take one op.
2. only add/multipy/min/max op supported for reduction function.


fengleitian:
@csullivan  There are several ways to do it.
1. use cpu
2. have our own kernel to do it and take the emitter the reduction function for the kernel. This will require some change to framework. And also the element-wised multi-op  kernel generator

fengleitian:
TESTNOW

csullivan:
@fengleitian, I see. So in the GPU case we detect each op in the Reduction Function subgraph and either (1) emit a fused cuda kernel for it, or (2) run it through something like the external_function::compile, to compile the subgraph, and then insert it into the larger execution graph. Basically the reverse of op fusion in the latter case; op expansion on the reduction function.

fengleitian:
@csullivan (2) is currently been used for CPU, which is hard to implement on GPU side. My thought is  we need to maybe use pattern match to disable it for GPU.

PR #988, opened 2018-05-10
Refactor CPUWorkspaceInsertion to simplify its use in MxNet

Krovatkin:


test/cpu_fusion.cpp
```diff
@@ -1213,13 +1215,126 @@ TEST(cpu_fusion, backwards_maxpool_with_indices_n4_c1_hw4_2x2_max)
...
+    // create the backward function
+    std::vector<std::shared_ptr<ngraph::op::Parameter>> param_adjoints;
+    for (auto n : adjoints)
+        param_adjoints.push_back(std::dynamic_pointer_cast<ngraph::op::Parameter>(n));
+    back_parameters.insert(back_parameters.begin(), param_adjoints.begin(), param_adjoints.end());
+
+    return {std::make_shared<ngraph::Function>(dYdXs, back_parameters), adjoints};
+}
+
+void OptimizeGraph(std::shared_ptr<ngraph::Function>& f, std::shared_ptr<ngraph::Function> bf)
```
diyessi: `optimize_graph`

Krovatkin:
> Okay except for a name violation.

fixed, resolved conflicts! Ready for a lift-off

test/cpu_fusion.cpp
```diff
@@ -1213,13 +1215,126 @@ TEST(cpu_fusion, backwards_maxpool_with_indices_n4_c1_hw4_2x2_max)
...
+    if (nv_cwi.size() > 0)
+    {
+        NodeVector new_outputs;
+        for (auto r : f->get_results())
+        {
+            new_outputs.push_back(r->get_argument(0));
+        }
+
+        new_outputs.insert(new_outputs.end(), nv_cwi.begin(), nv_cwi.end());
+        f = std::make_shared<ngraph::Function>(new_outputs, f->get_parameters());
```
mbrookhart: I think this should be a new Function, instead of overwriting f. Doesn't matter when reimplementing

Krovatkin: @mbrookhart indeed, this is pretty much copy-pasted from `ngraph-mxnet`. However, this is ngraph's test code and making `ngraph-mxnet` rely/call the *test* code doesn't quite give me warm fuzzy feelings. It precludes us making arbitrary changes to this test, extending it because of an introduced dependency on `ngraph-mxnet`. I'd rather have the two separate.

Krovatkin: @mbrookhart I wouldn't care much until I have to test something in the context of MxNet bridge at which point I'll pull yr changes manually. You don't seem to be updating this logic too often and my need to use it is even less frequent. Also, usually you introduce very minor changes to this particular logic which would be very really to port.

mbrookhart: :) Okay

Krovatkin: :yayfox

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -289,6 +289,9 @@ static const runtime::cpu::OpMap dispatcher{
     {TI(ngraph::op::Or), &runtime::cpu::CPU_Emitter::emit<op::Or>},
 };
 
+const size_t runtime::cpu::CPU_ExternalFunction::CPU_ExternalFunction::s_memory_pool_alignment =
+    4096;
+
```
jbobba: why?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -316,12 +319,15 @@ void runtime::cpu::CPU_ExternalFunction::compile()
 
     ngraph::pass::Manager pass_manager;
 
+    //nv_cwi is required only by some frontends
+    //in which case they should run this pass(CPUWorkspaceInsertion) explicitly
```
jbobba: Comment is not clear. Better to move it into the pass

src/ngraph/runtime/cpu/pass/cpu_workspace_insertion.cpp
```diff
@@ -52,7 +52,9 @@
 #include "ngraph/runtime/cpu/op/max_pool_with_indices.hpp"
 #include "ngraph/runtime/cpu/op/sigmoid.hpp"
 
-void ngraph::runtime::cpu::pass::CPUWorkspaceInsertion::construct_max_pool_with_indices()
+using namespace ngraph;
+
+static std::shared_ptr<pattern::Matcher> create_maxpool_with_indices_matcher()
```
jbobba: Do we need pattern matcher here? Why not just look for MaxPoolBackprop ops?

jbobba: "create_maxpool_backprop_pattern"

src/ngraph/runtime/cpu/pass/cpu_workspace_insertion.cpp
```diff
@@ -66,83 +68,106 @@ void ngraph::runtime::cpu::pass::CPUWorkspaceInsertion::construct_max_pool_with_
...
+    auto m_max_pool_bprop = std::dynamic_pointer_cast<op::MaxPoolBackprop>(m.get_match_root());
 
-        //find the original MaxPool now
-        std::shared_ptr<op::MaxPool> m_max_pool;
-        for (auto u : pattern_map[data]->get_users())
+    if (m_max_pool_bprop->get_shape().size() != 4 ||
+        m_max_pool_bprop->get_window_shape().size() != 2 ||
+        m_max_pool_bprop->get_input_element_type(0) != element::f32)
+    {
+        NGRAPH_DEBUG << "MKLDNN doesn't support inputs of given shape type";
```
jbobba: Expand on the debug statement. "Skipping workspace insertion on maxpool bprop"

src/ngraph/runtime/cpu/pass/cpu_workspace_insertion.cpp
```diff
@@ -66,83 +68,106 @@ void ngraph::runtime::cpu::pass::CPUWorkspaceInsertion::construct_max_pool_with_
...
+
+bool runtime::cpu::pass::CPUWorkspaceInsertion::transform(pattern::Matcher& m)
+{
+    auto data = std::dynamic_pointer_cast<pattern::op::Label>(m.get_pattern()->get_argument(0));
+    auto delta = std::dynamic_pointer_cast<pattern::op::Label>(m.get_pattern()->get_argument(1));
+    NGRAPH_DEBUG << "In a callback for construct_max_pool_with_indices against "
+                 << m.get_match_root()->get_name();
+
+    auto pattern_map = m.get_pattern_map();
+    auto m_max_pool_bprop = std::dynamic_pointer_cast<op::MaxPoolBackprop>(m.get_match_root());
```
jbobba: Style violation "m_" only class members. There are a few others like this.

src/ngraph/runtime/cpu/pass/cpu_workspace_insertion.hpp
```diff
@@ -32,15 +50,18 @@ namespace ngraph
...
+        , m_indices_list(indices_list)
     {
-        construct_max_pool_with_indices();
     }
 
+    virtual bool run_on_function(std::shared_ptr<ngraph::Function> f);
+
 private:
-    void construct_max_pool_with_indices();
+    ngraph::NodeVector& m_indices_list;
```
jbobba: Add comment here about m_indices_list

jbobba: Also make this optional.

src/ngraph/runtime/cpu/cpu_external_function.cpp
Krovatkin: oh boy. Because `register_pass` uses universal references and takes an address of `s_memory_pool_alignment`.  If  `s_memory_pool_alignment` defined in a class header it won't receive static storage which results in a linking error.

src/ngraph/runtime/cpu/pass/cpu_workspace_insertion.cpp
Krovatkin: It's being used here as a parameter object pattern i.e. it groups together multiple arguments to pass into `transform`.

src/ngraph/runtime/cpu/pass/cpu_workspace_insertion.hpp
Krovatkin: Then I will have to turn this into a pointer. This seems to be cleaner at an expense of a little bit of clunkiness in API

src/ngraph/runtime/cpu/pass/cpu_workspace_insertion.cpp
Krovatkin: it's less of a pattern more of a parameter object 😄  this is intentionally named differently from a fusion naming convention.

PR #994, opened 2018-05-11
Fix MNIST MLP model

okhat:
Removed ReLu activation present just before softmax.
(It's not a good idea to have ReLu > Softmax immediately. The test accuracy of the model is now much better.)


diyessi:
Thanks. This also needs a slight doc change. 
Due to the way our CI/merge works, I need to merge this the hard way.

diyessi:
Migrated to #1000

binarybana:
Thanks so much @okhat! Congrats on your first PR btw.

PR #999, opened 2018-05-11
[CS:GPU::Part4] Added op::ReplaceSlice and enabled respective tests.

csullivan:
- [x] [CS:GPU::Part 4] Note: do not merge until all parts[1-4] have been reviewed and approved. After approvals: merge part 4 into 3, then 3 -> 2, 2 -> 1, and finally 1 -> master after CI passes.

This PR implements ReplaceSlice using the coordinate transformation strategy. A thread for each tensor element of the input tensor is chosen and it's position in the source tensor coordinate system is calculated. If it is within the source slice, the source is loaded and written out, otherwise the input tensor is loaded. 
* Relevant tests are enabled.
* This op was refactored to utilize the new GPUAllocator and memory manager.


src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -676,6 +676,234 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
...
+            writer << "if (tid < nthreads)\n";
+            writer.block_begin();
+            {
+                writer << "int dim_product = tid;\n";
+                writer << "int data_idx = 0;\n";
+                writer << "for (int i = threadIdx.x; i < (rank - 1) * " << nthreads_per_block
+                       << "; i += " << nthreads_per_block << ")\n";
+                writer.block_begin();
+                {
+                    writer << "dimensions[i] = div64(dim_product, dim_magic[data_idx], "
```
fengleitian: nit:
is it possible to wrap this to a function with more readable name?

csullivan: Absolutely, I'll try to find something better.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -676,6 +680,236 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
...
+    size_t primitive_index = m_primitive_emitter->lookup(hash);
+    if (primitive_index != std::numeric_limits<size_t>::max())
+    {
+        return primitive_index;
+    }
+
+    constexpr const int nthreads_per_block = 32;
+
+    // if the kernel has not been compiled, build it
+    auto compiled_kernel = ctx->compiled_kernel_pool->get(kernel_name);
```
fengleitian: if kernel use different data type, the kernel_name still the same, it will get the same compiled kernel which sounds not right.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -676,6 +680,236 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
...
+        }
+        writer.block_end();
+        compiled_kernel = ctx->compiled_kernel_pool->set(kernel_name, writer.get_code());
+    }
+
+    std::vector<int> params;
+    params.reserve(tensor_shape.size() * 3 - 1 + lower_bounds.size() + upper_bounds.size() +
+                   slice_strides.size() * 3 + source_shape.size() * 2);
+
+    int dstr_offset = params.size();
```
fengleitian: some help function?

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -676,6 +680,236 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
...
+                writer << "out[tid] = in_bounds ? source[source_idx] : in[tid];\n";
+            }
+            writer.block_end();
+        }
+        writer.block_end();
+        compiled_kernel = ctx->compiled_kernel_pool->set(kernel_name, writer.get_code());
+    }
+
+    std::vector<int> params;
+    params.reserve(tensor_shape.size() * 3 - 1 + lower_bounds.size() + upper_bounds.size() +
```
rkimballn1: I think reserve() has a place but this is not that place. Not needed and adds two lines of code.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -676,6 +680,236 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
...
+        int shift;
+        std::tie(magic, shift) = idiv_magic_u64(static_cast<int>(dim_strides[i]));
+        dmagics.push_back(magic);
+        dshifts.push_back(shift);
+    }
+    std::copy(dmagics.begin(), dmagics.end(), std::back_inserter(params));
+    int dshift_offset = params.size();
+    std::copy(dshifts.begin(), dshifts.end(), std::back_inserter(params));
+    int lbound_offset = params.size();
+    for (auto const& lbound : lower_bounds)
```
rkimballn1: The more I see `auto` the more I don't like it. Reviewing this code I do not know what type `lbound` is. I can see it comes out of a Coordinate but now I have to go and find that class to find out what the Coordinate iterator outputs. Sigh. I think it is a size_t but I could be wrong. If it is a size_t then why not just put `size_t lbound` which is shorter and clearer. To me auto is for things like opaque handles or things you  simply pass around. Here you are casting it so you must really know what it is.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -676,6 +680,236 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
...
+    }
+
+    size_t nthreads = shape_size(tensor_shape);
+    int rank = tensor_shape.size();
+
+    // TODO: blending factors are not currently implemented
+    float alpha = 1.0f;
+    float beta = 0.0f;
+
+    std::unique_ptr<gpu::primitive> replace_slice(new gpu::primitive{[=](void** inputs,
```
rkimballn1: The lambda is very large here and embedding it in the function call really hides things. Define the lambda first then use this new variable to build the replace_slice

src/ngraph/runtime/gpu/cuda_emitter.cpp
csullivan: By using reserve one avoids many memcpy's under the hood of the std::vector. The size of the vector can be determined ahead of time and the needed space allocated with reserve so that when push_back is called we never exceed the std::vector's internal storage, which otherwise forces a malloc and memcpy.

src/ngraph/runtime/gpu/cuda_emitter.cpp
csullivan: I hear you, the reason I prefer auto for iterators is because if at some point we decide to change class Coordinate so that it inherits from std::vector<uint32_t> instead of std::vector<size_t>, then this iterator will still work and won't force an implicit cast that requires us to update this code.

Sidebar: It's a different discussion but the fact that Shape/Coordinate/Stride are all inheriting from std::vector<size_t> actually causes some pain points because size_t is 64bit wide. The fact that it is 64bit wide causes us to need to lop off the high bits with a static_cast all over in cudnn and cuda, since 64bit is much slower on the GPU and the API's require it for cudnn.

Will we ever have a tensor that has a size or stride that is larger than 2^32 ? That seems very unlikely.

src/ngraph/runtime/gpu/cuda_emitter.cpp
csullivan: You are correct. I'm still not sure if this is really ever a concern, switching types in the middle of a function, but best to be safe.

csullivan: I have added the dtypes to the kernel_name.

fengleitian: COOL. The kernel should be able to shared between functions.

PR #1003, opened 2018-05-14
Make sure that generating pyapi does not throw errors

indie:
due to directory structure, or linenos


doc/sphinx/source/python_api/index.rst
```diff
@@ -8,9 +8,9 @@ This section contains nGraph™ Python API documentation. Python API exposes
 nGraph™ C++ operations to Python users. For quick-start you can find an example
 of the API usage below.
 
-.. literalinclude:: /../../../python/examples/basic.py
+.. literalinclude:: ../../../../python/examples/basic.py
```
arogowie-intel: According to [doc](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude): 
,, The file name is usually relative to the current file’s path. __However, if it is absolute (starting with /), it is relative to the top source directory.__ ''

It's unnecessary to add additional `..` to file path, however it's correct.

indie: Being consistent with the rest of the documentation is important

indie:
@postrational :   The blank line is necessary or the documentation build throws an error locally and fails to import the example.py correctly due to the `literalinclude`

indie:
I tried again and it's okay.  2 blank lines were on the original PR even though only one shows up on Github

Issue #947, opened 2018-05-03
CUDA 9 support


csullivan:
Need to test compatibility, update CI docker, and update build scripts. Required for compatibility with clang 3.9. This issue is blocking PR #906.

WilliamTambellini:
Hi @csullivan 
The gpu runtime cmakelist file in ~/repos/ngraph/src/ngraph/runtime/gpu  is still imposing cuda 8 : 
find_package(CUDA 8 REQUIRED)
- Is it on purpose ?
- is cuda 9 support still unofficial ?
Cheers
W.

csullivan:
Hi @WilliamTambellini, it is supported. #906 updates the cmakelist with 
https://github.com/NervanaSystems/ngraph/pull/906/files. It is held up by an internal CI upgrade that is needed for testing. You shouldn't have any issues with CUDA 9 and ngraph.

PR #1009, opened 2018-05-14
Upgrade nGraph support to CUDA 9.1

csullivan:
Presently CUDA 8 is required and CUDA 9 is now functional but optional.

I've tested CUDA 9.1 with driver 390.12. When installing CUDA 9.1, it was found on Titan-XP11 that the driver had already been upgraded to 390.12. On the dev and CI nodes that run CUDA 8, the driver version is 384.98. CUDA 9.1 requires nvidia-drivers >= 387.26, and so these will need to be upgraded prior to upgrading to CUDA 9.1.

Additional notes:
* Added compiler log output to NVRTC compiler. 
* Added CUDA version dependent defguards to add CUDA 8 backward compatibility to some deprecated CUDA intrinsics.
* Unrelated: a few tests were disabled that should not have been in PR #983, which this PR fixes.

This PR addresses issue #947.


rkimballn1:
Do you recommend that we upgrade all lab machines to 390.12? Should we upgrade all cuda to 9.1? Should we run CI with multiple cuda versions?

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -483,7 +483,8 @@ size_t runtime::gpu::CUDAEmitter::build_avg_pool(const GPURuntimeContext* ctx,
                     if (include_pad == false)
                     {
                         // count the number of (non-padded) elements
-                        writer << "pool_size += __popc(__ballot(within_tensor_bounds));\n";
+                        writer << "pool_size += __popc(__ballot_sync(0xffffffff, "
```
fengleitian: Note, in the future, we can check if we can use mask to mask out some of the shift, memory load/store is not really full parallel. might save some work. No need to investigate for now.

csullivan:
@rkimballn1 
> Do you recommend that we upgrade all lab machines to 390.12? Should we upgrade all cuda to 9.1? Should we run CI with multiple cuda versions?

Yes, I recommend upgrading the driver to 390.12, and CUDA to 9.1. If we are going to support 8 as well, then we should have a CUDA 8 test, yes. Backwards compatibility is nice so that we don't alienate some older system targets which have compilers that can compile against 8 but not against 9.

PR #1010, opened 2018-05-14
give frontends some flexibility over fusions they would like to run

Krovatkin:
tagging @mbrookhart since this is probably going to be an API breaking change for MxNet.


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -323,7 +323,7 @@ void runtime::cpu::CPU_ExternalFunction::compile()
     pass_manager.register_pass<ngraph::pass::AlgebraicSimplification>();
     pass_manager.register_pass<ngraph::pass::CommonSubexpressionElimination>();
     pass_manager.register_pass<ngraph::pass::CoreFusion>();
-    pass_manager.register_pass<runtime::cpu::pass::CPUFusion>();
+    pass_manager.register_pass<runtime::cpu::pass::CPUFusion>(runtime::cpu::pass::CPUFusion::ALL);
```
jbobba: Can you make the default be ALL

src/ngraph/runtime/cpu/pass/cpu_fusion.hpp
```diff
@@ -35,21 +35,40 @@ namespace ngraph
 class ngraph::runtime::cpu::pass::CPUFusion : public ngraph::pass::GraphRewrite
 {
 public:
-    CPUFusion()
+    //30 different fusion groups
+    //that we can nest/mix&match/etc
+    //should be good enough
+    //for quite a while
```
jbobba: fewer linebreaks?

src/ngraph/runtime/cpu/pass/cpu_fusion.hpp
```diff
@@ -35,21 +35,40 @@ namespace ngraph
...
+    enum fusions
+    {
+        DIFFERENTIABLE_FUSIONS = 0x1,
```
jbobba: Add a comment "fusions that result in ops that are differentiable through ngraph's autodiff"

Krovatkin:
@jbobba addressed yr feedback

src/ngraph/runtime/cpu/pass/cpu_fusion.hpp
```diff
@@ -35,21 +35,40 @@ namespace ngraph
 class ngraph::runtime::cpu::pass::CPUFusion : public ngraph::pass::GraphRewrite
 {
 public:
```
diyessi: I don't like the 2^i enum with ors/ands. How about a regular enum and sets of values?
Also, `fusions` should be `Fusions`. If you make it `enum class Fusion` then `Fusion::differentiable` etc. will be easier to read than the C-like `DIFFERENTIABLE_FUSIONS`.

Krovatkin: We can definitely switch to `enum class`. Actually, my thinking to use regular enums was exactly the opposite: I didn't want introduce another namespace and `DIFFERENTIABLE_FUSIONS` seemed more readable to me than French's object + adjective :-) 

> I don't like the 2^i enum with ors/ands. How about a regular enum and sets of values?


This will probably result in code duplication since we won't be able to nest/reuse groups of fusions with CPUFusion as easily. It's pretty common even in higher-level languages like Java or c#. What I could do in case you really don't like this enum is convert it so something like this

```cpp
enum class Fusion
{
Regular,
Differentiable,
Invalid
}

std::bitset<Fusion::Invalid> fusion_groups;
```

then we could do ?

```cpp
fusion_groups.set(Fusion::Regular);
fusion_groups.set(Fusion::Differentiable);
```

PR #1011, opened 2018-05-15
[Py]Update unit tests for python API

tsocha:


python/test/ngraph/test_ops_unary.py
```diff
@@ -20,7 +20,6 @@
 from test.ngraph.util import run_op_numeric_data, run_op_node
 
 
-@pytest.mark.xfail(reason='Results mismatch when passing created Constant node from raw data.')
 @pytest.mark.parametrize('ng_api_fn, numpy_fn, range_start, range_end', [
     (ng.cos, np.cos, -100., 100.),
     (ng.sin, np.sin, -100., 100.),
```
arogowie-intel: I'd remove `test_unary_op_array_err` test since it's solved and update `test_unary_op_array` data ranges for respective functions.

python/test/ngraph/test_ops_unary.py
```diff
@@ -31,7 +30,7 @@ def test_unary_op_array_err(ng_api_fn, numpy_fn, range_start, range_end):
     expected = numpy_fn(input_data)
 
     result = run_op_numeric_data(input_data, ng_api_fn)
-    assert np.allclose(result, expected)
+    np.testing.assert_allclose(result, expected, rtol=0.001)
```
arogowie-intel: Please pass these keyword arguments: `rtol=1e-05, atol=1e-08, equal_nan=False`

python/test/ngraph/test_ops_unary.py
```diff
@@ -62,10 +61,10 @@ def test_unary_op_array(ng_api_fn, numpy_fn, range_start, range_end):
     expected = numpy_fn(input_data)
 
     result = run_op_node([input_data], ng_api_fn)
-    assert np.allclose(result, expected)
+    np.testing.assert_allclose(result, expected, rtol=0.001)
```
arogowie-intel: Same as above.

python/test/ngraph/test_ops_unary.py
```diff
@@ -62,10 +61,10 @@ def test_unary_op_array(ng_api_fn, numpy_fn, range_start, range_end):
...
 
     result = run_op_numeric_data(input_data, ng_api_fn)
-    assert np.allclose(result, expected)
+    np.testing.assert_allclose(result, expected, rtol=0.001)
```
arogowie-intel: Please pass these keyword arguments: `rtol=1e-05, atol=1e-08, equal_nan=False`

python/test/ngraph/test_ops_unary.py
postrational: If we want to pass these arguments in every test, perhaps we should make our own custom assert function.

PR #1020, opened 2018-05-18
add gpu reduce_window op

fengleitian:


src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -677,6 +678,95 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
...
+    auto compiled_kernel = ctx->compiled_kernel_pool->get(kernel_name.str());
+    if (compiled_kernel == nullptr)
+    {
+        codegen::CodeWriter writer;
+        CudaKernelBuilder::add_pod_typedefs(writer);
+        if (kernel)
+        {
+            CudaKernelBuilder::get_device_helper(writer, op, kernel, dtypes);
+        }
+
```
csullivan: nit: extra spaces on lines 721 and 723

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -677,6 +678,95 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
...
+        CudaKernelBuilder::get_reduce_window_op(writer, kernel_name.str(), op, dtypes, rank);
+
+        compiled_kernel = ctx->compiled_kernel_pool->set(kernel_name.str(), writer.get_code());
+    }
+    size_t nthreads = shape_size(output_shape);
+    auto input_strides = row_major_strides(input_shape);
+
+    std::shared_ptr<void*> input_strides_ptr = std::make_shared<void*>(
```
csullivan: Using shared_ptr here to avoid the need to free this allocation later? That won't work on a `void*` though. I would suggest using the new memory manager and the GPUAllocator as below. Replace lines 729 - 736 with:
```c++
    GPUAllocator allocator =
        external_function->get_primitive_emitter()->get_memory_allocator();
    size_t idx_input_strides = allocator.reserve_argspace(
        input_strides.data(), input_strides.size() * sizeof(size_t));
    size_t idx_output_shape = allocator.reserve_argspace(
        output_shape.data(), output_shape.size() * sizeof(size_t));
    size_t idx_reduce_window_shape = allocator.reserve_argspace(
        reduce_window_shape.data(), reduce_window_shape.size() * sizeof(size_t));
    size_t idx_reduce_window_strides = allocator.reserve_argspace(
        reduce_window_strides.data(), reduce_window_strides.size() * sizeof(size_t));
```

and then in the primitive get the device addresses with with:
```c++
    void* input_strides_d = runtime::gpu::invoke_memory_primitive(ctx, idx_input_strides);
    void* output_shape_d = runtime::gpu::invoke_memory_primitive(ctx, idx_output_shape);
    void* reduce_window_shape_d = runtime::gpu::invoke_memory_primitive(ctx, idx_reduce_window_shape);
    void* reduce_window_strides_d = runtime::gpu::invoke_memory_primitive(ctx, idx_reduce_window_strides);
```

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -677,6 +678,95 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
...
+        runtime::gpu::create_gpu_buffer(sizeof(size_t) * rank, output_shape.data()));
+    std::shared_ptr<void*> reduce_window_shape_ptr = std::make_shared<void*>(
+        runtime::gpu::create_gpu_buffer(sizeof(size_t) * rank, reduce_window_shape.data()));
+    std::shared_ptr<void*> reduce_window_stride_ptr = std::make_shared<void*>(
+        runtime::gpu::create_gpu_buffer(sizeof(size_t) * rank, reduce_window_strides.data()));
+
+    // create the launch primitive
+    std::unique_ptr<gpu::primitive> f(
+        new gpu::primitive{[=](void** inputs, void** outputs) mutable {
+            std::vector<void*> args_list;
```
csullivan: nit: Why use std::vector::push_back() here when the number of elements is fixed and known at compile time? std::vector<void*> args_list { ... } would be better so that you avoid the copying when push_back overflows the internal memory storage of the vector.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -677,6 +678,95 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const GPURuntimeConte
     return primitive_index;
 }
 
+size_t runtime::gpu::CUDAEmitter::build_reduce_window_helper(const GPURuntimeContext* ctx,
+                                                             const std::vector<std::string>& dtypes,
+                                                             const Shape& input_shape,
+                                                             const Shape& output_shape,
```
csullivan: Consider switching to GPUShape so that we are not using 64 bit words on the GPU unnecessarily.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -242,6 +242,66 @@ void runtime::gpu::CudaKernelBuilder::get_reverse_op(codegen::CodeWriter& writer
...
 
+void runtime::gpu::CudaKernelBuilder::get_reduce_window_op(
+    codegen::CodeWriter& writer,
+    const std::string& name,
+    const std::string& op,
+    const std::vector<std::string>& data_types,
+    const size_t rank)
+{
+    writer << "extern \"C\" __global__ void cuda_" << name << "(" << data_types[0] << "* in, "
+           << data_types[1] << "* out, size_t* input_strides, size_t* output_shape, size_t* "
```
csullivan: If you switch to GPUShape above we can declare all of these as arrays as `int` or `uint32_t` and avoid the performance loss of 64 bit words. Consider switching `size_t` throughout the kernel to 32 bits.

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: Here is actually by design, not a good design. 
It should have same lifetime as lambda function. It should not be deleted.

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: 32bit definitely will save some space and computation, one concern is do we have limitation on shape to 32bit?

csullivan: Yep! We will have that check with the GPUShape PR, see https://github.com/NervanaSystems/ngraph/pull/1031/files#diff-87aac37cceeb79e1d3bf95706a2a804eR82

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: memory allocation changes looks good.

fengleitian:
waiting for PR #1035.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -117,8 +118,13 @@ static const std::unordered_map<std::type_index, cudnnReduceTensorOp_t> reduce_m
     {TI(ngraph::op::Add), CUDNN_REDUCE_TENSOR_ADD},
     {TI(ngraph::op::Multiply), CUDNN_REDUCE_TENSOR_MUL},
     {TI(ngraph::op::Maximum), CUDNN_REDUCE_TENSOR_MAX},
-    {TI(ngraph::op::Minimum), CUDNN_REDUCE_TENSOR_MIN},
-};
+    {TI(ngraph::op::Minimum), CUDNN_REDUCE_TENSOR_MIN}};
+
```
csullivan: Are these static unordered_maps used in multiple functions? 

If not, thinking we could declare them (statically) inline in the function call, rather than having them be at global scope.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1429,6 +1434,106 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
                 return;
             }
 
+            template <>
```
csullivan: Would it be possible to put some comments in here as to what the variable names are in the kernel? In the below kernel, am I correct assuming in assuming that `idx_in_start_1` is the index to the beginning of the window along axis 1? What is `idx_init`? Overall looks good, but if you can add some comments to help us understand in the future, that would be great.

```
extern "C" __global__ void cuda_reduce_window_max_float_float_float3(float* in, float* out, int* input_strides, int* output_shape, int* reduce_window_shape, int* reduce_window_strides, size_t n)
{
    const int tid = blockIdx.x * blockDim.x + threadIdx.x;
    if (tid < n)
    {
        int idx_out = tid;
        int idx_init = 0;
        int idx_out_2 = idx_out % output_shape[2];
        int idx_in_start_2 = idx_out_2 * reduce_window_strides[2];
        int idx_in_end_2 = idx_in_start_2 + reduce_window_shape[2];
        idx_init += idx_in_start_2 * input_strides[2];
        idx_out /= output_shape[2];
        int idx_out_1 = idx_out % output_shape[1];
        int idx_in_start_1 = idx_out_1 * reduce_window_strides[1];
        int idx_in_end_1 = idx_in_start_1 + reduce_window_shape[1];
        idx_init += idx_in_start_1 * input_strides[1];
        idx_out /= output_shape[1];
        int idx_out_0 = idx_out % output_shape[0];
        int idx_in_start_0 = idx_out_0 * reduce_window_strides[0];
        int idx_in_end_0 = idx_in_start_0 + reduce_window_shape[0];
        idx_init += idx_in_start_0 * input_strides[0];
        idx_out /= output_shape[0];
        float result = in[idx_init];
        for(int i_0 = idx_in_start_0; i_0 < idx_in_end_0; i_0++)
        {
            for(int i_1 = idx_in_start_1; i_1 < idx_in_end_1; i_1++)
            {
                for(int i_2 = idx_in_start_2; i_2 < idx_in_end_2; i_2++)
                {
                    int idx_in = 0;
                    idx_in += i_0 * input_strides[0];
                    idx_in += i_1 * input_strides[1];
                    idx_in += i_2 * input_strides[2];
                    result = (idx_in == idx_init) ? result : max(result, in[idx_in]);
                }
            }
        }
        out[tid] = result;
    }
}
```

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: Good point, fixed.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: add some comments and change names to more meaningful ones.

PR #1027, opened 2018-05-18
fix building with mpi on local machine

rkimballn1:
On my local machine the cmake variable MPI_C_INCLUDE_PATH contains 4 paths. This causes the build to fail. The we also need to update the build-in compiler to split these 4 paths to add to the search path. Now it works on my local machine.


rkimballn1:
TESTNOW

DawnStone:
TESTNOW

rkimballn1:
TESTNOW

PR #1028, opened 2018-05-18
Loop Kernel Op + Tests

Krovatkin:


src/ngraph/runtime/cpu/op/loop_kernel.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include "ngraph/util.hpp"
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace cpu
+        {
+            namespace op
+            {
```
diyessi: This needs a description of what the op does. I think just computes outputs, the same as would have happened without it, only it does the part from args to outputs more efficiently.
Couldn't node_list be computed from outputs and args?

Krovatkin: Will add the description. 

> Couldn't node_list be computed from outputs and args?

Yes, but `node_list` will be anyways constructed in a topologically sorted order by a fusion pass. It seems more practical to simply "cache" it inside an op. This way we don't need more logic to do the traversals from the outputs to args and also we save ourselves a computation. I'd suggest we will leave it this way until I add the fusion pass.

Krovatkin:
TESTNOW

PR #1031, opened 2018-05-19
[CS:GPU::Part 1] Add GPUShape type, conversion operators, and generalized shape helpers.

csullivan:
- [x] [CS:GPU::Part 1] Note: do not merge until all parts[1-4] have been reviewed and approved. After approvals: merge part 4 into 3, then 3 -> 2, 2 -> 1, and finally 1 -> master after CI passes.

This PR adds type GPUShape. 
* Conversion operators ensure tensor dimensions are not greater than 2^32 elements for the GPU backend. 
* The Shape helpers `row_major_strides` and `shape_size` are now specialized to allow for generic operability with different shape types.
* Updated `row_major_strides` implementation to avoid reverse-copy.


src/ngraph/shape.hpp
```diff
@@ -68,11 +68,40 @@ namespace ngraph
...
+        return strides;
+    }
 
-    inline bool is_scalar(const Shape& shape) { return 0 == shape.size(); }
-    inline bool is_vector(const Shape& shape) { return 1 == shape.size(); }
+    template <typename ShapeType>
+    inline bool is_scalar(const ShapeType& shape)
+    {
+        return 0 == shape.size();
+    }
```
diyessi: Need a blank line here.

src/ngraph/runtime/gpu/gpu_shape.hpp
```diff
@@ -0,0 +1,115 @@
...
+        {
+        }
+
+        GPUShape() {}
+        GPUShape& operator=(const GPUShape& v)
+        {
+            static_cast<std::vector<uint32_t>*>(this)->operator=(v);
+            return *this;
+        }
+        GPUShape& operator=(GPUShape&& v)
```
diyessi: Please add a blank lone between methods. Thanks.

src/ngraph/shape.hpp
csullivan: Added line spacing as suggested

src/ngraph/shape.hpp
```diff
@@ -68,11 +68,41 @@ namespace ngraph
...
+
+    template <typename ShapeType>
+    inline bool is_scalar(const ShapeType& shape)
+    {
+        return 0 == shape.size();
+    }
 
-    inline bool is_scalar(const Shape& shape) { return 0 == shape.size(); }
-    inline bool is_vector(const Shape& shape) { return 1 == shape.size(); }
+    template <typename ShapeType>
```
rkimballn1: template parameters are UPPER_SNAKE_CASE

src/ngraph/shape.hpp
```diff
@@ -68,11 +68,41 @@ namespace ngraph
     };
 
     /// Number of elements in spanned by a shape
-    size_t shape_size(const Shape& shape);
+    template <typename ShapeType>
```
rkimballn1: template parameters are UPPER_SNAKE_CASE

src/ngraph/shape.hpp
```diff
@@ -68,11 +68,41 @@ namespace ngraph
...
+        for (auto d : shape)
+        {
+            size *= d;
+        }
+        return size;
+    }
 
     /// Row-major strides for a shape
-    Strides row_major_strides(const Shape& shape);
+    template <typename ShapeType>
```
rkimballn1: template parameters are UPPER_SNAKE_CASE

src/ngraph/shape.hpp
```diff
@@ -68,11 +68,41 @@ namespace ngraph
...
+        auto st = strides.rbegin();
+        for (auto d = shape.rbegin(); d != shape.rend(); d++, st++)
+        {
+            *st = s;
+            s *= *d;
+        }
+        return strides;
+    }
+
+    template <typename ShapeType>
```
rkimballn1: template parameters are UPPER_SNAKE_CASE

src/ngraph/runtime/gpu/gpu_shape.hpp
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    throw std::runtime_error(
+                        "Request for Strides which exceed the bitwidth available for GPUShapes "
+                        "(32)");
+                }
+                this->push_back(low);
+            }
+        }
+
+        GPUShape(const Coordinate& coord)
```
fengleitian: 32bit will definitely save space and computation. Do we have limitation on shape to 32bit? for example currently we treat all element-wised op as 1D, is it possible it will be more than 4G?

csullivan: There is no limitation on Shape, it can have dimension's of length >= 2^32. GPUShape removes support for shape sizes greater than 32 bits in the GPU transformer. If we want to support 64 bit descriptors in our kernels (CUDNN does not), we should do it as an option, but not by default because of the performance cost that you point out.

GPUShape makes the check that the shape is less than 32 bits in precision explicit. And it allows us to implicitly convert to 32 bit shapes in the transformer without needing to `static_cast` the high bits.

csullivan: >for example currently we treat all element-wised op as 1D, is it possible it will be more than 4G
Can you elaborate what you mean?

fengleitian: I mean it's possible to exceed 32bit.

diyessi: GPUShape only needs to handle shapes that GPU kernels use. The 6d things from a framework will have been down-ranked by the time they get to GPUShape.

diyessi: Any operation shape that would be larger than the GPU can handle would first need to be transformed into a series of operations that fit, so encountering a GPUShape that was too big would be a bug.

fengleitian: @diyessi @csullivan That's what I'm worried,  just an example, we have an 1K*1K*1K*1K 4D shape, when we transfer to 1D on GPU, it will be 1T.

fengleitian: @diyessi GPU can handle 64bit address, here we are trying to see if we can manually limit it to 32bit to improve performance.

diyessi: For that, we'd want to first automatically chop the graph into pieces that fit in a way that wasn't too bad for performance (although any performance would be better than not being able to run at all). But first we want to handle the cases that do fit, since those are the kinds of graphs we would be chopping into.

csullivan: @fengleitian Shape{1000,1000,1000,1000} and GPUShape{1000,1000,1000,1000} are both allowed. You are right with the shape_size template, below, we should there return std::vector<size_t>, but this is the only problem I see. The only time GPUShape would limit us is if we had a GPUShape with one dimension greater than 32 bits in width.

The problem with what we are doing right now is that we pretend we support the full Shape, but in reality, most of our ops do not -- all CUDNN ops are limited to 2**32 elements per dimensions.

src/ngraph/shape.hpp
```diff
@@ -68,11 +68,41 @@ namespace ngraph
     };
 
     /// Number of elements in spanned by a shape
-    size_t shape_size(const Shape& shape);
+    template <typename SHAPE_TYPE>
```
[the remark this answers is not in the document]
csullivan: So that `shape_size` can operate on both Shape and GPUShape.

csullivan:
Overall, the question is whether we will exceed 32 bits for a single axis. In our internal kernel implementations we can still utilize 64 bit where necessary. 

For example, in the above case of a 4 dimensional tensor with each axis having 1000 elements, the GPU allocation will exceed 32 bits in the total number of elements (i.e. its `shape_size` will be > 2^32) as @fengleitian points out. Internally that is ok, because we will store that as size_t and we can index into the dense flat array as usual. The only thing GPUShape does is to ensure that a single axis does not exceed 2^32 elements because this is not supported in cuDNN descriptors.

If we make replacement kernels for each of the cuDNN ops that we can fall back to in these cases, then GPUShape wouldn't be needed. This doesn't require any changes to current CUDA kernels which support 64 bit tensor descriptors.

fengleitian:
leave it for 32bit now, we can add 64bit support when we need. I was worried it will take more time to add 64bit support later than now.

rkimballn1:
TESTNOW

src/ngraph/runtime/gpu/gpu_call_frame.cpp
```diff
@@ -68,26 +66,11 @@ void runtime::gpu::GPU_CallFrame::call(
...
-    cudnnStatus_t cudnnStatus = cudnnCreate(&m_cudnn_handle);
-    if (cudnnStatus != CUDNN_STATUS_SUCCESS)
-    {
-        throw runtime_error("cuDnn create handle failed");
-    }
-
-    // Pass scalars as reference on the Device
-    cublasSetPointerMode(m_cublas_handle, CUBLAS_POINTER_MODE_DEVICE);
-
+    // add pointers to gpu primitives into the gpu runtime context
```
fengleitian: move this comment with code

PR #1032, opened 2018-05-19
Modernize cmake usage

rkimballn1:
This is the big cmake rewrite I have been wanting to do for a while. It does several things:
1) Each external project is given a library interface, so eigen is referred to as libeigen, llvm as libllvm, etc. The benefit of making the external projects as libraries is that a library carries with it it's include directory and link libraries. To set up dependencies, include paths, and link to the library you need to only add
```target_link_libraries(<my_project> PUBLIC libllvm)```
You don't need to add_dependencies or include_directories or anything else.
2) codegen and all backends are now libraries rather than conditionally being built directly into ngraph. This normalizes the build procedure for builtin backends, CPU, GPU and INTERPRETER and external backends like IE and NNP. If a backend needs codegen then it just adds codegen to it's `target_link_libraries`. No more "build the CPU backend because I need codegen". The locations of all of the source code is unchanged so developers should not notice a difference.
3) Conditional building of components is fully controlled by cmake. If you want to build the CPU backend for unit test you simply add cpu_backend to the project's `target_link_libraries`. Since the CPU depends on codegen then codegen is built, and codegen depends on llvm so llvm is built.
4) All backends are now optional, but CPU and INTERPRETER are built by default. It is now possible to build just ngraph with zero backends. This is used by external backend developers like IE.
5) All examples are now individual cmake projects.

### CODE REVIEW

File: src/ngraph/CMakeLists.txt [hunk 1]
```diff
@@ -381,20 +219,16 @@ install(DIRECTORY
...
-    install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/tbb_build/tbb_release/
-        DESTINATION ${NGRAPH_INSTALL_LIB}
-        FILES_MATCHING REGEX "/libtbb${CMAKE_SHARED_LIBRARY_SUFFIX}(\\.[0-9]+)*$"
-    )
-    install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/tbb_build/tbb_debug/
-        DESTINATION ${NGRAPH_INSTALL_LIB}
-        FILES_MATCHING REGEX "/libtbb_debug${CMAKE_SHARED_LIBRARY_SUFFIX}(\\.[0-9]+)*$"
-    )
-endif()
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "A most excellent graph library")
```
aprocter: :metal:

rkimballn1:
TESTNOW

silee2:
python ci is failing because dlopen() in libcodegen cannot find the backend so's
I've set CMAKE_INSTALL_RPATH to influence dlopen() and it seems to work.
[silee2/cmake_install_rpath](https://github.com/NervanaSystems/ngraph/commit/4d74e2447e156337cd89c012026a72eb653926ef)
python ci test results at:
https://aipg-rancher.intel.com/jenkins/algo/job/testing-ngraph-python-merge/1854/console
Can anyone think of an unwanted side effect this change would introduce?

rkimballn1:
The tests are able to find the backend's so file but that backend is unable to find the codegen.so which is odd because they are located in the same directory. Anyway, the python CI does not set LD_LIBRARY_PATH until AFTER the python tests are run. The python_ci test should be updated.
I prefer fixing the CI over changing cmake, but I'm open to what others think.

rkimballn1:
I merged in @silee2 change

PR #1038, opened 2018-05-22
Fuse conv+bias bprop 

nbpatel:


src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -50,6 +50,10 @@
 #include "ngraph/runtime/cpu/op/matmul_bias.hpp"
 #include "ngraph/runtime/cpu/op/sigmoid.hpp"
 
+#ifndef __clang__
+    #pragma GCC optimize ("-O0")
+#endif
+
```
jbobba: Please remove.

Krovatkin: this isn't needed anymore.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -753,6 +754,54 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias()
...
+                                                            conv_bprop->get_filters_shape(),
+                                                            bias_shape, pattern_map[delta],
+                                                            conv_bprop->get_window_movement_strides_forward(),
+                                                            conv_bprop->get_window_dilation_strides_forward(),
+                                                            conv_bprop->get_padding_below_forward(), 
+                                                            conv_bprop->get_padding_above_forward(),
+                                                            conv_bprop->get_data_dilation_strides_forward());
+               auto goe1 = std::make_shared<op::GetOutputElement>(conv_bias_bprop, 0);
+               auto goe2 = std::make_shared<op::GetOutputElement>(conv_bias_bprop, 1);
+               ngraph::replace_node(m.get_match_root(), goe1);
```
Krovatkin: could you please add two `NGRAPH_DEBUG` traces that specify which nodes are being replaced and with what?

PR #1039, opened 2018-05-23
add check to make sure we don't replace unreachable nodes

Krovatkin:


aprocter:
:frog: report: CI is failing because the test `cpu_fusion.fuse_lstm_cells` is triggering the check introduced by this PR.

PR #1041, opened 2018-05-23
Group Convolution

Krovatkin:


src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -248,6 +249,14 @@ namespace ngraph
...
                     auto arg0_shape = node->get_input_shape(0);
                     auto arg1_shape = node->get_input_shape(1);
+
+                    if (default_weights_format)
+                    {
+                        const size_t IC = 1;
+                        arg0_shape.at(IC) /=
+                            std::dynamic_pointer_cast<ngraph::op::GroupConvolution>(node)
+                                ->get_groups();
+                    }
```
jbobba: Dont we need arg1_shape to be 5D for mkldnn?

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -282,37 +291,65 @@ namespace ngraph
...
                     }
                     convolution_forward::primitive_desc prim_desc(*fwd_desc, cpu_engine);
                     prim_input_formats.push_back(static_cast<memory::format>(
                         prim_desc.src_primitive_desc().desc().data.format));
-                    prim_input_formats.push_back(static_cast<memory::format>(
-                        prim_desc.weights_primitive_desc().desc().data.format));
+
+                    if (default_weights_format)
+                    {
+                        prim_input_formats.push_back(prim_input_formats.back());
```
jbobba: What does this do? Are we saying that we need the data/input format on the weights?

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
Krovatkin: Not sure. Will it affect the data layout since it's the only one we care about here?
I remember you told me that the data tensor remains 4D, so I did the bare minimum to keep MKLDNN happy and give us the data layout.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
Krovatkin: We don't use it in Emitter. We do need **some** layout, so the number of layouts and nodes match. I **thought** that's what was exactly what you wanted me to do ¯\_(ツ)_/¯

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
jbobba: It might. Can you check with the MKLDNN team on this?

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
jbobba: This could lead to conversions being introduced on the weights tensor. We basically need prim_input_format[WEIGHTS] = input_format(arg1)

PR #1047, opened 2018-05-24
fix invalid context when run mxnet and nbench

fengleitian:
what happened?
throw invalid_context context error when run some full model use nbench or mxnet.

Cause:
The context will not automatically share between threads. 
Previously we will only explicitly set cotext for the first external function.

Solution:
So if an external function runs a new threads, we need to setCtxCurrent for using the same context.
Assume each external function will run on the same thread.


fengleitian:
https://devtalk.nvidia.com/default/topic/519087/cuda-context-and-threading/?offset=4

fengleitian:
@csullivan could you help test to see if it works on mxnet? Thank you.

csullivan:
Excellent! Will absolutely do so.

fengleitian:
@csullivan so far as I know, there is only one primary context for each device and will share between threads. So a singleton will be OK. Will modify it for multiple GPU when we need it.

PR #1050, opened 2018-05-25
Serializer Pass

Krovatkin:
@rkimballn1  This is useful for us in some scenarios where we have other tools that can use json files with graphs. For example, GraphViz can't visualize bigger graphs but we can use gephi for that. What do you think?


src/ngraph/pass/serialize.cpp
```diff
@@ -0,0 +1,45 @@
...
+pass::Serialization::Serialization(const string& output_file)
+    : m_prefix_filename{output_file}
+{
+}
+
+bool pass::Serialization::run_on_module(vector<shared_ptr<Function>>& functions)
+{
+    for (shared_ptr<Function> f : functions)
+    {
+        string js = serialize(f, 4);
```
rkimballn1: There is a serialize call that can serialize directly to a file.
`void serialize(const std::string& path,
                   std::shared_ptr<ngraph::Function> func,
                   size_t indent = 0);`

src/ngraph/pass/serialize.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "ngraph/file_util.hpp"
+#include "ngraph/pass/serialize.hpp"
+#include "ngraph/serializer.hpp"
+#include "ngraph/util.hpp"
+#include "nlohmann/json.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+pass::Serialization::Serialization(const string& output_file)
```
rkimballn1: The name should be `output_prefix` or something similar. And update the header too.

PR #1051, opened 2018-05-25
change get_shape().size() to get_size()

fengleitian:
we check the actual size, get_shape().size() is only dimension count. 

Dimension will size 1 is the tricky part, for example `1, 3, 1` vs `3,1`, `get_shape().size()` will be different. but when we emit with cudnn, both will be change to `1, 1, 3, 1`. cudnn will throw `CUDNN_STATUS_BAD_PARAM ` if input and output shape match.


Krovatkin:
TESTNOW

PR #1054, opened 2018-05-26
export LD_LIBRARY_PATH in import howto

philtomson:
There are two places in the docs currently where the user is instructed to export LD_LIBRARY_PATH to point to $HOME/ngraph_dist/lib -  both are after the instructions for importing ONNX files (ngraph/doc/sphinx/source/howto/imort.rst). There is a note (labeled "Important") in the Import HOWTO instructions to ensure you've followed the instructions in the Install guide, however, the Install guide does not mention anything about setting LD_LIBRARY_PATH to point to the ngraph_dist libs - that comes later in [Activating nGraph on generic frameworks](http://ngraph.nervanasys.com/docs/latest/optimize/generic.html?highlight=ld_library_path). You can easily get to the Import HOWTO without having set LD_LIBRARY_PATH if you were only following the path to Importing ONNX models. Without LD_LIBRARY_PATH  being set you'll get errors in Python when trying to import ngraph. This pull request remedies that problem and makes things a bit more newbie friendly. 

Another alternative would be to add the LD_LIBRARY_PATH setting to the [first Install doc](http://ngraph.nervanasys.com/docs/latest/install.html#installation-steps), though it might be best to have it in both locations since the Import HOWTO seems to recapitulate many of the Installation instructions already.


apwojcik:
Using LD_LIBRARY_PATH in production is incorrect. The purpose of LD_LIBRARY_PATH is only for debugging and testing. It is not meant to be used in production. Instead please add appropriate configuration file to `/etc/ld.so.conf.d` directory and run `ldconfig` in your installation script.

PR #1056, opened 2018-05-29
Minor bug fix in function outlining

jbobba:
I don't have good data to back this up at the moment. At the very least, we will need the change in cpu_external_function since it fixes a potential bug.


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -545,6 +545,10 @@ using namespace ngraph::runtime;
             string match_function_name;
             for (size_t j = i + 1; j < op_list.size(); j++)
             {
+                if (op_list[j]->is_constant() || op_list[j]->is_parameter())
```
diyessi: Can you add a comment about why we need this test? I think we will actually want a new virtual for what it's testing for since there might be some additional node types that need the same treatment when state shows up.

jbobba: Added a comment. These ops dont take part in Bob's outlining optimization and are not added to the node_cache.

diyessi: If any more nodes join this list, lets add an `is_cacheable` to `Node`.

jbobba:
[WIP] until we fix the stack alignment issue exposed by GPU codegen.

aprocter:
TESTNOW

PR #1058, opened 2018-05-29
Strided Convolution

Krovatkin:


Krovatkin:
This needs to be updated to Resnet_v2

PR #1063, opened 2018-05-30
Add checks for conv + bias bprop fusion

nbpatel:


src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -775,30 +775,35 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_bprop()
         auto conv_bprop =
             std::dynamic_pointer_cast<op::ConvolutionBackpropFilters>(m.get_match_root());
 
-        for (auto delta_user : pattern_map[delta]->get_users())
+        if (conv_bprop->get_input_shape(0).size() == 4 &&
```
Krovatkin: do we need to check for dilated_strides? @jbobba

jbobba: Yes, please add that. We need to add it to other fusion patterns that are missing this check as well.

nbpatel:
TESTNOW

PR #1064, opened 2018-05-31
gpu convolution asymmetric pad

fengleitian:
This PR add asymmetric padding to Convolution, ConvolutionBackpropData, and ConvolutionBackpropFilter.
1. Refactor the code to avoid emit and compile at runtime.
2. Check and add asymmetric padding.
3. enable convolution test with asymmetric padding.


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -269,51 +293,126 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                {
+                    writer << "std::vector<void*>{pad_buffer}.data()";
+                }
+                else
+                {
+                    writer << "std::vector<void*>{" << out[0].get_name() << "}.data()";
+                }
+                writer << ");\n";
+
+                // since we paadded output with another buffer, we neec to copy back to ouput
```
csullivan: Small typo in comment.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -177,6 +178,243 @@ size_t runtime::gpu::CUDNNEmitter::build_reduce_forward(const runtime::gpu::GPUR
...
+    }
+    }
+
+    // get an allocator for transient per kernel gpu memory
+    GPUAllocator allocator = this->m_primitive_emitter->get_memory_allocator();
+    // (lazy) allocation for kernel arguments
+    size_t workspace_idx = allocator.reserve_workspace(workspace_size_in_bytes);
+
+    std::unique_ptr<gpu::primitive> conv;
+    switch (direction)
```
csullivan: Convolution is a big enough op that I'm wondering if we should consider breaking its cudnn implementation up into `build_conv_forward` `build_conv_backward` etc. This is fine as is but the switch for the workspace is a bit tedious, and with Prop::BackwardFilter we now have to add that case to future ops.

fengleitian: Agree.

fengleitian: Let me change it.

fengleitian: done

PR #1067, opened 2018-06-01
Add StopGradient op to ngraph

ashokei:
Add StopGradient op to ngraph. This op allows us to implement BlockGrad symbol in mxnet, or for any use-case where this is needed :

`Stops the accumulated gradient of the inputs from flowing through this operator in the backward direction.`

https://mxnet.incubator.apache.org/api/python/symbol/symbol.html#mxnet.symbol.BlockGrad

### DIFF
--- a/src/ngraph/op/stop_gradient.cpp
+++ b/src/ngraph/op/stop_gradient.cpp
@@ -1,0 +1,49 @@
+/*******************************************************************************
+* Copyright 2017-2018 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "ngraph/op/stop_gradient.hpp"
+#include "ngraph/op/broadcast.hpp"
+#include "ngraph/op/constant.hpp"
+
+using namespace std;
+using namespace ngraph;
+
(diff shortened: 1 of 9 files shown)


src/ngraph/op/stop_gradient.cpp
```diff
@@ -0,0 +1,49 @@
...
+shared_ptr<Node> op::StopGradient::copy_with_new_args(const NodeVector& new_args) const
+{
+    if (new_args.size() != 1)
+    {
+        throw ngraph_error("Incorrect number of new arguments");
+    }
+    return make_shared<StopGradient>(new_args.at(0));
+}
+
+void op::StopGradient::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
```
mbrookhart: I think if you don't provide a generate_adjoints, ngraph will default to returning zeros? You probably don't need this function.

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -1495,6 +1496,14 @@ namespace ngraph
...
                 }
             }
+            template <>
+            void CPU_Emitter::EMITTER_DECL(ngraph::op::StopGradient)
+            {
+                writer.block_begin();
+                writer << "memcpy(" << out[0].get_name() << ", " << args[0].get_name() << ", "
+                       << out[0].get_size() * out[0].get_element_type().size() << ");\n";
+                writer.block_end();
+            }
```
mbrookhart: I think CPU has a no-op emitter you can use instead to avoid the memcpy?

ashokei:
thanks @mbrookhart for quick review, i removed adjoints gen, but i notice we need a copy as we need to pass through the inputs, if i use nop i do not see correct results.

src/ngraph/runtime/cpu/cpu_emitter.cpp
ashokei: nop does not seem to do pass through inputs (stop gradient expects identity/copy function on forward)

diyessi: Why not just pattern-match this into oblivion during passes shared by all transformers? Then you don't need anything for the transformers.

PR #1068, opened 2018-06-01
Floating point comparison with ULP

yxlao:
- Added `close_f` for comparing floating points
- Only converted selected unit test to use `all_close_f` for NNP


test/autodiff.in.cpp
```diff
@@ -1364,7 +1364,7 @@ NGRAPH_TEST(${BACKEND_NAME}, backwards_tan)
     auto pi = 3.14159f;
 
     // Stay away from the asymptotes at 6 and 12 o'clock.
-    auto slop = 0.1f;
+    auto slop = 0.2f;
```
yxlao: To stay further away from 6 and 12 o'clock

aprocter: No objection from me... `0.1f` was kind of arbitrary in the first place.

test/util/all_close.hpp
```diff
@@ -39,15 +39,17 @@ namespace ngraph
...
                        T atol = static_cast<T>(1e-8))
         {
+            bool rc = true;
             assert(a.size() == b.size());
             for (size_t i = 0; i < a.size(); ++i)
             {
                 if (std::abs(a[i] - b[i]) > atol + rtol * std::abs(b[i]))
                 {
-                    return false;
+                    NGRAPH_INFO << a[i] << " !≈ " << b[i];
```
aprocter: Minor nit: we should probably stay within 7-bit ASCII, which precludes using `≈`

test/util/all_close_f.cpp
```diff
@@ -0,0 +1,112 @@
...
+    if (a.size() != b.size())
+    {
+        throw ngraph_error("a.size() != b.size() for all_close comparison.");
+    }
+    for (size_t i = 0; i < a.size(); ++i)
+    {
+        bool is_close_f = close_f(a[i], b[i], mantissa_bits, tolerance_bits);
+        if (!is_close_f)
+        {
+            NGRAPH_INFO << a[i] << " !≈ " << b[i];
```
aprocter: Same minor nit as above on use of `≈`

test/util/test_tools.cpp
```diff
@@ -23,6 +23,74 @@
...
+        float_vec = vector<float>(vec.begin(), vec.end());
+    }
+    else if (element_type == element::u64)
+    {
+        vector<uint64_t> vec = read_vector<uint64_t>(tv);
+        float_vec = vector<float>(vec.begin(), vec.end());
+    }
+    else
+    {
+        throw ngraph_error("Unsupported ngraph element type.");
```
aprocter: n`G`raph :)

yxlao:
good point @ransford2011 , added `nan` and `inf` handling

PR #1071, opened 2018-06-02
ngraph-1676 batch dot fusion

louisfeng:
BatchDot fuses multiple dot operations.


src/ngraph/runtime/cpu/op/batch_dot.cpp
```diff
@@ -0,0 +1,66 @@
...
+    NGRAPH_DEBUG << "dot_dimension_a = " << dot_dimension_a
+                 << " , dot_dimension_b = " << dot_dimension_b;
+    NGRAPH_DEBUG << "a shape = " << vector_to_string(shape_a)
+                 << " , b shape = " << vector_to_string(shape_b);
+
+    if (shape_a.at(dot_dimension_a) != shape_b.at(dot_dimension_b))
+    {
+        throw ngraph_error("product dimensions are not equal while creating BatchDot");
+    }
+
```
pruthviIntel: Can we also add the check, if the element type for a and b are same?

louisfeng: Good point, I'll add that.

src/ngraph/runtime/cpu/pass/cpu_mat_fusion.cpp
```diff
@@ -199,3 +201,104 @@ bool runtime::cpu::pass::CPURnnMatFusion::run_on_function(std::shared_ptr<Functi
     }
     return modified;
 }
+
+#define TI(x) std::type_index(typeid(x))
+
+std::shared_ptr<Node> fuse_batch_dot(const std::shared_ptr<Node>& n)
+{
+    auto reshape_pred = [](const std::shared_ptr<Node>& r) {
```
Krovatkin: could we please use `has_class` to reduce the boiler-plate?

src/ngraph/runtime/cpu/pass/cpu_mat_fusion.cpp
```diff
@@ -199,3 +201,104 @@ bool runtime::cpu::pass::CPURnnMatFusion::run_on_function(std::shared_ptr<Functi
...
+    return {nullptr};
+}
+
+bool runtime::cpu::pass::CPUBatchDotFusion::run_on_function(std::shared_ptr<Function> func)
+{
+    bool modified = false;
+
+    for (auto n : func->get_ordered_ops())
+    {
+        const Node& node = *n;
```
Krovatkin: if we wud just do `std::dynamic_pointer_cast<op::Concat>` we could avoid all this `*n` and `TI` class comparisons 😃

src/ngraph/runtime/cpu/pass/cpu_mat_fusion.cpp
```diff
@@ -199,3 +201,104 @@ bool runtime::cpu::pass::CPURnnMatFusion::run_on_function(std::shared_ptr<Functi
...
+                }
+            }
+        }
+    }
+    if (fuse_input[0] && fuse_input[1])
+    {
+        return std::make_shared<op::BatchDot>(
+            fuse_input[0], fuse_input[1], transpose[0], transpose[1]);
+    }
+    return {nullptr};
```
Krovatkin: what do we need the brackets for?

src/ngraph/runtime/cpu/pass/cpu_mat_fusion.cpp
```diff
@@ -199,3 +201,104 @@ bool runtime::cpu::pass::CPURnnMatFusion::run_on_function(std::shared_ptr<Functi
...
+                        ++reshape_count[i];
+                        if (reshape_count[i] == num_expected_reshape_with_trans)
+                        {
+                            transpose[i] = true;
+                            break;
+                        }
+                    }
+                    // branch to either input 0 or 1 depending on which one we are traversing
+                    iter =
+                        iter->get_input_size() > 1 ? iter->get_argument(i) : iter->get_argument(0);
```
Krovatkin: this is cool  😄

src/ngraph/runtime/cpu/pass/cpu_mat_fusion.cpp
```diff
@@ -199,3 +201,104 @@ bool runtime::cpu::pass::CPURnnMatFusion::run_on_function(std::shared_ptr<Functi
...
+            // we found a match, determine whether we have to transpose for each input by
+            // counting the number of reshapes in each branch, if transpose is applied, there
+            // should be 3 reshapes.
+            for (int i = 0; i < num_op_branches; ++i)
+            {
+                auto iter = matcher->get_match_root();
+                auto& input_node = pattern_map[input[i]];
+                do
+                {
+                    if (std::dynamic_pointer_cast<op::Reshape>(iter) != nullptr)
```
Krovatkin: do we want to check what those reshapes actually do (i.e. transpose or reshape or god forbids both at the same time) ?

src/ngraph/runtime/cpu/pass/cpu_mat_fusion.cpp
louisfeng: Return type is `shared_ptr`

src/ngraph/runtime/cpu/pass/cpu_mat_fusion.cpp
louisfeng: Copy&paste our conversation: Dynamic cast actually works for derived class too. So there's potentially ambiguity in this check and we want `concat` exactly, not potentially a derived class of `concat`

PR #1073, opened 2018-06-04
IA Transformer - direct execution 1

jmenon:


src/ngraph/runtime/cpu/cpu_builder.cpp
```diff
@@ -0,0 +1,286 @@
...
+#ifdef NGRAPH_DISTRIBUTED
+#include <mpi.h>
+#include "ngraph/op/allreduce.hpp"
+#endif
+
+using namespace std;
+using namespace ngraph;
+
+// Per-type kernel macro
+#define SELECT_KERNEL(KV, ET, K)                                                                   \
```
jbobba: We will need kernel selection logic to consider input element types as well especially once we get to quantization.

jbobba:
TESTNOW

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -679,21 +683,21 @@ using namespace ngraph::runtime;
...
-            shared_ptr<descriptor::TensorView> otv = op->get_output_tensor_view();
-            vector<size_t>& al = output_alias_map[otv.get()];
-            al.push_back(output_index);
-            if (al.size() > 1)
-            {
-                aliases.push_back(output_index);
-            }
-            output_index++;
-        }
+          size_t output_index = 0;
```
jbobba: needs style check.

src/ngraph/runtime/cpu/cpu_builder.cpp
diyessi: This should get an "unsupported element type" kind of error if no kernel is found, rather than whatever you get when you try to call a function that hasn't been initialized.

src/ngraph/runtime/cpu/cpu_builder.cpp
```diff
@@ -0,0 +1,286 @@
...
+                    {
+                        dest.push_back(&tensor_data[result->get_output_tensor(0).get_name()]);
+                    }
+                }
+                auto& src = tensor_data[node->get_output_tensor(0).get_name()];
+                auto size = node->get_output_tensor(0).size();
+                auto functor = [&, dest, src, size](CPURuntimeContext* ctx) {
+                    for (auto p : dest)
+                    {
+                        memcpy(*p, src, size);
```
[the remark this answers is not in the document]
diyessi: I think @rkimballn1 has a way to do that.

src/ngraph/runtime/cpu/cpu_external_function.cpp
diyessi: @jbobba it's all commented out. I'd prefer a #if 0 ... #endif with a comment since the indentation doest get messed up like this and you don't need to worry about nested comments.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -956,6 +960,34 @@ using namespace ngraph::runtime;
         writer += "}\n\n";
     }
 
+    // TODO: Cleanup and make this a utility function
+    file_util::make_directory(s_output_dir);
```
diyessi: What is cleaning up these files?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -991,35 +1024,188 @@ using namespace ngraph::runtime;
...
+        if (!res->needs_copy())
+        {
+            shared_ptr<descriptor::TensorView> itv =
+                res->get_inputs().at(0).get_output().get_tensor_view();
+            function_output_index[itv->get_tensor().get_name()] = i;
+        }
+    }
+
+    // Intermediates
+    assert(tensor_data.empty());
```
diyessi: throw instead of assert.

src/ngraph/runtime/cpu/cpu_builder.hpp
```diff
@@ -0,0 +1,75 @@
...
+                }
+
+                static void nop(CPU_ExternalFunction* external_function,
+                                const ngraph::Node* node,
+                                const std::vector<TensorViewWrapper>& args,
+                                const std::vector<TensorViewWrapper>& out)
+                {
+                }
+
+                static void buildBatchNorm(CPU_ExternalFunction* external_function,
```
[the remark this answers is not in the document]
jmenon: Yes, copied from cpu_emitter.cpp. Don't have much for context for why it was done that way but the extra append_relu sounds like it

src/ngraph/runtime/cpu/cpu_builder.cpp
jmenon: Add a TODO item here then? The kernels currently assume a single source and output type so we'd need at least one kernel which supports that

jmenon: This would trigger a compile error which IMO is the right thing to do.

src/ngraph/runtime/cpu/cpu_external_function.cpp
jmenon: Thanks for catching that. Should I just remove it? Anyone want to make a case for why it should be there?

src/ngraph/runtime/cpu/cpu_external_function.cpp
jmenon: Changed

src/ngraph/runtime/cpu/cpu_external_function.cpp
jmenon: This is existing code which was moved to place it closer to the codegen path. 
`StaticInitializers() { ngraph::file_util::remove_directory(s_output_dir); }`

src/ngraph/runtime/cpu/cpu_external_function.cpp
jbobba: Remove it.

jbobba:
Replaced by #1092

PR #1078, opened 2018-06-04
Disable one_hot tests which throw exceptions.

csullivan:
The GPU transformer doesn't yet handle these special cases in op::OneHot and so these tests should be failing but are not on some systems. Disabling them until they are handled.


fengleitian:
Those test should not be enabled for GPU.

PR #1087, opened 2018-06-05
workaround for nGraph python runtime tensor write data corruption

adstraw:


python/ngraph/runtime.py
```diff
@@ -120,7 +120,9 @@ def _write_ndarray_to_tensor_view(value, tensor_view):
 
         buffer_size = Computation._get_buffer_size(
             tensor_view.element_type, tensor_view.element_count)
-        tensor_view.write(util.numpy_to_c(np.ascontiguousarray(value)), 0, buffer_size)
+
+        nparray = np.ascontiguousarray(value)
```
RanACohen: Does not look like a workaround. looks like a fix to me.
If you really want to avoid memory copy you need to add strides. but IE will not support it so the copy will happen anyhow.

PR #1088, opened 2018-06-06
Fix proxy settings in contrib/docker so that builds work both on the Internet and within Intel

crlishka:
Implements fix described in NGRAPH-1999

PR#1062 removed all proxy settings from ngraph/contrib/docker, since the proxy settings were preventing developers (external to Intel) from using the docker builds. Presumably, since nGraph builds had moved to cje-algo, nothing was using the build system in ngraph/contrib/docker.
Unfortunately, the ngraph-mxnet Jenkins automation was still using the ngraph/contrib/docker build system, and PR#1062 broke the ngraph-mxnet CI merge-check builds.

ChrisL came up with a quick switch that checks whether the hostname contains ".intel.com", and if so uses proxy settings that are appropriate for Intel's network. If not, then no proxies are used – which should work for external developers. Nguyen, Lam and Ketineni, Rama requested that this switch be added to ngraph/contrib/docker's build system, to unblock the broken ngraph-mxnet builds.

This PR implements the fix.

### DIFF
--- a/contrib/docker/make-dimage.sh
+++ b/contrib/docker/make-dimage.sh
@@ -49,7 +49,13 @@ DIMAGE_VERSION=`date -Iseconds | sed -e 's/:/-/g'`
 
 DIMAGE_ID="${DIMAGE_NAME}:${DIMAGE_VERSION}"
 
-cd ${CONTEXTDIR}
+# Auto-detect proxy settings when running from within Intel's network
+hostname | grep '\.intel\.com$'
+if [ $? = 0 ] ; then  # Within .intel.com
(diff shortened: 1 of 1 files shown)


crlishka:
Testing with ngraph-mxnet-unittest passes:

https://aipg-rancher.intel.com/jenkins/algo/job/ngraph-mxnet-unittest/367/

https://aipg-rancher.intel.com/jenkins/algo/job/ngraph-mxnet-unittest/374/ (in progress, but has made it successfully past the builds).

Manual build tests on command-line also pass, and CI merge-checks pass.

lnguyen-nvn:
👍

PR #1070, opened 2018-06-02
CUDA softmax kernel and broadcast kernel support for multiple non-consecutive axes

csullivan:
This PR introduces fused elementwise+collective kernels. op::Softmax is implemented in the GPU via two of these kernels: 
```
1. elementwise exponentiation fused with sum reduce
2. elementwise division fused with broadcast
```

Features:
* Adds softmax cuda kernel to handle activation along arbitrary axes.
* Adds a few new softmax tests
* Adds broadcast to the CUDAEmitter with support for all axis broadcasting (coordinate transformation function is shared in softmax and broadcast)

Limitations:
* The softmax kernel relies on atomic updates to global memory which can create contention. I began work on a hierarchical reduction in cs/gpu_softmax_cuda_shfl, but decided to postpone further efforts until later.

PR #1095, opened 2018-06-08
Zero init. GPU workspace allocations by default

csullivan:
[Please merge this prior to PR #1070]

With this change, every time a workspace memory address is requested (invoked at JIT runtime), the device memory will be zeroed by default. It can be turned off by setting zero_initialize=false in order to cache results between kernels.


csullivan:
Please merge this prior to PR #1070

PR #1097, opened 2018-06-11
Replace Check

Krovatkin:


src/ngraph/descriptor/input.cpp
```diff
@@ -37,6 +37,13 @@ void Input::replace_output(Output& new_output)
     new_output.add_input(this);
     m_output = &new_output;
     m_src_node = std::shared_ptr<Node>(new_output.get_node());
+
+    static const auto nerc = std::getenv("NGRAPH_ENABLE_REPLACE_CHECK");
+
+    if (nerc)
+    {
+        (this->m_node->copy_with_new_args(this->m_node->get_arguments()));
```
aprocter: Looks like the `shared_ptr` is dying immediately, which makes sense (we're just calling `copy_with_new_args` for its validating side effects). Could you throw in a comment to explain this?

Krovatkin:
@aprocter added a comment.

PR #1098, opened 2018-06-11
OS X support

megawidget:
* sane extensions for Mac
* llvm 8.1 compatibility (older version TensorFlow gets compiled with)


src/ngraph/CMakeLists.txt
```diff
@@ -345,7 +345,7 @@ endif()
 
 if (APPLE)
     set_property(TARGET ngraph PROPERTY PREFIX "lib")
-    set_property(TARGET ngraph PROPERTY OUTPUT_NAME "ngraph.so")
+    set_property(TARGET ngraph PROPERTY OUTPUT_NAME "ngraph.dylib")
```
rkimballn1: Not sure what these three lines do, but it looks like they construct the output name. Do we have to use this? What happens if we don't use these three lines?
If you want to use these then let's use them what looks to be the correct way. Set the SUFFIX on line 349 to be ${CMAKE_SHARED_LIBRARY_SUFFIX}. 
I suspect we can remove lines 346 - 350

src/ngraph/runtime/backend.cpp
```diff
@@ -43,7 +43,12 @@ runtime::Backend::~Backend()
 void* runtime::Backend::open_shared_library(const string& type)
 {
     void* handle = nullptr;
-    string name = "lib" + to_lower(type) + "_backend.so";
+    const string name = "lib" + to_lower(type) +
```
rkimballn1: Merge in the lastest master to get a better fix for this.

test/ngraph.cpp
```diff
@@ -29,7 +29,8 @@ using namespace std;
 TEST(NGraph, loadTest)
 {
     // load the triangle library
-    void* ngraphImplLib = dlopen("../src/libngraph.so", RTLD_LAZY);
+    void* ngraphImplLib = dlopen("../src/libngraph.so", RTLD_LAZY) ||
```
rkimballn1: We could do this the same way it is done in backend.cpp, that way it supports all targets.

megawidget:
needs backmerge from master, blocked by #1112

src/ngraph/runtime/backend.cpp
rkimballn1: Can you just pull this file from master. The code on master supports linux, macos and windows.

PR #1099, opened 2018-06-11
gpu deconvolution

fengleitian:
This PR add deconvolution support for` convolution`, `convolutionbackwarddata` and `convolutionbackwardfilter`, some changes as following:
1. Add padding with data dilation support. (NOTE: Due to the limitation of CUDNN.  we use a preprocessing to pad the input with data dilation. This is not the best way to do it. Will revisit it during the optimization pass)
2. Changed `get_padded_shape` to support `padding_interior` which is `data_dilation` for convolution.
3. `pad_dilation` is actually a full functional padding with `padding_interior`, will have another PR to consolidate with `pad` op.


src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -231,6 +231,106 @@ size_t runtime::gpu::CUDAEmitter::build_pad(const runtime::gpu::GPURuntimeContex
     return primitive_index;
 }
 
+size_t runtime::gpu::CUDAEmitter::build_pad_dilation(const runtime::gpu::GPURuntimeContext* ctx,
```
csullivan: `build_pad` actually supports interior padding and so this kernel and the other perform the same function. 

One interesting point to consider, `build_pad` statically compiles the padding op and so doesn't require passing of input shapes/strides/padding etc to the kernel, whereas `build_pad_dilation` does the same thing but dynamically, passing in that information. It would be interesting to do a performance comparison between the two in the future. I have been wondering for some time if we can't get some performance boost from emitting shape information directly into the kernels (statically) at the cost of compiling a new kernel for every shape. 

My thought is that if there are many different shapes in a model, the compile time would become expensive and a dynamic kernel, such as `build_pad_dilation` would be better. On the other hand, if an op is called a lot with the same shape (e.g. LSTM?), the static version is likely faster as there are no memory accesses for kernel arguments.

Perhaps it is worth keeping `build_pad_dialation` and renaming it to `build_pad_dynamic` and `build_pad` to `build_pad_static`, so that we can test this out in the optimization pass. If not, at least keep it around in a branch.

fengleitian: good point. Let's keep it here. We might need both for different purpose.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -162,69 +162,71 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                        input_shape, padding_below, padding_above, padding_interior);
+                    Shape input_padded_strides = row_major_strides(input_shape_padded);
                     auto temp_size =
                         shape_size(input_shape_padded) * args[0].get_element_type().size();
                     GPUAllocator allocator =
                         external_function->get_primitive_emitter()->get_memory_allocator();
                     size_t idx_workspace = allocator.reserve_workspace(temp_size);
                     writer << "void* pad_buffer = runtime::gpu::invoke_memory_primitive(ctx, "
                            << idx_workspace << ");\n";
+                    writer << "std::vector<" << args[0].get_type() << "> pad_buffer_host("
```
csullivan: Use GPUAllocator::reserve_argspace for this to avoid the memcpy at runtime.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -162,69 +162,71 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
...
+                           << shape_size(input_shape_padded) << ", 0);\n";
```
csullivan: Will the pad value always be zero? `build_pad` supports padding value other than 0.

src/ngraph/runtime/gpu/gpu_emitter.cpp
csullivan: Actually, I'm not sure that is even necessary for the 0 case, because the workspace is zero initialized. Is it required?

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -231,6 +231,106 @@ size_t runtime::gpu::CUDAEmitter::build_pad(const runtime::gpu::GPURuntimeContex
...
+                                                     const std::array<std::string, 2>& dtypes,
+                                                     const Shape& input_shape,
```
csullivan: Should these be `GPUShape`?

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: it can be set to any value.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: It depends. here it's necessary, since I don't handle padded area in kernel.

csullivan: In this case you just need to ensure the space is zeroed, which occurs when `runtime::gpu::invoke_memory_primitive` is called - so I'm not sure the memcpy of the zero-vector is needed. Unless the workspace is used multiple times within the kernel? It doesn't look to be that way though

src/ngraph/runtime/gpu/gpu_emitter.cpp
csullivan: Should it be part of the kernel?

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: we used two different approach for padding. 
1. In you case, you handle padded area in kernel. No need initialization. Number of threads will be the output size. 
2. In my case, I handle padded area during initialization. Number of threads will be input size.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: I can be. as described above, there are two methods to do it. I not sure which one is better yet.

src/ngraph/runtime/gpu/cuda_emitter.hpp
```diff
@@ -47,6 +47,13 @@ namespace ngraph
                                  GPUShape pad_interior,
                                  const std::string& pad_value = "");
 
+                size_t build_pad_dynamic(const runtime::gpu::GPURuntimeContext* ctx,
+                                         const std::array<std::string, 2>& dtypes,
+                                         const Shape& input_shape,
```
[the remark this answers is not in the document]
fengleitian: I did changed it to GPUShape in one branch, but that branch got messed up. will give another try.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -231,6 +231,106 @@ size_t runtime::gpu::CUDAEmitter::build_pad(const runtime::gpu::GPURuntimeContex
...
+    auto compiled_kernel = ctx->compiled_kernel_pool->get(kernel_name.str());
+    if (compiled_kernel == nullptr)
+    {
+        codegen::CodeWriter writer;
+        CudaKernelBuilder::add_pod_typedefs(writer);
+        CudaKernelBuilder::get_pad_dynamic_op(writer, kernel_name.str(), dtypes);
+        compiled_kernel = ctx->compiled_kernel_pool->set(kernel_name.str(), writer.get_code());
+    }
+
+    unsigned int rank = static_cast<unsigned int>(input_shape.size());
```
rkimballn1: We are using "unsigned int" here but things like int64_t a few lines down. I "unsigned int" a cuda thing? It would be nice to be consistent unless the cuda API uses these types.

fengleitian: @rkimballn1 Yes, this `unsigned int` will pass to cuda. We are trying to limit the index to 32bit on GPU side for now. But still use `int64_t` or `size_t` on CPU side.

PR #1101, opened 2018-06-12
Ubuntu 18 build support

rkimballn1:
Works for me ;)


cmake/external_mkldnn.cmake
```diff
@@ -48,6 +48,7 @@ set(MKLVERSION "2018.0.3.20180406")
 if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
     set(MKLPACKAGE "mklml_lnx_${MKLVERSION}.tgz")
     set(MKL_SHA1_HASH aea0d9ce65773cfcf5d8292b8db553bde965fc8f)
+    set(MKL_LIBS libiomp5.so libmklml_intel.so)
```
aprocter: I'm guessing a similar line will be needed for `Darwin` but we can deal with that later.

src/resource/main.cpp
```diff
@@ -94,18 +94,18 @@ int main(int argc, char** argv)
 #endif
     include_paths.push_back({NGRAPH_HEADERS_PATH, {}, true});
     include_paths.push_back({CLANG_BUILTIN_HEADERS_PATH, {}, true});
-    include_paths.push_back({"/Library/Developer/CommandLineTools/usr/include/c++/v1", {}});
+// include_paths.push_back({"/Library/Developer/CommandLineTools/usr/include/c++/v1", {}});
```
aprocter: Was this change intentional?

cmake/external_mkldnn.cmake
rkimballn1: yep

rkimballn1: and windows too

src/resource/main.cpp
rkimballn1: Yes. No longer embedding system headers in the resource file, just the headers that are from the ngraph directory. The system headers still exist after you delete the repo directory but the ngraph headers don't so they must be preserved. We COULD "install" the ngraph headers and solve the problem that way.

PR #1108, opened 2018-06-13
DEX Part 2

jmenon:


src/ngraph/runtime/cpu/cpu_builder.cpp
```diff
@@ -247,6 +288,151 @@ namespace ngraph
...
+                if (args.size() > 2)
+                {
+                    auto& arg2_tensor = tensor_data[args[2].get_name()];
+
+                    auto axes = mm->get_broadcast_axes();
+                    if (axes.size() == 1)
+                    {
+                        if (*(axes.begin()) == 0)
+                        {
+                            vector<float> ones_row(arg2_shape[0], 1.0f);
```
jbobba: when does this get cleaned up?

src/ngraph/runtime/cpu/cpu_builder.cpp
```diff
@@ -228,6 +231,44 @@ namespace ngraph
...
+                auto element_count = out[0].get_size();
+                auto& arg0_tensor = tensor_data[args[0].get_name()];
+                auto& out0_tensor = tensor_data[out[0].get_name()];
+
+                auto functor = [&, kernel, element_count](CPURuntimeContext* ctx) {
+                    kernel(arg0_tensor, out0_tensor, element_count);
+                };
+                functors.emplace_back(functor);
+            }
+
```
jbobba: How about a MACRO here for defining these functions?

src/ngraph/runtime/cpu/cpu_builder.cpp
jmenon: local one gets destroyed after end of scope, captured one when all call frames have gone out of scope

src/ngraph/runtime/cpu/cpu_builder.cpp
jmenon: Sure, maybe a helper function even. The select kernel by type and rank macros are already too unwieldy

src/ngraph/runtime/cpu/cpu_builder.cpp
jbobba: never mind, i missed the explicit capture.

aprocter:
TESTNOW

PR #1111, opened 2018-06-14
gpu function call

fengleitian:
Add gpu function call op.

### DIFF
--- a/src/ngraph/runtime/gpu/gpu_emitter.cpp
+++ b/src/ngraph/runtime/gpu/gpu_emitter.cpp
@@ -1004,8 +1004,8 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
 
                 writer.block_begin();
                 {
-                    vector<string> input_names;
-                    vector<string> output_names;
+                    std::vector<string> input_names;
+                    std::vector<string> output_names;
 
                     for (const runtime::gpu::GPU_TensorViewWrapper& input : args)
                     {
--- a/src/ngraph/runtime/gpu/gpu_emitter.cpp
+++ b/src/ngraph/runtime/gpu/gpu_emitter.cpp
@@ -1152,11 +1152,6 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
             }
 
             template <>
-            void GPU_Emitter::EMITTER_DECL(ngraph::op::FunctionCall)
-            {
-            }
-
-            template <>
             void GPU_Emitter::EMITTER_DECL(ngraph::op::Multiply)
             {
                 if (out[0].get_size() == 0)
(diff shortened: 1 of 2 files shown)


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -996,6 +996,45 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
...
 
+            template <>
+            void GPU_Emitter::EMITTER_DECL(ngraph::op::FunctionCall)
+            {
+                auto function_call = static_cast<const ngraph::op::FunctionCall*>(node);
+                shared_ptr<Function> function = function_call->get_functions()[0];
+
+                writer.block_begin();
+                {
+                    vector<string> input_names;
```
csullivan: nit: can we use std:: for standard lib items? I prefer it, personally, and we do so already in gpu_emitter.cpp, so it would be good to be consistent.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -996,6 +996,45 @@ CUDNN_SAFE_CALL(cudnnSetOpTensorDescriptor(opTensorDesc,
                 writer.block_end();
             }
 
+            template <>
+            void GPU_Emitter::EMITTER_DECL(ngraph::op::FunctionCall)
+            {
+                auto function_call = static_cast<const ngraph::op::FunctionCall*>(node);
+                shared_ptr<Function> function = function_call->get_functions()[0];
```
csullivan: Why only the first function?

csullivan:
Does this handle recursive function calls?

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: check the spec, it's actually a single function even it use vector. @bob correct me if I'm wrong.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: me too, I just did a copy and didn't change it.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: https://github.com/NervanaSystems/ngraph/blob/3d66cba498b0c8076a82b52d29a4d25a691698f5/src/ngraph/op/function_call.cpp#L63

csullivan: Ah I see, an overriding function that in this case it returns a singleton vector.

nit: could consider using `get_functions().back()` to suggest it's a unit length vector

fengleitian:
It should handle recursive function calls, the test is actually have depth 2.
https://github.com/NervanaSystems/ngraph/blob/3d66cba498b0c8076a82b52d29a4d25a691698f5/test/backend_test.in.cpp#L1597

PR #1112, opened 2018-06-14
Fix build for MacOS

rkimballn1:
Builds and unit tests pass for CPU and INTERPRETER backends. Did not test GPU on macos because I can't.

The change in loop_kernel.?pp was because m_outputs shadows the variable of the same name in class Node and I was getting a compiler warning.


src/ngraph/codegen/file_util.cpp
```diff
@@ -0,0 +1,349 @@
...
+        path = "/tmp";
+    }
+
+    return path;
+}
+
+vector<char> codegen::file_util::read_file_contents(const string& path)
+{
+    size_t file_size = get_file_size(path);
+    vector<char> data;
```
megawidget: why not just `vector<char> data(file_size);`?

src/ngraph/codegen/file_util.cpp
```diff
@@ -0,0 +1,349 @@
...
+            case DT_REG:
+            {
+                string file_name = file_util::path_join(path, name);
+                func(file_name, false);
+                break;
+            }
+            default: break;
+            }
+        }
+        closedir(dir);
```
megawidget: what if `func` throws? you'll leak the pointer?  should `func` be `noexcept`?

src/ngraph/codegen/file_util.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace ngraph
+{
+    namespace codegen
+    {
+        class file_util;
+    }
+}
+
+class ngraph::codegen::file_util
```
megawidget: i'm a bit confused why this is a class and not a namespace, but :okay:

src/ngraph/codegen/compiler.cpp
```diff
@@ -260,6 +278,21 @@ bool codegen::StaticCompiler::is_version_number(const string& path)
...
 
+static template <typename U, typename T>
+bool contains(const U& container, const T& obj)
+{
+    bool rc = false;
+    for (auto o : container)
+    {
+        if (o == obj)
+        {
+            rc = true;
```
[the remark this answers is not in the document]
rkimballn1: Yes, but I believe in single return value.

src/ngraph/codegen/file_util.cpp
rkimballn1: My understanding of noexcept is that it is meaningless. A noexcept function can throw exceptions.

megawidget: ok, so it needs a guard then?

avijit-nervana:
Replacing myself with @diyessi so that Scott can assign someone from the nGraph core team.

src/ngraph/codegen/compiler.cpp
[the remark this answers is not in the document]
diyessi: @megawidget 's is simpler, but even simpler is
`return std::find(container.begin(), container.end(), obj) != container.end()`

src/ngraph/codegen/file_util.cpp
diyessi: yes it does.

src/ngraph/codegen/file_util.cpp
```diff
@@ -0,0 +1,341 @@
...
+        rc = tmpname;
+    }
+
+    free(tmpname);
+    return rc;
+}
+
+std::string codegen::file_util::get_temp_directory()
+{
+    const vector<string> potential_tmps = {"NGRAPH_TMP", "TMPDIR", "TMP", "TEMP", "TEMPDIR"};
```
diyessi: This list seems like overkill.

src/ngraph/codegen/file_util.cpp
```diff
@@ -0,0 +1,341 @@
...
+    if (mkdtemp(tmpname))
+    {
+        rc = tmpname;
+    }
+
+    free(tmpname);
+    return rc;
+}
+
+std::string codegen::file_util::get_temp_directory()
```
diyessi: I think `get_temp_directory_path` is a bette name.

src/ngraph/codegen/file_util.cpp
```diff
@@ -0,0 +1,341 @@
...
+    {
+        path = getenv(var.c_str());
+        if (path != nullptr)
+        {
+            break;
+        }
+    }
+    if (path == nullptr)
+    {
+        path = "/tmp";
```
diyessi: This won't work on native windows will it?

src/ngraph/codegen/file_util.hpp
diyessi: I agree -- it should be a namespace.

src/ngraph/codegen/file_util.hpp
```diff
@@ -0,0 +1,59 @@
...
+{
+public:
+    static std::string get_file_name(const std::string&);
```
diyessi: These should all have short doc strings.

src/ngraph/codegen/compiler.cpp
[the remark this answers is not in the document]
rkimballn1: I removed this function as it exists elsewhere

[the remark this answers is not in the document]
rkimballn1: Also @diyessi your solution does not work for all instances of the templated function, it does not work with unordered_set. Less code is not always better nor more readable.

src/ngraph/codegen/file_util.cpp
rkimballn1: This list is to cover multiple operating systems. I would rather have too many than too few

src/ngraph/codegen/file_util.cpp
rkimballn1: Windows defines both TEMP and TMP so the environment variable defining the temp directory will be found before we get here.

rkimballn1:
TESTNOW

PR #1114, opened 2018-06-15
Backend/API: First IntelGPU backend based on clDNN with empty functions

shssf:
This is the first "empty" implementation of the backed which will use clDNN library.


shssf:
I resubmitted this to #1116

PR #1116, opened 2018-06-16
Backend/API: First IntelGPU backend based on clDNN with empty functions

shssf:
This is the first "empty" implementation of the backed which will use clDNN library.


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -0,0 +1,52 @@
+/*******************************************************************************
```
rkimballn1: This comment block should have all of the `*` at column zero. This is usually caused by an incorrect version for clang format. We use clang-format 3.9 (or 3.9.1 as I recall) to get consistent formatting. You will have to manually edit this as clang-format won't put it back they way we want it.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+#include "ngraph/runtime/intelgpu/intelgpu_backend.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+extern "C" void create_backend()
```
diyessi: `create_backend` should be in a C header included in an `extern "C"` by `intelgpu_backend.hpp`. Then you can leave it out here, and you don't need to worry about headers and implementations getting out of sync.

You also want it to be `create_backend(void)` since in C `create_backend()` means you can pass anything you want to the function.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -0,0 +1,52 @@
...
+{
+    runtime::Backend::register_backend("INTELGPU",
+                                       make_shared<runtime::intelgpu::IntelGPUBackend>());
+};
+
+shared_ptr<runtime::TensorView>
+    runtime::intelgpu::IntelGPUBackend::create_tensor(const element::Type& element_type,
+                                                      const Shape& shape)
+{
+    throw runtime_error("IntelGPUBackend::create_tensor: Not implemented yet");
```
rkimballn1: You can probably just steal the CPU implementation

src/ngraph/runtime/intelgpu/unit_test.manifest
```diff
@@ -0,0 +1 @@
+intelgpu_layouts
```
rkimballn1: If you want I can give you a complete list of all unit tests you need to exclude.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
shssf: I did this in the way the same as for all other backends. If you like refactor the code I could do this in separate PR.

CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ include(var_functions)
 option(NGRAPH_UNIT_TEST_ENABLE "Control the building of unit tests" TRUE)
 option(NGRAPH_TOOLS_ENABLE "Control the building of tool" TRUE)
 option(NGRAPH_CPU_ENABLE "Control the building of the CPU backend" TRUE)
+option(NGRAPH_INTELGPU_ENABLE "Control the building of the CPU backend with clDNN" TRUE)
```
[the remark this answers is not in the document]
shssf: I implement it as TRUE to involve the new code into regular build procedure for all other development work (in all incoming PRs). If we keep it switched FALSE it will be potentially broken by other development work due no regular build tests.
Anyway, I will keep in FALSE as you like.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
shssf: I will do this in separate PR with first implementation of the create_tensor() functions.

CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ include(var_functions)
 option(NGRAPH_UNIT_TEST_ENABLE "Control the building of unit tests" TRUE)
 option(NGRAPH_TOOLS_ENABLE "Control the building of tool" TRUE)
 option(NGRAPH_CPU_ENABLE "Control the building of the CPU backend" TRUE)
+option(NGRAPH_INTELGPU_ENABLE "Control the building of the Intel GPU backend with clDNN" FALSE)
```
fengleitian: `NGRAPH_INTELGPU`: Since this is based on CL, I suppose it will also support devices that can run CL, not just intel GPU. `NGRAPH_GPU` is not all GPU but only CUDA. maybe we should discuss about the name.

WilliamTambellini: I agree that the name should better be something like  NGRAPH_CL : for some RNN and some models of CPU, I have seen Intel OpenCL CPU drivers (for exple opencl-1.2-6.4.0.25) faster than Eigen : in other words that CL backend, could also be useful for some non gpu hardware.

shssf: Propose exact name for this variable please.

rkimballn1: I suggest that we name the backends for the APIs that they are programmed to, not what is behind that API. Currently we have GPU which is really CUDA with an Nvidia GPU behind it and should really be called CUDA: NGRAPH_CUDA_ENABLE, and backend::create("CUDA")
This backend is for OpenCL, specifically targeting the Intel GPU but can be so much more. We could call it CL or OCL: NGRAPH_CL_ENABLE and backend::create("CL") or OCL or ???
Do we get support for OpenCL CPU drivers?
Do we get support for AMD GPUs?
This is not a simple decision and I don't want to make any changes without consideration.
I also don't want this PR stuck in limbo over a name. We can change it later and we already need to rename GPU.

shssf:
Is it ready for merging or something else is expected? We can't continue without this PR because we need directories skeleton.

PR #1118, opened 2018-06-17
Bob/cmake

rkimballn1:


louisfeng:
thanks I'm going to give it a try right now.

louisfeng:
Also when the mkldnn patch fails, the build would hang.

aprocter:
I wonder if the `--forward` option for `patch` will get around this?

rkimballn1:
My very limited understanding is that the patch step is only supposed to be run once after the package is downloaded and unpacked. I will look into it

PR #1122, opened 2018-06-19
enable cudnn datatype support

fengleitian:
Add datatype support for CuDNN.
1.  was only `float` supported, add `int32`, `double` and `int8`.
2. backward pooling does not support `int32` on cudnn. add 5 tests to `unit_test.manifest`
```
[  FAILED  ] GPU.backwards_maxpool_n4_c1_hw4_2x2_max
[  FAILED  ] GPU.backwards_maxpool_n2_c1_hw5_3x3_str2_max
[  FAILED  ] GPU.backwards_avgpool_n1_c1_hw2x2
[  FAILED  ] GPU.backwards_avgpool_n1_c1_hw4x4
[  FAILED  ] GPU.backwards_avgpool_n2_c2_hw4x4
```
Those test were accidentally passed by set tensor type to `float`. My guess is`int32` and `float` happen to get the same result for max and avg when only use Mantissa parts.

3. refactor binary op using CuDNN to `cudnn_emitter`.
4. add comments to `unit_test.manifest`
5. remove `dot_4d_5d_multi_axis_big_fp64_VERY_SLOW` which is not in tests anymore.


fengleitian:
add datatype support alpha, beta. change back to WIP.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -112,13 +112,57 @@ runtime::gpu::CUDNNEmitter::CUDNNEmitter(GPUPrimitiveEmitter* emitter)
 {
 }
 
+cudnnDataType_t runtime::gpu::CUDNNEmitter::getCudnnDataType(std::string dtype)
```
rkimballn1: function names are snake case

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -112,13 +112,57 @@ runtime::gpu::CUDNNEmitter::CUDNNEmitter(GPUPrimitiveEmitter* emitter)
...
+{
+    // CUDNN_DATA_FLOAT
```
rkimballn1: The comments seem redundant. The map provides the same info. I am opening this for debate. Do you agree?

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -112,13 +112,57 @@ runtime::gpu::CUDNNEmitter::CUDNNEmitter(GPUPrimitiveEmitter* emitter)
...
+    {
+        std::string err = dtype + "is not supported by CuDNN";
+        throw std::runtime_error(err);
+    }
+    return p->second;
+}
+
+void* runtime::gpu::CUDNNEmitter::getDataByType(cudnnDataType_t data_type, double value)
+{
+    void* r = NULL;
```
rkimballn1: use nullptr, not NULL

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: agree.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: fixed.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: fixed.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: fixed.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -112,13 +112,49 @@ runtime::gpu::CUDNNEmitter::CUDNNEmitter(GPUPrimitiveEmitter* emitter)
 {
 }
 
+cudnnDataType_t runtime::gpu::CUDNNEmitter::get_cudnn_datatype(std::string dtype)
```
csullivan: v nice.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -112,13 +112,49 @@ runtime::gpu::CUDNNEmitter::CUDNNEmitter(GPUPrimitiveEmitter* emitter)
...
+{
+    static const std::unordered_map<std::string, cudnnDataType_t> datatype_map{
+        {"float", CUDNN_DATA_FLOAT},
+        {"double", CUDNN_DATA_DOUBLE},
+        {"int8_t", CUDNN_DATA_INT8},
+        {"int32_t", CUDNN_DATA_INT32}};
+    auto p = datatype_map.find(dtype);
+    if (p == datatype_map.end())
+    {
+        std::string err = dtype + "is not supported by CuDNN";
```
csullivan: Maybe it doesn't matter, not sure though: lower case C, cuDNN. Same for other references in this file.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -146,29 +184,30 @@ size_t runtime::gpu::CUDNNEmitter::build_reduce_forward(const runtime::gpu::GPUR
         *ctx->cudnn_handle, desc, input_desc, output_desc, &workspace_size));
     size_t workspace_idx = allocator.reserve_workspace(workspace_size);
 
+    void* alpha = get_data_by_type(data_type, 1.0);
```
csullivan: Why this instead of declaring on stack and using the captured variables address like before? Also, I don't see alpha and beta being freed/deleted anywhere.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -178,6 +217,56 @@ size_t runtime::gpu::CUDNNEmitter::build_reduce_forward(const runtime::gpu::GPUR
     return primitive_index;
 }
 
+size_t runtime::gpu::CUDNNEmitter::build_tensor_op(const GPURuntimeContext* ctx,
```
csullivan: Nice.

src/ngraph/runtime/gpu/unit_test.manifest
```diff
@@ -1,8 +1,12 @@
+#int64 is not supprted by cudnn
```
csullivan: Good idea on the comments.

csullivan:
This PR fixes the segfault I am seeing in the CUDNN descriptor creation for op::Add in MXNet.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: fixed in all comments.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
csullivan: Since cudnnDataType_t is a constant expression in these use cases. Can we do this at compile time instead of runtime? It's a little bit more code, but it offloads the resolution to the compiler instead of doing it at runtime and avoids the need for heap allocations/deallocations. Something like this:

```C++
template <cudnnDataType_t data_type>
struct type_from_cudnn_datatype
{
        static_assert(not (data_type == CUDNN_DATA_HALF), "CUDNN_DATA_HALF is not supported by cuDNN");
        static_assert(not (data_type == CUDNN_DATA_INT8x4), "CUDNN_DATA_INT8x4 is not supported by cuDNN");
};

template<>
struct type_from_cudnn_datatype<CUDNN_DATA_FLOAT>
{
        static float type(double value) { return static_cast<float>(value); }
};

template<>
struct type_from_cudnn_datatype<CUDNN_DATA_DOUBLE>
{
        static double type(double value) { return static_cast<double>(value); }
};

template<>
struct type_from_cudnn_datatype<CUDNN_DATA_INT32>
{
        static int32_t type(double value) { return static_cast<int32_t>(value); }
};

template<>
struct type_from_cudnn_datatype<CUDNN_DATA_INT8>
{
        static int8_t type(double value) { return static_cast<int8_t>(value); }
};

template<cudnnDataType_t data_type>
auto get_data_by_type(double value) -> decltype(type_from_cudnn_datatype<data_type>::type(value)) {
        return type_from_cudnn_datatype<data_type>::type(value);
}

int main()
{

        float float_variable = get_data_by_type<CUDNN_DATA_FLOAT>(42.5);
        int int_variable = get_data_by_type<CUDNN_DATA_INT32>(42.5);

        auto half_variable = get_data_by_type<CUDNN_DATA_HALF>(42.5); // compile-time error
        auto int8x4_variable = get_data_by_type<CUDNN_DATA_INT8x4>(42.5); // compile-time error

        std::cout << float_variable << " " << int_variable << std::endl;
}
```

fengleitian: My understanding is we don't know the datatype at compile time.

csullivan: Ah, now that I look at it again, I see you are right. I thought the types were being hard coded in a switch, within which we could know the type at compile time.

fengleitian: I'm trying to add some code to manage allocations/deallocations.

fengleitian: one way is to make the whole build function as template<datatype>, and invoke the function by using a switch for different datatype.

fengleitian: Add host parameters manager.

PR #1129, opened 2018-06-21
Support dimshuffle/transpose with MKLDNN

nbpatel:


src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -1622,14 +1622,13 @@ namespace ngraph
...
                     {
-                        writer.block_begin();
-                        writer << "mkl::MKL_Somatcopy('R', 'T', " << to_string(arg_shape[0])
-                               << ",\n"
-                               << "                   " << to_string(arg_shape[1]) << ", 1.0f,\n"
-                               << "                   " << args[0].get_name() << ", "
-                               << to_string(arg_shape[1]) << ",\n"
-                               << "                   " << out[0].get_name() << ", "
-                               << to_string(arg_shape[0]) << ");\n";
+                        writer.block_begin() writer
```
jbobba: fix this

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -1652,35 +1651,92 @@ namespace ngraph
...
                 else
-                {
-                    kernel::emit_reshape(writer,
-                                         args[0].get_element_type().c_type_string(),
-                                         args[0].get_name(),
-                                         out[0].get_name(),
-                                         args[0].get_shape(),
-                                         out[0].get_shape(),
-                                         reshape->get_input_order());
+                { //Dead Code : What should be done?
```
jbobba: this is still required for non-mkldnn cases - non fp32, larger dimensions etc. 
we will need to benchmark mkldnn vs eigen.

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -87,6 +87,35 @@ mkldnn::memory::desc MKLDNNEmitter::build_memory_descriptor(const Shape& shape,
...
 
+mkldnn::memory::desc
+    MKLDNNEmitter::build_blocked_memory_descriptor(const mkldnn::memory::dims& dim,
+                                                   const mkldnn::memory::dims& strides,
+                                                   mkldnn::memory::data_type dtype) const
+{
+    // We have to construct memory descriptor in a C style. This is not at all
+    // ideal but MKLDNN does not offer any API to construct descriptor in
+    // blocked format except a copy constructor that accepts
+    // mkldnn_memory_desc_t.
```
jbobba: Remove comment.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -473,6 +474,22 @@ namespace ngraph
                     }
                 }
 
+                template <>
+                void CPUAssignment::ASSIGN_DECL(ngraph::op::Reshape)
+                {
+                    auto reshape = static_cast<op::Reshape*>(node);
+
+                    if (node->get_input_element_type(0) == element::f32 &&
+                        node->get_input_shape(0).size() < 12 &&
```
jbobba: dont hardcode 12. instead use the macro value (https://github.com/intel/mkl-dnn/blob/master/include/mkldnn_types.h) from MKLDNN

src/ngraph/runtime/cpu/cpu_emitter.cpp
nbpatel: i mean the conditions we are checking for it to go through "cpu::kernel::reshape_3d_3d_float32" & "cpu::kernel::reshape_3d_3d_float32" are exactly same as what i check for it to be mkldnn kernel. So i dont see these two block of code getting executed ever.

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
nbpatel: Addressed all the other feedback.

src/ngraph/runtime/cpu/cpu_emitter.cpp
jbobba: agree. can you benchmark performance for these two options and see which one is faster?

nbpatel: Sure.

nbpatel:
Ping for review

src/ngraph/runtime/cpu/cpu_emitter.cpp
jbobba: lets use eigen for 3d and MKLDNN for higher dimensions.

nbpatel: should i keep the 3D test case or remove it?

PR #1132, opened 2018-06-21
move sigmoid to core fusion

sasadep:
* support for sigmoid on multiple backend
* sigmoid reference kernels for interpreter backend


src/ngraph/op/sigmoid.hpp
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+#include "ngraph/op/util/requires_tensor_view_args.hpp"
+#include "ngraph/util.hpp"
+
+namespace ngraph
+{
+    namespace op
+    {
+        class Sigmoid : public util::RequiresTensorViewArgs
```
aprocter: I'm thinking this should inherit from `UnaryElementwiseArithmetic`.

src/ngraph/op/sigmoid.hpp
```diff
@@ -0,0 +1,54 @@
...
+        {
+        public:
+            Sigmoid(std::shared_ptr<Node> input);
+            Shape get_input_shape() const { return m_shape_input; }
```
aprocter: This function is not necessary; we can always use `get_input_shape(0)` (from `Node`)

src/ngraph/op/sigmoid.hpp
```diff
@@ -0,0 +1,54 @@
...
+            virtual std::shared_ptr<Node>
+                copy_with_new_args(const NodeVector& new_args) const override;
+            virtual void generate_adjoints(autodiff::Adjoints& adjoints,
+                                           const NodeVector& deltas) override;
+
+        private:
+            Shape m_shape_input;
```
aprocter: This field is not necessary.

sasadep:
@Krovatkin all of points that @aprocter mentioned are addressed

sasadep:
TESTNOW

src/ngraph/op/sigmoid.cpp
```diff
@@ -0,0 +1,69 @@
...
+        throw ngraph_error("Incorrect number of new arguments");
+    }
+
+    return make_shared<Sigmoid>(new_args.at(0));
+}
+
+op::Sigmoid::Sigmoid(shared_ptr<Node> arg)
+    : UnaryElementwiseArithmetic("Sigmoid", {arg})
+{
+    set_value_type_checked(arg->get_element_type(), arg->get_shape());
```
aprocter: It won't hurt anything, but this call is not necessary. (`UnaryElementwise`'s constructor does it for you.)

PR #1133, opened 2018-06-21
remove dlclose in backend destructor

rkimballn1:
It is causing segfault on exit with python bindings


rkimballn1:
I agree that we are replacing one bug with another. We can properly fix this with ngraph_initialize and ngraph_finalize functions which is a separate PR. This PR gets Travis and ONNX working again.

PR #1137, opened 2018-06-22
constant broadcast folding

adstraw:


adstraw:
Accidentally branched from adstraw/constant-prop instead of master.  Assuming this will be OK as we will squash and merge but let me know if I need to fix.

PR #1140, opened 2018-06-22
one should not inherit from standard containers

apwojcik:
The thing is standard containers **do not have virtual destructors**. One should never use something polymorphically that does not have a virtual destructor. There's no guarantee for the proper cleanup in derived classes.


diyessi:
The justification for this PR is incorrect.

Krovatkin:
@apwojcik it shouldn't be an issue for us because a) we don't use those containers in a **polymorphic** way b) we don't define any new **data** members or manage OS resources or do anything that could leak memory or result in UB.  This is really to strengthen static type checking, so we don't use `Coordinate` where `Shape` is expected.

PR #1143, opened 2018-06-22
add gpu timer

fengleitian:
Add detailed gpu timer.

### DIFF
--- a/src/ngraph/runtime/gpu/gpu_backend.cpp
+++ b/src/ngraph/runtime/gpu/gpu_backend.cpp
@@ -52,6 +52,7 @@ bool runtime::gpu::GPU_Backend::compile(shared_ptr<Function> func)
     {
         FunctionInstance instance;
         instance.m_external_function = make_shared<GPU_ExternalFunction>(func);
+        instance.m_external_function->m_emit_timing = instance.m_performance_counters_enabled;
         auto cf = instance.m_external_function->make_call_frame();
         instance.m_call_frame = dynamic_pointer_cast<GPU_CallFrame>(cf);
         m_function_map.insert({func, instance});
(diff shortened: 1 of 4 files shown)


src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -356,6 +361,59 @@ using namespace std;
...
+                    names.push_back(node->get_name());
+                    m_name_index_map.insert({node->get_name(), index++});
+                }
+            }
+        }
+        writer << "ngraph::stopwatch timers[" << names.size() << "];\n";
+        writer << "extern \"C\" size_t get_debug_timer_count() { return " << names.size()
+               << "; }\n";
+        writer << "extern \"C\" const char* get_debug_timer_name(size_t index)\n";
+        writer << "{\n";
```
rkimballn1: Use block begin/end thing here and following.

fengleitian: fixed.

src/ngraph/runtime/gpu/gpu_backend.cpp
```diff
@@ -67,20 +67,62 @@ bool runtime::gpu::GPU_Backend::call(shared_ptr<Function> func,
 
     validate_call(func, outputs, inputs);
 
-    auto it = m_function_map.find(func);
-    if (it == m_function_map.end())
+    FunctionInstance& instance = m_function_map[func];
```
csullivan: Pretty sure calling `m_function_map::operator[](func)` will default construct an entry in m_function_map if one does not exist. Why not use find?

fengleitian: Copied from CPU code. Since you need to create one anyway if it's not exist. I think code is more clean in this way.

fengleitian: Another reason is `enable_performance_data` can also create an key in `m_function_map`, so could not use find itself to determine if it's already compile.

fengleitian:
TESTNOW

PR #1144, opened 2018-06-22
move/rename is_unreachable to ngraph_util.cpp

Krovatkin:


Krovatkin:
@pruthviIntel we actually had a discussion on this topic this Tue. Everyone unanimously agreed that `is_unreachable` was a very ambiguous name and `is_used` was a much better option. Don't shoot the messenger 😄

PR #1151, opened 2018-06-25
Reshape bias to 1D for cpufusion of conv+bias bprop

nbpatel:


nbpatel:
Sorry for that. Done

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -783,7 +783,19 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_bprop()
...
+                    if (bias_shape.size() > 1)
+                    {
+                        NGRAPH_DEBUG
+                            << "mpattern = " << m.get_match_root()->get_name()
+                            << "conv_bias bias shape != 1, requires reshape to match filter count.";
+                        ngraph::AxisVector order(bias_shape.size());
+                        std::iota(begin(order), end(order), 0);
+                        auto bias_reshape = std::make_shared<op::Reshape>(
+                            bias, order, Shape{conv_bprop->get_filters_shape()[0]});
+                        bias_shape = bias_reshape->get_shape();
```
jbobba: Cant we compute the correct shape without creating an op?

Krovatkin: I think the shape would be `conv_bprop->get_filters_shape()[0]` we probably need a reshape because of a check in `ConvolutionBiasBackprop` which means we might be doing a copy (for Reshape) even though we could avoid one if `ConvolutionBiasBackprop` was a little bit more flexible and just checked for `shape_size(bias) ==  conv_bprop->get_filters_shape()[0]}`

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -813,7 +813,9 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_bprop()
                     ngraph::replace_node(m.get_match_root(), goe1);
                     NGRAPH_DEBUG << "Replacing bias and adding it as a second o/p of "
                                     "ConvolutionBiasBackpropFiltersBias";
-                    ngraph::replace_node(delta_user, goe2);
+                    auto goe2_reshape =
```
Krovatkin: we should check if this is needed... the same way we check above if we have to reshape `bias` into 1D

louisfeng:
@nbpatel Could you please merge my branch `https://github.com/NervanaSystems/ngraph/tree/louisfeng/convbias-shape_check` into yours for this PR? It includes some validation code that should help us catch these errors sooner. Thanks.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
nbpatel: My bad. Done

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
Krovatkin: @jbobba i misread that code... the right shape always seems to be `Shape{conv_bprop->get_filters_shape()[0]}` @nbpatel ?

rkimballn1:
TESTNOW

PR #1154, opened 2018-06-25
get_output_elements

Krovatkin:


src/ngraph/op/get_output_element.hpp
```diff
@@ -39,6 +39,8 @@ namespace ngraph
             size_t get_n() const { return m_n; }
             virtual NodeVector get_arguments() const override;
 
+            static NodeVector get_get_output_elements(std::shared_ptr<Node> mon);
```
diyessi: Move this up out of the class and call it `get_output_elements` and make `mon` be `const std::shared_ptr<Node>&`

PR #1155, opened 2018-06-26
Backend/API: cmake module to find Intel clDNN

shssf:
This module will be used with Intel GPU backend.
how to use it (at configure stage):
`cmake -DCLDNN_ROOT_DIR=${path_to_cldnn_sources_dir} -DNGRAPH_INTELGPU_ENABLE=TRUE
`


apwojcik:
You forgot to assign reviewers. Anyway, nGraph is using Modern CMake. This means it works on **targets** and **properties** (not variables!). Check this [link](https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/). At the end of the article, you will find a simple example of FindRapidJson.cmake. Long story short the find module script shall model dependencies through properties of the target (the library to be found). [Here](https://cliutils.gitlab.io/modern-cmake/) is the introduction to modern CMake (inside you will find links to useful resources, too).

shssf:
@apwojcik I'm new member and I don't know who could be a reviewer for such thing.
I did all thing with examples are located in the same directory.  But if you like I will try to redevelop this.
Thank you.

PR #1156, opened 2018-06-26
Updates towards building on windows native

rkimballn1:


CMakeLists.txt
```diff
@@ -148,7 +148,9 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=${NGRAPH_TARGET_ARCH}")
 
 # flags required for SDL-3
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2 -fPIE -D_FORTIFY_SOURCE=2")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wformat -Wformat-security")
+if (NOT WIN32)
```
Krovatkin: in future, we should probably set those depending on a compiler rather than a platform.

src/ngraph/runtime/interpreter/CMakeLists.txt
```diff
@@ -28,5 +28,8 @@ if (NGRAPH_INTERPRETER_ENABLE)
     target_link_libraries(interpreter_backend PUBLIC ngraph)
     set_target_properties(interpreter_backend PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${NGRAPH_BUILD_DIR})
 
-    install(TARGETS interpreter_backend LIBRARY DESTINATION ${NGRAPH_INSTALL_LIB})
+    install(TARGETS interpreter_backend
+        LIBRARY DESTINATION "${NGRAPH_INSTALL_LIB}"
+        ARCHIVE DESTINATION "${NGRAPH_INSTALL_LIB}"
```
Krovatkin: a quick question: why is `ARCHIVE DESTINATION`  important for windows?

PR #1160, opened 2018-06-26
move static backend functions to BackendCache so there is only one copy

rkimballn1:
This PR just moves all of the static functions and variables from the Backend class, which is the base class for all backends, to a new BackendCache class. The BackendCache class lives in the ngraph library and is not inherited by the backends so there will be only one copy of the static variables used internally.


jbobba:
@rkimballn1 are you still interested in pushing this? Minor nit, i would prefer BackendStore to BackendCache

rkimballn1:
This is no longer needed

PR #1164, opened 2018-06-26
remove custom install path

rkimballn1:
Remove NGRAPH_INSTALL_PREFIX and use CMAKE_INSTALL_PREFIX instead. This is the normal way to do cmake. This breaks build that rely on the default behavior of NGRAPH_INSTALL_PREFIX which sets to install directory to \~/ngraph_dist. To get the same old behavior use -DCMAKE_INSTALL_PREFIX=~/ngraph_dist


rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

raramer01:
TESTNOW

PR #1166, opened 2018-06-26
Collect matched nodes

Krovatkin:


src/ngraph/pattern/matcher.hpp
```diff
@@ -124,6 +124,20 @@ namespace ngraph
             friend op::Label; //TODO: refine to match_class
 
         protected:
+            size_t add_node(std::shared_ptr<Node> node)
+            {
```
diyessi: It would be cleaner to have `add_node` just add the node and not return anything.
You could just set `watermark` to `m_matched_list.size()`.

src/ngraph/pattern/matcher.hpp
```diff
@@ -124,6 +124,20 @@ namespace ngraph
...
+                m_matched_list.push_back(node);
+                return m_matched_list.size() - 1;
+            }
+            bool adjust_list(size_t watermark, bool matched)
+            {
```
diyessi: You are essentially doing transaction semantics -- the match succeeds and you commit, or the match fails and you abort, so `abort_match` might be a better name than `adjust_list` and could apply to all kinds of match.

PR #1167, opened 2018-06-27
Jbobba/conv sum cleanup

jbobba:


src/ngraph/graph_util.hpp
```diff
@@ -52,6 +52,8 @@ namespace ngraph
     std::list<std::shared_ptr<Node>>
         topological_sort(const std::list<std::shared_ptr<Node>>& nodes);
 
+    bool is_post_dominated(Node* X, Node* Y);
```
diyessi: Can you copy the comment from the implementation to here?

src/ngraph/graph_util.hpp
```diff
@@ -132,5 +134,8 @@ namespace ngraph
 
     bool is_one(std::shared_ptr<Node> reduce_constant);
 
-    bool is_used(std::shared_ptr<Node> node);
+    bool is_used(Node* node);
```
diyessi: This could use a short comment too, since there are different ways for a node to be used.

src/ngraph/graph_util.cpp
```diff
@@ -174,6 +174,33 @@ std::list<std::shared_ptr<ngraph::Node>>
...
+bool ngraph::is_post_dominated(Node* X, Node* Y)
+{
+    std::unordered_set<Node*> visited;
+    std::deque<Node*> stack;
+    stack.push_front(X);
+
+    while (stack.size() > 0)
+    {
+        ngraph::Node* curr = stack.front();
+        visited.insert(curr);
```
Krovatkin: why isn't `visited` used? we will probably look at some nodes more than once

src/ngraph/graph_util.cpp
```diff
@@ -435,15 +462,15 @@ bool ngraph::is_one(std::shared_ptr<Node> reduce_constant)
     return result_bool;
 }
 
-bool ngraph::is_used(std::shared_ptr<ngraph::Node> node)
+bool ngraph::is_used(Node* node)
```
Krovatkin: is there a reason why we are switching to raw pointers? It adds more clutter and makes code less readable.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -1155,18 +1155,19 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_add()
...
+            return false;
         }
 
-        if (inplace_input->get_users().size() > 1)
+        if (!is_post_dominated(inplace_input.get(), add_m.get()))
         {
-            NGRAPH_DEBUG << "Add has more than one user. Convolution Add might use an in-place "
-                            "destructive kernel";
-            // return false;
+            NGRAPH_DEBUG << "Unsafe to use in-place kernel since add's in-place input has "
```
Krovatkin: this error message is vague and doesn't describe the actual issue here, but I can't seem to come up with a better one.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -1155,18 +1155,19 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_add()
             return false;
         }
 
-        if (conv_m->get_users().size() > 1)
+        if (get_user_count(conv_m.get()) > 1)
         {
+            std::cout << "Convolution has more than one user: " << get_user_count(conv_m.get())
```
Krovatkin: please save this `get_user_count` into a temp 😄 it's an expensive call

Krovatkin: change `std::cout` to `NGRAPH_DEBUG`

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -826,8 +826,33 @@ using namespace ngraph::runtime;
                     }
                     return false;
                 };
-                // Always enable nodes computing output tensors
-                if (computes_output())
+
+                auto possibly_overwritten = [&]() {
```
Krovatkin: I'd suggest we move this to user `util.cpp` or `graph_util.cpp` It's a quite general routine, but I'm fine with keeping it as a lambda.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -498,8 +498,12 @@ namespace ngraph
                         auto op_annotations =
                             std::make_shared<ngraph::runtime::cpu::CPUOpAnnotations>();
                         op_annotations->set_mkldnn_op(true);
-                        std::map<size_t, size_t> oi_pairs = {{0, 0}};
-                        op_annotations->set_in_place_oi_pairs(oi_pairs);
+                        if (get_user_count(node->get_argument(0).get()) == 1)
```
Krovatkin: I think this is overly conservative. Multiple users do not preclude us from re-using the storage. It's up to liveness to detect the **last** use and re-use the storage in that case. I'll bug you tomorrow about this.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -826,8 +826,33 @@ using namespace ngraph::runtime;
...
+                    for (const descriptor::Output& output : node->get_outputs())
+                    {
+                        for (const descriptor::Input* input : output.get_inputs())
```
Krovatkin: At this point, we should be able to confidently tell if an input was overwritten or not. Do we need to be conservative here?

src/ngraph/graph_util.hpp
```diff
@@ -132,5 +135,10 @@ namespace ngraph
 
     bool is_one(std::shared_ptr<Node> reduce_constant);
 
-    bool is_used(std::shared_ptr<Node> node);
+    // Returns true if `node` is live in the graph i.e. a result op
+    // transitively uses this `node`
```
Krovatkin: that's an awesome description!

src/ngraph/graph_util.cpp
jbobba: shared_ptr is in general slower since it will need to increment a ref count atomically. I prefer using it only if we need it or if the compiler can eliminate it.

PR #1168, opened 2018-06-27
IntelGPUBackend: create_tensor functionality implementation with Intel clDNN

shssf:
Sources are formatted by
`${ngraph_nervana_build}/llvm/bin/clang-format -style=file *.hpp`

CMake module **CLDNN** implemented in PR #1155


src/ngraph/runtime/intelgpu/CMakeLists.txt
```diff
@@ -16,12 +16,16 @@
...
 set(SRC
     intelgpu_backend.cpp
+    intelgpu_tensor_view.cpp
 )
 
 if (NGRAPH_INTELGPU_ENABLE)
     add_library(intelgpu_backend SHARED ${SRC})
     set_target_properties(intelgpu_backend PROPERTIES VERSION ${NGRAPH_VERSION} SOVERSION ${NGRAPH_API_VERSION})
     target_link_libraries(intelgpu_backend PUBLIC ngraph)
+    find_package(CLDNN REQUIRED)
```
rkimballn1: Do we need so specify a minimum version?

src/ngraph/runtime/intelgpu/intelgpu_tensor_view.cpp
```diff
@@ -0,0 +1,104 @@
...
+void runtime::intelgpu::IntelGPUTensorView::write(const void* source,
+                                                  size_t tensor_offset,
+                                                  size_t n)
+{
+    if (tensor_offset + n > ocl_memory->size())
+    {
+        throw out_of_range("write access past end of tensor");
+    }
+
+    auto ptr = ocl_memory->pointer<char>();
```
rkimballn1: Can this line or the next return nullptr or empty objects? Do we need to check? A nullptr target will cause a segfault.

src/ngraph/runtime/intelgpu/CMakeLists.txt
shssf: I'm not sure. I didn't find any limitation on the interface. Perhaps we will do that later, if we use a specific, version dependant things.

PR #1172, opened 2018-06-27
gpu slice optimization

fengleitian:
1. move slice op to cuda_emitter.
2. change size_t to uint32_t in kernel
3. take rank out of slice kernel, unroll loop
3. change block_size from 1 to 64 for ops in cuda_emitter.

before:
```
Slice         34,658us
Broadcast        176us
Add               85us
Divide            46us
Exp               43us
Multiply          35us
Tanh              34us
Negative          32us
Reshape           27us
Result            23us
Dot               20us
```

after:
```
Broadcast     85us
Slice         40us 
Add           36us  
Divide        30us
Exp           25us
Tanh          25us
Multiply      14us
Negative      12us
Reshape       11us
Result         9us
Dot            8us
```


fengleitian:
TESTNOW

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1349,3 +1460,8 @@ __device__ __forceinline__ int64_t  load(const int64_t*  __restrict__ in, int i=
 )";
     return ss.str();
 }
+
```
csullivan: nit:
This is ceiling_divide(threads, block_size) * block_size right?

It's possible that for large tensors `value + alignment` can overflow. We could use idiv_ceil from the convolution PR: https://github.com/NervanaSystems/ngraph/pull/1131/files#diff-35e76a29b9a82763acee985ab907327cR190

There the `n/d` and `n%d` are computed together so it is equivalent complexity but avoids the overflow possibility.

Also, the name `align` confused me at first as I thought it had to do with memory alignment. Maybe `block_align/block_alignment`?

fengleitian: There is no overflow and performance concern here:
1. the max block size is 1024 and max grid size for x is 2^31-1, we are using uint32_t, there will be no overflow.
2. Since we are using 2^n as alignment, the better way is using `<<` and `>>`, but I think there is no performance, I can change to `idiv_ceil` when your PR get merged.

good point for the name. I'll change it now.

csullivan: I was just looking at lines 1295 - 1298 where `value = shape_size(result_shape)` and `alignment = 64`. In that case it could overflow if nthreads > 2^32-64 - 1, right?

Good point about using shifts. If you want to add a 2^n alignment function for this instead that works too.

fengleitian: cuda only support max grid size 2^31-1. 

if shape_size >= 2^31, we need to make either of the following changes:
1. make it into 2D
2. calculate multiple  data in a thread to reduce threads number, which will be in my next PR.

csullivan: Yes, I know that, I'm referencing the specific use of `align` above. You changed: ` **int** nthreads = static_cast<int>(shape_size(result_shape));` to `size_t nthreads = shape_size(result_shape);` And so `static_cast<uint32_t>(nthreads) + 64` could overflow inside the align function. I'm not commenting on the specific launch parameters.

csullivan: > if shape_size >= 2^31, we need to make either of the following changes.

That is correct, in the GPUShape PR I changed the internal size to be int32_t. Maybe this can help us check?

csullivan: > calculate multiple data in a thread to reduce threads number, which will be in my next PR.

Yes, this is the best option, but we still have the issue until we handle it for all ops. Anyway, it is a nitpick and is part of a larger issue, as you point out, and so better addressed elsewhere.

fengleitian: add gridsize check in the align_to_blocksize function, partially fix the problem. Will need a better way to check grid size and block size. Overflow is just part of it.

PR #1175, opened 2018-06-28
check cudnn version

fengleitian:
Since those datatype was not supportted before cudnn version 7100, need to add cudnn version.
```
CUDNN_DATA_UINT8
CUDNN_DATA_UINT8x4
```


fengleitian:
TESTNOW

PR #1184, opened 2018-07-02
DEX Part 3

jmenon:


src/ngraph/runtime/cpu/cpu_builder.hpp
```diff
@@ -29,6 +29,134 @@
...
+    else if (R == 12)                                                                              \
+        KV = K<ET, 12>;                                                                            \
+    else if (R == 13)                                                                              \
+        KV = K<ET, 13>;                                                                            \
+    else if (R == 14)                                                                              \
+        KV = K<ET, 14>;                                                                            \
+    else if (R == 15)                                                                              \
+        KV = K<ET, 15>;                                                                            \
+    else if (R == 16)                                                                              \
+        KV = K<ET, 16>;
```
jbobba: Why not KV = K<ET, R>?

src/ngraph/runtime/cpu/cpu_builder.hpp
```diff
@@ -29,6 +29,134 @@
...
+    }                                                                                              \
+    else if (ET == element::u32)                                                                   \
+    {                                                                                              \
+        SELECT_RANK(KV, uint32_t, R, K);                                                           \
+    }                                                                                              \
+    else if (ET == element::u64)                                                                   \
+    {                                                                                              \
+        SELECT_RANK(KV, uint64_t, R, K);                                                           \
+    }
+
```
jbobba: i'm assuming that this will be used in future PRs?

jmenon: Yes, just wanted to get the header change in earlier with the refactor

src/ngraph/runtime/cpu/cpu_builder.hpp
jmenon: To allow for cases where R is not a constexpr, and the compiler should eliminate these when it is

PR #1186, opened 2018-07-02
Refactored GPU backend state into BackendContext 

csullivan:
[Note: This PR depends on commits from #1191 and #1195 - which will be duplicated in the review tab until they are merged with master - thus, please ignore any changes in cuda_emitter.h/cpp]

This PR refactors the GPU backend state into BackendContext and exports it to the GPU_Backend. CUDNN and CUBLAS handles, as well as the GPUPrimitiveEmitter, and the GPURuntimeContext are now all managed by the GPU_Backend.

This required the GPUMemoryManager to support a list of argspace and workspace buffers which are each allocated at GPU_ExternalFunction compile time.


csullivan:
TESTNOW

src/ngraph/runtime/gpu/gpu_call_frame.cpp
```diff
@@ -61,15 +61,15 @@ void runtime::gpu::GPU_CallFrame::call(
         outputs.push_back(tv->m_allocated_buffer_pool);
     }
 
-    m_compiled_function(inputs.data(), outputs.data(), m_external_function->m_ctx.get());
+    m_compiled_function(inputs.data(), outputs.data(), m_external_function->ctx().get());
```
fengleitian: The change is good. My question is are we still using `gpu_call_frame`? Is it in `gpu_backend' now? if so, shall we delete this file?

src/ngraph/runtime/gpu/gpu_backend.hpp
```diff
@@ -59,6 +67,17 @@ namespace ngraph
                 std::vector<PerformanceCounter>
                     get_performance_data(std::shared_ptr<Function> func) const override;
 
+                class BackendContext
+                {
+                public:
+                    BackendContext();
+                    ~BackendContext();
+                    cublasHandle_t m_cublas_handle;
```
fengleitian: Since `m_cublas_handle` and `m_cudnn_handle` are already in `m_ctx`, looks redundant here for me.

src/ngraph/runtime/gpu/gpu_memory_manager.hpp
```diff
@@ -75,10 +76,15 @@ namespace ngraph
...
-                void* m_workspace;
-                size_t m_allocation_size;
 
+                struct allocation
+                {
+                    void* ptr;
+                    size_t size;
+                };
+
+                std::list<allocation> m_argspace_mem;
```
fengleitian: If my understanding is correct, we have a allocation for each function, is it better to use `std::unordered_map<string, allocation>`, where string is the unique function name. It will be more clear for `gpu_memory_manager` to handle.

src/ngraph/runtime/gpu/gpu_external_function.hpp
```diff
@@ -56,13 +57,14 @@ namespace ngraph
 
             public:
                 GPU_ExternalFunction(const std::shared_ptr<ngraph::Function>& function,
+                                     std::shared_ptr<GPU_Backend::BackendContext>& shared_context,
                                      bool release_function = true);
                 ~GPU_ExternalFunction();
                 std::shared_ptr<ngraph::runtime::gpu::GPU_CallFrame> make_call_frame();
                 std::unique_ptr<runtime::gpu::GPURuntimeContext>& ctx();
```
fengleitian: I'm not sure if we should still have a public member function to get `ctx()` and `get_primitive_emitter()` since those are not own by `GPU_ExternalFunction` now.

src/ngraph/runtime/gpu/gpu_external_function.hpp
```diff
@@ -100,10 +102,11 @@ namespace ngraph
                 bool m_release_function;
                 bool m_is_compiled;
 
-                cublasHandle_t m_cublas_handle;
-                cudnnHandle_t m_cudnn_handle;
-                std::unique_ptr<GPUPrimitiveEmitter> m_primitive_emitter;
-                std::unique_ptr<GPURuntimeContext> m_ctx;
+                // cublasHandle_t m_cublas_handle;
```
fengleitian: remove?

src/ngraph/runtime/gpu/gpu_backend.hpp
csullivan: Good point, they are removed and allocated on the heap, the pointers for which are stored in the runtime context.

src/ngraph/runtime/gpu/gpu_external_function.hpp
csullivan: Agreed. This required a significant refactor of the runtime context use. Now the emitters maintain a non-owning pointer to the runtime context `ctx` which is given at construction from the BackendContext.

src/ngraph/runtime/gpu/gpu_external_function.hpp
csullivan: Done.

src/ngraph/runtime/gpu/gpu_memory_manager.hpp
csullivan: This would require us to correlate the reservation with the allocation. In order to do that, we would need to pass the unique function name as an argument to the GPUAllocator reservations.

csullivan: This is doable, but is it worth it? Would require a not insignificant API change to all the emitters in order to pass in the function name. Would also have to handle the case when there are multiple functions for a single external function.

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
```diff
@@ -86,17 +115,16 @@ size_t runtime::gpu::GPUAllocator::reserve_argspace(const void* data, size_t siz
     // add parameter data to host buffer that will be transfered to device
     size = pass::MemoryManager::align(size, runtime::gpu::GPUMemoryManager::alignment);
     size_t offset = m_manager->queue_for_transfer(data, size);
-    // required to capture m_manager pointer
-    // directly rather than `this` pointer
-    auto manager = m_manager;
+    auto local = --m_manager->m_argspace_mem.end();
```
fengleitian: nit: use `.back()`?

src/ngraph/runtime/gpu/gpu_memory_manager.hpp
fengleitian: Just for the `m_argspace_mem` and `m_workspace_meme`. Let's not handle it for now. Currently it should be fine.

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
csullivan: `local` needs to be an iterator to the specific element in the list which will be lazily updated prior to the memory primitive being invoked. However, I updated it to use std::prev instead of operator--() for readability.

src/ngraph/runtime/gpu/gpu_memory_manager.hpp
csullivan: > Just for the m_argspace_mem and m_workspace_meme

Yes, this is actually what I was referring to. Because GPUAllocator creates reservations which are eventually "looked up" in the lambda closure, the lookup key for a map would need to be available at the time of reservation so that it could be captured and later used in the primitive. It's doable, but it would require us to touch quite a bit more than just the argspace and workspace buffers.

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
```diff
@@ -115,7 +115,7 @@ size_t runtime::gpu::GPUAllocator::reserve_argspace(const void* data, size_t siz
     // add parameter data to host buffer that will be transfered to device
     size = pass::MemoryManager::align(size, runtime::gpu::GPUMemoryManager::alignment);
     size_t offset = m_manager->queue_for_transfer(data, size);
-    auto local = --m_manager->m_argspace_mem.end();
+    auto local = std::prev(m_manager->m_argspace_mem.end());
     // return a lambda that will yield the gpu memory address. this
```
csullivan: Worth noting that `.rbegin()` would be nice here, however the standard does not guarantee reverse_iterators will remain stable upon insertion. It only makes that guarantee with forward iterators (c.f. https://stackoverflow.com/a/10086036)

PR #1188, opened 2018-07-03
Use mkldnn reorder only for transpose/dimshuffles.

nbpatel:
The mkldnn reorder works only when the output axis is a permutation of the input axis.


src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -554,12 +554,25 @@ namespace ngraph
...
                 {
                     auto reshape = static_cast<op::Reshape*>(node);
+                    auto arg0_shape = node->get_input_shape(0);
+                    auto result_shape = node->get_output_shape(0);
+                    auto axis_order = reshape->get_input_order();
+                    bool flag = true;
 
+                    for (size_t i = 0; i < axis_order.size(); i++)
+                    {
+                        if (arg0_shape[axis_order[i]] == result_shape[i])
```
mbrookhart: You could rid of the else if you did `!=`

nbpatel:
@mbrookhart thanks for adding the test case

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -554,17 +554,32 @@ namespace ngraph
...
-                        auto op_annotations =
-                            std::make_shared<ngraph::runtime::cpu::CPUOpAnnotations>();
-                        op_annotations->set_mkldnn_op(true);
-                        reshape->set_op_annotations(op_annotations);
+                        for (size_t i = 0; i < axis_order.size(); i++)
+                        {
+                            if (arg0_shape[axis_order[i]] != result_shape[i])
+                            {
+                                flag = false;
+                                break;
```
Krovatkin: you could've probably just `return;` at this point 😄  might have been cleaner. I dunno

cconvey: The first rule of @mbrookhart PR Fight Club is _don't talk about early returns_.  :)

mbrookhart:
TESTNOW

mbrookhart:
TESTNOW

PR #1189, opened 2018-07-03
gpu_external_function and gpu constant memory refactor

fengleitian:
1. refactor the huge `compile` function small functions.
2. refactor constants to memory manager to manager constant manager on GPU
3. removed alias output, which synced with CPU code.
4. removed `using namespace std;`
5. removed unused parameters, comments and code.


src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -104,23 +97,16 @@
...
-#include "ngraph/pass/memory_layout.hpp"
 #include "ngraph/runtime/gpu/gpu_backend.hpp"
-#include "ngraph/runtime/gpu/gpu_call_frame.hpp"
 #include "ngraph/runtime/gpu/gpu_cuda_kernel_emitters.hpp"
 #include "ngraph/runtime/gpu/gpu_emitter.hpp"
 #include "ngraph/runtime/gpu/gpu_external_function.hpp"
 #include "ngraph/runtime/gpu/gpu_kernel_emitters.hpp"
 #include "ngraph/runtime/gpu/gpu_runtime_context.hpp"
 
-using namespace std;
```
rkimballn1: Why remove this? If you left it in (most of our code has it in c++ files) then the number of changes would be much smaller. Put it back. and remove the `std::` that you added.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -377,77 +343,116 @@ using namespace std;
...
-                writer << "{\n";
-                writer.indent++;
-                writer << emit_string_array(c_value_strings, 100 - writer.indent * 4);
-                writer.indent--;
-                writer << "\n};\n\n";
+                shared_ptr<descriptor::TensorView> tv =
+                    node->get_outputs()[0].get_tensor_view();
+                // get an allocator for transient per kernel gpu memory
+                GPUAllocator allocator = this->m_primitive_emitter->get_memory_allocator();
+                size_t idx = allocator.reserve_argspace(
```
csullivan: Nice.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -377,77 +343,116 @@ using namespace std;
...
+                    c->get_data_ptr(),
+                    tv->get_tensor().size() * tv->get_tensor().get_element_type().size());
+                m_writer << "static size_t " << tv->get_tensor().get_name() << "_idx = " << idx
+                         << ";\n";
+                m_writer << "static " << tv->get_tensor().get_element_type().c_type_string() << "* "
+                         << tv->get_tensor().get_name() << " = nullptr;\n";
                 m_variable_name_map[tv->get_tensor().get_name()] = tv->get_tensor().get_name();
             }
         }
     }
```
csullivan: I'm a little bit worried about these `static` declarations in the JIT. I have found in several instances that there were errors in application teardown post return in `main` when memory was statically declared in the JIT code. Is it necessary for these to be static? I'm guessing so, in which case it would be useful to test this in a production-like environment to understand if there are any side effects.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -377,77 +343,116 @@ using namespace std;
...
+            }
+            m_writer << "is_constant_mem_ptr_null = false;\n";
+        }
+        m_writer.block_end();
     }
-    writer << "\n";
+    m_writer.block_end();
+}
 
+void runtime::gpu::GPU_ExternalFunction::emit_declare_functions()
```
csullivan: nit: emit_declare_functions -> emit_function_declarations

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -360,15 +321,20 @@ using namespace std;
...
-        writer << "// Declare debug timers\n";
+        m_writer << "// Declare debug timers\n";
         vector<string> names;
         size_t index = 0;
-        for (shared_ptr<Function> current_function : pass_manager.get_state().get_functions())
+        for (shared_ptr<Function> current_function :
+             m_pass_manager.get_state().get_functions())
         {
-            for (shared_ptr<Node> node : function_ordered_ops.at(current_function))
+            for (shared_ptr<Node> node : m_function_ordered_ops.at(current_function))
```
csullivan: Nice.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -481,13 +486,62 @@ using namespace std;
...
+                m_writer << emitted_function << "\n";
             }
-            node_function_map.insert({&node, match_function_name});
+            m_node_function_map.insert({&node, match_function_name});
         }
     }
+}
 
-    for (shared_ptr<Function> current_function : pass_manager.get_state().get_functions())
+void runtime::gpu::GPU_ExternalFunction::emit_allocate_temp_mem_pool(
```
csullivan: nit: emit_allocate_temp_mem_pool -> emit_temp_mem_pool_allocation / emit_function_memory_allocation

(If you change this one, then also change the corresponding release function below)

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -496,267 +550,202 @@ using namespace std;
...
-            shared_ptr<descriptor::TensorView> otv = op->get_output_tensor_view();
-            vector<size_t>& al = output_alias_map[otv.get()];
-            al.push_back(output_index);
-            if (al.size() > 1)
-            {
-                aliases.push_back(output_index);
+                //it should be safe to assign both descriptors to one output*
+                //since needs_copy == false makes `op::Result` an nop
+                auto res = dynamic_pointer_cast<ngraph::op::Result>(op);
+                if (!res->needs_copy())
```
csullivan: Nice, it looks like you have added support for `ngraph::pass::ResultCopyElimination`. Have you tried adding `    pass_manager.register_pass<ngraph::pass::ResultCopyElimination>();` to see if it functions as expected?

src/ngraph/runtime/gpu/gpu_external_function.cpp
fengleitian: *The way it works now need to be static.
*There is a way to change it but will need to get buffer address each time, which I think should be fine. 
*I don't understand why static variable should not be a problem in application teardown if use correctly.

src/ngraph/runtime/gpu/gpu_external_function.cpp
fengleitian: let me try it.

fengleitian: Looks like it works.

fengleitian: added.

PR #1195, opened 2018-07-03
Fix incorrect hash strings for softmax and 1d maxpool.

csullivan:


src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -515,37 +515,37 @@ size_t runtime::gpu::CUDAEmitter::build_1d_max_pool(const GPURuntimeContext* ctx
...
+
     std::string kernel_name = "maxpool_" + join(dtypes, "_") + "_iw" + std::to_string(input_width) +
                               "_ow" + std::to_string(output_width) + "_ww" +
                               std::to_string(window_width) + "_wst" + std::to_string(window_stride);
     std::replace(kernel_name.begin(), kernel_name.end(), ' ', '_');
 
-    // primitive hash and kernel name are equivalent for maxpool_1d
-    auto hash = kernel_name;
-
+    std::string hash = kernel_name + "_nc" + std::to_string(output_shape[0] * output_shape[1]);
```
fengleitian: should it be `nthreads` which is `shape_size(output_shape)`, or it fixed to 2d?

csullivan: Using `nthreads` is overspecifying the requirement since the 1d width is already specified in the kernel_name, but I think it is more intuitive so I have changed it as you suggest.

fengleitian: `void* args_list[] = {&inputs[0], &outputs[0], &nthreads};`
My thought is it should match what's in the args_list.

csullivan: You are right. This is a special case where N*C would be sufficient for the primitive hash, but it should match the args_list in general.

PR #1197, opened 2018-07-04
onnx [3]: add 'constant' operator

apwojcik:


src/ngraph/frontend/onnx_import/ops_bridge.cpp
```diff
@@ -39,6 +41,20 @@ namespace ngraph
             } // namespace error
 
             NodeVector add(const Node& node) { return op::add(node); }
+            NodeVector constant(const Node& node)
+            {
+                NodeVector output_nodes;
+                for (const auto& attribute : node.attributes())
```
postrational: More attributes may be added to `Constant` in the future. Please select the `value` attribute by name.
Let's use a helper function for selecting attributes by name here.

src/ngraph/frontend/onnx_import/ops_bridge.cpp
```diff
@@ -39,6 +41,20 @@ namespace ngraph
...
+                {
+                    if (!attribute.is_tensor())
+                    {
+                        throw onnx_import::error::attribute::invalid_data{attribute};
```
postrational: What error message will this generate? Can we Add the name od the node to the error message, so we can see which node caused the problem?

apwojcik: No need here, it will be reported at operators bridge level and it knows what node caused the problem.

apwojcik: Anyway extracting attribute by name removed this check and exception.

src/ngraph/frontend/onnx_import/attribute.hpp
```diff
@@ -20,6 +20,9 @@
 #include "onnx.pb.h"
 #include "tensor.hpp"
 
+#define likely(__x)         __builtin_expect(!!(__x), 1)
```
postrational: If we use `#define` in this header file, we will potentially be affecting code using this header.

apwojcik: Only ONNX importer use this header, no one else. It not supposed to be used by others.

src/ngraph/frontend/onnx_import/op/constant.hpp
```diff
@@ -0,0 +1,112 @@
+/*******************************************************************************
```
diyessi: Bob says these /* headers cause problems, so can you // them?

apwojcik: done, according to review by @rkimballn1 #1169

PR #1199, opened 2018-07-05
Backend/API: Implementation of the call method for IntelGPU

dmyershov:
@rkimballn1 please review.


src/ngraph/runtime/intelgpu/intelgpu_tensor_view.hpp
```diff
@@ -54,6 +54,7 @@ class ngraph::runtime::intelgpu::IntelGPUTensorView : public ngraph::runtime::Te
     /// @param n Number of bytes to read, must be integral number of elements.
     void read(void* p, size_t tensor_offset, size_t n) const override;
 
+    cldnn::memory* get_data_ptr() { return ocl_memory.get(); }
```
rkimballn1: This should probably be private

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -59,5 +64,43 @@ bool runtime::intelgpu::IntelGPUBackend::call(
...
+
+    // Process input parameters. Correctness of parameters was validated by validate_call.
+    // Since we have no correlation between Function::m_parameters and inputs, there is
+    // we try to match them by index number in vectors.
+    for (size_t i = 0; i < inputs.size(); i++)
+    {
+        shared_ptr<runtime::intelgpu::IntelGPUTensorView> tv =
+            static_pointer_cast<runtime::intelgpu::IntelGPUTensorView>(inputs[i]);
+        const op::ParameterVector& input_params = func->get_parameters();
+        network->set_input_data(input_params[i]->get_output_tensor().get_name(),
```
rkimballn1: You could replace `get_name()` with something that is not a string. Can you just use its position in the input vector? I just don't like strings for performance reasons but it is probably not an issue.

rkimballn1: Does this copy the data?

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -59,5 +64,43 @@ bool runtime::intelgpu::IntelGPUBackend::call(
...
+    std::map<cldnn::primitive_id, cldnn::network_output> result = network->execute();
+
+    // Process output parameters. Correctness of parameters was validated by validate_call.
+    // Since we have no correlation between Function::m_results and outputs, there is
+    // we try to match them by index number in vectors.
+    for (size_t i = 0; i < func->get_output_size(); i++)
+    {
+        shared_ptr<runtime::intelgpu::IntelGPUTensorView> ngraph_res =
+            static_pointer_cast<runtime::intelgpu::IntelGPUTensorView>(outputs[i]);
+        *(ngraph_res->get_data_ptr()) =
```
rkimballn1: This does not look like it is copying the data. Don't you need to move the results from the igpu to host memory? I know nothing about how cldnn works.

src/ngraph/runtime/intelgpu/intelgpu_tensor_view.hpp
dmyershov: I did this in the same manner as other backends.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
dmyershov: We should use std::string here because we have to provide the name to cldnn::set_input_data. clDNN uses std::string for  primitive_id also.

We aren't sure but it doesn't matter because method network.execute() is followed the input data.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
dmyershov: Fixed.

PR #1204, opened 2018-07-06
make cuda call async, add cuda async timer

fengleitian:
Found cuda version `negative` is 6x slower than cudnn version.  By dig into the problem I found cudnn is running on async mode but cuda is running on sync mode. So the cpu timer for `timing-detail` does not make sense. Here we can both cuda call and gpu timer to support async mode.

Why we change to async calls.
Performance 
some reference here:
https://www.cs.virginia.edu/~mwb7w/cuda_support/kernel_overhead.html

Why we can change to async calls.
1. since currently all ops running on GPU in default stream, it guaranteed to be run in order. 
2. the only time need to sync is the data copy between device and host. and we are currently using sync copy(I could be wrong here, need to verify).

Result (since only the total time is correct for previous version, we only compare that)
- before  without timing-detail:
```
enchmarking test/models/mxnet/LSTM_backward.json, GPU backend, 10 iterations.
compile time: 13,936ms
77.3ms per iteration

---- Aggregate times per op type/shape ----
```
- after without timing-detail:
```
Benchmarking test/models/mxnet/LSTM_backward.json, GPU backend, 10 iterations.
compile time: 13,603ms
61.4ms per iteration

---- Aggregate times per op type/shape/count ----
```

`timing-detail` after change, a `count` is been add to the list after shape, which shows the total number of the op run in each iteration. 

- Note that `timing-detail` will make both compile and run slower, but it will be useful if we compare apple to apple for future performance optimization:
```
Benchmarking test/models/mxnet/LSTM_backward.json, GPU backend, 10 iterations.
compile time: 28,752ms
87.9ms per iteration

ReplaceSlice     15,388us
Multiply         10,041us
Reshape           9,206us
Add               8,495us
Dot               8,077us
Sum               4,615us
Negative          2,910us
Divide            2,592us
Slice             2,496us
Broadcast         1,040us
Cosh                981us
Result               95us

---- Aggregate times per op type/shape/count ----
Multiply {32, 200} 1678           10,041us
ReplaceSlice {32, 60, 200} 60      9,848us
ReplaceSlice {32, 800} 480         5,540us
Add {32, 200} 596                  3,735us
Dot {200, 32} 238                  3,641us
Sum {800} 240                      3,480us
Negative {32, 200} 720             2,910us
Add {800, 200} 236                 2,796us
Divide {32, 200} 600               2,592us
Dot {200, 800} 240                 2,586us
Reshape {800, 200} 240             1,953us
Reshape {800, 32} 238              1,839us
Slice {32, 200} 360                1,811us
Reshape {32, 200} 298              1,782us
Add {800} 236                      1,606us
Reshape {200, 800} 240             1,491us
Reshape {200, 32} 240              1,409us
Sum {10000} 1                      1,135us
Cosh {32, 200} 240                   981us
Broadcast {32, 800} 120              953us
Dot {200, 10000} 1                   935us
Dot {200, 1920} 1                    915us
Slice {32, 1, 200} 119               685us
Reshape {10000, 1920} 1              366us
Add {32, 1, 200} 59                  358us
Reshape {32, 1, 200} 60              246us
Broadcast {32, 60, 200} 1             87us
Reshape {200, 10000} 1                43us
Result {10000, 200} 1                 42us
Reshape {10000, 200} 1                41us
Result {800, 200} 4                   24us
Result {800} 4                        16us
Reshape {1920, 200} 1                 13us
Reshape {200, 1920} 1                 12us
Reshape {32, 60, 200} 1               11us
Result {32, 60, 200} 1                 8us
Result {10000} 1                       5us
```


src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -205,8 +205,7 @@ size_t runtime::gpu::CUDAEmitter::build_pad(const runtime::gpu::GPURuntimeContex
                                           0,
                                           NULL, // shared mem and stream
                                           args_list,
-                                          0));  // arguments
-            CUDA_SAFE_CALL(cuCtxSynchronize()); // Retrieve and print output.
+                                          0)); // arguments
```
rkimballn1: If you see lines like this you should revert them so they don't show up in the review. The only thing that changed on this like is number of spaces before the comment char.

fengleitian: This is from clang format.

src/ngraph/runtime/gpu/gpu_util.cpp
```diff
@@ -192,3 +192,62 @@ uint32_t runtime::gpu::idiv_ceil(int n, int d)
     // compiler fused modulo and division
     return n / d + (n % d > 0);
 }
```
csullivan: The gpu::stopwatch class crosses the nGraph - JIT CXX ABI boundary. The class will be compiled by the nGraph static compiler but declared in the JIT. This will be unsafe if the two compilers have different ABIs. For that reason, I would suggest making this a header-only class (and marking some of those functions as `inline` for guidance)

fengleitian:
@csullivan good point, we can add sync in debug mode.

src/ngraph/runtime/gpu/gpu_util.cpp
fengleitian: Change to inline might increase compile time a lot. Make nGraph strict ABI compatible is a hard task.

fengleitian:
TESTNOW

src/ngraph/runtime/gpu/gpu_util.cpp
fengleitian: moved timer to runtime_context

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -319,7 +320,13 @@ void runtime::gpu::GPU_ExternalFunction::emit_timer_functions()
...
             }
         }
-        m_writer << "ngraph::stopwatch timers[" << names.size() << "];\n";
+
+        if (m_shared_context->m_runtime_context->stopwatch_pool == nullptr)
+        {
+            m_shared_context->m_runtime_context->stopwatch_pool = new StopWatchPool;
+        }
+        m_offset = m_shared_context->m_runtime_context->stopwatch_pool->size();
+        m_shared_context->m_runtime_context->stopwatch_pool->allocate(names.size());
```
csullivan: nit:Any reason why the size of the stopwatch pool can't be sent in as an argument to the constructor instead of allocating the pool separately?

src/ngraph/runtime/gpu/gpu_runtime_context.hpp
```diff
@@ -40,13 +40,19 @@ namespace ngraph
                 gpu::primitive* const* gpu_primitives;
                 const gpu::memory_primitive* gpu_memory_primitives;
                 CudaFunctionPool* compiled_kernel_pool;
+                StopWatchPool* stopwatch_pool = nullptr;
```
csullivan: The stopwatch pool is now in the runtime context, but it is being called from the JIT runtime which will still relies on the virtual function table to correctly dispatch to the member functions - which is ABI dependent. The CPU transformer handles this by only passing the `op_durations` array through the ctx. 

https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/cpu_runtime_context.hpp#L48

csullivan: I'm not sure we need to take the same approach, but I don't believe what we have is ABI safe yet. We can add a ticket it to address it later if you prefer, but I would say it probably should be done prior to the next version tagging.

src/ngraph/runtime/gpu/gpu_external_function.cpp
fengleitian: since we put the pool in the shared context, functions can share the same pool, my thought is it will be more convenient to `allocate` instead of `construct`.

csullivan: Ah, ok got it, thanks.

src/ngraph/runtime/gpu/gpu_runtime_context.hpp
fengleitian: My understanding is the interface are the following functions:
```
   void start_stopwatch(GPURuntimeContext* ctx, size_t idx);
   void stop_stopwatch(GPURuntimeContext* ctx, size_t idx);
   size_t count_stopwatch(GPURuntimeContext* ctx, size_t idx);
   size_t us_stopwatch(GPURuntimeContext* ctx, size_t idx);
```
only `GPURuntimeContext*` pointer and `size_t` been passed through interface, which should be potable.

csullivan: Got it, misunderstood, thought these were member functions. LGTM.

Issue #1196, opened 2018-07-04
`RuntimeError: alpha must be constant for bounded relu` while bounded relu is not used.


tsocha:
When I'm trying to run this model:
```json
{
  "name": "Function_0",
  "ops": [
    {
      "element_type": "float",
      "inputs": [],
      "name": "Parameter_0",
      "op": "Parameter",
      "outputs": [
        "Parameter_0_0"
      ],
      "shape": [
        3,
        4,
        5
      ]
    },
    {
      "element_type": "float",
      "inputs": [],
      "name": "Constant_5",
      "op": "Constant",
      "outputs": [
        "Constant_5_0"
      ],
      "shape": [],
      "value": [
        "3.40282e+38"
      ]
    },
    {
      "element_type": "float",
      "inputs": [],
      "name": "Constant_1",
      "op": "Constant",
      "outputs": [
        "Constant_1_0"
      ],
      "shape": [],
      "value": [
        "0"
      ]
    },
    {
      "axes": [],
      "inputs": [
        "Parameter_0"
      ],
      "name": "Broadcast_2",
      "op": "Broadcast",
      "outputs": [
        "Broadcast_2_0"
      ],
      "shape": [
        3,
        4,
        5
      ]
    },
    {
      "axes": [
        0,
        1,
        2
      ],
      "inputs": [
        "Constant_5"
      ],
      "name": "Broadcast_7",
      "op": "Broadcast",
      "outputs": [
        "Broadcast_7_0"
      ],
      "shape": [
        3,
        4,
        5
      ]
    },
    {
      "axes": [
        0,
        1,
        2
      ],
      "inputs": [
        "Constant_1"
      ],
      "name": "Broadcast_3",
      "op": "Broadcast",
      "outputs": [
        "Broadcast_3_0"
      ],
      "shape": [
        3,
        4,
        5
      ]
    },
    {
      "inputs": [
        "Broadcast_2",
        "Broadcast_3"
      ],
      "name": "Maximum_4",
      "op": "Maximum",
      "outputs": [
        "Maximum_4_0"
      ]
    },
    {
      "axes": [],
      "inputs": [
        "Maximum_4"
      ],
      "name": "Broadcast_6",
      "op": "Broadcast",
      "outputs": [
        "Broadcast_6_0"
      ],
      "shape": [
        3,
        4,
        5
      ]
    },
    {
      "inputs": [
        "Broadcast_6",
        "Broadcast_7"
      ],
[...]

PR #1205, opened 2018-07-06
use predicate for alpha in Bounded Relu

pruthviIntel:
fix for #1196


src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -1237,7 +1237,11 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_bounded_relu()
     auto relu = std::make_shared<op::Relu>(relu_input);
     auto iconst1 = op::Constant::create(element::f32, Shape{}, {1});
     auto alpha = std::make_shared<pattern::op::Label>(iconst1);
-    auto min = std::make_shared<op::Minimum>(relu, alpha);
+    auto broadcast_pred = [](std::shared_ptr<Node> n) {
```
Krovatkin: use `has_class` https://github.com/NervanaSystems/ngraph/blob/d18a9faf531f4cf4027c76a4b0d62cae822f984c/src/ngraph/pass/algebraic_simplification.cpp#L87

pruthviIntel: @Krovatkin just wondering is their a benefit using  `has_class` v/s ` lambda` function like above?

Krovatkin: readability 😄

PR #1206, opened 2018-07-06
Reshape inplace without copy data if possible

louisfeng:
This uses the new inplace API to avoid copy input data to output data for reshape when possible under certain conditions. One thing I'm not sure is how the input/output layout will be affected. Looking at the reshape emitter implementation, it seems only the mkldnn code path is doing something special about layout. @jbobba please advice when the assumption is valid.


src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -1652,6 +1652,14 @@ namespace ngraph
             void CPU_Emitter::EMITTER_DECL(ngraph::op::Reshape)
             {
                 auto reshape = static_cast<const ngraph::op::Reshape*>(node);
+                auto annotation = reshape->get_op_annotations();
+                if (annotation && annotation->get_in_place_oi_pairs().size() > 0)
+                {
```
jbobba: Safer to check if out[0].get_name() == args[0].get_name() 
Memory allocation doesn't guarantee in-place buffers. Especially if args[0] is an input parameter.

Krovatkin: second that

PR #1208, opened 2018-07-07
Support for multiple precompiled header files

rkimballn1:
This offers a significant performance improvement when running unit tests, about 20% faster when running a build that has CPU and GPU.


rkimballn1:
Is there some dashboard that shows build times? I timed this on a server but I would love to see actual CI times.

PR #1210, opened 2018-07-08
Liveness optimizations

rkimballn1:
Faster liveness.

### DIFF
--- a/src/ngraph/pass/liveness.cpp
+++ b/src/ngraph/pass/liveness.cpp
@@ -106,7 +106,7 @@ bool pass::Liveness::run_on_function(shared_ptr<ngraph::Function> function)
                 // this is the last node that value is seen in
                 // delete it at the end of the op
                 currently_live.insert(tensor_decl);
-                if (output_tensors.count(tensor_decl) == 0)
+                if (output_tensors.find(tensor_decl) == output_tensors.end())
                 {
                     // Don't free output tensors
                     free_tensor_decls.insert(tensor_decl);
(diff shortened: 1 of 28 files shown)


src/ngraph/pass/liveness.cpp
```diff
@@ -102,79 +101,31 @@ bool pass::Liveness::run_on_function(shared_ptr<ngraph::Function> function)
...
         for (descriptor::Tensor* tensor_decl : all_tensor_decls)
         {
-            if (!contains(currently_live, tensor_decl))
+            if (currently_live.find(tensor_decl) == currently_live.end())
             {
                 // this is the last node that value is seen in
                 // delete it at the end of the op
                 currently_live.insert(tensor_decl);
-                free_tensor_decls.insert(tensor_decl);
+                if (output_tensors.find(tensor_decl) == output_tensors.end())
```
Krovatkin: this needs  a comment. Why would we see the def of a tensor before a use (or in the absence of the use)?

src/ngraph/pass/liveness.cpp
```diff
@@ -102,79 +101,31 @@ bool pass::Liveness::run_on_function(shared_ptr<ngraph::Function> function)
...
-            }
-        }
-    }
-
-    // validate_liveness(ops);
     return false;
 }
-
-void pass::Liveness::validate_liveness(const list<Node*>& ops)
-{
```
Krovatkin: i think we do need some sanity checks. We could just keep track of dead tensors and check that no node is using them as inputs later on?

PR #1211, opened 2018-07-08
add pass profiler

rkimballn1:
To enable the pass profiler set the environment varialble `NGRAPH_PROFILE_PASS_ENABLE`

### DIFF
--- a/src/ngraph/pass/manager.cpp
+++ b/src/ngraph/pass/manager.cpp
@@ -15,6 +15,8 @@
 *******************************************************************************/
 
 #include <algorithm>
+#include <cxxabi.h>
+#include <iomanip>
 #include <iostream>
 #include <memory>
 
@@ -27,6 +29,7 @@
 #include "ngraph/pass/pass.hpp"
 #include "ngraph/pass/serialize.hpp"
 #include "ngraph/pass/visualize_tree.hpp"
+#include "ngraph/util.hpp"
 
 using namespace std;
 using namespace ngraph;
@@ -55,6 +58,8 @@ void ngraph::pass::Manager::initialize_default_passes()
 
 void ngraph::pass::Manager::run_passes(shared_ptr<Function> func)
 {
+    bool profile_enabled = getenv("NGRAPH_PROFILE_PASS_ENABLE") != nullptr;
+
     // find all functions
     vector<shared_ptr<Function>> fs;
     traverse_functions(func, [&](shared_ptr<Function> f) { fs.push_back(f); });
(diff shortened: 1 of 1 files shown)


src/ngraph/pass/manager.cpp
```diff
@@ -122,6 +131,19 @@ void ngraph::pass::Manager::run_passes(shared_ptr<Function> func)
             }
         }
         index++;
+        pass_timer.stop();
+        if (profile_enabled)
+        {
+            PassBase* p = pass.get();
+            string name = typeid(*p).name();
+            int status;
+            name = abi::__cxa_demangle(name.c_str(), 0, 0, &status);
```
Krovatkin: This is so awesome!! I didn't realize we can call a demangler programmatically.

src/ngraph/pass/manager.cpp
```diff
@@ -122,6 +131,19 @@ void ngraph::pass::Manager::run_passes(shared_ptr<Function> func)
             }
         }
         index++;
+        pass_timer.stop();
+        if (profile_enabled)
+        {
+            PassBase* p = pass.get();
+            string name = typeid(*p).name();
```
Krovatkin: do you think we could get rid of `m_pass_names` for good and just use this logic to get pass names?

src/ngraph/pass/manager.cpp
```diff
@@ -122,6 +131,19 @@ void ngraph::pass::Manager::run_passes(shared_ptr<Function> func)
             }
         }
         index++;
+        pass_timer.stop();
+        if (profile_enabled)
```
Krovatkin: this is a bit inconsistent with `m_visualize` and `m_serialize`. I don't know if we want consistency here like `m_profile` which is initialized in `PassManager::PassManager`?

src/ngraph/pass/manager.cpp
```diff
@@ -55,6 +58,8 @@ void ngraph::pass::Manager::initialize_default_passes()
 
 void ngraph::pass::Manager::run_passes(shared_ptr<Function> func)
 {
+    bool profile_enabled = getenv("NGRAPH_PROFILE_PASS_ENABLE") != nullptr;
```
Krovatkin: we could probably make it `static const` to signify that this value won't change?

src/ngraph/pass/manager.cpp
```diff
@@ -15,6 +15,8 @@
 *******************************************************************************/
 
 #include <algorithm>
+#include <cxxabi.h>
```
Krovatkin: I wonder if this will work on macosx, let alone windows?

src/ngraph/pass/manager.cpp
rkimballn1: My goal is to blow your mind

src/ngraph/pass/manager.cpp
rkimballn1: yes

src/ngraph/pass/manager.cpp
rkimballn1: const is ok, static is meh

src/ngraph/pass/manager.cpp
rkimballn1: No, too big a change. Needs a separate PR.

src/ngraph/pass/manager.cpp
rkimballn1: This is only used in this one method.

PR #1212, opened 2018-07-08
Memory Layout pass optimizations

rkimballn1:


src/ngraph/pass/memory_layout.hpp
```diff
@@ -55,7 +55,8 @@ class ngraph::pass::MemoryManager
     enum class allocation_scheme
     {
         FIRST_FIT,
-        BEST_FIT
+        BEST_FIT,
+        SIMPLE
```
Krovatkin: `NO_REUSE`, mb?

src/ngraph/pass/memory_layout.hpp
```diff
@@ -68,7 +69,7 @@ class ngraph::pass::MemoryManager
         block_state m_state;
     };
 
-    MemoryManager(size_t alignment = 1);
+    MemoryManager(size_t alignment = 1, bool disable_reuse = false);
```
Krovatkin: I was wondering if we could maybe add a test case for `SIMPLE`. It doesn't seem to be used anywhere

PR #1215, opened 2018-07-09
update version and add glossary defs

indie:


doc/sphinx/source/glossary.rst
```diff
@@ -120,5 +120,42 @@ Glossary
...
+      See :term:`ANN`.
+
+   ANN
+
+      :abbr:`ANN (Artificial Neural Network)` is an acronym for "Artificial 
+      Neural Network". ANN models are used to simulate possible combinations of 
+      binary logic processing and multi-layer (multi-dimensional) paths through 
+      which a :term:`data-flow graph` may be mapped or computed. An ANN does not 
+      have centralized storage; rather, an ANN manifests as information stored 
+      as patterns throughout the network structure. ANNs may be **Recurrent**
```
diyessi: Recurrent networks are just ones that unroll along some underlying structure in the input; typically a sequence or a (parse) tree.
Residual neural networks have a feedback mechanism (the residual); for example resnet.
Feed forward/backwards is a completely different thing having to do with whether you are trying to reduce a loss or not, where the loss is how much your results differ from the results you want; the feed backwards is used to compute how to adjust the weights to lower the loss.

doc/sphinx/source/glossary.rst
```diff
@@ -120,5 +120,42 @@ Glossary
...
+      storage, this loopback is the means by which the ANN can "learn" or be 
+      trained.  There are several sub-categories of RANNs. An :term:`LSTM`, for 
+      example, is a specific kind of "recurrent" ANN with a particular way of 
+      implementing long/short-term memory. For brevity, the RANN acronym is 
+      often abbreviated as `RNN`.
+
+
+   LSTM
+
+      :abbr:`LSTM (Long Short-Term Memory)` is an acronym for "Long Short-Term
```
diyessi: LSTMs add in mechanisms to keep chains of backprops from blowing up or going to 0, as they tend to do if you just use something simply for your cells.

doc/sphinx/source/glossary.rst
indie: Thanks, I read a bunch of academic papers from the late 90's to get some of these defs, since I thought it would make sense to use the established lexicology, but maybe it has changed ... ?

indie: Adding the "Artificial" part anyhow.  And, WRT to "Recurrent" being anything with a loopback, I read a few articles and looked up an old slack thread where @aprocter tried to explain, and kind of merged them to get this.  I am trying to get the docs prepped for some of the dynamic graph docs we will be working on soon, since i am thinking these kinds of RNNs will need to be explained thoroughly regarding how they are different from each other.

diyessi: TF is putting the dynamic in the graph, rather than dealing with it in python. This gives more opportunity to the compiler. I'm in the process of grokking what they are doing right now.
https://arxiv.org/abs/1805.01772

doc/sphinx/source/glossary.rst
```diff
@@ -120,5 +120,42 @@ Glossary
    export
    
       The serialized version of a trained model that can be passed to
-      one of the nGraph backends for computation.      
+      one of the nGraph backends for computation.
 
+   NN
```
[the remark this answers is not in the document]
indie: Yup, will do. See the later comment.

doc/sphinx/source/glossary.rst
```diff
@@ -120,5 +120,41 @@ Glossary
...
+
+      :abbr:`Recurrent Artificial Neural Network (RANN)`, often abbreviated as 
+      :term:`RNN`.
+
+
+   RNN 
+    
+      A :abbr:`Recurrent Neural Network (RNN)` is a variety of :term:`NN` where 
+      output nodes from a layer on a data-flow graph have loopback to nodes that 
+      created an earlier layer. Since the RNN has no "centralized" storage, this
```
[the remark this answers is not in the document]
indie: Excellent suggestion; done.

doc/sphinx/source/glossary.rst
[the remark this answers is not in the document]
indie: Done.

doc/sphinx/source/glossary.rst
```diff
@@ -120,5 +120,44 @@ Glossary
...
+      :term:`RNN`.
+
+
+   RNN 
+    
+      A :abbr:`Recurrent Neural Network (RNN)` is a variety of :term:`NN` where 
+      output nodes from a layer on a data-flow graph have loopback to nodes that 
+      comprise an earlier layer. Since the RNN has no "centralized" storage, this 
+      loopback is the means by which the ANN can "learn" or be trained. There are 
+      several sub-categories of RNNs. An :term:`LSTM`, for example, is a
```
mbrookhart: I would remove the LSTM example in this category and replace with the traditional RNN, i.e., s_t = tanh(dot(W,x_(t-1)) + dot(U, s_(t-1)), where x is the input data and s is the output/memory.

doc/sphinx/source/glossary.rst
mbrookhart: Something along the lines of: LSTMs extend on the traditional RNN by providing a number of ways to forget the memory of the previous time step via a set of learnable gates. These gates help avoid the problem of exploding or vanishing gradients that occur in the traditional RNN.

doc/sphinx/source/glossary.rst
indie: done

doc/sphinx/source/conf.py
```diff
@@ -68,9 +68,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '0.2.0'
+version = '0.4'
```
rkimballn1: Now at 0.5

doc/sphinx/source/framework-integration-guides.rst
```diff
@@ -17,19 +17,19 @@ MXNet\*
 Compile MXNet with nGraph
 --------------------------
 
-.. important:: These instructions pick up from where the :doc:`install`
-   installation instructions left off, so they presume that your system already
-   has the library installed at ``$HOME/ngraph_dist`` as the default location.
-   If the |nGl| code has not yet been installed to your system, please go back
-   and return here to finish compiling MXNet with ``libngraph``.
+.. important:: As of version |version|, these instructions presume that your 
+   system already has the Library installed to the default location, as outlined
```
rkimballn1: I think we should change the install docs to have the user set `-DCMAKE_INSTALL_PREFIX=~/ngraph_dist` as the default location. The cmake default install location is to install ngraph into the system folders as defined by each OS.

doc/sphinx/source/framework-integration-guides.rst
```diff
@@ -17,19 +17,19 @@ MXNet\*
...
+   MXNet with ``libngraph``.
 
 
 #. Set the ``LD_LIBRARY_PATH`` path to the location where we built the nGraph 
    libraries:
 
    .. code-block:: bash
 
-      export LD_LIBRARY_PATH=$HOME/ngraph_dist/lib/
+      export LD_LIBRARY_PATH=path/to/ngraph_dist/lib/
```
rkimballn1: It is unfortunate if we have to tell the user to use `LD_LIBRARY_PATH`.  Is there a make flag that we can add to mxnet to tell it where ngraph lives? I also see on line 53 where we specify `NGRAPH_DIR`. Too many environment variables. This is not a problem with the documents but with our build process.

doc/sphinx/source/framework-integration-guides.rst
```diff
@@ -119,22 +119,19 @@ Detailed info about neon's features and functionality can be found in the
...
-   location):
+#. Set the ``NGRAPH_CPP_BUILD_PATH`` and the ``LD_LIBRARY_PATH``. You can use 
+   the ``env`` command to see if these paths have been set already and if they 
+   have not, they can be set with something like: 
 
    .. code-block:: bash
 
-      export NGRAPH_CPP_BUILD_PATH=$HOME/ngraph_dist/
-      export LD_LIBRARY_PATH=$HOME/ngraph_dist/lib/       
+      export NGRAPH_CPP_BUILD=path/to/ngraph_dist
```
rkimballn1: Can we normalize TF and mxnet to use the same environment variable?

doc/sphinx/source/howto/import.rst
```diff
@@ -57,14 +57,16 @@ skip ahead to the next section, :ref:`install_ngonnx`.
...
+   be sure to export the ``NGRAPH_CPP_BUILD_PATH`` where the ``ngraph_dist`` was 
+   installed. 
 
    .. code-block:: console
 
       $ cd ngraph/python
       $ git clone --recursive -b allow-nonconstructible-holders https://github.com/jagerman/pybind11.git
       $ export PYBIND_HEADERS_PATH=$PWD/pybind11
-      $ export NGRAPH_CPP_BUILD_PATH=$HOME/ngraph_dist
+      $ export NGRAPH_CPP_BUILD_PATH=path/to/ngraph_dist
```
rkimballn1: And now `NGRAPH_CPP_BUILD_PATH` joins the fray. It's almost like we lack a system architect to keep everything consistent. Now we have three environment variables pointing to the same thing.

doc/sphinx/source/install.rst
```diff
@@ -40,16 +40,19 @@ compatible with a gcc 4.8-based build.)
 Installation Steps
 ==================
 
-The CMake procedure installs ``ngraph_dist`` to the installing user's ``$HOME`` 
-directory as the default location. See the :file:`CMakeLists.txt` file for 
-details about how to change or customize the install location.
+.. important:: The default cmake procedure will the install ``ngraph_dist`` to 
+   a system install location like ``/usr/bin/ngraph_dist``.  If you instead want 
+   to install it to a user's ``$HOME`` directory, you can run the cmake command
```
rkimballn1: How about `~/ngraph_dist` so it matches other references to ngraph_dist

doc/sphinx/source/conf.py
indie: fixed

doc/sphinx/source/framework-integration-guides.rst
indie: fixed, we now specify manual compilation

doc/sphinx/source/framework-integration-guides.rst
indie: @rkimballn1  The reason the documentation was originally structured this way is that setting the `LD_LIBRARY_PATH` is the first invocation where we let the "framework" be in charge of adding new stuff to whatever it needs to compile / be compiled.  Since this is a framework-specific setting (the first thing a framework needs to do is make the libraries available to the framework), we parsed this out and let this be a basic starting point for *any* framework, so that other framework developers not using TF or MXNet can be in charge of their own destinies.  Additional `LD_LIBRARY_PATH` config options are beyond the scope of nGraph since we already built nGraph in the previous install doc.

doc/sphinx/source/framework-integration-guides.rst
indie: I'm not sure we even need the CPP build path any more for neon?

doc/sphinx/source/install.rst
indie: Done

PR #1216, opened 2018-07-09
Faster Liveness pass.

cconvey:
On the DS2 torture-test subgraph, this reduces the Liveness avg. running time from 4.2 seconds to 2.1 seconds on skx12.


src/ngraph/pass/liveness.cpp
```diff
@@ -33,98 +36,136 @@
...
             }
-        }
 
-        unordered_set<descriptor::Tensor*> output_tensor_decls;
-        for (size_t i = 0; i < node->get_output_size(); ++i)
-        {
-            descriptor::Tensor& tensor = node->get_output_tensor(i);
-            if (persistent_tensors.find(&tensor) == persistent_tensors.end())
+            const auto iter = tensor_refcounts.find(t);
+            assert(iter != tensor_refcounts.end());
```
rkimballn1: remove all asserts

src/ngraph/pass/liveness.cpp
```diff
@@ -33,98 +36,136 @@
...
+            tensors.insert(&t);
         }
     }
+}
 
-    unordered_set<descriptor::Tensor*> currently_live;
-    for (auto it = ops.rbegin(); it != ops.rend(); it++)
+using OpsSchedule = vector<Node*>;
+
+static OpsSchedule get_ops_schedule(ngraph::Function& f)
```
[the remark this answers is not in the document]
cconvey: By "schedule" I meant the specific execution sequence we'd chosen for the graph's ops.  Will make the change.

src/ngraph/pass/liveness.cpp
```diff
@@ -33,98 +36,136 @@
...
-        for (descriptor::Input& input_decl : node->get_inputs())
+        ops_schedule.push_back(n.get());
+    }
+
+    return ops_schedule;
+}
+
+bool pass::Liveness::run_on_function(shared_ptr<ngraph::Function> function)
+{
+    assert(function);
```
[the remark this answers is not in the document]
cconvey: Thanks, didn't realize it violated a coding standard.  Will remove.

The check is probably overkill anyway.

cconvey:
@rkimballn1 
> Since this is not used in the CPU backend anymore and there are no unit test for this I would suggest not doing this PR. 

Could you please clarify?  AFAICT the Liveness pass is still being used by the CPU backend: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/cpu_external_function.cpp#L367

src/ngraph/pass/liveness.cpp
```diff
@@ -14,117 +14,173 @@
...
+    // This pass uses a backwards-dataflow analysis, very similar to what one would find in textbook
+    // descriptions of liveness analysis.
+    // Note that we visit the ops in *reverse* execution order.  So we visit the final use (if there
+    // is one) of each non-persistent tensor before visiting the op that generates that tensor.
+    //------------------------------------------------------------------------------------------------
+
+    ngraph::Function& f = *(function.get());
+
+    // FIXME: Ideally this should be computed just once and passed in to the function...
+    const vector<Node*> ops_execution_order = get_ordered_ops_vector(f);
```
Krovatkin: nearly every single optimization we have invalidates this list, that's why it's always recomputed. if `replace_node` or `replace_output` had access to `Function` we could invalidate it automatically and use a cached version otherwise

cconvey: > nearly every single optimization we have invalidates this list, that's why it's always recomputed.

So following that logic, and optimization which alters that list _also_ invalidates the liveness analysis, right?

If so, I wonder if we should make that more obvious to pass- and/or back-end- developers?

rkimballn1:
The Liveness pass is `called` by the CPU backend, but the results are not used by the CPU backend. The purpose of liveness is to assist in memory layout for tensor reuse. The CPU currently disables tensor reuse. This means that the CPU is not using liveness, but it is calling it. 
Memory layout was designed to use the information generated by liveness in order to reduce memory usage. Since the CPU backend is interested in runtime performance over memory usage it ignores the liveness information when doing memory layout. Memory layout ends up being a simple matter of allocating the memory required to support all temporary tensors in memory at the same time. It would be much more efficient to just write a new memory layout pass that does this simple layout rather than call liveness and then ignore the results for memory layout.

Krovatkin:
We use liveness **at least** for a few inplace optimizations. Could we
please keep it and MemoryLayout enabled for now until we come up with a
data-flow alternative robust enough for parallel execution? Also, it's now
beautifully commented. Even if we eventually get rid of it it would be nice
to have Christian's version for posteriority i.e. GitHub history

rkimballn1:
Testing this against what is currently on master I see that master takes 2236ms and this PR takes 2825ms, so master is faster. I made a followup PR to my last one where I replaced `shared_ptr<Node>` with `const shared_ptr<Node>&` in the for loops.

rkimballn1:
@Krovatkin Liveness will not go away and the CPU will need to have a mode where tensor sharing is enabled. There are models where memory usage is more important that speed.

Krovatkin:
@rkimballn1 @cconvey can we at least keep the comments from this PR?
They would be extremely useful for the others since it might not be intuitive at all (for non-compiler devs) why liveness is typically implemented as a backwards dataflow problem or the fact that we don't need use counts to compute it.

PR #1220, opened 2018-07-10
Bob/backend list

rkimballn1:


src/ngraph/runtime/backend.cpp
```diff
@@ -62,16 +51,24 @@ void* runtime::Backend::open_shared_library(string type)
...
+    version = regex_replace(version, regex("\\."), "\\.");
+    version = regex_replace(version, regex("\\+"), "\\+");
+    regex reg("lib" + to_lower(type) + "_backend.*" + version);
+    smatch result;
+
+    auto f = [&](const string& file, bool is_dir) {
+        string name = file_util::get_file_name(file);
+        if (regex_match(name, result, reg))
+        {
+            handle = dlopen(file.c_str(), RTLD_NOW | RTLD_GLOBAL);
```
aprocter: Would be good to keep the `runtime_error` here if `dlopen` fails, and perhaps throw rather than returning `nullptr` if the .so is not found.

src/ngraph/runtime/backend.cpp
```diff
@@ -62,16 +51,24 @@ void* runtime::Backend::open_shared_library(string type)
     {
         type = type.substr(0, colon);
     }
-    string lib_name = "lib" + to_lower(type) + "_backend" + ext;
+
+    vector<string> backend_list;
+    string version = NGRAPH_VERSION;
+    version = regex_replace(version, regex("\\."), "\\.");
+    version = regex_replace(version, regex("\\+"), "\\+");
+    regex reg("lib" + to_lower(type) + "_backend.*" + version);
```
aprocter: Maybe stick a `^` at the beginning and a `\.so$` (or `\.dylib$` for Mac, which is a pain) at the end?

rkimballn1: Done. I had defined a compiler def SHARED_LIB_EXT because cmake knows the correct extension.

src/ngraph/CMakeLists.txt
```diff
@@ -178,10 +178,12 @@ add_subdirectory(codegen)
 add_subdirectory(runtime)
 
 add_library(ngraph SHARED ${SRC})
-add_definitions("-DSHARED_LIB_EXT=\"${CMAKE_SHARED_LIBRARY_SUFFIX}\"")
-add_definitions("-DLIBRARY_VERSION=\"${NGRAPH_VERSION}\"")
+target_compile_definitions(ngraph PRIVATE SHARED_LIB_EXT="${CMAKE_SHARED_LIBRARY_SUFFIX}")
+# add_definitions("-DSHARED_LIB_EXT=\"${CMAKE_SHARED_LIBRARY_SUFFIX}\"")
```
diyessi: Remove these commented out ones.

src/ngraph/runtime/backend.cpp
```diff
@@ -37,22 +38,10 @@ static string find_my_file()
...
-//         NGRAPH_INFO << name;
-//     }
-//     return rc;
-// }
-
 void* runtime::Backend::open_shared_library(string type)
 {
     string ext = SHARED_LIB_EXT;
-    string ver = LIBRARY_VERSION;
+    // string ver = LIBRARY_VERSION;
```
diyessi: Remove

PR #1223, opened 2018-07-11
Added more convolution variants to DEX

jbobba:


src/ngraph/runtime/cpu/builder/convolution.cpp
```diff
@@ -143,16 +108,40 @@ namespace ngraph
...
+                        cpu::mkldnn_utils::set_memory_ptr(ctx, deps[0], arg0_tensor);
+                        cpu::mkldnn_utils::set_memory_ptr(ctx, deps[1], arg1_tensor);
+                        cpu::mkldnn_utils::set_memory_ptr(ctx, deps[2], out_tensor);
+                        cpu::mkldnn_utils::mkldnn_invoke_primitive(ctx, conv_index);
+                    };
+                    functors.emplace_back(functor);
+                }
+                else
+                {
+                    throw ngraph_error("ConvolutionBias is only supported with MKLDNN kernel.");
```
nbpatel: ConvolutionRelu?

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -262,6 +265,188 @@ size_t MKLDNNEmitter::build_convolution_forward(const mkldnn::memory::desc& inpu
...
+
+size_t MKLDNNEmitter::build_convolution_relu(const ngraph::Node* node,
+                                             const std::vector<TensorViewWrapper>& args,
+                                             const std::vector<TensorViewWrapper>& out)
+{
+    auto convolution = static_cast<const ngraph::op::ConvolutionRelu*>(node);
+
+    // For dilation, MKLDNN wants to know how many elements to insert between, not how far
+    // apart to space the elements like nGraph. So we have to subtract 1 from each pos.
+    Strides window_dilation_strides_adjusted;
```
nbpatel: Can we have a function(build_conv) which does the initial stride check as well as getting the formats and building descriptors ? We can call that routine for every build_convolution_* rather than  duplicating that code in every build_convolution_*

nbpatel:
LGTM

PR #1224, opened 2018-07-12
Refactoring MMB

Krovatkin:


louisfeng:
@Krovatkin An after thought, could you please add a comment in the code to explain why you did two matmul (second time for bias)? It would be good to know it's for avoid broadcasting and potentially better performance.

Krovatkin:
@diyessi  rebased... i'll need to change cpu_builder implementation which I'll do in a separate PR.. this one is already quite complicated.

PR #1226, opened 2018-07-12
Modify TBB graph nodes creation and deletion

ayzhuang:
* Add a graph* member to CPURuntimeContext.
* Create nodes the first time a function is called, all the following calls only exectue the computation.
* Delete nodes when cleanup_runtime_context is called.

### DIFF
--- a/src/ngraph/runtime/cpu/cpu_external_function.cpp
+++ b/src/ngraph/runtime/cpu/cpu_external_function.cpp
@@ -428,6 +428,10 @@ using namespace ngraph::runtime;
 
     if (m_use_tbb)
     {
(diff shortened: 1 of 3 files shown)


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -684,6 +682,19 @@ using namespace ngraph::runtime;
 
         writer << "bool* t_en = (bool*)" << current_function->get_name() << "_t_en;\n";
 
+        if (m_use_tbb)
+        {
+            // TODO: This should be static but we don't codegen statics correctly yet
```
jbobba: Remove.

PR #1230, opened 2018-07-13
Fix a segfault in the strided conv optimization

aprocter:
This fixes an issue we're observing in RN-50 on the TensorFlow bridge. This function, further down, seems to assume that all the users of `m_eltwise` are `Convolution` nodes, and tries to `dynamic_pointer_cast` accordingly. The patch avoids this by bailing out ahead of time if the users are not all `Convolution`.

The change here doesn't seem to cause any performance regression in the TF bridge.


jbobba:
i've seen this show up before where we have dead nodes. Can we insert a check for liveness before bailing out?

aprocter:
@jbobba Do you mean we ignore dead users (i.e., it should work as long as there are two _live_ users and they are both `Convolution`) or do you just want to log this condition when we bail?

jbobba:
the former.

PR #1231, opened 2018-07-13
Preallocate intermediate buffers

csullivan:
Use GPUMemoryManager and GPUAllocator to preallocate tensor pools used by compiled functions.


src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -486,11 +511,9 @@ void runtime::gpu::GPU_ExternalFunction::emit_temp_mem_pool_allocation(
     }
     if (m_temporaries_used)
     {
-        size_t temp_pool_size = current_function->get_temporary_pool_size();
         m_writer << "// Allocate the memory pool\n";
-        // TODO memory pool malloc.
-        m_writer << "void* pool_base_ptr = ngraph::runtime::gpu::create_gpu_buffer("
-                 << temp_pool_size << ");\n";
+        m_writer << "void* pool_base_ptr = ngraph::runtime::gpu::invoke_memory_primitive(ctx, "
```
fengleitian:
```
+        // reserve memory buffers for intermediate results
 +        size_t temp_pool_size = current_function->get_temporary_pool_size();
 +        if (temp_pool_size)
 +        {
 +            size_t pool_idx = allocator.reserve_workspace(temp_pool_size, false);
 +            m_memory_buffers[current_function->get_name()] = pool_idx;
 +        }
```
should be here?

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -254,6 +254,31 @@ runtime::gpu::GPU_ExternalFunction::~GPU_ExternalFunction()
 {
 }
 
+void runtime::gpu::GPU_ExternalFunction::assemble_and_reserve()
+{
+    GPUAllocator allocator = m_shared_context->m_primitive_emitter->get_memory_allocator();
```
fengleitian: may add this part to another`cache_ordered_list()`.

rkimballn1: Shouldn't from line 259 through 273 be a pass? Seems like you are working hard to make it not a pass. The pass manager state is really intended as an communications channel between passes. You could make this a function pass if you wanted. Since this code runs immediately after the pass manager just make this a pass.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -254,6 +254,31 @@ runtime::gpu::GPU_ExternalFunction::~GPU_ExternalFunction()
...
+
+    for (shared_ptr<Function> current_function : m_pass_manager.get_state().get_functions())
+    {
+        // cache list of ordered ops
+        m_function_ordered_ops.insert({current_function, current_function->get_ordered_ops()});
+
+        // reserve memory buffers for intermediate results
+        size_t temp_pool_size = current_function->get_temporary_pool_size();
```
fengleitian: this should go to `emit_temp_mem_pool_allocation`,

csullivan: The reservations for the pool need to be done prior to the op emissions. The function bodies are emitted in `collect_unique_functions` which occurs prior to `emit_temp_mem_pool_allocation`. By splitting up the reservation and the invocation in the way I have, the correct order is ensured.

I think I agree with @rkimballn1 now that I look at it. After separating out the reservation and the invocation, it looks like it would work naturally as a pass.

src/ngraph/runtime/gpu/gpu_external_function.cpp
fengleitian: like this:
```
 // reserve memory buffers for intermediate results
  size_t temp_pool_size = current_function->get_temporary_pool_size();
  size_t pool_idx = allocator.reserve_workspace(temp_pool_size, false);
  m_writer << "void* pool_base_ptr = ngraph::runtime::gpu::invoke_memory_primitive(ctx," <<  pool_idx << ");\n"
```

src/ngraph/runtime/gpu/gpu_external_function.cpp
csullivan: #1246 is a WIP towards converging on a way in which all runtime allocations (tensor pool and kernel-based memory) can be exported to the graph. In offline discussions we decided to iterate on it a bit more and to use #1231 until we converge on how to best achieve this.

src/ngraph/runtime/gpu/gpu_external_function.cpp
csullivan: I think this is the comment as below. Since the common function elimination is calling the emitters twice, these had to be separated so that the allocation reservations are made prior to the emitters being called.

If @rkimballn1 comes up with a solution for the CFE pass to only call the emitters once (avoiding the double allocations - until we formally fix that in #1246 - then we can do as you suggest and put the reservations into `emit_temp_mem_pool_allocation`.

src/ngraph/runtime/gpu/gpu_external_function.cpp
csullivan: Is line 271 kind of what you were meaning? Currently storing the buffers into cache with a lookup key based on the function name.
https://github.com/NervanaSystems/ngraph/pull/1231/files/74df2123da1095fe42544f0587fb19e405ff768a#diff-d68b46ecac06f2a6346a24a2af11f87aR271

src/ngraph/runtime/gpu/gpu_external_function.cpp
fengleitian: Make sense.

src/ngraph/runtime/gpu/gpu_external_function.cpp
fengleitian: Make sense.

csullivan:
@rkimballn1's CFE pass PR (#1236) moved the emission of static functions to a pass. Some of these have allocations. As a result, the function memory allocations needed also to be moved to a pass that runs prior to the CFE pass. I have re-added @fengleitian for an additional review as there are significant enough changes to warrant it.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -255,6 +257,49 @@ runtime::gpu::GPU_ExternalFunction::~GPU_ExternalFunction()
 {
 }
 
+void runtime::gpu::GPU_ExternalFunction::optimize_and_assemble()
```
fengleitian: https://en.wikipedia.org/wiki/Single_responsibility_principle

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
```diff
@@ -62,9 +63,15 @@ runtime::gpu::GPUMemoryManager::~GPUMemoryManager()
 
 void runtime::gpu::GPUMemoryManager::allocate()
 {
+    if (m_open_allocators)
+    {
```
fengleitian: I feel this can not ensure all reservations is done.

fengleitian: for one allocator, you can multiple reserve or no reserve.

fengleitian: is this check necessary?

csullivan: This is essentially a liveness check to ensure that the pass::MemoryManager has freed all of its allocations prior to actually making real allocations on the device. It doesn't mean there won't be any more reservation, it's just that there cannot be any that are _in progress_. In other words, a new allocator must be made. 

This way, the external function tensor pool workspace is the first allocation reserved, and then all kernel reservations are "on top" of that, and when the external function memory pool is freed, the kernel workspaces better have already been freed. That is essentially what this check is on, that the reservation tree was release correctly.

This is a bit confusing though and so I replaced the check with the following equivalent check
  ```  if (m_workspace_manager.get_node_list().size() != 1)  ```. Thus the open allocators variable isn't needed and so is removed.

src/ngraph/runtime/gpu/gpu_external_function.cpp
csullivan: Fair point, it was done so that the GPUAllocator would correctly go out of scope when the reservation requests had expired (after the pool reservation and all kernel allocations were made). Instead, I added GPUAllocator::close() to force the closure of the allocator so the device allocations can proceed without the requiring GPUAllocator to explicitly deconstruct.

src/ngraph/runtime/gpu/gpu_external_function.hpp
```diff
@@ -117,6 +116,7 @@ namespace ngraph
                 std::string m_function_name;
                 std::string m_pch_header_source;
 
+                std::shared_ptr<std::unordered_map<std::string, size_t>> m_memory_buffers;
```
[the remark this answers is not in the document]
csullivan: Temporary tensor memory buffers. I changed it to `m_tensor_memory_buffers`.

src/ngraph/runtime/gpu/pass/function_memory_reservation.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "ngraph/node.hpp"
+#include "ngraph/pass/manager_state.hpp"
+
+#include "ngraph/graph_util.hpp"
+#include "ngraph/runtime/gpu/gpu_memory_manager.hpp"
+#include "ngraph/runtime/gpu/pass/function_memory_reservation.hpp"
+
+using namespace ngraph;
+
+bool ngraph::runtime::gpu::pass::FunctionMemoryReservation::run_on_function(
```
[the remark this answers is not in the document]
csullivan: Changed the name to TensorMemoryReservation to reflect that the pass reserves memory for intermediate tensors.

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
```diff
@@ -62,9 +63,15 @@ runtime::gpu::GPUMemoryManager::~GPUMemoryManager()
...
+        throw std::runtime_error(
+            "Attempt to allocate memory while reservations are inprogress. Ensure all "
+            "GPUAllocators are closed before allocating.");
+    }
     if (m_buffer_offset)
     {
-        m_buffer_offset = pass::MemoryManager::align(m_buffer_offset, alignment);
+        m_buffer_offset = ngraph::pass::MemoryManager::align(m_buffer_offset, alignment);
```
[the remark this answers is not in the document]
csullivan: Changed so that the full path to the static alignment variable is used throughout.

src/ngraph/runtime/gpu/pass/function_memory_reservation.cpp
```diff
@@ -0,0 +1,45 @@
...
+    std::shared_ptr<Function> f)
+{
+    auto allocator = m_allocator.lock();
+    auto buffers = m_memory_buffers.lock();
+    if (allocator && buffers)
+    {
+        size_t mem_pool_size = f->get_temporary_pool_size();
```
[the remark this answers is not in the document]
csullivan: This is the maximum memory _needed_ by the function `f`, as calculated by liveness. It isn't a memory limit.

src/ngraph/runtime/gpu/pass/function_memory_reservation.cpp
```diff
@@ -0,0 +1,45 @@
...
+        if (mem_pool_size)
+        {
+            size_t pool_idx = allocator->reserve_workspace(mem_pool_size, false);
+            buffers->insert({f->get_name(), pool_idx});
+        }
+        return true;
```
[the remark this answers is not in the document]
csullivan: Good catch, the true return should be inside the conditional. Code is updated.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -635,28 +629,41 @@ void runtime::gpu::GPU_ExternalFunction::compile()
...
+        m_shared_context->m_primitive_emitter->get_memory_allocator());
+
+    m_pass_manager.register_pass<ngraph::pass::ResultCopyElimination>();
+
+    m_pass_manager
+        .register_pass<ngraph::pass::AssignLayout<descriptor::layout::DenseTensorViewLayout>>();
+
+    m_pass_manager.register_pass<ngraph::pass::Liveness>();
+
+    m_pass_manager.register_pass<ngraph::pass::MemoryLayout>(64);
```
rkimballn1: We should probably have the magic number 64 defined as a constant or something

I just found `runtime::gpu::GPUMemoryManager::alignment`

src/ngraph/runtime/gpu/gpu_external_function.hpp
```diff
@@ -80,7 +80,7 @@ namespace ngraph
                 EntryPoint m_compiled_function;
 
             private:
-                void collect_unique_functions();
+                void optimize_and_assemble();
```
fengleitian: nid: a more meaningful name than `optimize`

csullivan: Oops, that function's definition is removed but not it's declaration. I'll fix that. Thanks

fengleitian: I see.

src/ngraph/runtime/gpu/gpu_external_function.cpp
csullivan: Done, added a static variable for the pool alignment.

PR #1239, opened 2018-07-17
Bob/emit op

rkimballn1:


src/ngraph/node.cpp
```diff
@@ -329,3 +329,15 @@ NodeVector Node::get_users() const
...
     return result;
 }
+
+string Node::emit_op(const vector<runtime::TensorView>& args,
+                     const vector<runtime::TensorView>& out) const
+{
+    return "";
+}
+
+size_t Node::get_workspace_size(const vector<runtime::TensorView>& args,
```
fengleitian: `args` and `out` information is already in `Node` instance?

fengleitian: same question for the other added function.

PR #1244, opened 2018-07-18
IntelGPUBackend: BatchNorm 5x1 operation

shssf:
BatchNorm variant 3x3 is not supported due to things are not implemented in clDNN yet.

### DIFF
--- a/src/ngraph/runtime/intelgpu/intelgpu_op_batchnorm.cpp
+++ b/src/ngraph/runtime/intelgpu/intelgpu_op_batchnorm.cpp
@@ -27,7 +27,7 @@ using namespace std;
 using namespace ngraph;
 
 // This function converts Shape dimension id into cldnn::concatenation id
-cldnn::concatenation::concatenation_axis get_cldnn_axis(size_t tensor_channel)
+static cldnn::concatenation::concatenation_axis get_cldnn_axis(size_t tensor_channel)
 {
     switch (tensor_channel)
     {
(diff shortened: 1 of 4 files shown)


src/ngraph/runtime/intelgpu/intelgpu_op_batchnorm.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+#include "ngraph/runtime/intelgpu/intelgpu_op_batchnorm.hpp"
+
+#include "ngraph/op/batch_norm.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+// This function converts Shape dimension id into cldnn::concatenation id
+cldnn::concatenation::concatenation_axis get_cldnn_axis(size_t tensor_channel)
```
rkimballn1: All of the functions in this file should be static so they are not visible outside the file.

src/ngraph/runtime/intelgpu/intelgpu_op_batchnorm.cpp
```diff
@@ -0,0 +1,149 @@
...
+        mean_name = get_batch_norm_mean(topology, input_name);
+    }
+
+    string variance_name = variance_name_inp;
+    if (variance_name_inp.empty())
+    {
+        variance_name = get_batch_norm_variance(topology, input_name, mean_name);
+    }
+
+    const string input_splitted_name = do_matrix_split(topology, input_name, split_offsets);
```
rkimballn1: Can we change the part of the name `splitted` to `split`, which is the more correct past tense of split. This applies to all variables with `splitted` in the name.

PR #1247, opened 2018-07-19
Add TBB flow graphs to DEX.

ayzhuang:


ayzhuang:
test now

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1209,27 +1211,117 @@ void runtime::cpu::CPU_ExternalFunction::build()
...
+                                if (!arg->is_parameter() && !arg->is_constant())
+                                {
+                                    is_head = false;
+                                    tbb::flow::make_edge(*(nodename_tbbnode_map[arg->get_name()]),
+                                                         *(nodename_tbbnode_map[n->get_name()]));
+                                }
+                            }
+                            if (is_head)
+                            {
+                                dependence_graph_heads.emplace_back(n.get());
```
jbobba: Instead of collecting into dependence_graph_heads, and subsequently creating a flowgraph edge, directly create the edge between flowgraph_node_start and this node here. 

change codegen logic as well to not skip dependence_graph_heads.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1209,27 +1211,117 @@ void runtime::cpu::CPU_ExternalFunction::build()
...
+                }
             }
         }
         first_iteration = false;
     };
 
     m_is_built = true;
 
-    if (m_release_function)
+    if (m_release_function && !m_use_tbb)
```
jbobba: @jmenon do we need this in the executor or should this be done at the end of the build function outside the executor.

ayzhuang: The executor needs m_function to build the tbb flow graph at the first iteration. It is released after tbb flow graph is built.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -23,6 +23,8 @@
 #include <typeinfo>
 #include <unordered_map>
 
+#define __TBB_PREVIEW_LIGHTWEIGHT_POLICY 1
```
jbobba: you will probably have the same style issue on CI as last time. You might want to temporarily disable some checks around this header

ayzhuang: Last time I saw warnings when I built it before moving the header to the top. Now I do not see warnings when I build it. Will CI build have warnings?

PR #1251, opened 2018-07-20
More efficient sum for some cases

louisfeng:
Sum was a hot spot on sockeye transformer model. We tried a few different ways to speed it up (including using Eigen), but our modified omp parallel for and simd was much faster (~10X) than the Eigen implementation on that particular model. To generalize that special case, I needed to modify the placement (to avoid slowness in batch size of 1 case) of the parallel for loop from the original code as well as adding simd/reduction clause etc. Which led to thread safety issues and a fair amount of work to optimize for different cases. It seems to work well for the few test cases I have (unit test and various benchmark scripts). 

There's still more work can be done, but will introduce more complexities. What I have here will enable good performance for things we have seen so far. If there are cases down the road that demands more optimization. We can revisit this.


rkimballn1:
TESTNOW

PR #1252, opened 2018-07-22
IntelGPU backend: broadcast operation

shssf:
This is partially implemented "Broadcast" operation.
First commit:
- Broadcast operation
- GetOutputElement operation
- small code refactoring to avoid code duplication (get_cldnn_axis() moved to IntelGPULayout interface)

Second commit:
- more tests for Broadcast operation
- test.manifest aligned

Please ignore "second commit" if you think it is wrong, useless or implemented in wrong way.


test/backend_test.in.cpp
```diff
@@ -1908,6 +1908,152 @@ NGRAPH_TEST(${BACKEND_NAME}, broadcast_vector_rowwise_int64)
     EXPECT_EQ((vector<int64_t>{1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4}), read_vector<int64_t>(result));
 }
 
+#define BROADCAST                                                                                  \
```
rkimballn1: This would be much better as a static function.

PR #1256, opened 2018-07-23
Work around some buggy (and deprecated) rpath directives

jmenon:


jbobba:
Do we need both set_properties and set_target_properties?

jmenon:
I wasn't sure what the other one was trying to do so I left it there in case it's needed for macos or something. CMake clears the rpath at install time so this lets mkldnn find its dependents.

PR #1257, opened 2018-07-23
is_contained_match

Krovatkin:


src/ngraph/pattern/matcher.cpp
```diff
@@ -71,6 +72,27 @@ namespace ngraph
...
+                {
+                    //leaf label
+                    if (entry.first->get_inputs().empty())
+                    {
+                        label_exclusions.push_back(entry.second);
+                    }
+                }
+                return ngraph::get_subgraph_outputs(
+                           get_matched_nodes(), label_exclusions, ignore_unused)
+                           .size() < 2;
```
jbobba: Minor nit. "<=1" instead of 2

PR #1262, opened 2018-07-26
Build ngraph as static libraries

rkimballn1:
This is the change to build as much of ngraph as static libraries as we can. This includes ngraph core, codegen, cpu, gpu, igpu, and interpreter. All these are generated as static libraries and linked into the executable.
I also changed the `ngraph` target to be `ngraph_core` then added a new interface target `ngraph`. This allows the backends to depend on `ngraph_core` and then `ngraph` depends on `ngraph_core` and each of the backends. The "user" such as unit-test or a framework bridge links `ngraph` and gets core and all backends. Fun.
We still have support for using shared libraries so we can add backends like nnp by copying those shared libs to the lib directory. 
The new BackendManager is responsible for maintaining the registry of backends and opening shared libs.
There are a few random changes in here related to static ctors of vectors that contain element::Type. Since we are linking everything together as a big blob static initialization order is changed and these things broke, like in backend_test.in.cpp and mkldnn_utils.cpp.


cconvey:
> Since we are linking everything together as a big blob static initialization order is changed and these things broke, like in backend_test.in.cpp and mkldnn_utils.cpp.

@rkimballn1 : As of this PR, does nGraph have any remaining assumptions about static-initialization order across different source files?

rkimballn1:
@cconvey I'm sure there are more assumptions about static init order in there, but none that currently break the build

src/ngraph/runtime/gpu/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ set(SRC
     )
 
 if (NGRAPH_GPU_ENABLE)
-    add_library(gpu_backend SHARED ${SRC})
+    add_library(gpu_backend STATIC ${SRC})
```
WilliamTambellini: I guess it s ok for ngraph_core to be static lib but by imposing the gpu_backend to be static linked, you would create a hard dependency between ngraph and nvidia libs making a ngraph exec impossible to run if nvidia libs not available on the target machines. That is making production deployment very hard : that is why dynamic loading (implemented few weeks ago) is a good idea.  
1- Are you aware of that big drawback ?
2- Could you at least create a CMake option in order to let us build ngraph with either static or dynamic lib backends ? 
Kind
W.

cconvey: @WilliamTambellini : I think you raise a good point.  I have to wonder though if this is a problem for our target audience.  

At least with MXnet, it looks like the question of CUDA dependencies is addressed by having users [download the appropriate Pip Wheel](https://mxnet.apache.org/install/index.html?platform=Linux&language=Python&processor=CPU).

WilliamTambellini: @cconvey  Tks. My usecase is a little different than mxnet : I have only 1 release (1 executable), that must be able to run on any target machines, whatever the config (CPU, cuda GPU, opencl GPU, ...). In other words, in order for the exec to be able to run, it doesnt have to link with any proprietary vendor specific libs like nvidia, meaning ngraph to try to dynamically load the backends at runtime. Full static linking is a red flag for me preventing to use ngraph at all. I guess @rkimballn1 has some good reasons to try to do a full static build of ngraph. But keeping the dynamic loading approach seems like a must have at least for some of us.

avijit-nervana: I tend to agree with @WilliamTambellini. Backends may come and go - and since we do have a factory interface to instantiate the backend at runtime - why not use the dlopen. However, the CPU backend needs to be baked in (i.e., statically linked in) as that is the default backend

rkimballn1:
This dog won't hunt
Life has passed this PR by so I'm closing it.

PR #1267, opened 2018-07-27
IntelGPU backend: Pad operation

shssf:


src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -0,0 +1,130 @@
...
+                                      const string& scalar_name,
+                                      const string& output_name,
+                                      const Shape& output_shape,
+                                      const element::Type& output_type,
+                                      const Shape& pad_below,
+                                      const Shape& pad_interior)
+{
+    const size_t input_count = shape_size<Shape>(output_shape);
+    const string entry_point_name = "op_pad_kernel";
+    ostringstream kernel_code;
```
rkimballn1: In the codegen library there is a CodeWriter class which makes emitting nice, readable code easy. It does indenting and has helpters for code block begin and end.
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/codegen/code_writer.hpp
It is just a stringstream with some helpers.

PR #1270, opened 2018-07-27
Optionally get rid of codegen from the CPU backend

jmenon:
I'd have liked this to be cleaner but if this changeset seems intrusive (IMO) it demonstrates architectural and abstraction issues both in the buildsystem as well as backend design. I can document and discuss these with @jbobba but in the meantime, this is holding up Halide integration


raramer01:
TESTNOW

CMakeLists.txt
```diff
@@ -215,11 +220,15 @@ include(cmake/external_gtest.cmake)
 include(cmake/external_json.cmake)
 include(cmake/external_eigen.cmake)
 include(cmake/external_mkldnn.cmake)
-if (NGRAPH_USE_PREBUILT_LLVM OR DEFINED LLVM_TARBALL_URL)
-    include(cmake/external_llvm_prebuilt.cmake)
-else()
-    include(cmake/external_llvm.cmake)
+
+if (NOT NGRAPH_DEX_ONLY)
```
rkimballn1: DEX and the use of the GPU are mutually exclusive? Including llvm does not build llvm. It is only build if something depends on it. We need to be able to use DEX and the GPU in the same build. If the CPU make does not depend on codegen then codegen and llvm won't be built.

jmenon: > DEX and the use of the GPU are mutually exclusive?

I kept the DEX design general so the GPU backend could certainly implement it. But currently there's no DEX GPU backend so they are mutually exclusive.

>  Including llvm does not build llvm.  It is only build if something depends on it.

Exactly. Since nothing depends on it in a DEX build, we don't include it.

> We need to be able to use DEX and the GPU in the same build.

When the GPU backend implements DEX, we should definitely do that. Otherwise, this conflicts with our actual goal, which is to remove LLVM as a dependency so I can start pushing DEX Halide integration.

> If the CPU make does not depend on codegen then codegen and llvm won't be built.

The goal was to excise LLVM completely so no backend/component drags it back in. DEX uses Halide so it cannot co-exist with any backend that uses LLVM

louisfeng: Can we link to LLVM shared library rather than static linked? Would that enable us to allow multiple backends to link to LLVM?

jmenon: I don't know the answer to that question but I'd say that's unrelated to this PR since I'm not planning on using the same version of LLVM with Halide anyway

PR #1276, opened 2018-07-29
IntelGPU backend: Sum and redeveloped Broadcast operation

shssf:
Implemented Sum operation.
The Broadcast operation redeveloped completely and the code reused with Sum operation.
These operations are quite similar despite of the fact that data flow directions are different.


rkimballn1:
TESTNOW

rkimballn1:
TESTNOW

PR #1279, opened 2018-07-30
nbench: add option to run all models in a directory

rkimballn1:


src/tools/nbench/nbench.cpp
```diff
@@ -157,8 +172,34 @@ OPTIONS
...
+        for (const string& m : models)
+        {
+            try
+            {
+                shared_ptr<Function> f = deserialize(m);
+                cout << "Benchmarking " << m << ", " << backend << " backend, " << iterations
+                     << " iterations.\n";
+                run_benchmark(f, backend, iterations, timing_detail);
+            }
+            catch (...)
```
csullivan: If one of the models throws an exception in the benchmark, we should let it emit rather than ignoring it with this catch.

rkimballn1: Do we want to continue on? I think we should. We can catch, emit, and continue.

rkimballn1: The plan is to aggregate everything into tables in a web page generated by the test.

csullivan: Sure, continuing on is fine. The catch can update the GUI/std::err/out with a notice of which model failed.

csullivan: 👍 , this one is good to go.

PR #1281, opened 2018-07-30
fake PR to test Jenkins trigger job modifications

raramer01:


raramer01:
Disabled trigger, found no difference in git api rate limit of the nervana-bot user, now re-enabling it...

raramer01:
TESTNOW

Issue #1124, opened 2018-06-20
Mismatch template compile error


DavidBrayford:
When trying to build ngraph with gcc version 7 and icc version 18 I get the following template mismatch error.

cd /home/hpc/pr28fa/di72giz/NGRAPH/ngraph/build/src/ngraph && /lrz/sys/compilers/gcc/7.3.0/bin/g++   -DLIBRARY_VERSION=\"0.3.0+9441ea0\" -DSHARED_LIB_EXT=\".so\" -Dngraph_EXPORTS -I/home/hpc/pr28fa/di72giz/NGRAPH/ngraph/src/ngraph -I/home/hpc/pr28fa/di72giz/NGRAPH/ngraph/src -isystem /home/hpc/pr28fa/di72giz/NGRAPH/ngraph/build/json/src/ext_json/include  -DEIGEN_MPL2_ONLY -std=c++11 -march=native -O2 -fPIE -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -fstack-protector-strong -DNGRAPH_CPU_ENABLE -DIN_NGRAPH_LIBRARY -DGRAPHVIZ_FOUND -fPIC   -o CMakeFiles/ngraph.dir/autodiff/adjoints.cpp.o -c /home/hpc/pr28fa/di72giz/NGRAPH/ngraph/src/ngraph/autodiff/adjoints.cpp
In file included from /home/hpc/pr28fa/di72giz/NGRAPH/ngraph/src/ngraph/op/constant.hpp:25:0,
                 from /home/hpc/pr28fa/di72giz/NGRAPH/ngraph/src/ngraph/autodiff/adjoints.cpp:29:
/home/hpc/pr28fa/di72giz/NGRAPH/ngraph/src/ngraph/util.hpp:232:9: error: template-id ‘check_fp_values<double, std::isinf>’ for ‘void ngraph::check_fp_values(const char*, const double*, size_t)’ does not match any template declaration
         check_fp_values<double, std::isinf>(const char* name, const double* array, size_t n);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/hpc/pr28fa/di72giz/NGRAPH/ngraph/src/ngraph/util.hpp:213:10: note: candidate is: template<class T, bool (* func)(T)> void ngraph::check_fp_values(const char*, const T*, size_t)
     void check_fp_values(const char* name, const T* array, size_t n)
          ^~~~~~~~~~~~~~~
[...]

PR #1284, opened 2018-07-31
Fix SUSE build and run errors

rkimballn1:
There were two problems, one a compile issue and one a runtime error with codegen and header search paths.
1) The compile error has to do with template deduction and std::isinf/isnan functions. With SUSE the compiler tries to use `isinf(float)` for the `double` template specialization. Same for isnan.
2) The SUSE header search path for codegen was missing. Reworked the header search path creation to look for `x86_64-*-linux` so we should handle other variants in the future.
3) I also got annoyed with using string concat to join paths and also update path_join to handle joining multiple pieces in a single call. Spiffy!

CI tests won't test this but I am running tests locally and will solicit help from original issue author.

Addresses https://github.com/NervanaSystems/ngraph/issues/1124


src/ngraph/codegen/compiler.cpp
```diff
@@ -511,3 +518,25 @@ string codegen::CompilerCore::find_header_version(const string& path)
...
+string codegen::CompilerCore::find_os_specific_path(const string& path)
+{
+    string rc;
+    auto f = [&](const std::string& file, bool is_dir) {
+        if (is_dir)
+        {
+            const string prefix = "x86_64-";
+            const string suffix = "-linux";
+            string path = file_util::get_file_name(file);
+            if (path.size() > prefix.size() && path.size() > suffix.size() &&
```
diyessi: Should be `path.size() > prefix.size() + suffix.size()` or your substring below could end up with the start after the end.

PR #1286, opened 2018-07-31
IntelGPU backend: Power, Sigmoid and ReluBackprop operations

aalbersk:
Power, Sigmoid and ReluBackprop operations added for IntelGPU.

### DIFF
--- a/src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
+++ b/src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
@@ -448,6 +448,69 @@ bool runtime::intelgpu::IntelGPUBackend::compile(shared_ptr<Function> func)
                 arguments_check(op, 5, 1); // throw exception in this case
             }
         }
+        else if ("Convolution" == op->description())
+        {
+            arguments_check(op, 2, 1);
+
+            const std::string& conv_name = op->get_outputs().begin()->get_tensor().get_name();
+            const std::string& image_name = op->get_inputs().at(0).get_tensor().get_name();
+            const std::string& weight_name = op->get_inputs().at(1).get_tensor().get_name();
+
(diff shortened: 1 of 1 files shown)


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -96,6 +97,23 @@ static void do_unary_operation(cldnn::topology& topology,
     topology.add(cldnn_unary);
 }
 
+static void do_backprop_operation(cldnn::topology& topology,
```
shssf: Is this function necessary in case of no one else use it? It is ok if you plan to use it in future. If not, place the code in function call location.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -96,6 +97,23 @@ static void do_unary_operation(cldnn::topology& topology,
...
+                                  const shared_ptr<Node>& op,
+                                  cldnn_activation_grad_func mode,
+                                  const cldnn_activation_additional_params& param = {0.f, 0.f})
+{
+    arguments_check(op, 2, 1);
+
+    const string& input = op->get_inputs()[0].get_tensor().get_name();
```
shssf: Please don't use [0] because if the container empty it creates the element. Please use .at(0) because if no element exists it throw an exception.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -96,6 +97,23 @@ static void do_unary_operation(cldnn::topology& topology,
...
+    const string& input_grad = op->get_inputs()[1].get_tensor().get_name();
+
```
shssf: I think this is unnecessary empty line.

rkimballn1:
TESTNOW

PR #1287, opened 2018-07-31
gpu element op optimize

fengleitian:
Optimize binary operation, now cuda version is 40% faster than cudnn version.
25338us -> 17610us

before:
```
compile time: 12,643ms
316.7ms per iteration

Softmax       135,839us
Broadcast      55,390us
Add            25,338us
Slice          19,824us
Concat         18,523us
Dot            18,452us
Multiply        5,854us
Divide          3,615us
Exp             3,167us
Negative        3,093us
Reshape         2,732us
Result          2,312us
Tanh            2,136us
```

after:
```
compile time: 13,993ms
305ms per iteration

Softmax       133,966us
Broadcast      54,882us
Slice          19,913us
Concat         18,711us
Dot            18,361us
Add            17,610us
Divide          3,590us
Multiply        3,299us
Exp             2,889us
Negative        2,884us
Reshape         2,711us
Result          2,284us
Tanh            1,863us
```


src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1117,11 +1117,12 @@ size_t runtime::gpu::CUDAEmitter::build_elementwise_n_to_1(const std::vector<std
...
-    size_t nthreads = shape_size(tensor_shape);
+    uint32_t nthreads = static_cast<uint32_t>(shape_size(tensor_shape));
     //TODO: currently we set it to 64, will add tuning method later
-    uint32_t block_size_x = 64;
-    uint32_t aligned_grid_size_x =
-        align_to_block_size(static_cast<uint32_t>(nthreads), block_size_x);
+    uint32_t block_size_x = 512;
+    int num_SMs;
+    CUDA_RT_SAFE_CALL(cudaDeviceGetAttribute(&num_SMs, cudaDevAttrMultiProcessorCount, 0));
+    uint32_t aligned_grid_size_x = fmin(num_SMs * 32, align_to_block_size(nthreads, block_size_x));
```
csullivan: We could add this to a method that can be called specifically for thread coarsened kernels.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -33,14 +33,13 @@ void runtime::gpu::CudaKernelBuilder::get_elementwise_op(codegen::CodeWriter& wr
...
+           << "uint32_t n)\n";
+    writer.block_begin();
     {
-        writer << "size_t tid = blockIdx.x * blockDim.x + threadIdx.x; \n";
-        writer << "if (tid < n)\n";
-        writer << "{\n";
-        writer.indent++;
+        writer << "uint32_t tid = blockIdx.x * blockDim.x + threadIdx.x; \n";
+        writer << "uint32_t step = gridDim.x * blockDim.x; \n";
+        writer << "for ( ;tid < n; tid += step)\n";
```
csullivan: Have you experimented with unrolling this loop? It would require an extra parameter to come into the cuda_kernel_builder function.

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: Yes. I'm thinking those will go into tuning frame.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
fengleitian: Here it's hard to unroll, since the initial value of `tid` is not fixed. So thread will have different number of loop. Plus `step` can also change.

csullivan: Not too hard, the information is known from when we calculate the number of blocks in the grid. So the emitter could take an extra parameter that would indicate the number of blocks, and from that you can calculate the number of steps needed.

It would have the side effect of making the kernel semi-shape dependent, but it could improve the performance. It's not critical for this PR, but I am interested in the impact it might have.

csullivan: Ah, but you maybe need to either pad or add a branch. Anyway, approved, just wondered if you had experimented with it at all.

PR #1289, opened 2018-07-31
Implement trigonometric ops for direct execution.

ayzhuang:


src/ngraph/runtime/cpu/cpu_builder.cpp
```diff
@@ -83,7 +83,15 @@
 #include "ngraph/runtime/cpu/kernel/and.hpp"
 #include "ngraph/runtime/cpu/kernel/broadcast.hpp"
 #include "ngraph/runtime/cpu/kernel/ceil.hpp"
+#include "ngraph/runtime/cpu/kernel/cwise_acos.hpp"
```
jmenon: Could you rename these to just acos.hpp etc? Similar to how you named tanh
cwise_pow() is different from pow() so the prefix captures that. These don't need that prefix

ayzhuang: tanh is implemented similar to abs. All the others are implemented similar to cwise_power,usin unaryExpr. So they don't need the prefix?

jmenon: They're all implemented like `abs` in the sense that they're all unary element-wise ops (build steps are identical). Trigonometric ops don't need the cwise prefix, yes.

jbobba:
TESTNOW

rkimballn1:
TESTNOW

PR #1290, opened 2018-08-01
softmax & convolution memory primitive cacheing

csullivan:
This moves the temporary workspace/argspace allocations behind the primitive hash lookup to avoid repeat allocations when called multiple times, and greatly simplifies the gpu_emitter logic.

The main interface difference this introduces is a `build_primitive(op::Type* node)` function overload for each of the ops that need to have their emitter refactored. This increases the coupling between ngraph-core, but it means that the primitives can be built without the need of the TensorViewWrapper. As we move toward emitter objects, this helps us as we can then emit in a pass prior to the construction of the tensor view wrappers, allowing inline calculation of memory requirements. With that functionality, these memory requirements can more easily be exported to the graph.


csullivan:
TESTNOW

PR #1294, opened 2018-08-01
Fix first_iteration

jmenon:
***Please do not merge***


jbobba:
Can you change the _init in codegen to use ctx->first_iteration as well?

PR #1301, opened 2018-08-01
Quantization, Dequantization and INT8 Conv support for IA backend

nbpatel:


nbpatel:
Currently the min max values outputted are const floats, to have them as a ngraph constant op would be the way forward me thinks.Thoughts?

jbobba:
@nbpatel Yes, they should be constant ops.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -168,6 +174,105 @@ namespace ngraph
...
+                {
+                    if (node->get_input_element_type(0) == element::u8 ||
+                        node->get_input_element_type(0) == element::i8)
+                    {
+                        auto dequantize = static_cast<op::Dequantize*>(node);
+                        auto op_annotations =
+                            std::make_shared<ngraph::runtime::cpu::CPUOpAnnotations>();
+                        op_annotations->set_mkldnn_op(true);
+                        dequantize->set_op_annotations(op_annotations);
+                    }
```
pruthviIntel: I guess we need to throw exception in the `else` since for input dtype to Dequantized Op always expect them to be `u8` && `i8`

pruthviIntel:
screen shot 2018-08-06 at 10 29 16 am

I was going through this quantized graph from tensor flow, here each of the Quantized Op takes 3 inputs and 3 outputs. but in the PR all the quantized ops have just 1 I/p and the min and max are stored as attributes. If I understand correctly either we will have core Ops which will take 3 I/ps so we can directly emit ngraph quantized core op (and the current cpu specific op are just placeholders)  from the bridge or the bridge will do additional data marshaling before emiting the ngraph Op. is it correct?

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
nbpatel: I throw an ngraph error in cpu_emitter if the doesnt qualify as mkldnn op currently.

nbpatel:
@pruthviIntel The min and max will be constant ngraph ops

diyessi:
I was thinking we might add a member to element type for max value and add qint/quint types that use them. Like qint(8.4) or quint(10.5). If the inputs for something like add don't match, then we need to add a scaling (maybe just use the convert op) to get them matched. Some ops will also need one or more attributes added to specify the output element type -- for example, a dot would specify an output of q[u]int16/32. The quantized gemm pattern would have the output of the final sum quantized back to 8. I guess add is also an op that we need to be able to specify an element type for; we might add two 32 bit quants to get an 8 bit quant.

PR #1304, opened 2018-08-02
IntelGPU backend: Slice operation

shssf:
Also, unit_test.manifest updated for current passing state.


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -186,6 +187,39 @@ bool runtime::intelgpu::IntelGPUBackend::compile(shared_ptr<Function> func)
...
+            const string& output_name = op->get_outputs().begin()->get_tensor().get_name();
+            const Shape& output_shape = op->get_outputs().begin()->get_shape();
+            const element::Type& output_type =
+                op->get_outputs().begin()->get_tensor().get_element_type();
+            const shared_ptr<op::Slice> elem = static_pointer_cast<op::Slice>(op);
+            const Coordinate& lower_bounds = elem->get_lower_bounds();
+            const Coordinate& upper_bounds = elem->get_upper_bounds();
+            const Strides& strides = elem->get_strides();
+
+            if (input_shape.empty() || output_shape.empty() || lower_bounds.empty() ||
```
rkimballn1: So if I have upper and lower bounds but strides is empty then I can't do slice? There are a lot of OR operations in this conditional. Is this correct?

shssf: Yes, I programmed this in this way - if something is empty, it means no transformations\computation needed.
I didn't find any explanation in documentation on this. In tests I see only one test with all empty inputs which expect "do nothing".
I'm glad to change this code if you give me an advise.

aprocter: If do_equal_propagation means “pass the input straight to the output unmodified”, then this is correct. These tests correspond to the case where we are slicing a scalar. This is allowed, but it is always an identity op. We could get by with just testing input_shape.empty() but the extra tests won’t hurt anything.

shssf: @aprocter Yes, _do_equal_propagation_ means "copy input to the output". I believe this is not real copy but just name of the parameter.

PR #1308, opened 2018-08-02
Added DEX execution support for Lstm & Rnn

pruthviIntel:


pruthviIntel:
@jbobba addressed your comments.

pruthviIntel:
@jbobba how about now? ^^

src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -391,6 +392,85 @@ namespace ngraph
                                                 const mkldnn::memory::desc& dweights_desc,
                                                 const double eps);
 
+                template <typename OP>
+                size_t build_rnn(const ngraph::Node* node,
```
jbobba: looks good. but dont see a need for templatized method.

pruthviIntel: ya but `OP` can be either `ngraph::op::Lstm` or `ngraph::op::Rnn` so I templatized it

PR #1311, opened 2018-08-02
Implement new CI trigger to alleviate API rate issues

crlishka:
To be filled in later...


crlishka:
TESTNOW

crlishka:
TESTNOW

crlishka:
TESTNOW

nlsapp:
TESTNOW

nlsapp:
TESTNOW

.ci/jenkins/jenkins-trigger.groovy
```diff
@@ -0,0 +1,44 @@
...
+Integer TIMEOUTTIME = "3600"
+// BRANCH parameter is no loner needed
+// TRIGGER_URL parameter is no longer needed
+
+// Constants
+JENKINS_DIR="."
+
+env.MB_PIPELINE_CHECKOUT = true
+
+node("bdw && nogpu") {
```
nlsapp: To improve debugging/readability of the console log, I'd suggest wrapping this node block with a timestamps block: 

```
timestamps {
    // some block
}
```

This would prefix console log lines with a time stamp, and can be used to help measure node transition timings (i.e.: as you shift from node to node, you could observe time delays in how long it took to acquire an executor). I'd place this at the top-most layer, above the node

.ci/jenkins/jenkins-trigger.groovy
```diff
@@ -0,0 +1,44 @@
...
+// Parameters which ngraph-unittest uses:
+String  PR_URL = CHANGE_URL
+String  PR_COMMIT_AUTHOR = CHANGE_AUTHOR
+String  JENKINS_BRANCH = "chrisl/new-ci-trigger"
+Integer TIMEOUTTIME = "3600"
+// BRANCH parameter is no loner needed
+// TRIGGER_URL parameter is no longer needed
+
+// Constants
+JENKINS_DIR="."
```
nlsapp: You may want to use `env.WORKSPACE` here instead of `.`

.ci/jenkins/jenkins-trigger.groovy
nlsapp: For your node labels, I'd suggest shifting to `trigger` so that we can load this top-level/trigger executor slot onto commodity hardware (that may or may not be a bdw). We can label a few nodes in Jenkins to properly hold this

.ci/jenkins/jenkins-trigger.groovy
```diff
@@ -0,0 +1,44 @@
...
+
+env.MB_PIPELINE_CHECKOUT = true
+
+node("bdw && nogpu") {
+
+    deleteDir()  // Clear the workspace before starting
+
+    // Clone the cje-algo directory which contains our Jenkins groovy scripts
+    git(branch: JENKINS_BRANCH, changelog: false, poll: false,
```
nlsapp: I agree that `poll` does not belong here, but `changelog` might. This would render the list of changes/commits from the jenkins repo on the build page in addition to the ones from this repo. Not sure if that would be valuable for folks to see...

crlishka: Let's leave "changelog: false".  In a perfect world, only the git revision for ngraph would be displayed on the job page, because there is no labelling on what git revision is for what repo and having three unlabelled git SHAs is too confusing.  That said, even with "changelog: false", the git revision for cje-algo *does* appear on the job page (unfortunately).

.ci/jenkins/jenkins-trigger.groovy
crlishka: Change made

.ci/jenkins/jenkins-trigger.groovy
crlishka: Change made

crlishka: Change made

crlishka:
TESTNOW

crlishka:
"ci" failure was an internal Jenkins issue.  Rerunning with TESTNOW.

crlishka:
TESTNOW

PR #1312, opened 2018-08-02
Propagate input buffers for passthrough kernels

jbobba:


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -973,6 +974,41 @@ using namespace ngraph::runtime;
...
+
+            if (auto op_annotations = c_op->get_op_annotations())
+            {
+                for (auto oi_pair : op_annotations->get_in_place_oi_pairs())
+                {
+                    if (oi_pair.input == input->get_index() && !oi_pair.destructive)
+                    {
+                        size_t output_index = oi_pair.output;
+                        auto& output_tensor = c_op->get_outputs().at(output_index).get_tensor();
+
```
Krovatkin: Could you please add an `NGRAPH_DEBUG` statement specifying outputs "inplaced"?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -973,6 +974,41 @@ using namespace ngraph::runtime;
...
+                        m_variable_name_map[output_tensor.get_name()] = input_name;
+                        it = &c_op->get_outputs().at(output_index);
+                        propagate_further = true;
```
Krovatkin: if we have more than one non-destructive users we will only propagate parameter's storage through one of them? I think we might want consider using a work stack here.

src/ngraph/runtime/cpu/cpu_external_function.cpp
jbobba: Will sneak it in one of the other PRs.

src/ngraph/runtime/cpu/cpu_external_function.cpp
jbobba: good point. will address

PR #1319, opened 2018-08-03
DEX BatchDot

Krovatkin:


src/ngraph/runtime/cpu/builder/matmul_bias.cpp
```diff
@@ -172,7 +176,177 @@ namespace ngraph
...
+                void*& data_a;
+                void*& data_b;
+                void*& data_c;
+                int64_t group_count;
+
+                void call(CPURuntimeContext* ctx)
+                {
+                    std::vector<float*> a_array;
+                    std::vector<float*> b_array;
+                    std::vector<float*> c_array;
```
jbobba: we know the sizes of the vectors. might be better to instantiate one with the right size instead of risking reallocs later in line 217.

Krovatkin: hhhmmmkaaaayyy 😄

Krovatkin:
@jbobba addressed yr feedback... merged your changes

Krovatkin:
@jbobba Addressed yr feedback

PR #1323, opened 2018-08-03
Undo changes to Get Output Elements

varun-intel:


test/autodiff.in.cpp
```diff
@@ -1584,32 +1584,6 @@ NGRAPH_TEST(${BACKEND_NAME}, backwards_maxpool_n2c1h5w5_kh3kw3_sh2sw2)
     ASSERT_TRUE(read_vector<float>(output) == expected);
 }
```
rkimballn1: Don't remove this test

varun-intel:
@rkimballn1 I don't think we need this after @Krovatkin 's pr.

PR #1326, opened 2018-08-03
Auto. gen. kernel signatures and argument expansion

csullivan:
Functionality:
Introduce GPUKernelArgs helper class. Each argument that is added will be added to a kernel argument signature string and to the argument list. The kernel signature is used to generate the kernel declaration and an argument list which can be bound to the launch primitive. 

Performance:
In addition, GPUKernelArgs expands array arguments into individual registers for faster access. As registers are sent directly, this removes the need for GPUAllocator `argspace` reservations.

Notes:
Because this helper auto generates the kernel parameter signature based on the registered arguments, bugs such as that recently found with Broadcast (uint32_t arg passed to a size_t parameter in the nvrtc kernel) will no longer be possible.


src/ngraph/runtime/gpu/gpu_kernel_args.cpp
```diff
@@ -0,0 +1,101 @@
...
+    {TI(double), "double"},
+    {TI(int8_t), "int8_t"},
+    {TI(int16_t), "int16_t"},
+    {TI(int32_t), "int32_t"},
+    {TI(int64_t), "int64_t"},
+    {TI(uint8_t), "uint8_t"},
+    {TI(uint16_t), "uint16_t"},
+    {TI(uint32_t), "uint32_t"},
+    {TI(uint64_t), "uint64_t"},
+    {TI(int), "int"},
```
rkimballn1: Aren't these old style types covered in int32_t, etc.? If not do we need short and unsigned short, char and unsigned char?

src/ngraph/runtime/gpu/gpu_kernel_args.hpp
```diff
@@ -0,0 +1,190 @@
...
+            struct has_const_iterator
+            {
+            private:
+                typedef struct
+                {
+                    char x;
+                } true_type;
+                typedef struct
+                {
+                    char x, y;
```
rkimballn1: Don't use comma. One def per line

src/ngraph/runtime/gpu/gpu_kernel_args.cpp
```diff
@@ -0,0 +1,141 @@
...
+                                                                          const std::string& name)
+{
+    validate();
+    m_argument_list.push_back(nullptr);
+    m_placeholder_positions.push_back(true);
+    add_to_signature(type + "*", name);
+    return *this;
+}
+
+void** runtime::gpu::GPUKernelArgs::get_argument_list(std::vector<void*> arg_list)
```
fengleitian: Nit: 
My feeling is don't combine get_argument_list with resolve_placeholder, it's a little bit confuse when using it.  It took me a while to figure out it in #1322. My thought is it will be more clear in this way:
```
args.resolve_placeholder(...);
args.get_argument_list();
```
If we set args in place, we don't need placeholder, but will add some runtime.

src/ngraph/runtime/gpu/gpu_kernel_args.cpp
csullivan: I checked, and they are covered.

src/ngraph/runtime/gpu/gpu_kernel_args.hpp
csullivan: Fixed.

src/ngraph/runtime/gpu/gpu_host_parameters.hpp
```diff
@@ -0,0 +1,102 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#pragma once
+#include <list>
```
rkimballn1: Insert space between pragma and include

src/ngraph/runtime/gpu/gpu_kernel_args.hpp
```diff
@@ -0,0 +1,164 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#pragma once
+#include <memory>
```
rkimballn1: insert line again

PR #1327, opened 2018-08-03
Remove validation checks from performance critical code paths and ski…

jbobba:
…p layout propagation to inputs

validate_call adds about 200us to resnet-50. 
propagate_layouts is about 150us.


src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -80,8 +80,6 @@ bool runtime::cpu::CPU_Backend::call(shared_ptr<Function> func,
 {
     bool rc = true;
 
-    validate_call(func, outputs, inputs);
```
rkimballn1: Should we perhaps put this in a conditional compile or runtime conditional? It has a purpose and we may want it for debugging a model.

jbobba: this seems too low-level for a compilation flag. i was contemplating adding an optional argument to the call method. would you be ok with that?

rkimballn1: Perfect. thanks

PR #1336, opened 2018-08-06
IntelGPU backend: AvgPool operation(partially)

aalbersk:


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -279,6 +280,44 @@ bool runtime::intelgpu::IntelGPUBackend::compile(shared_ptr<Function> func)
...
+            const Shape& pad_above = avg_pool->get_padding_above();
+
+            vector<cldnn::tensor::value_type> offset({0, 0, 0, 0});
+            size_t ridx = 4;
+            for (auto i = pad_below.crbegin(); i != pad_below.crend() && ridx > 0; ++i, --ridx)
+            {
+                offset.at(ridx - 1) = -(*i);
+            }
+
+            const cldnn::tensor input_offset(
```
shssf: To avoid this code duplication, could you please look into `IntelGPULayout::create_cldnn_tensor()` function and create `IntelGPULayout::create_cldnn_offset()` with same approach? 
After this, please use the function in MaxPool and AvgPool operations.

PR #1337, opened 2018-08-06
IntelGPU backend: And, Or operations

aalbersk:


src/ngraph/runtime/intelgpu/intelgpu_layout.cpp
```diff
@@ -82,10 +82,10 @@ cldnn::tensor runtime::intelgpu::IntelGPULayout::create_cldnn_tensor(const Shape
 {
     vector<size_t> idx(4, 1);
     size_t index = 0;
-    const size_t total_zise = shape_size<Shape>(element_shape);
+    const size_t total_size = shape_size<Shape>(element_shape);
```
shssf: good catch!

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -617,10 +617,20 @@ void runtime::intelgpu::do_logic_kernel(cldnn::topology& topology,
     const string entry_point_name = "logic_" + output_name;
     codegen::CodeWriter writer;
 
-    writer << "__kernel void " << entry_point_name << "(const __global float inputA"
-           << array_dims(inputA_shape) << ", const __global float inputB"
-           << array_dims(inputB_shape) << ", __global char output" << array_dims(output_shape)
-           << ")\n";
+    if (operation == " && " || operation == " || ")
```
shssf: Please avoid string comparison as much as possible. This is not important here but it is better to be avoided.
I would like to ask you to redevelop/refactor the code a bit. As I can see you would like to use diffrent input parameters type for this kernel. I would propose to use `op->inputs().at(0).get_tensor().get_element_type().c_type_string()`  to get the type of input parameters in `do_logical_operation()`, add types to the function interface and use them in the proper line generation.
I expect it will generate `bool input[][]` instead `char` for boolean datatype including cases already covered by the function.
Please check tests for all affected operations before commit.

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -634,8 +644,17 @@ void runtime::intelgpu::do_logic_kernel(cldnn::topology& topology,
             ++var_idx;
         }
 
-        writer << "if (inputA" << access_dims(inputA_shape) << operation << "inputB"
-               << access_dims(inputB_shape) << ")\n";
+        if (operation == " && " || operation == " || ")
+        {
+            writer << "if (inputA" << access_dims(inputA_shape) << "!= 0" << operation << "inputB"
```
shssf: I don't think you have to use `!=0` here. At least after input/output data types changed to `bool`. And you will not have to use additional `if` above.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -119,9 +119,13 @@ static void do_logical_operation(cldnn::topology& topology,
 
     const string& inputA_name = op->get_inputs().at(0).get_tensor().get_name();
```
rkimballn1: Our naming convention is for snake_case for local variables. `inputA_name` does not conform. Perhaps we can call these `input0_name` and `input1_name`.  We can do this in a separate PR. There is plenty of refactoring to do later...

PR #1338, opened 2018-08-06
IntelGPU backend: Reverse operation implementation

dmyershov:


src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.hpp
```diff
@@ -80,10 +80,19 @@ namespace ngraph
...
+                                      const std::string& output_name,
+                                      const Shape& output_shape,
+                                      const element::Type& output_type,
+                                      const AxisSet& reversed_axes);
+
             // Helper functions used in cldnn::custom_gpu_primitive kernels
             std::vector<cldnn_arg> get_kernel_args(size_t input, size_t output);
             std::string array_dims(const Shape& dimentions);
             std::string access_dims(const Shape& dimentions, const AxisSet& axis = {});
+            std::string reverse_dims(const Shape& dimentions, const AxisSet& axis = {});
```
shssf: Why do we need this function in public interface?
I think it should be `static` in `.cpp`

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -83,6 +83,32 @@ string runtime::intelgpu::access_dims(const Shape& dimentions, const AxisSet& ax
     return buffer;
 }
 
+string runtime::intelgpu::reverse_dims(const Shape& dimentions, const AxisSet& axis)
```
shssf: I see many code duplicated. Maybe it is better to merge this with `access_dims()` with additional parameter with default empty value?

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -666,3 +692,72 @@ void runtime::intelgpu::do_logic_kernel(cldnn::topology& topology,
...
+
+    writer << "__kernel void " << entry_point_name << "(const __global float input"
+           << array_dims(input_shape) << ", __global float output" << array_dims(output_shape)
+           << ")\n";
+
+    writer.block_begin();
+    {
+        size_t var_idx = 0;
+        // Current clDNN implementation supports only 3 dimension of global work sizes
+        for (auto const& i : output_shape)
```
shssf: Could you please make a `static vector<size_t> generate_loops_begin()` to generate this sequence of the initial loops?

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -666,3 +692,72 @@ void runtime::intelgpu::do_logic_kernel(cldnn::topology& topology,
...
+                                             const Shape& input_shape,
+                                             const string& output_name,
+                                             const Shape& output_shape,
+                                             const element::Type& output_type,
+                                             const AxisSet& reversed_axes)
+{
+    const cldnn::layout layout = IntelGPULayout::create_cldnn_layout(output_type, output_shape);
+    const string entry_point_name = "reverse_" + output_name;
+    codegen::CodeWriter writer;
+    std::vector<size_t> gws;
```
shssf: You don't need `std::` here since we have `using namespace std;` at this file.

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -666,3 +692,72 @@ void runtime::intelgpu::do_logic_kernel(cldnn::topology& topology,
...
+                writer.block_begin();
+            }
+            ++var_idx;
+        }
+        writer << "output" << access_dims(output_shape) << " = input"
+               << reverse_dims(output_shape, reversed_axes) << ";\n";
+
+        // Closing brackets for extra loops
+        var_idx = 0;
+        for (auto const& i : output_shape)
```
shssf: The same as above. Please create `static vector<size_t> generate_loops_close()` for this.

PR #1340, opened 2018-08-06
Add OneHot op to direct execution.

ayzhuang:


src/ngraph/runtime/cpu/kernel/one_hot.hpp
```diff
@@ -0,0 +1,90 @@
...
+            {
+                template <typename ElementType>
+                void one_hot_rank_0(void* arg,
+                                    void* out,
+                                    const Shape& arg_shape,
+                                    const Shape& out_shape,
+                                    size_t one_hot_axis)
+
+                {
+                    eigen::EigenVector<ElementType> arg_vector(
```
jmenon: DEX shouldn't be using eigen core since that brings no real benefit. Could you switch to Eigen::Tensor?
If this isn't used in current or future models we care about (@jbobba should know this, I think?), then we could go with this

src/ngraph/runtime/cpu/kernel/one_hot.hpp
```diff
@@ -0,0 +1,90 @@
...
+                        static_cast<ElementType*>(arg), eigen::fmt::V(shape_size(arg_shape)));
+                    eigen::EigenMatrix<ElementType> out_vector(
+                        static_cast<ElementType*>(out), eigen::fmt::M(out_shape, out_strides));
+
+                    out_vector.setZero();
+                    for (size_t i = 0; i < arg_shape[0]; i++)
```
jmenon: Instead of hardcoding a loop like this where the inner loop probably isn't packetized, it's probably easier to just not use eigen at all. Previous comment applies as well

src/ngraph/runtime/cpu/kernel/one_hot.hpp
```diff
@@ -0,0 +1,99 @@
...
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace cpu
+        {
+            namespace kernel
+            {
+                template <typename ElementType>
+                void one_hot_rank_0(void* arg,
```
jmenon: Maybe we could avoid using eigen for this?

src/ngraph/runtime/cpu/kernel/one_hot.hpp
```diff
@@ -0,0 +1,99 @@
...
+                    out_dims[1] = out_shape[1];
+                    in_dims[0] = arg_shape[0];
+
+                    Eigen::TensorMap<Eigen::Tensor<ElementType, 2, Eigen::RowMajor>> out_tensor(
+                        static_cast<ElementType*>(out), out_dims);
+                    Eigen::TensorMap<Eigen::Tensor<ElementType, 1, Eigen::RowMajor>> in_tensor(
+                        static_cast<ElementType*>(arg), in_dims);
+
+                    out_tensor.setZero();
+                    for (size_t i = 0; i < arg_shape[0]; i++)
```
jmenon: You'll want to use generators for this

PR #1343, opened 2018-08-06
reduce fprop cache outputs

mbrookhart:
The previous version of fprop cache was returning every intermediate value in fprop due to some logic changes several months ago. This ensures that only the values needed by bprop are returned by fprop.


src/ngraph/util.cpp
```diff
@@ -270,23 +251,90 @@ ngraph::FpropCache ngraph::cache_fprop(std::shared_ptr<ngraph::Function> fprop,
...
     }
 
-    // add the cached fprop nodes as inputs to bprop
-    for (auto x : fprop_cache.fprop_output_nodes)
+    auto traverse_from_results = [](ResultVector results,
+                                    std::function<void(std::shared_ptr<Node>)> f) {
+        std::unordered_set<std::shared_ptr<Node>> instances_seen;
+        std::deque<std::shared_ptr<Node>> stack;
+
+        for (auto r : results)
```
Krovatkin: definitely, let's move it to `ngraph_utils.cpp` either as an extra flag to `traverse_nodes` or as a standalone function

src/ngraph/util.cpp
```diff
@@ -270,23 +251,90 @@ ngraph::FpropCache ngraph::cache_fprop(std::shared_ptr<ngraph::Function> fprop,
         cloned_results.push_back(result);
     }
 
-    // get clone bprop parameters
-    op::ParameterVector bprop_input_params;
-    for (auto param : bprop_inputs)
+    auto get_bprop_params = [&bprop_inputs, &fprop_cache]() {
```
Krovatkin: can we make it a member function?

src/ngraph/util.cpp
```diff
@@ -268,25 +250,64 @@ ngraph::FpropCache ngraph::cache_fprop(std::shared_ptr<ngraph::Function> fprop,
             throw ngraph_error("Expected op::Result values for op::Result keys in node_param_map");
         }
         cloned_results.push_back(result);
+        result_nodes.push_back(result);
     }
 
-    // get clone bprop parameters
-    op::ParameterVector bprop_input_params;
-    for (auto param : bprop_inputs)
+    auto get_bprop_params = [&bprop_inputs, &fprop_cache]() {
```
rkimballn1: Someday I want to move the fprop cache out of util.?pp into something like fprop_cache.?pp. Not needed now, just someday...

src/ngraph/graph_util.cpp
```diff
@@ -46,17 +46,29 @@ void ngraph::traverse_nodes(const std::shared_ptr<const Function> p,
...
     for (auto param : p->get_parameters())
     {
-        stack.push_front(param);
+        nodes.push_back(param);
+    }
+
+    traverse_nodes(nodes, f);
+}
+
+void ngraph::traverse_nodes(const NodeVector& results, std::function<void(std::shared_ptr<Node>)> f)
```
rkimballn1: The name `results` does not seem right as it contains both results and parameters. Perhaps we could call it `io_nodes` or something that more closely reflects what it contains.

src/ngraph/graph_util.cpp
```diff
@@ -46,17 +46,29 @@ void ngraph::traverse_nodes(const std::shared_ptr<const Function> p,
 
 void ngraph::traverse_nodes(const Function* p, std::function<void(std::shared_ptr<Node>)> f)
```
rkimballn1: Can you put as a comment whta your wrote in the description of this PR? Also is there a way you can add a unit test so that the next time someone makes a change it does not again produce unsatisfactory results?

src/ngraph/util.cpp
```diff
@@ -270,23 +251,90 @@ ngraph::FpropCache ngraph::cache_fprop(std::shared_ptr<ngraph::Function> fprop,
...
+                {
+                    stack.push_front(arg);
+                }
+            }
+        }
+    };
+
+    traverse_from_results(
+        cloned_results,
+        [&cloned_bprop_inputs, &fprop_cache, &inverted_node_map](std::shared_ptr<Node> node) {
```
Krovatkin: this needs a comment

PR #1345, opened 2018-08-06
Simplify result copy elimination

Krovatkin:


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -677,11 +675,11 @@ using namespace ngraph::runtime;
             //it should be safe to assign both descriptors to one output*
             //since needs_copy == false makes `op::Result` an nop
```
jbobba: remove comment

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -977,6 +975,29 @@ using namespace ngraph::runtime;
     }
 }
 
+bool runtime::cpu::CPU_ExternalFunction::is_output(const ngraph::descriptor::Tensor& tensor)
+{
+    return m_variable_name_map[tensor.get_name()].find("output") != std::string::npos;
+}
```
jbobba: Would be good to create a const static string for "outputs" and use it here and at line 672, so folks dont accidentally break this. Also rename function to 'is_output_buffer'?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -977,6 +975,29 @@ using namespace ngraph::runtime;
...
+
+bool runtime::cpu::CPU_ExternalFunction::computes_result(Node* node)
+{
+    if (node->is_output())
+    {
+        return true;
+    }
+
+    // Check if node feeds a result node that has been copy eliminated
```
jbobba: remove/update comment.

jbobba:
Just realized that DEX also uses computes_result. you can use tensor_alias to check for outputs.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -533,10 +533,12 @@ void runtime::gpu::GPU_ExternalFunction::emit_functions()
                 ss << "((" << type << "*)(outputs[" << i << "]))";
                 m_variable_name_map[tv->get_tensor().get_name()] = ss.str();
 
-                //it should be safe to assign both descriptors to one output*
-                //since needs_copy == false makes `op::Result` an nop
                 auto res = dynamic_pointer_cast<ngraph::op::Result>(op);
-                if (!res->needs_copy())
+                //keep assigning different out to a result descriptor
+                //emitter will check if in and out descs are the same
+                //skip a copy
```
jbobba: gpu_external_function gets a new comment but not cpu_external_function!

PR #1352, opened 2018-08-07
DEX Micro-optimizations: Use the mapped-value reference capture idiom…

jmenon:
… everywhere

### DIFF
--- a/src/ngraph/runtime/cpu/cpu_external_function.hpp
+++ b/src/ngraph/runtime/cpu/cpu_external_function.hpp
@@ -24,6 +24,7 @@
 #include <typeindex>
 #include <typeinfo>
 #include <unordered_map>
+#include <utility>
 #include <vector>
 
 #include "ngraph/codegen/code_writer.hpp"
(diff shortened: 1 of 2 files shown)


src/ngraph/runtime/cpu/cpu_external_function.hpp
```diff
@@ -178,8 +179,11 @@ namespace ngraph
                     executor;
                 std::unordered_map<std::string, void*> tensor_data;
                 std::unordered_map<std::string, bool> tensor_stale;
-                std::unordered_map<std::string, size_t> intermediates_offsets;
-                std::unordered_map<std::string, size_t> function_input_index, function_output_index;
+                std::list<std::pair<std::reference_wrapper<void*>, size_t>> intermediates_offsets;
+                std::list<
+                    std::tuple<std::reference_wrapper<void*>, size_t, std::reference_wrapper<bool>>>
```
jbobba: We will need to change to a struct at some point.

PR #1354, opened 2018-08-07
Revert changes to gpu shape and update 

csullivan:
Principally this PR reverts the changes to GPUShape that were merged with the nd-convolution PR, namely that GPUShape::value_type was int32_t, instead of uint32_t.

I also am proposing a name change of the shape type to NVShape. I originally had NVGPUShape, but that made the argument lists for many of our emitters even bulkier. That said, I'm open to it if you think it's better.

Also adds NVDiff which is similar to NVShape but has a signed value_type.


fengleitian:
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
There are some `int` and `int32_t` inside kernel. let's handle it later.

PR #1355, opened 2018-08-07
Added DEX support for BoundedRelu

pruthviIntel:


src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -403,6 +404,19 @@ namespace ngraph
                                              const mkldnn::memory::desc& result_desc,
                                              int softmax_axis);
 
+                template <typename OP>
+                size_t build_bounded_relu(const ngraph::Node* node,
+                                          const std::vector<TensorViewWrapper>& args,
+                                          const std::vector<TensorViewWrapper>& out)
```
jbobba: why do we need to templatize this? Are you planning to share this with the regular relu op

PR #1361, opened 2018-08-08
Backend support for static and dynamically linked backends

rkimballn1:
This is the code that handles both statically and dynamically linked libraries. I wanted to get this merged in regardless of which way we decide to go, but I feel that we will be supporting both options. There is some code in here that allows building on windows. Probably should be a separate PR, what do you  think?
I broke up backend.?pp to backend.?pp and backend_manager.?pp so that classes that derive from backend.hpp don't get the manager code in there too.


src/ngraph/runtime/backend.hpp
```diff
@@ -17,83 +17,82 @@
...
+    static bool register_backend(const std::string& name, std::shared_ptr<Backend>);
+
+    template <typename OP_TYPE>
+    bool is_supported(element::Type t)
+    {
+        return is_op_supported(OP_TYPE::get_op_name(), t);
+    }
+
+protected:
+    virtual bool is_op_supported(const std::string& name, element::Type t) const
```
aprocter: This is new, rather than having been moved from elsewhere in the file like the rest of this stuff. Should we do it in this PR? (I think it'd be fine.)

src/ngraph/runtime/backend_manager.cpp
```diff
@@ -0,0 +1,226 @@
...
+#include "ngraph/file_util.hpp"
+#include "ngraph/runtime/backend.hpp"
+#include "ngraph/runtime/backend_manager.hpp"
+#include "ngraph/util.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+#ifdef WIN32
+#define OPEN_LIBRARY(a, b) LoadLibrary(a)
```
aprocter: You don't seem to be using this macro.

src/ngraph/runtime/backend.hpp
rkimballn1: Grrr, that was not supposed to be in this PR. That is the next PR. It must have crept in from the merging work I was doing. It is protected so users can't call it.

src/ngraph/runtime/backend_manager.cpp
rkimballn1: Will remove

src/ngraph/runtime/backend_manager.hpp
```diff
@@ -0,0 +1,61 @@
...
+    namespace runtime
+    {
+        class Backend;
+        class BackendManager;
+
+        using new_backend_t = std::function<Backend*(const char* config)>;
+    }
+}
+
+class ngraph::runtime::BackendManager
```
avijit-nervana: Need to provide comments explaining how to use backend, how to specify one etc.

src/ngraph/runtime/backend.hpp
```diff
@@ -17,83 +17,75 @@
...
+class ngraph::runtime::Backend
+{
+public:
+    virtual ~Backend();
+    /// @brief Create a new Backend object
+    /// @param type The name of a registered backend, such as "CPU" or "GPU".
+    ///   To select a subdevice use "GPU:N" where s`N` is the subdevice number.
+    /// @returns shared_ptr to a new Backend or nullptr if the named backend
+    ///   does not exist.
+    static std::shared_ptr<Backend> create(const std::string& type);
```
avijit-nervana: Shouldn't it be a function of the Backend manager class?

src/ngraph/runtime/backend.hpp
```diff
@@ -17,83 +17,75 @@
...
+                      const std::vector<std::shared_ptr<runtime::TensorView>>& outputs,
+                      const std::vector<std::shared_ptr<runtime::TensorView>>& inputs) = 0;
+
+    virtual void remove_compiled_function(std::shared_ptr<Function> func);
+
+    virtual void enable_performance_data(std::shared_ptr<Function> func, bool enable) {}
+    virtual std::vector<PerformanceCounter>
+        get_performance_data(std::shared_ptr<Function> func) const;
+
+    static bool register_backend(const std::string& name, std::shared_ptr<Backend>);
```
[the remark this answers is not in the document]
rkimballn1: The backend has two APIs, one for the frameworks and one for the backend implementations. I wanted to keep all of the APIs in backend.hpp rather than split it across Backend and BackendManager. BackendManager is something that I want hidden behind the scene, not something visible to either frameworks or backend implementations. runtime::Backend is really just the API.

src/ngraph/runtime/backend_manager.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include "ngraph/util.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+#ifdef WIN32
+#define CLOSE_LIBRARY(a) FreeLibrary(a)
+#define DLSYM(a, b) GetProcAddress(a, b)
+#else
+// #define OPEN_LIBRARY(a, b) dlopen(a, b)
```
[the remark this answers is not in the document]
rkimballn1: :( I would swear I had fixed this. Done.

src/ngraph/runtime/backend_manager.cpp
```diff
@@ -0,0 +1,225 @@
...
+#define CLOSE_LIBRARY(a) dlclose(a)
+#define DLSYM(a, b) dlsym(a, b)
+#endif
+
+unordered_map<string, runtime::new_backend_t>& runtime::BackendManager::get_registry()
```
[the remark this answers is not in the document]
rkimballn1: It is a private static member function. Do you want it to just be a static function in backend_manager.cpp?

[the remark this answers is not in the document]
avijit-nervana: I am sorry - my bad. This is fine as you've done it.

src/ngraph/runtime/backend.hpp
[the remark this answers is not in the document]
avijit-nervana: Spliting the API into two files has some value in my opinion. One is the usage simplicity i.e., you don't need to know which functions are for you - rather all of them can be used by the framework or a specific backend. The other is the decoupling of the dependencies. Which actually is something needs to be address at the nGraph design level but for now I would recommend thinking of splitting this into two parts.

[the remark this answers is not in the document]
rkimballn1: I removed the register_backend from Backend and we will use the one in BackendManager. I agree with you on this.

src/ngraph/runtime/backend_manager.hpp
rkimballn1: I added comments for all of the public API

PR #1365, opened 2018-08-08
Jmenon/nbench

jmenon:


ashokei:
agree with @jbobba , warmup is more common in DL benchmarks. We noticed significant variation if we dont do warmup of atleast 10 iterations (mkldnn team recommends a warmup of 100.. upto 1000 for training measurements)

jmenon:
I agree, warmup semantics and name are better. Will change

PR #1370, opened 2018-08-08
update MKLDNN version from v0.14 -> v0.16

pruthviIntel:
updated the mkldnn version to v0.15, however v0.15 has padding requirement for kernels like convolution etc, which needs to be taken in to account while doing memory allocation (which will be addressed in the separate PR) . This PR is to verify if the MKLDNN V0.15 works , without any memory allocation changes). if the CI passes, hope we can safely bump the version


jbobba:
Can you check with @avijit-nervana and @mbrookhart on performance with the new version?

pruthviIntel:
will do

pruthviIntel:
@jbobba , can you review this... I will check the perf or if something is breaking on the bridge side with this update.

pruthviIntel:
@sharathns93 repotrted 1ms of latency improvement and throughput of 5-10images/s on resnet50 with this patch.

cmake/external_mkldnn.cmake
```diff
@@ -82,7 +82,7 @@ foreach(LIB ${MKL_LIBS})
 endforeach()
 
 set(MKLDNN_GIT_REPO_URL https://github.com/intel/mkl-dnn)
-set(MKLDNN_GIT_TAG "0e7ca73")
+set(MKLDNN_GIT_TAG "64e03a1")
```
jbobba: v0.16 is out. We might want to jump to that and skip v0.15

src/ngraph/pass/memory_layout.cpp
```diff
@@ -62,9 +64,13 @@ bool pass::MemoryLayout::run_on_function(shared_ptr<ngraph::Function> function)
...
         for (descriptor::Tensor* tensor : node->liveness_new_list)
         {
+            // get the primary tennsor view descriptors from the Tensor
+            auto tv_desc = tensor->get_primary_tensor_view();
+            auto tv_layout = tv_desc->get_tensor_view_layout();
+
             size_t offset = in_place_outputs.count(tensor)
                                 ? in_place_outputs.at(tensor)->get_pool_offset()
-                                : mm.allocate(tensor->size());
+                                : mm.allocate(tv_layout ? tv_layout->size() : tensor->size());
```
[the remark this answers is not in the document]
pruthviIntel: this is to appease some current memory manger unit tests. those doesn't have layouts.

src/ngraph/descriptor/tensor.cpp
```diff
@@ -62,6 +53,11 @@ size_t descriptor::Tensor::get_pool_offset() const
     return m_pool_offset;
 }
 
+size_t descriptor::Tensor::size() const
+{
+    return m_size;
```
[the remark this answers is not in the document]
pruthviIntel: I had this changes before, but if I don't provide the definition here..it `seg fault` on  the following test case(which doesn't create a backend and don't have layoutdesc).

https://github.com/NervanaSystems/ngraph/blob/master/test/pass_memory_layout.cpp#L207. and it breaks some of the GPU uni test use, so I though add this definition, will make it less intrusive

cmake/external_mkldnn.cmake
pruthviIntel: sure, make sense.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -173,6 +176,7 @@ void runtime::cpu::pass::CPULayout::set_native_layouts(
     for (descriptor::Input& input : node->get_inputs())
     {
         const auto& output = input.get_output();
+        auto input_node = output.get_node();
```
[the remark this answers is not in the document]
pruthviIntel: yup unused variable, will delete.

src/ngraph/runtime/cpu/cpu_layout_descriptor.hpp
```diff
@@ -39,6 +39,8 @@ namespace ngraph
                 LayoutDescriptor(const ngraph::descriptor::TensorView& tv);
                 ~LayoutDescriptor() override {}
                 size_t get_size() override { return m_size; }
+                virtual size_t size() override { return m_mkldnn_memory_size; }
+                void compute_mkldnn_memory_size(const mkldnn::memory::desc md);
```
jbobba: Remove.

pruthviIntel:
@diyessi @jmenon can you review this PR?

src/ngraph/descriptor/layout/tensor_view_layout.hpp
```diff
@@ -50,7 +50,7 @@ namespace ngraph
                 ///
                 /// When we support non-linear buffers, this will need to be something other than size_t.
                 virtual size_t get_size() = 0;
-
+                virtual size_t size();
```
diyessi: In STL, `size()` is element size, not byte size, so change this to `get_byte_size()` or `get_allocated_size()`.

src/ngraph/descriptor/tensor.cpp
```diff
@@ -29,12 +30,8 @@ descriptor::Tensor::Tensor(const element::Type& element_type,
...
     , m_next_view_id{0}
 {
-    size_t size = 1;
-    for (size_t s : primary_tensor_view->get_tensor_view_type()->get_shape())
-    {
-        size *= s;
-    }
-    m_size = size * m_element_type.size();
+    auto tvt = primary_tensor_view->get_tensor_view_type();
+    m_size = shape_size(tvt->get_shape()) * m_element_type.size();
```
diyessi: Don't cache the size.

src/ngraph/descriptor/tensor.hpp
```diff
@@ -49,24 +49,23 @@ class ngraph::descriptor::Tensor
     Tensor(const element::Type& element_type,
            PrimaryTensorView* tensor_view,
            const std::string& name);
-
+    size_t m_size;
```
Krovatkin: is this used anywhere?

jbobba:
Agree with @Krovatkin

pruthviIntel:
agree, we can do this as a part for codegen refactoring in the ticket we have for this sprint

PR #1372, opened 2018-08-08
Support reverse sequence for floating point sequence lengths

jmenon:


jmenon:
Closing this since it's a framework bug which will be fixed in the bridge

PR #1375, opened 2018-08-08
added DEX support for BatchNormRelu

pruthviIntel:


src/ngraph/runtime/cpu/builder/batch_norm.cpp
```diff
@@ -324,7 +325,17 @@ namespace ngraph
...
             }
 
+            template <>
+            void Builder::BUILDER_DECL(ngraph::op::BatchNormRelu)
+            {
+                if (!mkldnn_utils::use_mkldnn_kernel(node))
+                {
+                    throw ngraph_error("BatchNormRelu is only supported with 4-D MKLDNN kernel.");
+                }
+                build_batch_norm(external_function, node, args, out, true);
```
jbobba: you will need to templatize build_batch_norm

pruthviIntel: jayaram, isn't it by flipping relu  in `build_batch_norm ` flag we can control whether to execute Reluas post op, still do we need to tempaltize it?

jbobba: There is a static_cast to op type in build_batch_norm

pruthviIntel: ah thanks for pointing , i overlooked that part :). will fix it.

PR #1376, opened 2018-08-08
Dex non-mkldnn version of clipped relu

jbobba:


src/ngraph/runtime/cpu/builder/bounded_relu.cpp
```diff
@@ -56,6 +51,19 @@ namespace ngraph
...
                     functors.emplace_back(functor);
                 }
+                else
+                {
+                    std::function<decltype(runtime::cpu::kernel::bounded_relu<float>)> kernel;
+
+                    SELECT_KERNEL(
+                        kernel, out[0].get_element_type(), runtime::cpu::kernel::bounded_relu);
+
+                    auto alpha = dynamic_cast<const op::BoundedRelu*>(node)->get_alpha();
```
jmenon: static_cast?

src/ngraph/runtime/cpu/kernel/relu.hpp
```diff
@@ -45,6 +45,22 @@ namespace ngraph
                     out.device(eigen::global_thread_pool_device) = in0.cwiseMax(ElementType(0));
                 }
 
+                template <typename ElementType>
+                void bounded_relu(void* input0, void* output, ElementType alpha, size_t count)
+                {
+                    Eigen::array<Eigen::Index, 1> out_dims, in_dims;
+
+                    out_dims[0] = in_dims[0] = count;
```
jmenon: Thanks for keeping it consistent. In a future refactor we could just use in_dims in places like this and the cwise macros

jbobba: sounds good.

PR #1381, opened 2018-08-09
IntelGPU backend: Sin, cos, log, exp operations

shssf:
Implementation of these layers are done in clDNN (https://github.com/intel/clDNN/pull/36) and we can use it in nGraph.
The latest version of clDNN required to build.


rkimballn1:
TESTNOW

PR #1384, opened 2018-08-09
Add cudnn host data param case for CUDNN_DATA_INT8x32

csullivan:


src/ngraph/runtime/gpu/cudnn_host_parameters.hpp
```diff
@@ -60,12 +60,17 @@ namespace ngraph
                         break;
                     case CUDNN_DATA_HALF:
                     case CUDNN_DATA_INT8x4:
+                    case CUDNN_DATA_INT8x32:
```
fengleitian: add `#if CUDNN_VERSION >= 7200`

src/ngraph/runtime/gpu/cudnn_host_parameters.hpp
```diff
@@ -58,14 +58,10 @@ namespace ngraph
...
-#if CUDNN_VERSION >= 7100
-                    case CUDNN_DATA_UINT8:
-                    case CUDNN_DATA_UINT8x4:
-#endif
-                        std::string err = "datatype is not supported by cuDNN";
-                        throw std::runtime_error(err);
+                    default:
+                        throw std::runtime_error(
+                            "Encountered unhandled cudnnDataType_t during compilation.");
+                        break;
```
rkimballn1: You don't need a break after throw.

PR #1389, opened 2018-08-09
DEX Sigmoid Multiply

Krovatkin:


src/ngraph/runtime/cpu/builder/sigmoid.cpp
```diff
@@ -109,8 +114,132 @@ namespace ngraph
...
+
+                const size_t index =
+                    static_cast<size_t>(sigmoid_mul->get_input_func_type(0)) *
+                        static_cast<size_t>(ngraph::op::SigmoidMultiply::FunctionType::NumTypes) +
+                    static_cast<size_t>(sigmoid_mul->get_input_func_type(1));
+
+                switch (index)
+                {
+                case 0 /*Logistic|Logistic*/:
+                    functor = [&, tensor_size](CPURuntimeContext* ctx) {
```
jmenon: Does this op enforce an element type of float? Otherwise you'll want to disallow other types

src/ngraph/runtime/cpu/builder/sigmoid.cpp
```diff
@@ -109,8 +114,132 @@ namespace ngraph
...
+                        auto in0 = wrap_into_tensor_map<float>(arg0_tensor, tensor_size);
+                        auto in1 = wrap_into_tensor_map<float>(arg1_tensor, tensor_size);
+                        auto out_tm = wrap_into_tensor_map<float>(out_tensor, tensor_size);
+                        auto c = (in0.exp() * in1.exp()) / ((in0.exp() + 1.f) * (in1.exp() + 1.f));
```
jmenon: Some of these expressions could probably be optimized but this looks like something we'd JIT anyway I guess...

src/ngraph/runtime/cpu/builder/sigmoid.cpp
```diff
@@ -109,8 +114,132 @@ namespace ngraph
...
+                auto& arg0_tensor = tensor_data[args[0].get_name()];
+                auto& arg1_tensor = tensor_data[args[1].get_name()];
+                auto& out_tensor = tensor_data[out[0].get_name()];
+                auto tensor_size = shape_size(args[0].get_shape());
+
+                auto sigmoid_mul = static_cast<const ngraph::op::SigmoidMultiply*>(node);
+                function<void(CPURuntimeContext*)> functor;
+
+                const size_t index =
+                    static_cast<size_t>(sigmoid_mul->get_input_func_type(0)) *
```
jmenon: All this casting looks excessive. Maybe you didn't want them to be enum classes to begin with?
That being said, it's not too important

src/ngraph/runtime/cpu/builder/sigmoid.cpp
Krovatkin: yes... we only fuse for floats.

src/ngraph/runtime/cpu/builder/sigmoid.cpp
Krovatkin: 😄 It looks much nicer for `cpu_codegen` .  Here, I have to consider all 9 combinations since we don't have a compiler to build code for an arbitrary combination at runtime. When we get some way of jitting either with halide or somehow else, these casts should go away and we will end up with a nice typed enum the way @louisfeng envisioned it.

src/ngraph/runtime/cpu/builder/sigmoid.cpp
Krovatkin: I'd just leave it as is. SigmoidMultiply doesn't seem to be the most important fusion, so this is probably good enough for now. I don't want to spend a lot of time on the code that we are going to scratch in the near future as soon as we get a compiler.

src/ngraph/runtime/cpu/builder/sigmoid.cpp
jmenon: Cool. So the framework can't create instantiate this op all by itself, is that right?

src/ngraph/runtime/cpu/builder/sigmoid.cpp
jmenon: Sounds good

src/ngraph/runtime/cpu/builder/sigmoid.cpp
jmenon: Yes, that is the tradeoff for not compiling every single time :) I'll look forward to that PR

PR #1377, opened 2018-08-08
[ONNX] Support for Split op

apwojcik:

PR #1396, opened 2018-08-10
[ONNX] Batchnorm operation

arogowie-intel:
Please merge after #1377


test/onnx_import.cpp
```diff
@@ -112,3 +113,33 @@ TEST(onnx, model_split_variable_parts_2d)
...
+    // bias (3)
+    inputs.emplace_back(std::vector<float>{0., 1.});
+    // mean (3)
+    inputs.emplace_back(std::vector<float>{0., 3});
+    // var (3)
+    inputs.emplace_back(std::vector<float>{1., 1.5});
+
+    // shape (1, 2, 1, 3)
+    auto expected_output = ngraph::test::NDArray<float, 4>(
+                               {{{{-0.999995, 0., 0.999995}}, {{-0.22474074, 1., 2.2247407}}}})
```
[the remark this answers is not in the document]
pruthviIntel: i guess u need append each of the float values with   `f` {{{{-0.999995f, 0.0f, 0.999995f}}, {{-0.22474074f, 1.0f, 2.2247407f}}}})`

test/onnx_import.cpp
```diff
@@ -112,3 +113,33 @@ TEST(onnx, model_split_variable_parts_2d)
...
+                               .get_vector();
+
+    auto result_vectors = execute(function, inputs, "CPU");
```
rkimballn1: These tests should be run against all available backends, not just CPU. You should use NGRAPH_TEST instead of TEST and add this source file to the list of configured source files. Look at backend_test.in.cpp. If you want to address this later it would be ok or I can help you get this setup.

test/onnx_import.cpp
```diff
@@ -112,3 +113,33 @@ TEST(onnx, model_split_variable_parts_2d)
...
+    EXPECT_EQ(expected_output, result_vectors.front());
```
rkimballn1: You can't directly compare floats. You need to use the ngraph::test::all_close_f function which will compare the float arrays to within the specified precision. The default precision is good and should work with bfloat16 number too. This should apply to all tests where you directly compare floats.

src/ngraph/frontend/onnx_import/op/batch_norm.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include "ngraph/node_vector.hpp"
+#include "ngraph/op/batch_norm.hpp"
+
+namespace ngraph
+{
+    namespace onnx_import
+    {
+        namespace op
+        {
+            inline NodeVector batch_norm(const Node& node, const NodeVector& inputs)
```
rkimballn1: This seems like a large function to inline. Does it make any measurable difference to inline this?

PR #1402, opened 2018-08-12
IntelGPU backend: Convolution workaround operations (partially)

shssf:
@dmyershov FYI
Implemented "ConvolutionBackpropFilters", "ConvolutionBackpropData" and a cases with "Convolution" which are not supported by clDNN yet.

In this version the `data_dilation` is implemented but doesn't work yet.
Perhaps I understood the algorithm in a wrong way. This feature is not used in models we use and I didn't find clear algorithm anywhere.

`window_dilation`, `padding`, `strides` are implemented and works well.


src/ngraph/runtime/intelgpu/intelgpu_op_convolution.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+// Padding, Strides and dilation are quite nice explained
+// with animations here https://github.com/vdumoulin/conv_arithmetic
+//
+// batch axes for both input data and output data are 0
+// input channel axes for both input data and filters are 1
+// output channel axes for filters is 0
+// output channel axis for output data is 1
+// Example (Convolution):
+//   data[ 2, 1, 3, 5, 8 ]
```
rkimballn1: This is a little hard to read. I think if you indented the sample tensors like on lines 63-65 then it would be easier to read.

shssf: @rkimballn1 I think I didn't understand. Could you please point an example?
Did you mean, better format might be:
```C++
// output channel axis for output data is 1
//
// Example (Convolution):
//       data[              2,              1, 3, 5, 8 ]
//     filter[              2,              1, 2, 2, 3 ]
//     output[              2,              2, 2, 4, 6 ]
// it is like
//       data[          batch,   data_channel, 3, 5, 8 ]
//     filter[ output_channel,   data_channel, 2, 2, 3 ]
//     output[          batch, output_channel, 2, 4, 6 ]
//
```

rkimballn1: I was thinking something a little easier. No need to line up the numbers
```
// output channel axis for output data is 1
//
// Example (Convolution):
//           data[ 2, 1, 3, 5, 8 ]
//         filter[ 2, 1, 2, 2, 3 ]
//         output[ 2, 2, 2, 4, 6 ]
// it is like
//           data[ batch, data_channel, 3, 5, 8 ]
//         filter[ output_channel, data_channel, 2, 2, 3 ]
//         output[ batch, output_channel, 2, 4, 6 ]
//
```

shssf: Fixed in PR #1412

PR #1417, opened 2018-08-14
test new theme

indie:
Let us see how this goes;


rkimballn1:
So what is the plan for this review? I have looked at the output and it is not as pretty as our old theme but it does address the issue with the licenses. At this point are we just validating that the theme is functional or are we going deeper?

indie:
Alternative solution approved

PR #1419, opened 2018-08-14
[ONNX] Link nGraph statically against Protocol Buffers

apwojcik:
The change is to allow nGraph link statically against Protocol Buffers. Static linking solves Protobuf issue of the same symbol registration from different components (e.g. shared object and application).
The source code is pulled from GitHub (currently the most recent version of branch 3.5.x, which is bugfix branch for the version).
It is still possible to use whatever the operating system is offering by enabling flag NGRAPH_USE_SYSTEM_PROTOBUF when at the same time flag NGRAPH_ONNX_IMPORT_ENABLE is enabled.


CMakeLists.txt
```diff
@@ -208,13 +212,18 @@ set(NGRAPH_BUILD_DIR ${CMAKE_CURRENT_BINARY_DIR}/src/ngraph)
...
-
 if(NOT DEFINED EXTERNAL_PROJECTS_ROOT)
     set(EXTERNAL_PROJECTS_ROOT ${CMAKE_CURRENT_BINARY_DIR})
 endif()
+
+if (NGRAPH_ONNX_IMPORT_ENABLE)
+    if (NOT NGRAPH_USE_SYSTEM_PROTOBUF)
+        include (cmake/external_protobuf.cmake)
+    else()
+        find_package(Protobuf 2.6.1 REQUIRED)
```
postrational: nGraph would be compatible with more models if we started to use 3.x series of Protobuf. Latest version seems to be 3.6.1, should we use that?

postrational:
On a Mac I needed to: `brew install automake`

You should add a note about this to:
https://github.com/NervanaSystems/ngraph/blob/master/doc/sphinx/source/install.rst#macos-development

.ci/travis/ubuntu/Dockerfile
```diff
@@ -11,11 +11,13 @@ RUN apt-get update && apt-get install -y \
...
   git \
   curl \
-  protobuf-compiler \
-  libprotoc-dev \
   zlib1g \
   zlib1g-dev \
-  libtinfo-dev && \
+  libtinfo-dev \
+  unzip \
+  autoconf \
```
postrational: Please add information about these new requirements to the docs.

cmake/external_protobuf.cmake
```diff
@@ -0,0 +1,121 @@
...
+set(Protobuf_SRC_ROOT_FOLDER ${SOURCE_DIR})
+
+set(Protobuf_PROTOC_EXECUTABLE ${Protobuf_SRC_ROOT_FOLDER}/src/protoc)
+set(Protobuf_INCLUDE_DIR ${Protobuf_SRC_ROOT_FOLDER}/src)
+set(Protobuf_LIBRARY ${Protobuf_SRC_ROOT_FOLDER}/src/.libs/libprotobuf.a)
+set(Protobuf_LIBRARIES ${Protobuf_LIBRARY})
+
+if (NOT TARGET protobuf::libprotobuf)
+    add_library(protobuf::libprotobuf UNKNOWN IMPORTED)
+    set_target_properties(protobuf::libprotobuf PROPERTIES
```
rkimballn1: use target_include_directories like this https://github.com/NervanaSystems/ngraph/blob/master/cmake/external_gtest.cmake#L77

cmake/external_protobuf.cmake
```diff
@@ -0,0 +1,121 @@
...
+        INTERFACE_INCLUDE_DIRECTORIES "${Protobuf_INCLUDE_DIR}"
+        IMPORTED_LOCATION "${Protobuf_LIBRARY}")
+    add_dependencies(protobuf::libprotobuf ext_protobuf)
```
rkimballn1: use target_link_libraries as in external_gtest. You should not need to set Protobuf_INCLUDE_DIR or any of the other things you set on lines 103-106. Then to use the protobuf library all you need is to include protobuf::libprotobuf in your target_link_libraries of the object that wants to use this lib. Look at how gtest was done as an external project.

cmake/external_protobuf.cmake
apwojcik: It will result in error message (see below). The "namespace::target" semantics needs 'set_target_properties' instead. I use the same API FindProtobuf module is exposing.
```
CMake Error at cmake/external_protobuf.cmake:110 (target_include_directories):
  target_include_directories called with non-compilable target type
```

cmake/external_protobuf.cmake
apwojcik: There are two reasons why I did such way:
1. It recreates portions of the FindProtobuf API. Soon we will add libonnx which depends on libprotobuf. The module uses find_package() macro to configure protobuf. If the macro will not detect variables i.e. Protobuf_INCLUDE_DIR and Protobuf_LIBRARY it will not honor the cache and it will pull the Protobuf from the system.
2. The include directory ${Protobuf_INCLUDE_DIR} as well as ${Protobuf_LIBRARIES} does not exist at the build configuration state. Use of target_link_libraries with i.e. protobuf::libprotobuf results in error:
```
CMake Error in src/ngraph/frontend/onnx_import/CMakeLists.txt:
  Imported target "protobuf::libprotobuf" includes non-existent path

    "/home/awojcik/Snapshots/ngraph/build.debug/protobuf/src/src"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

.ci/travis/ubuntu/Dockerfile
apwojcik: Done

PR #1420, opened 2018-08-15
Jbobba/tensor roles

jbobba:


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1141,6 +1161,34 @@ void runtime::cpu::CPU_ExternalFunction::build()
...
+                m_tensor_roles[tensor->get_name()] = CPUTensorRole::INTERMEDIATE;
+            }
+        }
+    }
+
+    // Constants
+    for (auto& node : m_function->get_ordered_ops())
+    {
+        const auto c = dynamic_cast<ngraph::op::Constant*>(node.get());
+        if (c)
```
Krovatkin: `is_constant()`

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1151,6 +1199,7 @@ void runtime::cpu::CPU_ExternalFunction::build()
             function_input_index.emplace_back(tensor_data[tv->get_tensor().get_name()],
                                               arg_index,
                                               tensor_stale[tv->get_tensor().get_name()]);
+            m_tensor_roles[tv->get_tensor().get_name()] = CPUTensorRole::INPUT;
```
Krovatkin: I'm wondering if we could do `.at()` instead since it throws an exception if a role for some tensor isn't defined?

PR #1422, opened 2018-08-15
Fix broken dependencies

jmenon:


rkimballn1:
The correct solution to this problem is to add a dependency of ext_mkldnn to the libmkldnn interface in external_mkldnn.cmake

PR #1423, opened 2018-08-15
collapse dims

jbobba:
Adds CPUCollapseDims pass to compress/squeeze contiguous dimensions. 

E.g., 
Broadcast([1,1,2], AxisSet{1, 2}, [1, 4, 2, 1, 2])

will be replaced with 

Reshape([1,1,2], AxisOrder{0,1,2},[1,2])
Broadcast([1,2], AxisSet{1}, [1,8,2])
Reshape([1,8,2],AxisOrder{0,1,2},[1,4,2,1,2])

Review cpu_collapse_dims.cc. Other changes are coming in from a separate PR #1420 

This can be a device-independent pass if other backends have inexpensive Reshapes.


src/ngraph/runtime/cpu/pass/cpu_collapse_dims.cpp
```diff
@@ -0,0 +1,155 @@
...
+        if (std::dynamic_pointer_cast<op::Broadcast>(n))
+        {
+            auto node = std::dynamic_pointer_cast<op::Broadcast>(n).get();
+            auto input_shape = node->get_argument(0)->get_shape();
+            auto output_shape = node->get_shape();
+            auto operated_axes = node->get_broadcast_axes();
+
+            struct CollapsedDims cdims;
+
+            collapse_dims(output_shape, operated_axes, cdims, true);
```
Krovatkin: `collapse_dims` should just return `CollapsedDims cdims`

src/ngraph/runtime/cpu/pass/cpu_collapse_dims.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+            if (cdims.axis_set.size() == 0)
+            {
+                // Null broadcast operation, replace with reshape
```
Krovatkin: `Nop` Broadcast; null sounds ambiguous

src/ngraph/runtime/cpu/pass/cpu_collapse_dims.cpp
```diff
@@ -0,0 +1,155 @@
...
+// Skip size 1 dimensions.
+static void collapse_dims(std::vector<size_t>& output_shape,
+                          std::set<size_t> operated_axes,
+                          struct CollapsedDims& cdims,
+                          bool collapse_operated_axes)
+{
+    size_t collapse_size = 1;
+    bool operated_axes_run = false;
+    bool collapsing = false;
+    for (int output_idx = static_cast<int>(output_shape.size()) - 1; output_idx >= 0; output_idx--)
```
Krovatkin: I wonder if it might be a bit simpler to start with `output_shape().size() - 2`
```
operated_axes_run = operated_axes.count(output_shape().size() -1)
collapse_size = output_shape.at(output_shape().size()-1)
```
This way we should have no need for `collapsing`?

PR #1424, opened 2018-08-15
Enable DEX only build of ngraph

jbobba:


CMakeLists.txt
```diff
@@ -91,6 +93,9 @@ option(NGRAPH_DISTRIBUTED_ENABLE "Add distributed mode to the CPU backend" FALSE
 option(NGRAPH_DEBUG_ENABLE "Enable output for NGRAPH_DEBUG statements" FALSE)
 option(NGRAPH_ONNX_IMPORT_ENABLE "Enable ONNX importer" FALSE)
 
+cmake_dependent_option(NGRAPH_DEX_ONLY "Remove codegen from the build" FALSE
```
rkimballn1: DEX does not depend on Intel or Nvidia GPU. They are independent. We should not limit the options based on DEX_ONLY. I would rename DEX_ONLY to something more descriptive to what it actually is.

CMakeLists.txt
```diff
@@ -219,11 +224,15 @@ include(cmake/external_gtest.cmake)
 include(cmake/external_json.cmake)
 include(cmake/external_eigen.cmake)
 include(cmake/external_mkldnn.cmake)
-if (NGRAPH_USE_PREBUILT_LLVM OR DEFINED LLVM_TARBALL_URL)
-    include(cmake/external_llvm_prebuilt.cmake)
-else()
-    include(cmake/external_llvm.cmake)
+
+if (NOT NGRAPH_DEX_ONLY)
```
rkimballn1: Always include LLVM. If the CPU backend does not depend on it then it won't be built.

src/CMakeLists.txt
```diff
@@ -18,7 +18,10 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DIN_NGRAPH_LIBRARY")
 
 include_directories(ngraph)
 
-add_subdirectory(resource)
+if (NOT NGRAPH_DEX_ONLY)
```
rkimballn1: Remove this check. Always add resource. If it is not needed then it won't be built.

src/ngraph/CMakeLists.txt
```diff
@@ -183,7 +183,9 @@ endif()
 
 include_directories("${NGRAPH_INCLUDE_PATH}")
 
-add_subdirectory(codegen)
+if (NOT NGRAPH_DEX_ONLY)
```
rkimballn1: Always include codegen. If it is not needed it won't be built.

src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -136,7 +145,11 @@ if (NGRAPH_CPU_ENABLE)
         target_link_libraries(cpu_backend PRIVATE ${MPI_CXX_LIBRARIES})
     endif()
 
-    target_link_libraries(cpu_backend PUBLIC ngraph codegen libmkldnn libeigen libjson libtbb)
+    add_dependencies(cpu_backend ext_mkldnn ext_eigen)
```
rkimballn1: This add_dependencies should not be needed because the target_link_libraries already depends on libmkldnn and libeigen. I recall that there is a problem with libmkldnn where it does not depend on ext_mkldnn and it should depend on that, but that would be a change in external_mkldnn.cmake.

test/CMakeLists.txt
```diff
@@ -63,7 +63,10 @@ add_subdirectory(files)
 add_subdirectory(util)
 
 if(NGRAPH_CPU_ENABLE)
-    set(SRC ${SRC} backend_performance.cpp codegen.cpp cpu_fusion.cpp cpu_test.cpp)
+    set(SRC ${SRC} backend_performance.cpp cpu_fusion.cpp cpu_test.cpp)
```
rkimballn1: I'm all for just removing the file backend_performance.cpp from the test directory.

src/tools/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
 # ******************************************************************************
 
-add_subdirectory(compile_benchmark)
+if (NOT NGRAPH_DEX_ONLY)
```
rkimballn1: compile_benchmark has nothing to do with DEX and should be built.

test/CMakeLists.txt
```diff
@@ -63,7 +63,10 @@ add_subdirectory(files)
...
+    if (NOT NGRAPH_DEX_ONLY)
```
rkimballn1: codegen has nothing to do with DEX. There is more than just the CPU backend. Leave this test in.

CMakeLists.txt
avijit-nervana: If LLVM is always included - then cmake will need to check if it needs to update and then download if needed even if it's not used by the backend - no?

rkimballn1: It will only try to download if something depends on it, so for DEX_ONLY it will not download

PR #1433, opened 2018-08-17
Fix Macos compilaton errors

diyessi:
CPU dot gets compilation errors on MacOS.


src/ngraph/runtime/cpu/builder/dot.cpp
```diff
@@ -183,7 +183,7 @@ namespace ngraph
                                                    i * offset_a);
                                 }
                                 const float** a_array = a.data();
-                                int64_t lda_array = std::max(1L, k);
+                                int64_t lda_array = std::max(int64_t{1}, k);
```
rkimballn1: Don't you get a warning about old-style casts?

jbobba:
group_count is being used inside the lambda. Can we move it into the lambda to keep the compiler happy.  @jmenon ?

diyessi:
If it were being modified, then I think the compiler would not complain.

diyessi:
@jmenon there are actually two warnings, `offset_b` and `group_count`:
`/Users/dscypher/code/ngraph/src/ngraph/runtime/cpu/builder/dot.cpp:175:43: warning: 
      lambda capture 'offset_b' is not required to be captured for this use
      [-Wunused-lambda-capture]
                            [&, offset_a, offset_b, offset_c, m, n, k, g...
                                          ^
/Users/dscypher/code/ngraph/src/ngraph/runtime/cpu/builder/dot.cpp:175:84: warning: 
      lambda capture 'group_count' is not required to be captured for this use
      [-Wunused-lambda-capture]
  ...[&, offset_a, offset_b, offset_c, m, n, k, group_size, group_count](
                                                            ^
2 warnings generated.
`

diyessi:
Why not just use [&] and let the compiler decide what it inlines and what it gets from the enclosing frame? It reads better with everything defined together the way you wrote it.

jbobba:
If the compiler decides to not inline, wouldnt' capturing by reference lead to undefined behavior?

diyessi:
Lisp is so much simpler...

diyessi:
I am closing this for now since it just warns and it takes 20 minutes to try a change on my computer.

PR #1434, opened 2018-08-17
IntelGPU backend: Code refactoring and optimization

aalbersk:
- add gen_fun_def to generate function definitions
- refactor names of variables and if statements
- optimize code by adding generate_loops() function when it's possible
- erase tests from manifest which pass and were not deleted previously

**No algorithms changed.**

### DIFF
--- a/src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
+++ b/src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
@@ -92,27 +92,26 @@ string
     return buffer;
 }
 
-void runtime::intelgpu::generate_function_definition(codegen::CodeWriter& writer,
-                                                     const string& entry_point_name,
-                                                     const vector<string>& inputs,
-                                                     const string& output_type,
-                                                     const Shape& output_shape)
+void runtime::intelgpu::gen_func_def(codegen::CodeWriter& writer,
+                                     const string& entry_point_name,
+                                     const vector<string>& input_types,
+                                     const vector<Shape>& input_shapes,
+                                     const string& output_type,
+                                     const Shape& output_shape)
 {
     writer << "__kernel void " << entry_point_name << "(";
 
-    const size_t inputs_number = inputs.size();
+    const size_t inputs_number = input_types.size();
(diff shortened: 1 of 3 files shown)


src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -92,6 +92,29 @@ string
     return buffer;
 }
 
+void runtime::intelgpu::generate_function_definition(codegen::CodeWriter& writer,
```
shssf: I would propose to change the interface.
It might be better (discussable) to use following:
```
gen_func_def(
codegen::CodeWriter& writer,
const string& entry_point_name,
const vector<string>& input_types, 
const vector<string>& input_shapes, 
const string& output_type, 
const string& output_shape, 
)
```
Assume all inputs are started with `"input"` string and a number (input0, input1, etc.) added in a loop.

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -823,55 +823,36 @@ void runtime::intelgpu::do_logic_kernel(cldnn::topology& topology,
...
     codegen::CodeWriter writer;
+    vector<size_t> gws;
 
-    writer << "__kernel void " << entry_point_name << "(const __global " << inputA_type << " inputA"
-           << array_dims(inputA_shape) << ", const __global " << inputB_type << " inputB"
-           << array_dims(inputB_shape) << ", __global char output" << array_dims(output_shape)
-           << ")\n";
+    runtime::intelgpu::gen_func_def(writer,
+                                    entry_point_name,
+                                    {2, input0_type},
```
shssf: Are you sure that types will be the same in all cases?
I mean, may be better to use `{input0_type, input1_type}`

aalbersk: According to docs, e.g. http://ngraph.nervanasys.com/docs/latest/ops/greater_eq.html or http://ngraph.nervanasys.com/docs/latest/ops/and.html or http://ngraph.nervanasys.com/docs/latest/ops/equal.html

PR #1435, opened 2018-08-17
ArgMin

Krovatkin:


src/ngraph/op/arg_minmax.cpp
```diff
@@ -0,0 +1,69 @@
...
+    else
+    {
+        output_shape.erase(output_shape.begin() + axis);
+    }
+
+    set_value_type_checked(is_int64 ? element::i64 : element::i32, output_shape);
+}
+
+void op::IndexReduction::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
+{
```
diyessi: No need for this method; backprop won't come here.

src/ngraph/op/arg_minmax.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "ngraph/op/broadcast.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+op::IndexReduction::IndexReduction(const std::string& node_type,
+                                   const std::shared_ptr<Node>& arg,
+                                   size_t axis,
+                                   bool keep_dimensions,
+                                   bool is_int64)
```
aprocter: This is the first op to have an opinion on what type an index should have. Morally it seems to me like `uint64` is the right choice since we usually use `size_t` for stuff like this in nGraph, but I suppose if we go with that we'll be forcing an extra conversion onto the bridges that backends will have to optimize away, so... oh well.

Will int32 and int64 suffice? (I know those are all the options where TensorFlow is concerned.)

I think maybe we should pass this in as an `element::Type` parameter and check that it's in the allowed set, rather than pass a bool here. That way if we change our minds later and need to support argmax/argmin on a wider set of output types there won't be as big an issue where backwards-compat is concerned.

src/ngraph/op/arg_minmax.hpp
```diff
@@ -0,0 +1,107 @@
...
+#include "ngraph/axis_set.hpp"
+#include "ngraph/graph_util.hpp"
+#include "ngraph/op/util/arithmetic_reduction.hpp"
+#include "ngraph/op/util/requires_tensor_view_args.hpp"
+
+namespace ngraph
+{
+    namespace op
+    {
+        /// \brief Tensor IndexReduction operation.
```
diyessi: Put the hairy description in the .rst file in the doc so we don't need to move it their later.

src/ngraph/runtime/cpu/builder/arg_minmax.cpp
```diff
@@ -0,0 +1,114 @@
...
+                auto& out_tensor = tensor_data[out[0].get_name()];
+                bool is_int64 = argmin->is_int64();
+                auto axis = argmin->get_axis();
+                auto in_shape = args[0].get_shape();
+                auto out_shape = out[0].get_shape();
+
+                auto element_type = args[0].get_element_type();
+                if (element_type == element::f32)
+                {
+                    if (is_int64)
```
diyessi: Just compute the 64-bit arg and then assign it to the appropriate result instead of duplicating the computing.

src/ngraph/op/arg_minmax.cpp
```diff
@@ -0,0 +1,69 @@
...
+    , m_axis(axis)
+    , m_is_int64(is_int64)
+{
+    auto rank = arg->get_shape().size();
+    if (rank < 1)
+    {
+        throw ngraph_error("Tensor's rank should be at least 1");
+    }
+
+    if (rank <= axis)
```
aprocter: `TYPE_CHECK_ASSERT` for this

src/ngraph/op/arg_minmax.hpp
```diff
@@ -0,0 +1,107 @@
...
+            /// \brief Constructs a IndexReduction operation.
+            ///
+            /// \param arg The tensor view to be IndexReduction.
+            /// \param axis The axis along which to compute an index for minimum/maximum
+            /// \param keep_dimensions keep the axis dimension.
+            /// \param is_int64 produce indices in int64 or int32.
+            IndexReduction(const std::string& node_type,
+                           const std::shared_ptr<Node>& arg,
+                           size_t axis,
+                           bool keep_dimensions,
```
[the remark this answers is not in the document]
diyessi: Agreed. It should be easier to optimize if all reshapes are reshapes, instead of some hiding in other ops.

src/ngraph/op/util/index_reduction.cpp
```diff
@@ -0,0 +1,49 @@
...
+    : RequiresTensorViewArgs(node_type, {arg})
+    , m_axis(axis)
+{
+    auto rank = arg->get_shape().size();
+
+    TYPE_CHECK_ASSERT(this, rank >= 1) << "Tensor's rank should be at least 1";
+    TYPE_CHECK_ASSERT(this, axis < rank) << "Axis " << axis << "is greater than rank of" << rank;
+    TYPE_CHECK_ASSERT(this,
+                      index_element_type == element::i32 || index_element_type == element::i64)
+        << "Index element type should be i64 or i32";
```
aprocter: `s/should/must/`

src/ngraph/op/util/index_reduction.cpp
```diff
@@ -0,0 +1,49 @@
...
+op::util::IndexReduction::IndexReduction(const std::string& node_type,
+                                         const std::shared_ptr<Node>& arg,
+                                         size_t axis,
+                                         const element::Type& index_element_type)
+    : RequiresTensorViewArgs(node_type, {arg})
+    , m_axis(axis)
+{
+    auto rank = arg->get_shape().size();
+
+    TYPE_CHECK_ASSERT(this, rank >= 1) << "Tensor's rank should be at least 1";
```
aprocter: `s/should/must/`

src/ngraph/op/util/index_reduction.cpp
```diff
@@ -0,0 +1,49 @@
...
+    TYPE_CHECK_ASSERT(this, axis < rank) << "Axis " << axis << "is greater than rank of" << rank;
```
aprocter: Spaces around the `"is greater than rank of"` string

src/ngraph/op/util/index_reduction.hpp
```diff
@@ -0,0 +1,45 @@
...
+namespace ngraph
+{
+    namespace op
+    {
+        namespace util
+        {
+            class IndexReduction : public util::RequiresTensorViewArgs
+            {
+            public:
+                size_t get_axis() const { return m_axis; }
```
aprocter: `get_reduction_axis` would be more consistent with the existing reduction ops

src/ngraph/runtime/cpu/builder/argmin.cpp
```diff
@@ -0,0 +1,110 @@
...
+                            ngraph::runtime::reference::argmin<float, int64_t>(
+                                static_cast<float*>(arg_tensor),
+                                static_cast<int64_t*>(out_tensor),
+                                in_shape,
+                                out_shape,
+                                axis);
+                        };
+                    }
+                    else
+                    {
```
diyessi: if the out[0] element type were f32 you would still go down this branch. Instead, check for get_output_element_type(0) being element::i64 and element::i32 and throw if it's neither case.

Krovatkin: It should happen. There's a check in ArgMin c-tor that ensures that output type is either `i32` or `i64`

diyessi: But that's a long way away from this, so if someone added int16 in the ctor, strange things would happen here. It's safer to check for the cases that work instead of assuming it's one or the other.

Krovatkin: this is the reason why I preferred a flag rather an explicit type(which is while being more extensible probably will not be extended in this case) to not give a maintainer an opportunity to mess up :-( *sighs* I'll change it.

PR #1439, opened 2018-08-17
Use the Eigen OpenMP thread pool device

jmenon:


jmenon:
Closing till someone fixes the CI

PR #1443, opened 2018-08-19
IntelGPU backend: BatchNormBackprop operation

shssf:


src/ngraph/runtime/intelgpu/intelgpu_op_batchnorm.cpp
```diff
@@ -118,7 +124,7 @@ void runtime::intelgpu::do_create_variance(cldnn::topology& topology,
 {
     if (input_shape.size() < 2 || input_shape.size() > 4)
     {
-        throw invalid_argument("intelgpu::do_create_mean_variance() wrong input shapes.");
+        throw invalid_argument("intelgpu::do_create_variance() wrong input shapes.");
```
rkimballn1: The message would be more useful if it said what was the wrong shape and what is expected.

PR #1445, opened 2018-08-20
Control dependencies

Krovatkin:


src/ngraph/function.cpp
```diff
@@ -85,24 +85,28 @@ void Function::init()
...
+                       {
+                           auto it = std::find(m_parameters.begin(), m_parameters.end(), node);
+                           if (it == m_parameters.end())
+                           {
+                               throw ngraph_error("Function references undeclared parameter");
+                           }
+                       }
+                   },
+                   true /*include control dependencies*/);
 }
```
diyessi: Why not add the control dependencies as an arg to `Function::get_ordered_ops` and pass it along to the sort?

test/control_dependencies.cpp
```diff
@@ -0,0 +1,211 @@
...
+            set_output_type(0, args.at(0)->get_element_type(), args.at(0)->get_shape());
+        }
+        else
+        {
+            auto dn = *(deps.begin());
+            set_output_type(0, dn->get_element_type(), dn->get_shape());
+        }
+    }
+};
+
```
diyessi: Check that something terrible doesn't happen if you introduce a circular dependency with control deps, something which can happen easier than you might think.

PR #1444, opened 2018-08-20
Statically link cpu backend into ngraph shared library

rkimballn1:

PR #1449, opened 2018-08-21
[ONNX] Fix build for CentOS 7

apwojcik:
This is just the build fix. There will be separate PR for unit-test enablement.


rkimballn1:
#1444 has been reverted so this is no longer needed

rkimballn1:
#1444 has been reverted so this is no longer needed

PR #1452, opened 2018-08-21
get_default_order

Krovatkin:


src/ngraph/util.cpp
```diff
@@ -14,6 +14,7 @@
 * limitations under the License.
 *******************************************************************************/
 
+#include <algorithm>
```
diyessi: Need to also include `<numeric>` for `iota`. This is one of those files that some compilers implicitly include and some don't.

src/ngraph/util.cpp
```diff
@@ -449,3 +450,10 @@ void ngraph::check_fp_values_isnan(const char* name, const double* array, size_t
         }
     }
 }
+
+AxisVector ngraph::get_default_order(size_t rank)
```
diyessi: You always call this on `shape.size()` so why not make it take `Shape` as its arg?
And maybe it should be `AxisVector AxisVector::default_order(Shape)`

PR #1453, opened 2018-08-21
ArgMax

Krovatkin:


Krovatkin:
The ultimate sequel would be `ArgSort` but that one is trickier :P

PR #1455, opened 2018-08-21
nvgpu reduction optimization

fengleitian:
Sum 46,878us -> 10,191us

before:
```
Benchmarking tfl/1024/mxnet-ngraph-Function_4-CompileSubgraph-fprop_cache.bprop.json, GPU backend, 10 iterations.
compile time: 13,305ms
181.7ms per iteration

Sum              46,787us
Dot              36,627us
Slice            18,563us
Multiply         17,284us
ReplaceSlice     13,277us
Add               8,717us
Reshape           8,547us
Subtract          5,942us
Negative          5,684us
Divide            3,511us
Broadcast         2,797us
Result              298us
OneHot               30us
```

after:
```
Benchmarking tfl/1024/mxnet-ngraph-Function_4-CompileSubgraph-fprop_cache.bprop.json, GPU backend, 10 iterations.
compile time: 13,897ms
145.7ms per iteration

Dot              35,221us
Slice            18,837us
Multiply         17,870us
ReplaceSlice     13,481us
Sum              10,191us
Add               9,033us
Reshape           8,559us
Subtract          5,944us
Negative          5,730us
Divide            3,584us
Broadcast         2,875us
Result              297us
OneHot               30us
```


src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -181,6 +181,86 @@ void runtime::gpu::CudaKernelBuilder::get_ew_collective_op(
...
+                    writer << "reduce_idx += idx" << j << " * reduce_strides" << j << ";\n";
+                }
+                writer << "int idx" << rr - 1 << " = 0;\n";
+                writer << "uint32_t step = reduce_strides" << rr - 1 << ";\n";
+                //unroll last reduction axis
+                writer << "for(; idx" << rr - 1 << "< (reduce_shape" << rr - 1 << " >> 3); idx"
+                       << rr - 1 << "++)\n";
+                writer.block_begin();
+                {
+                    for (int k = 0; k < 8; k++)
```
csullivan: Since you always reduce on the last reduction axis, you could make the length of that axis part of the kernel and unroll it completely to avoid the remainder loop. Would help most when the length is less than 8.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -851,8 +851,7 @@ namespace ngraph
                         {
                             kernel::emit_memcpyDtD(writer, out[0], args[0]);
                         }
-                        // descriptors for tensors  with <= 4 dimensions
-                        else
+                        else if (out[0].get_shape().size() == 0)
```
csullivan: Is this intentionally left here to allow cudnn to handle the reduction to scalar case? I assume then it was faster than the cuda impl.? In general, do you have an understanding as to why cudnn is slower for the model you tested?

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
fengleitian: Theoretically it will help. But I actually  tested it, it make no performance difference. and the way to completely unroll will create a lot of kernels instead of only one here.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: Yes, It's intentionally left here, from my test, cudnn didn't do well for non-scalar case. And I will have another PR that optimized scalar case so we don't need cudnn for reduce anymore.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -181,6 +181,86 @@ void runtime::gpu::CudaKernelBuilder::get_ew_collective_op(
...
+    writer << "extern \"C\" __global__ void cuda_" << name << args.get_input_signature();
+    writer.block_begin();
+    {
+        writer << "uint32_t tid = blockIdx.x * blockDim.x + threadIdx.x; \n";
+        writer << "if (tid < nthreads)\n";
+        writer.block_begin();
+        {
+            if (out_rank > 0)
+            {
+                writer << "uint32_t out_idx = tid;\n";
```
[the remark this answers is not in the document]
fengleitian: changed to 'dim_idx_generator`

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
csullivan: In that case, have you tried removing the unroll completely? Wondering if the reason you didn't see any difference is that it is spending all the time waiting on those strided loads anyway.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -181,6 +181,86 @@ void runtime::gpu::CudaKernelBuilder::get_ew_collective_op(
...
+            }
+            writer << "uint32_t in_idx = 0;\n";
+            writer << data_types[1] << " r = 0;\n";
+
+            //loop through all reduction axis
+            int64_t i = 0;
```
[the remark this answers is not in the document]
fengleitian: fixed.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -181,6 +181,86 @@ void runtime::gpu::CudaKernelBuilder::get_ew_collective_op(
...
+                {
+                    for (int k = 0; k < 8; k++)
+                    {
+                        writer << "r = " << reduce_op << "(r , in[reduce_idx]);\n";
+                        writer << "reduce_idx += step;\n";
+                    }
+                }
+                writer.block_end();
+                writer << "idx" << rr - 1 << " <<= 3;\n";
+                writer << "for(; idx" << rr - 1 << "< reduce_shape" << rr - 1 << "; idx" << rr - 1
```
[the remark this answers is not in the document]
fengleitian: fixed.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -181,6 +181,86 @@ void runtime::gpu::CudaKernelBuilder::get_ew_collective_op(
...
+            {
+                writer << "uint32_t reduce_idx = in_idx;\n";
+                for (int64_t j = 0; j < rr - 1; j++)
+                {
+                    writer << "reduce_idx += idx" << j << " * reduce_strides" << j << ";\n";
+                }
+                writer << "int idx" << rr - 1 << " = 0;\n";
+                writer << "uint32_t step = reduce_strides" << rr - 1 << ";\n";
+                //unroll last reduction axis
+                writer << "for(; idx" << rr - 1 << "< (reduce_shape" << rr - 1 << " >> 3); idx"
```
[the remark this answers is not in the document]
fengleitian: fixed.

src/ngraph/runtime/gpu/gpu_emitter.cpp
csullivan: It's still a bit surprising to me that its performance is much much worse. 4x improvement over cudnn reduce doesn't seem reasonable - even though I'm happy about it. Makes me wonder if there are some cudnnReduce options we missed originally.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1323,6 +1323,111 @@ size_t runtime::gpu::CUDAEmitter::build_softmax_divide(const std::vector<std::st
...
+    NVShape non_reduce_in_strides;
+    NVShape reduce_shape;
+    NVShape reduce_strides;
+    NVShape input_strides = row_major_strides(input_shape);
+    for (int i = 0; i < rank; i++)
+    {
+        if (reduce_flag[i] != 0)
+        {
+            reduce_shape.push_back(input_shape[i]);
+            reduce_strides.push_back(input_strides[i]);
```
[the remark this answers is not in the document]
fengleitian: changed `non_reduce_in_strides` to `non_reduce_strides`

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: I did check the options, no something I'm aware.

fengleitian: Case might be cudnn might be optimized for some case but not for others, and possible things may happen to our optimization. That's why we need an tuning framework.

csullivan: 💯

PR #1456, opened 2018-08-22
Fix compilation issues after #1444

apwojcik:


postrational:
Travis reports:
```
C++ exception with description "Backend 'INTERPRETER' not registered" thrown in the test body.
```

rkimballn1:
#1444 has been reverted so this is no longer needed

PR #1459, opened 2018-08-22
RNN fusion (inference)

csullivan:
This provides 4 graph fusions: Sigmoid, LSTMFusion, RNNFusion, MultiLayerRNNFusion.

In each case, the RNN inference kernels are instrumented via cuDNN. 

Roughly 4x performance improvement on rnn-10-step-fusion. Will be improved further when pass::AlgebraicSimplification is updated to remove the `Slice->Reshape->Concat`s which are residuals from RNN fusion.

Before:
```
Benchmarking ../../ngraph_dev/clang/rnn_models/rnn-10-step-fusion-test.json, GPU backend, 10 iterations.
compile time: 7,196ms
2.5ms per iteration
```
After:
```
Benchmarking ../../ngraph_dev/clang/rnn_models/rnn-10-step-fusion-test.json, GPU backend, 10 iterations.
compile time: 5,133ms
0.6ms per iteration
```

Limitations:
* Currently only implement static RNN evaluation. A persistent RNN plan can be added for evaluating sequence and layer cells concurrently.
* Fused dropout layers within RNN are supported by cuDNN and could be added to RNN pattern matcher.
* cuDNN supports input projection layer which is currently unsupported
* cuDNN supports RNN modes: Tanh, Relu, Gru, LSTM. This PR only enables support for LSTM.


src/ngraph/runtime/gpu/CMakeLists.txt
```diff
@@ -41,6 +41,9 @@ set(SRC
     pass/gpu_layout.cpp
     pass/tensor_memory_reservation.cpp
     gpu_kernel_args.cpp
+    pass/gpu_rnn_fusion.cpp
+    ../gpu/op/lstm.cpp
```
rkimballn1: replace the `../gpu` with something more elegant like ${CMAKE_CURRENT_SOURCE_DIR}

src/ngraph/runtime/gpu/cudnn_descriptors.hpp
```diff
@@ -135,6 +135,19 @@ namespace ngraph
                 }
             };
 
+            template <>
```
rkimballn1: Specializations can be moved to the cpp file. This makes the header much cleaner

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1545,6 +1547,56 @@ namespace ngraph
                 }
                 writer.block_end();
             }
+
+            template <>
+            void GPU_Emitter::EMITTER_DECL(ngraph::op::gpu::Lstm)
+            {
+                // auto lstm = static_cast<const ngraph::op::gpu::Lstm*>(node);
```
rkimballn1: How about throwing an exception in here for now?

src/ngraph/runtime/gpu/op/lstm.cpp
```diff
@@ -0,0 +1,185 @@
...
+                                 new_args.at(1),
+                                 new_args.at(2),
+                                 new_args.at(3),
+                                 new_args.at(4),
+                                 new_args.at(5),
+                                 new_args.at(6));
+    }
+    else
+    {
+        if (new_args.size() != 5 && m_fused_inputs)
```
rkimballn1: For all of these test/throw check you should use Adam Proctor's new exception class. It makes it easy to make much more expressive errors, like adding what the actual arg count is and what is expected in this example.
This comment applies to all such tests in this file.

src/ngraph/runtime/gpu/op/lstm.hpp
```diff
@@ -0,0 +1,103 @@
...
+{
+    namespace op
+    {
+        namespace gpu
+        {
+            class Lstm : public util::RequiresTensorViewArgs
+            {
+            public:
+                // INPUTS:
+                // [0] - xt, input tensor of layout TNC, Shape{sequence length*batch_size, feature_size}
```
rkimballn1: Use doxygen comments like `/// \brief`, `/// \param`
If you use the right syntax then the compiler will complain if you do something wrong or forget to document a parameter

src/ngraph/runtime/gpu/op/lstm.cpp
```diff
@@ -0,0 +1,185 @@
...
+    {
+        throw ngraph_error("input_xt_1 size is not equal t*n*c");
+    }
+
+    if (i2h_bias->get_shape()[0] != i2h_weights->get_shape()[0] ||
+        h2h_bias->get_shape()[0] != h2h_weights->get_shape()[0])
+    {
+        throw ngraph_error("bias and weights_shape are not compatible");
+    }
+
```
pruthviIntel: Can you also add a check , to verify if the feature_size/shape of ht and ct are same?.

test/gpu_fusion.cpp
```diff
@@ -0,0 +1,330 @@
...
+        EXPECT_TRUE(test::all_close(gpu_results.at(i), int_results.at(i), 1.0e-4f, 1.0e-4f));
+    }
+}
+
+TEST(gpu_fusion, rnn_fusion_inter_vs_gpu_1rnn_layer_3lstm_cell)
+{
+    const std::string file_name("mxnet/1rnn_layer_3lstm_cell.json");
+    auto gpu_f = make_function(file_name);
+    auto int_f = make_function(file_name);
+    test::Uniform<float> rng(0.0f, 1.0f);
```
pruthviIntel: same here

test/gpu_fusion.cpp
```diff
@@ -0,0 +1,330 @@
...
+    auto rnn_count = count_ops_of_type<op::gpu::Rnn>(func);
+    EXPECT_EQ(ref_rnn_count, rnn_count);
+}
+
+TEST(gpu_fusion, fuse_rnn_across_2layer_1timestep)
+{
+    const std::string file_name("mxnet/2rnn_layer_1timestep.json");
+    auto gpu_f = make_function(file_name);
+    auto int_f = make_function(file_name);
+    test::Uniform<float> rng(0.0f, 1.0f);
```
pruthviIntel: same here as above.

test/gpu_fusion.cpp
```diff
@@ -0,0 +1,330 @@
...
+        EXPECT_TRUE(test::all_close(gpu_results.at(i), int_results.at(i), 1.0e-4f, 1.0e-4f));
+    }
+}
+
+TEST(gpu_fusion, rnn_fusion_inter_vs_gpu_2rnn_layer_3lstm_cell)
+{
+    const std::string file_name("mxnet/2rnn_layer_3lstm_cell.json");
+    auto gpu_f = make_function(file_name);
+    auto int_f = make_function(file_name);
+    test::Uniform<float> rng(0.0f, 1.0f);
```
pruthviIntel: same here

src/ngraph/runtime/gpu/pass/gpu_rnn_fusion.cpp
```diff
@@ -0,0 +1,848 @@
...
+                                                   pattern_map[weights_h2h],
+                                                   pattern_map[input_xt],
+                                                   pattern_map[weights_i2h],
+                                                   pattern_map[bias_h2h],
+                                                   pattern_map[bias_i2h],
+                                                   pattern_map[ct_1]);
+        }
+        else if (pattern_map[hidden_ht]->get_arguments().size() &&
+                 pattern_map[ct_1]->get_arguments().at(0)->get_instance_id() ==
+                     pattern_map[hidden_ht]->get_arguments().at(0)->get_instance_id())
```
pruthviIntel: thanks for this check, i will port the same thing to CPU PM :)

src/ngraph/runtime/gpu/pass/gpu_rnn_fusion.cpp
```diff
@@ -0,0 +1,848 @@
...
+        // this still has a bug vector: if the hidden input ht is a non-broadcasted constant
+        // it will be misclassified as input data xt
```
pruthviIntel: can we have case, where it will be not broadcasted from constant?, usually the layout will be `tnc` or `ntc` . so i guess`constant` will be end up getting broadcasted.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -289,6 +307,24 @@ cudnnFilterDescriptor_t& runtime::gpu::CUDNNEmitter::get_cudnn_filter_descriptor
     return filter_descriptor;
 }
 
+cudnnFilterDescriptor_t& runtime::gpu::CUDNNEmitter::get_nd_filter_descriptor(
```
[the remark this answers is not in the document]
csullivan: Same as above, for RNNs we need a filter descriptor that represents a 3d filter tensor. The smallest dimension filter supported by the linked function is 4d.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -82,6 +82,24 @@ cudnnTensorDescriptor_t& runtime::gpu::CUDNNEmitter::tensor_descriptor_from_shap
     return desc;
 }
 
+cudnnTensorDescriptor_t& runtime::gpu::CUDNNEmitter::get_nd_tensor_descriptor(
```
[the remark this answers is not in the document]
csullivan: Originally, I tried altering these functions to support Nd for dimensions less than 4, but I had forgotten that some of our other ops depend on smaller tensors being 4d, padded with 1s.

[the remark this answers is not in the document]
fengleitian: Right, for now looks like better to separate the pad_dimension and get_descritpor.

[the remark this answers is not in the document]
fengleitian: I'm good for now.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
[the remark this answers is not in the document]
fengleitian: I see.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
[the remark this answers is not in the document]
csullivan: Yes, agreed, they should be separated and then a composite function used where padded 4d tensors are needed

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1545,6 +1547,56 @@ namespace ngraph
...
+                // for (size_t i = 1; i < out.size(); i++)
+                // {
+                //     writer << ", " << out[i].get_name();
+                // }
+                // writer << "}.data()";
+                // writer << ");\n";
+            }
+
+            template <>
+            void GPU_Emitter::EMITTER_DECL(ngraph::op::gpu::Rnn)
```
fengleitian: As you already mentioned it's request newer version of CUDNN. we should add a version guard somewhere. 

Since currently  we don't support old CUDNN version, we can fall back to the normal ops for old CUDNN version.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -93,11 +93,13 @@
 #include "ngraph/op/sum.hpp"
 #include "ngraph/op/tan.hpp"
 #include "ngraph/op/tanh.hpp"
+#include "ngraph/runtime/cpu/op/rnn.hpp"
```
fengleitian: why it request `cpu/op/rnn.hpp`

src/ngraph/runtime/gpu/cudnn_descriptors.hpp
csullivan: I believe these specializations actually need to be defined in the header because the template function build<T> requires their definition. The forward declaration of 
```
            template <typename T>
            struct cudnn_descriptor;
``` 
doesn't suffice in this case.

csullivan:
```
cudnn_descriptors.hpp:56:41: error: implicit instantiation of undefined template 'ngraph::runtime::gpu::cudnn_descriptor<cudnnTensorStruct *>'
                        descriptor_() { cudnn_descriptor<T>::create(stored_descriptor); }
```

test/gpu_fusion.cpp
```diff
@@ -0,0 +1,330 @@
...
+    shared_ptr<Function> func = ngraph::deserialize(ss);
+    return func;
+}
+
+TEST(gpu_fusion, rnn_fusion_inter_vs_gpu_1lstm_cell)
+{
+    const std::string file_name("mxnet/1_lstm_cell_forward.json");
+    auto gpu_f = make_function(file_name);
+    auto int_f = make_function(file_name);
+    test::Uniform<float> rng(0.0f, 1.0f);
```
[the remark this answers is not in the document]
csullivan: I'm also seeing numerical correctness issues for negative input values with the fused cuDNN RNN kernel.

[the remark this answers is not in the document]
pruthviIntel: hmm, I am seeing the same on the MKLDNN side as well. however, when I run MXNET unit test through ngraph, its numpy reference values matches to the  INTERPRETER but not with the fused MKLDNN RNN kernel.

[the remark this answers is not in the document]
csullivan: Right, that is the same behavior I see. If I use the reference graph on GPU it matches the INTERPRETER with negative values. However, with the fused cuDNN RNN kernel does not match the unfused INTERPRETER result for negative values..

How can I enable the MKLDNN RNN code to compare it to the cuDNN RNN impl.? Would be informative to see if they are both incorrect in the same way.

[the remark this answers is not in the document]
pruthviIntel: you can enable this passes, https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/cpu_external_function.cpp#L375-#L381

test/gpu_fusion.cpp
```diff
@@ -0,0 +1,422 @@
...
+    vector<float> expected_ct;
+    for (size_t i = 0; i < 10 * 100; i++)
+    {
+        if (i < 1000)
+        {
+            expected_ct.push_back(0.964028f);
+        }
+        else
+        {
+            expected_ct.push_back(2.0f);
```
fengleitian: won't get here?

PR #1463, opened 2018-08-23
Validate/infer types as a virtual function

diyessi:
First part of moving op validation/inference out of constructors into a separate method. In order to not need to convert every op in one large PR, this PR allows for the separate method to be gradually added to ops and called by the constructor. When all ops have the new method and everything works the new way, we can remove the call from the constructors.

This also replaces several ways of adding outputs for a node with one new way to define outputs for a node.

BroadcastLike, ScalarConstantLike
Set output shape separately from output types
Remove some obsolete tests
Replace *Like ops with fixed type versions


src/ngraph/node.cpp
```diff
@@ -46,32 +46,47 @@ Node::Node(const std::string& node_type, const NodeVector& arguments)
             m_inputs.emplace_back(this, i++, output);
         }
     }
+    set_output_size(output_size);
 }
 
-void Node::set_value_type_checked(const element::Type& element_type, const Shape& shape)
+static bool in_transition = true;
```
Krovatkin: I think this deserves a comment. Took me a bit to remember what it meant.

src/ngraph/op/batch_norm.hpp
```diff
@@ -93,6 +95,16 @@ namespace ngraph
                                            const NodeVector& deltas) override;
 
         private:
+            enum
```
Krovatkin: I think I'm worried more about code duplication for BatchNormBackprop rather than keeping this `enum` private

src/ngraph/op/convolution.cpp
```diff
@@ -328,8 +357,8 @@ op::Convolution::Convolution(const shared_ptr<Node>& data_batch,
                   filters,
                   window_movement_strides,
                   window_dilation_strides,
-                  default_padding(data_batch),
-                  default_padding(data_batch))
+                  CoordinateDiff(),
```
Krovatkin: Okay, now I see what you meant.

src/ngraph/runtime/interpreter/int_backend.cpp
```diff
@@ -63,6 +64,7 @@ bool runtime::interpreter::INTBackend::compile(shared_ptr<Function> function)
     {
         instance.m_is_compiled = true;
         pass::Manager pass_manager;
+        pass_manager.register_pass<pass::NopElimination>();
```
Krovatkin: okay, adding optimization passes to Interpreter which is supposed to be our reference backend is officially scary.

src/ngraph/node.cpp
```diff
@@ -46,32 +46,47 @@ Node::Node(const std::string& node_type, const NodeVector& arguments)
...
+{
+    if (!in_transition)
+    {
+        validate_and_infer_types();
+    }
+}
+
+void Node::set_output_size(size_t n)
+{
+    m_outputs.clear();
```
Krovatkin: wonder if it makes to first check if for all outputs the element type is `unspecified` meaning this is the first time running this function. This way we could try catching any situations resulting in overriding already set types?

src/ngraph/op/parameter.hpp
```diff
@@ -45,12 +45,16 @@ namespace ngraph
...
 
+            void validate_and_infer_types() override;
+
             bool get_cacheable() const { return m_cacheable; }
             virtual std::shared_ptr<Node>
                 copy_with_new_args(const NodeVector& new_args) const override;
 
         protected:
             bool m_cacheable;
+            Shape m_shape;
```
Krovatkin: I wonder if it would be possible to set these on an output tensor descriptor. A user can always override it later?

src/ngraph/autodiff/adjoints.cpp
```diff
@@ -28,34 +28,35 @@
...
+    std::shared_ptr<Node> zero = std::make_shared<op::ScalarConstantLike<double>>(node, 0.0);
+    std::shared_ptr<Node> bzero = std::make_shared<op::BroadcastLike>(zero, node, AxisSet{});
+    return bzero;
 }
 
 NodeVector make_zeros(std::shared_ptr<Node> x)
 {
     NodeVector zeros;
-    for (auto& output : x->get_outputs())
+    auto& outputs = x->get_outputs();
```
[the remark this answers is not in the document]
diyessi: Not sure how to do what you're saying.

src/ngraph/autodiff/adjoints.cpp
```diff
@@ -161,11 +156,13 @@ void autodiff::Adjoints::add_delta(const std::shared_ptr<Node>& x,
                                    const std::shared_ptr<Node>& delta,
                                    size_t output_index)
 {
+#if 0
```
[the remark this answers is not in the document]
diyessi: I thought I had. Will remove.

src/ngraph/node.cpp
diyessi: ok

src/ngraph/node.cpp
diyessi: If validate... gets called a second time (like after a pass), it wants to reset things.
It would be nice to have a way to just do the new nodes spliced in, but that seems hard to do without a lot of error-prone checks in every op.

src/ngraph/op/parameter.hpp
diyessi: There isn't one until validate runs.

src/ngraph/node.cpp
```diff
@@ -46,32 +45,48 @@ Node::Node(const std::string& node_type, const NodeVector& arguments)
...
+    for (size_t i = m_outputs.size(); i < n; ++i)
     {
-        throw ngraph_error("Setting value type to a different ValueType");
+        auto tensor_view_descriptor = make_shared<descriptor::TensorView>(
+            element::unspecified, Shape(), ngraph::descriptor::Tensor::make_tensor_name(this, i));
```
aprocter: Will the ambiguity between scalar and shape-not-yet-specified be a problem here?

src/ngraph/node.cpp
```diff
@@ -217,7 +232,10 @@ const Shape& Node::get_shape() const
 {
     if (get_output_size() != 1)
     {
-        throw ngraph_error("get_shape() must be called on a node with exactly one output.");
+        stringstream es;
```
aprocter: FWIW we could use `NGRAPH_ASSERT` here:

```
NGRAPH_ASSERT(get_output_size() == 1)
   << "get_shape() must be called on a node with exactly one output; node has "
   << get_output_size()
   << "outputs (node: " << description() << ")";
```

aprocter: (Note that if we do, the exception class thrown will be `AssertionFailure`)

src/ngraph/descriptor/tensor_view.hpp
```diff
@@ -35,35 +35,22 @@ namespace ngraph
...
-            {
-                return m_tensor_view_type;
-            }
+            const Tensor& get_tensor() const;
+            Tensor& get_tensor();
 
-            virtual void set_tensor_view_type(const element::Type& element_type,
-                                              const Shape& shape);
+            const std::string& get_name() const { return m_name; }
+            void set_tensor_view_type(const element::Type& element_type, const Shape& shape);
```
aprocter: Should we break it up into `set_element_type` and `set_shape`?

src/ngraph/function.cpp
```diff
@@ -70,10 +70,18 @@ Function::Function(const std::shared_ptr<Node>& result,
                    const std::string& name)
     : Function(NodeVector{result}, parameters, name)
 {
+    validate_nodes_and_infer_types();
```
aprocter: Is the duplication between here and `init()` just to cover the "before transition/after transition" cases?

src/ngraph/op/op.cpp
```diff
@@ -20,12 +20,14 @@
...
+#include "ngraph/type/element_type.hpp"
 
 using namespace ngraph;
 using namespace std;
 
 op::Op::Op(const std::string& node_type, const NodeVector& args)
     : Node(node_type, args)
 {
+    // Will replace with real type
+    set_output_type(0, element::Type{}, Shape());
```
aprocter: Not sure why we need this here.

src/ngraph/runtime/interpreter/int_backend.cpp
aprocter: Agreed. Any comment on why we need this?

src/ngraph/shape.hpp
```diff
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <cstdio>
+#include <iostream>
```
aprocter: fun fun fun

src/ngraph/runtime/interpreter/int_backend.cpp
diyessi: We need to convert the constant and broadcast that take their shape from another op back into regular constant and broadcast.

src/ngraph/autodiff/adjoints.cpp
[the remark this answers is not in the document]
diyessi: The semantics of GOE are vague. I don't want a GOE in there for a single-output node. Do I need a GOE for output 0 of a multi-output node?

src/ngraph/descriptor/tensor.hpp
```diff
@@ -33,22 +35,19 @@ namespace ngraph
...
         class TensorView;
-        class PrimaryTensorView;
         class Tensor;
     }
 }
 
 class ngraph::descriptor::Tensor
 {
-    friend class PrimaryTensorView;
+    friend class TensorView;
```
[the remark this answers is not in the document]
diyessi: Once TensorView and Tensor merge, they won't be friends.

src/ngraph/descriptor/tensor_view.hpp
diyessi: So far, I have never seen a case where one was set and the other was not.

src/ngraph/function.cpp
diyessi: Comment added; it's a transient part of achieving the transition; it doesn't do anything while we are doing validate in construction.

src/ngraph/node.cpp
diyessi: We will need some Shape changes. For now we are okay.

src/ngraph/descriptor/tensor_view.hpp
aprocter: Sure, because before `set_tensor_view_type` was the only way.  ;)

changes.md
```diff
@@ -1,10 +1,14 @@
 # API Changes
 
+## Passes
+* `LikeReplacement` pass must be run by all transformers.
```
aprocter: I think it'd be better to have reference implementations for `BroadcastLike` and `ScalarConstantLike`, and that other backends run the elimination pass if they want to. `INTERPRETER` should run on a structurally unmodified graph.

src/ngraph/op/constant.hpp
```diff
@@ -240,8 +255,58 @@ namespace ngraph
...
+        protected:
+            ScalarConstantLikeBase(const std::string& name, const NodeVector& args)
+                : Constant(name, args)
+            {
+            }
+        };
+
+        /// \brief A scalar constant whose element type is the same as like.
+        template <typename T>
+        class ScalarConstantLike : public ScalarConstantLikeBase
```
aprocter: I'm kind of leery of a templated op class here. We've managed to avoid it for `Constant`.

changes.md
diyessi: These are not really ops. For example, the op ordering for the Like argument is only needed for shape inference, not execution order, which is why it goes away afterwards. They could be derived from Node directly once we're no longer calling things in the constructors.

test/type_prop.cpp
```diff
@@ -247,77 +227,6 @@ TEST(type_prop, concat_deduce)
...
-    {
-        EXPECT_EQ(error.what(), std::string("Setting value type to a different ValueType"));
-    }
-    catch (...)
-    {
-        FAIL() << "Deduced type check failed for unexpected reason";
-    }
-}
-
-TEST(type_prop, concat_deduce_wrong_rank)
```
aprocter: Did you mean to delete `concat_deduce_wrong_rank` and `concat_deduce_wrong_shape`? I think we should keep them.

src/ngraph/op/batch_norm.hpp
diyessi: There's no guarantee the backprop args are in the same order as the fprop.

src/ngraph/op/constant.hpp
diyessi: The problem is that we don't know the element type until later, so we need a way to hold onto the original value until we know the element type.

src/ngraph/op/convolution.cpp
```diff
@@ -241,17 +241,45 @@ op::Convolution::Convolution(const shared_ptr<Node>& data_batch,
...
+}
+
+void op::Convolution::validate_and_infer_types()
+{
+    auto& data_batch_shape = get_input_shape(0);
+    auto& data_batch_et = get_input_element_type(0);
+    auto& filters_shape = get_input_shape(1);
+    auto& filters_et = get_input_element_type(1);
+
+    if (m_data_dilation_strides.size() == 0)
```
Krovatkin: these are pre-real-validation attributes?

src/ngraph/op/max_pool.cpp
```diff
@@ -30,13 +30,44 @@ op::MaxPool::MaxPool(const shared_ptr<Node>& arg,
...
     , m_window_movement_strides(window_movement_strides)
     , m_padding_below(padding_below)
     , m_padding_above(padding_above)
 {
+    constructor_validate_and_infer_types();
+}
+
+void op::MaxPool::validate_and_infer_types()
+{
+    if (m_inputs.size() != 1)
```
Krovatkin: `check_single_output_args` should be enough, no?

src/ngraph/op/result.cpp
```diff
@@ -17,24 +17,29 @@
...
-    : RequiresTensorViewArgs("Result", {arg})
+    : Op("Result", check_single_output_args({arg}))
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::Result::validate_and_infer_types()
 {
-    if (arg->get_outputs().size() != 1)
+    if (get_input_size() != 1)
```
Krovatkin: isn't this covered by `check_single_output_args` ?

PR #1469, opened 2018-08-23
[ONNX] CentOS disable unit tests

apwojcik:
We have issues with Protobuf in CentOS. This PR temporarily disables unit testing of ONNX for CentOS until we resolve the issues. Unit tests for Ubuntu are not affected.


CMakeLists.txt
```diff
@@ -101,6 +101,10 @@ endif()
 # Installation logic...
 #-----------------------------------------------------------------------------------------------
 
+if (NOT APPLE AND NOT WIN32)
```
diyessi: Please select which systems you want, not which you don't want.

test/CMakeLists.txt
```diff
@@ -47,7 +47,15 @@ set(SRC
 )
 
 if (NGRAPH_ONNX_IMPORT_ENABLE)
-    list(APPEND SRC onnx_import.cpp)
+    if (APPLE OR WIN32)
+        list(APPEND SRC onnx_import.cpp)
+    else()
```
diyessi: Change else to if linux.

apwojcik: There is no `if(LINUX)` in CMake. Only `if(UNIX)`, but is set for both Linux and Apple.

CMakeLists.txt
apwojcik: There is no `if(LINUX)` in CMake. Only `if(UNIX)`, but is set for both Linux and Apple.

diyessi: No wonder no one used CMAKE back in the old days.
`if (unix and not apple)` isn't any better.

PR #1474, opened 2018-08-23
cmake: use GNUInstallDirs to autodetect installation directories (Linux only)

apwojcik:
This PR satisfies requirements regarding installation directories from GNU Coding Standards (https://www.gnu.org/prep/standards/html_node/Directory-Variables.html).


apwojcik:
The change allows address correctly installation to directories which names depend on Linux distribution - i.e. `/usr/lib/` vs `/usr/lib64/` for CentOS, SLES, and releated; and `/usr/lib/x86_64-linux-gnu/` for Ubuntu/Debian.

CMakeLists.txt
```diff
@@ -101,16 +101,25 @@ endif()
 # Installation logic...
 #-----------------------------------------------------------------------------------------------
 
+if (NOT APPLE AND NOT WIN32)
```
rkimballn1: If this is for linux only then specify `if(LINUX)` instead of `NOT OTHER STUFF`?

apwojcik: There is no such definition as `LINUX`. The definition `UNIX` does exist. However, it is set for both Linux and for `APPLE`, too. Either `if (NOT APPLE AND NOT WIN32)` or `if(UNIX AND NOT APPLE)`.

rkimballn1: How about we put at the top level cmake file
```
if (UNIX AND NOT APPLE)
    set(LINUX TRUE)
endif()
```
Then we can do the `if (LINUX)`

PR #1475, opened 2018-08-23
ignore CLion's project files and directories

apwojcik:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,7 @@ python/share/*
 
 python/pybind11/
 
+# Clion project files and directories
+\.idea
+build\.*
+cmake-*


.gitignore
```diff
@@ -111,3 +111,7 @@ python/share/*
 
 python/pybind11/
 
+# Clion project files and directories
+\.idea
+build\.*
```
rkimballn1: We have directories call build. Put these in your global .gitignore file, not in ngraph. I can help you setup global gitignore.

apwojcik: This is "build.*" directory. Anyway, I will add to the global gitignore.

PR #1477, opened 2018-08-23
Revert MKLDNN ver to 0.14

Krovatkin:


cmake/external_mkldnn.cmake
```diff
@@ -166,7 +165,6 @@ add_custom_command(TARGET ext_mkldnn POST_BUILD
 )
 
 add_library(libmkldnn INTERFACE)
-add_dependencies(libmkldnn ext_mkldnn)
```
jbobba: why are we removing this?

PR #1478, opened 2018-08-23
make CPU backend auto-register

rkimballn1:


avijit-nervana:
OK - looks like the codegen is not installed at all! Given the combination of features and flags - there should be unit tests to catch this.

rkimballn1:
There was an issue with codegen where it would not be installed even if it was depended on by some other library. This was fixed in PR #1482 
If nothing in the build depends on codegen then it is not built nor installed. This is the intended behavior.

rkimballn1:
I just checked and after merging master into this PR codegen is built and installed if it is needed and not built or installed if it is not needed. This is the correct behavior.

PR #1484, opened 2018-08-24
Add support for Quantize op via mkldnn for IA backend (codegen + DEX)

nbpatel:
This PR(and the following ones) is op-wise breakdown of #1301. Additionally this PR also adds DEX support.


src/ngraph/runtime/cpu/op/quantize.cpp
```diff
@@ -0,0 +1,40 @@
+/*******************************************************************************
+* Copyright 2017-2018 Intel Corporation
```
jbobba: Use copyright 2018 in all the new files

src/ngraph/runtime/cpu/op/quantize.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+ngraph::op::Quantize::Quantize(std::shared_ptr<Node> input,
+                               std::shared_ptr<Node> min,
+                               std::shared_ptr<Node> max,
+                               const element::Type& type)
+    : RequiresTensorViewArgs("Quantize", {input, min, max})
+    , m_element_type(type)
+{
+    add_output(type, input->get_shape());
+    add_output(element::f32, Shape{1});
```
jbobba: Shouldn't this be element_type of input?

src/ngraph/runtime/cpu/quantization_util.hpp
```diff
@@ -0,0 +1,82 @@
...
+                    auto quantize = static_cast<const ngraph::op::Quantize*>(node);
+                    auto min_const_op =
+                        std::dynamic_pointer_cast<ngraph::op::Constant>(quantize->get_argument(1));
+                    auto max_const_op =
+                        std::dynamic_pointer_cast<ngraph::op::Constant>(quantize->get_argument(2));
+                    float input_min_range =
+                        *(static_cast<float const*>(min_const_op->get_data_ptr()));
+                    float input_max_range =
+                        *(static_cast<float const*>(max_const_op->get_data_ptr()));
+                    // begin code copied and pasted (and modified) from mkl_quantize_op.cc
```
jbobba: This is the original source: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/quantize_op.cc

src/ngraph/runtime/cpu/op/quantize.cpp
nbpatel: Its not that generic at the moment. It'l only do float --> some xyz datatype.

src/ngraph/runtime/cpu/op/quantize.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+ngraph::op::Quantize::Quantize(std::shared_ptr<Node> input,
+                               std::shared_ptr<Node> min,
+                               std::shared_ptr<Node> max,
+                               const element::Type& type)
+    : RequiresTensorViewArgs("Quantize", {input, min, max})
+    , m_element_type(type)
+{
+    add_output(type, input->get_shape());
+    add_output(input->get_element_type(), Shape{1});
```
diyessi: This is a scalar, right? Should be `Shape{}`.
Also, `set_output_type(1, get_input_element_type(0), Shape{})` will save one of us some conflict resolution.

src/ngraph/runtime/cpu/op/quantize.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "ngraph/runtime/cpu/op/quantize.hpp"
+
+ngraph::op::Quantize::Quantize(std::shared_ptr<Node> input,
+                               std::shared_ptr<Node> min,
+                               std::shared_ptr<Node> max,
+                               const element::Type& type)
+    : RequiresTensorViewArgs("Quantize", {input, min, max})
+    , m_element_type(type)
+{
```
aprocter: If this goes into the core opset we'll need to add a check for the shapes/ETs of `min` and `max`. We'll also need to decide what element types we want to allow, both for the unquantized input (TF just allows float32) and the quantized output (TF has several allowed types).

For experimental purposes in the meantime, I'd be okay with targeting this directly (keeping things under `runtime/cpu`), and putting some safety logic in the bridge.

test/quantize_cpu.cpp
```diff
@@ -0,0 +1,113 @@
...
+                                   op::ParameterVector{A});
+    auto backend = runtime::Backend::create("CPU");
+    // Create some tensors for input/output
+    auto a = backend->create_tensor(element::f32, shape_a);
+    copy_data(a, a_data);
+    auto result = backend->create_tensor(element::u8, shape_a);
+    auto result_min = backend->create_tensor(element::f32, Shape{});
+    auto result_max = backend->create_tensor(element::f32, Shape{});
+    backend->call(f, {result, result_min, result_max}, {a});
+    EXPECT_EQ((vector<uint8_t>{0, 0, 6, 30, 45}), read_vector<uint8_t>(result));
```
aprocter: Question only: it looks like when quantizing to unsigned, we set the output range to [0,max(abs(input))] but then truncate negative input values to 0 before actually quantizing?

src/ngraph/runtime/cpu/op/quantize.cpp
nbpatel: Yes agreed, the ops dont have proper checks for ETs/shapes,  because my understanding was to have some temporary cpu specific ops until core has these ops, just so that i can break dependency from core and test my code. I am not sure what the plan is now. Maybe @jbobba can answer better if we want the bridge to call into cpu ops directly or the core can use these ops as a prototype and improve on it.

test/quantize_cpu.cpp
nbpatel: Yes, negative values should map to 0

src/ngraph/runtime/cpu/op/quantize.cpp
nbpatel: I dont see anything like 'set_output_type' in ngraph on my branch. Is it recently added? Or do you mean I should create one?

diyessi: I guess it's just in my PR then, so I'll deal with it since you'll be merged long before me.

nbpatel:
TESTNOW

nbpatel:
TESTNOW

nbpatel:
TESTNOW

Krovatkin:
@nbpatel a few things I forgot to mention:

* I wonder if we could refactor the logic common across dex and cpu_codegen into `mkldnn_emitter::build_quantization`?
* Do we want to add this to `cpu_layout` so we propagate the layouts or it doesn't make too much sense for quantization?

src/ngraph/runtime/cpu/quantization_util.hpp
```diff
@@ -0,0 +1,85 @@
...
+                    auto quantize = static_cast<const ngraph::op::Quantize*>(node);
+                    auto min_const_op =
+                        std::dynamic_pointer_cast<ngraph::op::Constant>(quantize->get_argument(1));
+                    auto max_const_op =
+                        std::dynamic_pointer_cast<ngraph::op::Constant>(quantize->get_argument(2));
+                    float input_min_range =
+                        *(static_cast<float const*>(min_const_op->get_data_ptr()));
+                    float input_max_range =
+                        *(static_cast<float const*>(max_const_op->get_data_ptr()));
+                    // begin code copied and pasted from
```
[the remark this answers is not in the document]
kimjanik: Um, you can't do that.  Remember those developer attestations you did?

PR #1487, opened 2018-08-24
[ONNXIFI] public interface stubs

apwojcik:


src/ngraph/frontend/onnxifi/onnxifi.cpp
```diff
@@ -0,0 +1,109 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <cstdint>
+#include <onnxifi.h>
+
+extern "C" {
+
+onnxStatus onnxGetBackentIDs(onnxBackendID* backendIDs, std::size_t* numBackends)
```
postrational: Typo in `onnxGetBackentIDs` -> `onnxGetBackendIDs`

apwojcik: Done

src/ngraph/frontend/onnxifi/onnxifi.cpp
```diff
@@ -0,0 +1,109 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include <cstdint>
+#include <onnxifi.h>
+
+extern "C" {
+
+onnxStatus onnxGetBackendIDs(onnxBackendID* backendIDs, std::size_t* numBackends)
```
rkimballn1: Function names are snake_case

postrational: @rkimballn1 We are implementing the ONNXIFI interface, which forces us to use these camelCase names. All names listed here:
https://github.com/onnx/onnx/blob/master/onnx/onnxifi.h#L932

postrational: @apwojcik Shouldn't we be appending `NGRAPH` to all the function names, so this should be `onnxGetBackendIDsNGRAPH`?

apwojcik: No, the documentation is not up to date. Please take a look at loader implementation
https://github.com/onnx/onnx/blob/master/onnx/onnxifi_loader.c

PR #1491, opened 2018-08-25
nvgpu reduce to scalar optimization 

fengleitian:
Sum     2,617us  ->    1,495us
Faster reduce to scalar than CUDNN version. So we don't need CUDNN for reduction now.
- it's about the same performance as CUDNN version for data size around 65536,

- 2-3x faster for small or large data size.


src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -260,6 +261,129 @@ void runtime::gpu::CudaKernelBuilder::get_reduce_op(codegen::CodeWriter& writer,
...
+        writer.block_begin();
+        {
+            writer << "r = " << reduce_op << "(r , in[in_idx]);\n";
+            writer << "in_idx += step;\n";
+        }
+        writer.block_end();
+        writer << "sdata[tid] = r;\n";
+        writer << "__syncthreads();\n";
+
+        if (block_size_x > 512)
```
rkimballn1: ALL code blocks get curly braces. That includes all of the single-line blocks in this file. Sorry :(

fengleitian: fixed.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1422,12 +1421,237 @@ size_t runtime::gpu::CUDAEmitter::build_reduce(const std::vector<std::string>& d
...
+                                      std::vector<void*>{outputs[0]}.data());
+            }});
+        primitive_index = this->m_primitive_emitter->insert(std::move(reduce));
+    }
+    else
+    {
+        uint32_t nthreads = static_cast<uint32_t>(shape_size(input_shape));
+        //if the data size is large, call reduce_to_scalar_acc first and then reduce_to_scalar.
+        //other wise, call reduce to scalar directly.
+        if (nthreads > nthreads_acc * 9)
```
csullivan: Why 9? Seems arbitrary. Can we estimate an upper bound from a device query on available shared memory or other resources? Likely a job for tuning, but even better if it is deterministic based on hardware.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -260,6 +261,134 @@ void runtime::gpu::CudaKernelBuilder::get_reduce_op(codegen::CodeWriter& writer,
...
+        writer << "sdata[tid] = 0;\n";
+        writer << "uint32_t in_idx = tid;\n";
+        writer << data_types[1] << " r = 0;\n";
+        writer << "if(in_idx < nthreads)\n";
+        writer.block_begin();
+        writer << "r = in[in_idx];\n";
+        writer << "in_idx += step;\n";
+        writer.block_end();
+        //accumulate reduction to blockDim.x threads
+        writer << "while(in_idx + (step * 7) < nthreads)\n";
```
csullivan: Can this 7 be emitted from the builder, rather than hardcoded in the emitter? If not, let's make it a #define in the kernel so it's more obvious what it is for, and that it will need to be changed if tuning is added that requires an update to the kernel.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -260,6 +261,134 @@ void runtime::gpu::CudaKernelBuilder::get_reduce_op(codegen::CodeWriter& writer,
...
+    const std::string& reduce_op,
+    uint32_t block_size_x)
+{
+    writer << "extern \"C\" __global__ void cuda_" << name << args.get_input_signature();
+    writer.block_begin();
+    {
+        writer << "extern __shared__ " << data_types[1] << " sdata[1024];\n";
+        writer << "uint32_t tid = threadIdx.x; \n";
+        writer << "uint32_t step = blockDim.x; \n";
+        writer << "sdata[tid] = 0;\n";
```
csullivan: Note that not all elements of sdata are zeroed. I _think_ you have it covered with your conditionals below, but I wasn't exhaustive in my traces.

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: not arbitrary, it's first read + one unrolled loop of 8.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
fengleitian: All used `sdata` already covered.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1422,12 +1421,237 @@ size_t runtime::gpu::CUDAEmitter::build_reduce(const std::vector<std::string>& d
...
+    }
+
+    std::unique_ptr<gpu::primitive> reduce(
+        new gpu::primitive{[=](void** inputs, void** outputs) mutable {
+            void** args_list = args.resolve_placeholder(0, &inputs[0])
+                                   .resolve_placeholder(1, &outputs[0])
+                                   .get_argument_list();
+
+            CUDA_SAFE_CALL(cuLaunchKernel(
+                *compiled_kernel.get(), 1, 1, 1, block_size_x, 1, 1, 0, NULL, args_list, 0));
```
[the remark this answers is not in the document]
fengleitian: good point.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
fengleitian: changed to host code.

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: changed unroll size to host code.

src/ngraph/runtime/gpu/cuda_emitter.cpp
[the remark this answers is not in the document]
fengleitian: changed.

test/backend_test.in.cpp
```diff
@@ -3485,6 +3487,32 @@ NGRAPH_TEST(${BACKEND_NAME}, sum_to_scalar)
...
+    auto f = make_shared<Function>(make_shared<op::Sum>(A, AxisSet{0}), op::ParameterVector{A});
+
+    auto backend = runtime::Backend::create("${BACKEND_NAME}");
+
+    // Create some tensors for input/output
+    vector<float> v_a(1000000, 0);
+    double r = 0;
+    for (int i = 0; i < 1000000; i++)
+    {
+        v_a[i] = static_cast<float>(random_generator() % 255);
```
rkimballn1: You need to seed your random number generator so the same values are used every time it run.

src/ngraph/runtime/gpu/cuda_emitter.cpp
csullivan: nit:It's my feeling that this 9 is dependent on the unroll size used in the kernel, and so it should be calculated from the unroll_size which is passed into the kernel builder as an argument - because if the unroll size changes, so should this.

fengleitian: you are right.

PR #1497, opened 2018-08-27
IntelGPU backend: AvgPoolBackprop operation

aalbersk:
- add AvgPoolBackprop operation
- add gen_window_loop() to generate window loop and input condition


src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -427,6 +446,159 @@ void runtime::intelgpu::do_max_pool_backprop_operation(cldnn::topology& topology
...
+                                                       const Shape& win_shape,
+                                                       const Shape& win_stride,
+                                                       const Shape& pad_below,
+                                                       const bool include_padding)
+{
+    const string entry_point_name = "op_avg_pool_backprop_" + output_name;
+    codegen::CodeWriter writer;
+    vector<size_t> gws;
+
+    const Shape delta_data(delta_shape.cbegin() + 2, delta_shape.cend());
```
shssf: Could you please check the size of the Shape before this construction?
Please use `static Shape get_channel_shape(const Shape& shape, const string& function_name)` as an example from `intelgpu_batch_norm.cpp`. Let's make this function a bit more general by adding `size_t min_size` parameter for the Shape limit. 
Also, please add these checks to `do_max_pool_backprop_operation` - I forgot to add them previously.

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -427,6 +446,159 @@ void runtime::intelgpu::do_max_pool_backprop_operation(cldnn::topology& topology
...
+                    ++var_idx;
+                }
+
+                writer << "output[i0][i1]";
+                // Additional dimentions for output
+                for (size_t i = 0; i < output_data.size(); ++i)
+                {
+                    writer << "[j" << i << "]";
+                }
+                writer << " = 0.0;\n";
```
shssf: = 0.0f;

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -252,6 +252,52 @@ void runtime::intelgpu::do_pad_operation(cldnn::topology& topology,
...
+
+        writer << "if (";
+        // Generate input coordinate condition
+        for (size_t i = 0; i < win_shape.size(); ++i)
+        {
+            if (i)
+            {
+                writer << " && ";
+            }
+            writer << "(win_idx" << i << " >= 0) && (win_idx" << i << " < "
```
shssf:
```
                            if ((win_idx0 >= 0) && (win_idx0 < 4) && (win_idx1 >= 0) && (win_idx1 < 4))
                            {
                                computed_val = input0[i0][i1][i2][i3]/win_elems_size;
                                output[i0][i1][win_idx0][win_idx1] += computed_val;
                            }
```
We don't need to check `(win_idx0 >= 0)` because we use `uint` and the value never be negative.

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -427,6 +446,159 @@ void runtime::intelgpu::do_max_pool_backprop_operation(cldnn::topology& topology
...
+                    // Closing brackets for window shape loop
+                    gen_window_loop(writer, output_shape, win_shape, win_stride, pad_below, false);
+                }
+
+                // Loop over window shape
+                // input coordinate condition
+                gen_window_loop(writer, output_shape, win_shape, win_stride, pad_below, true);
+
+                writer << "computed_val = input0" << access_dims(delta_shape)
+                       << "/win_elems_size;\n";
```
shssf: Spaces. `"/win_elems_size;\n"` -> `" / win_elems_size;\n"`

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -427,6 +446,159 @@ void runtime::intelgpu::do_max_pool_backprop_operation(cldnn::topology& topology
...
+        {
+            is_zero_pad = false;
+            break;
+        }
+    }
+
+    size_t win_elems_size = 1;
+    if (include_padding || is_zero_pad)
+    {
+        for (auto const& i : win_shape)
```
shssf: Perhaps it is better to use `shape_size<Shape>(input0_shape)` as it done in `do_1d_scalar_mul`

shssf:
Also, I suspect we have additional issues:
- The variable `win_elems_size` will continuously grow if padding exists by extra calculation loop. I'm not sure is it requires by the algorithm or not.
- In nGraph test system I see `const bool include_padding` is always zero. Perhaps it is a bug in nGraph or I missed something. If it is a bug in nGraph, please don't use this variable if it possible.

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
aalbersk: Shape of delta is checked in the definition of the operation, so is it necessary? (https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/avg_pool.cpp#L235 )

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -427,6 +445,148 @@ void runtime::intelgpu::do_max_pool_backprop_operation(cldnn::topology& topology
...
+    codegen::CodeWriter writer;
+    vector<size_t> gws;
+
+    const Shape delta_data(delta_shape.cbegin() + 2, delta_shape.cend());
+    const Shape output_data(output_shape.cbegin() + 2, output_shape.cend());
+
+    size_t win_elems_size = 1;
+    if (include_padding || shape_size<Shape>(pad_below) == 0)
+    {
+        for (auto const& i : win_shape)
```
shssf: `win_elems_size = shape_size<Shape>(win_shape);` insted of the loop?

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.hpp
```diff
@@ -142,6 +152,12 @@ namespace ngraph
                               const std::vector<Shape>& input_shapes,
                               const std::string& output_type,
                               const Shape& output_shape);
+            void gen_window_loop(codegen::CodeWriter& writer,
```
shssf: Forgot to mention it before. Do we really need this function exposed externally? I think it can be static?

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.hpp
```diff
@@ -53,6 +53,16 @@ namespace ngraph
                                                 const Shape& win_shape,
                                                 const Shape& win_stride,
                                                 const Shape& pad_below);
+            void do_avg_pool_backprop_operation(cldnn::topology& topology,
```
shssf: empty line before function might be more readable.

PR #1498, opened 2018-08-27
Shape implemented not as derived from vector<size_t>

rkimballn1:
Ultimately it would be nice to replace all types that are just specializations of vector<size_t> with real classes, giving us better control.

Shape is not a general purpose container and it is not intended to be. The only reason it has as many methods as it does now is to be backwards compatible with the previous implementations where it was simply a `vector<size_t>`. In the future we should move towards making it look and act like a proper Shape class and not an STL container.

Shape was originally simply a `using Shape = vector<size_t>`. This was no end of trouble because Stride was defined the same way and they could and were incorrectly used in multiple instances in the code. There is no type safety.

The next step was to change Shape to `class Shape : public vector<size_t>` which provided much better type safety than the `using Shape` method, uncovering multiple bugs/misuses in the code.

Some of the new things we will need to address are:
1) Placeholders in Shape. Values for all dimensions may not be known when we are constructing the Shape. As an example users can define models where the batch size is not defined until runtime. We need a way to have placeholder values in Shape, something better than `-1 == placeholder`.
2) We need a way to have an empty Shape. An empty `vector<size_t>` is a scalar and there is no sane way with `vector<size_t>` to define an empty Shape.


apwojcik:
Why not just `using Shape = std::vector<size_t>;` or `typedef std::vector<std::size_t> Shape_t;`. Do you plan to extend the Shape class with additional data members and function methods?

diyessi:
We will be extending shape to allow for unspecified lengths. Having a real type also prevents non-shape vectors from accidentally being passed as shapes, as has happened when args were passed in the wrong order.

src/ngraph/shape.hpp
```diff
@@ -25,46 +25,80 @@
 namespace ngraph
 {
     /// \brief Shape for a tensor.
-    class Shape : public std::vector<size_t>
+    class Shape
```
apwojcik: This class is missing traits. Standard `algorithms` use traits to operate on containers.
```
using value_type = std::size_t;
using allocator_type = typename std::vector<std::size_t>::allocator_type;
using size_type = std::size_t;
using difference_type = std::ptrdiff_t
using reference  = value_type&;
using const_reference = const value_type&;
using pointer = std::allocator_traits<allocator_type>::pointer;
using const_pointer = std::allocator_traits<allocator_type>::const_pointer;
using iterator = typename std::vector<value_type>::iterator;
using const_iterator = const iterator;
using reverse_iterator = std::reverse_iterator<iterator>;
using const_reverse_iterator = std::reverse_iterator<const_iterator>;
```

apwojcik:
Ok, this makes sense.

src/ngraph/shape.hpp
```diff
@@ -25,46 +25,80 @@
...
+        {
+            m_data.insert(pos, first, last);
+        }
+        void push_back(size_t value) { m_data.push_back(value); }
+        std::vector<size_t>::iterator erase(std::vector<size_t>::iterator pos)
+        {
+            return m_data.erase(pos);
         }
+        bool empty() const { return m_data.empty(); }
+        const std::vector<size_t>& get_value() const { return m_data; }
```
apwojcik: in case we want to be standard library compliant the method name should be `data()`.

src/ngraph/shape.hpp
```diff
@@ -25,46 +25,80 @@
...
-            return *this;
+            return m_data.insert(pos, value);
+        }
+        void insert(std::vector<size_t>::iterator pos,
+                    std::vector<size_t>::iterator first,
+                    std::vector<size_t>::iterator last)
+        {
+            m_data.insert(pos, first, last);
+        }
+        void push_back(size_t value) { m_data.push_back(value); }
```
apwojcik: Missing `emplace_back`. ONNX importer uses both `push_back` and `emplace_back`.
```
template <typename... Args>
void emplace_back(Args&&... args) {
    m_data.emplace_back(std::forward<Args>(args)...);
}
```

aprocter:
I like the general idea of switching to "contains-a" instead of "is-a", and the change seems less disruptive than I would have expected within `ngraph-tf` proper. We should see how bad it is for the bridges.

src/ngraph/shape.hpp
rkimballn1: This is not intended to be an stl container, just a Shape. If we don't need something then why go to the effort of implementing and testing it?

src/ngraph/shape.hpp
rkimballn1: Yes, I will fix onnx as there is no reason to use emplace_back for this

rkimballn1:
@avijit-nervana This PR is intended to be a discussion on this. It will merge to master if and when it is ready and we all agree and not before. Your best bet is to click `Request Changes` now to prevent it from merging until you approve.

src/ngraph/shape.hpp
```diff
@@ -25,46 +25,80 @@
...
-            static_cast<std::vector<size_t>*>(this)->operator=(v);
+            m_data = v.m_data;
             return *this;
         }
-        Shape& operator=(Shape&& v)
+
+        size_t size() const { return m_data.size(); }
+        operator std::vector<size_t>() const { return m_data; }
+        bool operator==(const Shape& other) const { return m_data == other.m_data; }
+        bool operator!=(const Shape& other) const { return m_data != other.m_data; }
```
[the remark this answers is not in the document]
rkimballn1: The goal should be readability and no redundant code. If your comparison operator is complex then I would agree with you, but in this case the comparison is trivial and the implementation I have is more straightforward and clearer then what you suggest.

src/ngraph/shape.hpp
rkimballn1: Again, this is not a stl container, it is a custom class Shape. Just because it used to be an stl container does not mean that we should continue to be one. The purpose of this PR is to make Shape NOT an stl container.

src/ngraph/shape.hpp
```diff
@@ -25,46 +25,80 @@
...
+        using value_type = size_t;
         Shape(const std::initializer_list<size_t>& axis_lengths)
-            : std::vector<size_t>(axis_lengths)
+            : m_data(axis_lengths)
         {
         }
 
         Shape(const std::vector<size_t>& axis_lengths)
-            : std::vector<size_t>(axis_lengths)
+            : m_data(axis_lengths)
```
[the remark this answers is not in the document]
rkimballn1: In ngraph we have no rule one way or the other for this. Until we have a rule then anything goes. This works.

rkimballn1:
obsolete

PR #1500, opened 2018-08-27
address klocwork warning

rkimballn1:
conditional with constant == bad according to klocwork


test/ref_generators/generate_convolution_ref.py
```diff
@@ -268,50 +266,52 @@ def emit_test(t,f):
                          data_str(output_batch_data),
                          bprop));
 
+bprop_check_result = "    EXPECT_TRUE(autodiff_numeric_compare<float>(backend, make_graph, {a, b}, .01f, .01f));"
```
aprocter: I think the rationale for the `if(true)`/`if(false)` in the first place was to make it easy to enable these checks by hand if one was doing some deep debugging. Could we instead include this line, but commented out, when `bprop?` (which should be changed to a proper boolean) is `False`?

PR #1501, opened 2018-08-27
Add support for Dequantize op via mkldnn for IA backend (codegen + DEX)

nbpatel:
Base branch: #1484.
Dequantize from fp32 --> i8/u8

### DIFF
--- a/src/ngraph/runtime/cpu/builder/dequantize.cpp
+++ b/src/ngraph/runtime/cpu/builder/dequantize.cpp
@@ -1,0 +1,85 @@
+/*******************************************************************************
+* Copyright 2018 Intel Corporation
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "ngraph/runtime/cpu/op/dequantize.hpp"
+#include <vector>
+#include "ngraph/op/constant.hpp"
+#include "ngraph/runtime/cpu/cpu_builder.hpp"
+#include "ngraph/runtime/cpu/mkldnn_invoke.hpp"
+#include "ngraph/runtime/cpu/mkldnn_utils.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace cpu
+        {
+            template <>
+            void Builder::BUILDER_DECL(ngraph::op::Dequantize)
+            {
+                if (runtime::cpu::mkldnn_utils::use_mkldnn_kernel(node))
+                {
+                    auto dequantize = static_cast<const ngraph::op::Dequantize*>(node);
+                    auto& functors = external_function->get_functors();
+
+                    auto& arg_tensor = external_function->get_tensor_data(args[0].get_name());
+                    auto& out_tensor = external_function->get_tensor_data(out[0].get_name());
+
+                    auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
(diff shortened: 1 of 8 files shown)


src/ngraph/runtime/cpu/op/dequantize.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "ngraph/op/constant.hpp"
+
+ngraph::op::Dequantize::Dequantize(std::shared_ptr<Node> input,
+                                   std::shared_ptr<Node> min,
+                                   std::shared_ptr<Node> max,
+                                   const element::Type& type)
+    : RequiresTensorViewArgs("Dequantize", {input, min, max})
+    , m_element_type(type)
+{
+    add_output(element::f32, input->get_shape());
```
aprocter: As with #1484, if we elevate this to core we'll need to check the element types and sizes of `min`/`max`, check the element type of `input` to make sure it is a valid type for a quantized, and make sure that `type` is supported.

test/quantize_cpu.cpp
```diff
@@ -111,3 +112,43 @@ TEST(quantize_cpu, quantize_to_int8)
     EXPECT_EQ((vector<float>{-127}), read_vector<float>(result_min));
     EXPECT_EQ((vector<float>{127}), read_vector<float>(result_max));
 }
+
+template <typename T>
+void DequantizeTest(
+    int input, float min, float max, float expected_output, const element::Type& type)
```
aprocter: You could use `element::Type::from<T>()` instead of passing `type` here.

src/ngraph/runtime/cpu/builder/dequantize.cpp
```diff
@@ -0,0 +1,85 @@
...
+                    auto& functors = external_function->get_functors();
+
+                    auto& arg_tensor = external_function->get_tensor_data(args[0].get_name());
+                    auto& out_tensor = external_function->get_tensor_data(out[0].get_name());
+
+                    auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
+                    auto input_desc = mkldnn_utils::get_input_mkldnn_md(node, 0);
+                    auto result_desc = mkldnn_utils::get_output_mkldnn_md(node, 0);
+
+                    auto min_const_op = std::dynamic_pointer_cast<ngraph::op::Constant>(
```
Krovatkin: I've just noticed but this logic that pre-computes values for reorder seems to be similar across `cpu_codegen` and `dex`. Maybe we should refactor it into `mkldnn_emitter::build_dequantization`?

src/ngraph/runtime/cpu/op/dequantize.cpp
Krovatkin: we at least should check that `min` and `max` are the same type and they are both scalars and maybe that the output type is f32/f64 and the inputs are integral types?

src/ngraph/runtime/cpu/builder/dequantize.cpp
jbobba: Why not static_pointer_cast?

src/ngraph/runtime/cpu/op/dequantize.cpp
jbobba: Also check that min and max are constants.

jbobba: Shouldn't we be using type instead of element::f32?

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -2740,6 +2741,49 @@ namespace ngraph
...
+            template <>
+            void CPU_Emitter::EMITTER_DECL(ngraph::op::Dequantize)
+            {
+                auto dequantize = static_cast<const ngraph::op::Dequantize*>(node);
+                auto arg0_shape = args[0].get_shape();
+                auto result_shape = out[0].get_shape();
+                if (runtime::cpu::mkldnn_utils::use_mkldnn_kernel(node))
+                {
+                    auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
+                    auto input_data_desc = mkldnn_utils::get_input_mkldnn_md(node, 0);
```
jbobba: need to support this op in cpu_layout

nbpatel:
Have addressed all the feedback in #1565. Closing this one since it is based of quantization op branch, which needs some licensing stuff to be resolved. Sorry for the confusion.

PR #1511, opened 2018-08-28
Extend concat elimination to fold slice + reshape

Krovatkin:
Let's try this again.


csullivan:
Please review PR #1532 which points to this branch. It adds support to the pass to handle fusion of slice + reshape + concat when the net result of the fusion requires an axis reordering.

test/algebraic_simplification.cpp
```diff
@@ -425,6 +425,26 @@ TEST(algebraic_simplification, concat_parameter_slice)
     pass_manager.register_pass<pass::AlgebraicSimplification>();
     pass_manager.register_pass<pass::VisualizeTree>("after.pdf");
 
+    auto f = std::make_shared<Function>(ngraph::NodeVector{concat}, op::ParameterVector{a});
+    pass_manager.run_passes(f);
+    ASSERT_EQ(f->get_results().at(0)->get_argument(0), a);
+}
+
+TEST(algebraic_simplification, concat_parameter_slices_reversed)
```
csullivan: Nice

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -97,53 +103,125 @@ static bool simplify_concat(std::shared_ptr<Node> n)
...
 
-        auto it = carg;
-        while (it != goe)
+        auto slice = std::dynamic_pointer_cast<op::Slice>(matcher->get_pattern_map()[lslice]);
+
+        if (branch_tip)
         {
-            if (auto rcarg = std::dynamic_pointer_cast<op::Reshape>(carg))
+            auto cur_lower_bounds = slice->get_lower_bounds();
+            if (cur_lower_bounds < prev_lower_bounds)
```
csullivan: I guess the only bug vector here, albeit unlikely, is if one of the slices is left out and the included slices are all in the correct order wrt their lower_bounds.

pruthviIntel:
I am hitting an exception "C++ exception with description "Input axis order for reshape is not a permutation of argument's axes" thrown in the test body" for the below 3 layered 10 tilmestep LSTM model. Will wait for the fix before approval.
[Function_0_001_N6ngraph7runtime3cpu4pass9RNNFusionE.pdf](https://github.com/NervanaSystems/ngraph/files/2376124/Function_0_001_N6ngraph7runtime3cpu4pass9RNNFusionE.pdf)

csullivan:
Great catch @pruthviIntel. There was a small bug when a transpose is required and `branch_shape.size() >= transposed_shape.size()` which I have fixed and tested that it solves the issue (unfused INTERPRETER result == fused GPU result for `rnn-10-step-fusion-test.json`). 

However, I had to disable the following check, is it required? 
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/pass/cpu_rnn_fusion.cpp#L728

csullivan:
Right. But after the algebraic simplification pass all of the slices are simplified into a single tensor. That said, I just verified and this check is fine as long as `AlgebraicSimplification` is run after all of the RNN fusion related passes. @pruthviIntel, please review the changes of !b43dda6 and verify fusions on CPU.

PR #1517, opened 2018-08-29
[ONNX] re-enable unit tests for CentOS

apwojcik:
GCC 4.5.x has a bug, and it generates invalid code. Here's an example C++ code.
```
class Foo {
public:
  Foo() { }
  ~Foo() { }
};

class Bar {
  Bar() = delete;
  Bar(const Bar&) = delete;
  Bar(Bar&&) = default;
  const Foo& tf;
public:
  Bar(const Foo& f) : tf{f} { }
  ~Bar(){ }
};
```
Here's how the C++ 4.5.x compiler should generate the constructor `Bar(const Foo&f)`
```
Bar::Bar(Foo const&):
        push    rbp
        mov     rbp, rsp
        mov     QWORD PTR [rbp-8], rdi
        mov     QWORD PTR [rbp-16], rsi
        mov     rax, QWORD PTR [rbp-8]
        mov     rdx, QWORD PTR [rbp-16]
        mov     QWORD PTR [rax], rdx
        pop     rbp
        ret
```
Here's what actually the C++ 4.5.x compiler is doing.
```
Bar::Bar(Foo const&):
        push    rbp
        mov     rbp, rsp
        sub     rsp, 32
        mov     QWORD PTR [rbp-24], rdi
        mov     QWORD PTR [rbp-32], rsi
        mov     rax, QWORD PTR [rbp-24]
        lea     rdx, [rbp-1]
        mov     QWORD PTR [rax], rdx
        lea     rax, [rbp-1]
        mov     rdi, rax
        call    Foo::~Foo() [complete object destructor]    // <---------- ??????
        leave
        ret
```
The workaround to this problem is to use `const T*` instead of `const T&`. In modern C++ the raw pointer is treated as reference, too. Everything which has raw pointer to other object does not own it and does not mange it lifetime. It only references the object.

There's no need for `ptr != nullptr`, because it is impossible to instantiate core ONNX class with an invalid Protobuf object (constructors get `const T&`). 

### DIFF
--- a/src/ngraph/frontend/onnx_import/core/attribute.cpp
+++ b/src/ngraph/frontend/onnx_import/core/attribute.cpp
@@ -23,10 +23,10 @@ namespace ngraph
     {
         std::vector<Graph> Attribute::get_graph_array() const
         {
-            return {std::begin(m_attribute_proto.graphs()), std::end(m_attribute_proto.graphs())};
+            return {std::begin(m_attribute_proto->graphs()), std::end(m_attribute_proto->graphs())};
         }
 
-        Graph Attribute::get_graph() const { return Graph{m_attribute_proto.g()}; }
+        Graph Attribute::get_graph() const { return Graph{m_attribute_proto->g()}; }
     } // namespace onnx_import
 
 } // namespace ngraph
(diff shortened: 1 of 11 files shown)


aprocter:
@apwojcik Any particular aspect you want me to review?

apwojcik:
@aprocter Long story short I need approval that there's no other well-known solution to this problem. This kind of bug we hit for the first time and maybe there are known solutions e.g. not affecting the code?

PR #1518, opened 2018-08-29
Add nbench support for processing all models in a directory

rkimballn1:
Timing detail aggregates all results from all models when run on a directory.


src/tools/nbench/nbench.cpp
```diff
@@ -33,6 +35,121 @@
...
+
+vector<PerfShape> to_perf_shape(shared_ptr<Function> f,
+                                const vector<runtime::PerformanceCounter>& perf_data)
+{
+    vector<PerfShape> result;
+    auto node_map = get_node_map(f);
+    for (const runtime::PerformanceCounter& p : perf_data)
+    {
+        auto node = node_map[p.name()];
+        Shape shape = node->get_outputs()[0].get_shape();
```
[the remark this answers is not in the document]
rkimballn1: I was planning on addressing that in a separate PR

src/tools/nbench/nbench.cpp
```diff
@@ -198,27 +316,35 @@ OPTIONS
...
-                run_benchmark(f, backend, iterations, timing_detail, warmup_iterations);
+                // cout << "Benchmarking " << m << ", " << backend << " backend, " << iterations
+                //      << " iterations.\n";
+                auto perf_data =
+                    run_benchmark(f, backend, iterations, timing_detail, warmup_iterations);
+                auto perf_shape = to_perf_shape(f, perf_data);
+                aggregate_perf_data.insert(
+                    aggregate_perf_data.end(), perf_shape.begin(), perf_shape.end());
             }
             catch (exception e)
```
csullivan: In a separate PR a check here on the exception type to discern whether it was an ngraph::unimplemented exception (for example) would be useful.

src/tools/nbench/nbench.cpp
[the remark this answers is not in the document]
fengleitian: OK

src/tools/nbench/nbench.cpp
```diff
@@ -33,6 +35,121 @@
...
+
+multimap<size_t, string> aggregate_timing_details(const vector<PerfShape>& perf_data)
+{
+    unordered_map<string, size_t> timing;
+    unordered_map<string, size_t> count;
+    for (const PerfShape& p : perf_data)
+    {
+        string op = p.name().substr(0, p.name().find('_'));
+        string shape_name = " {" + join(p.shape) + "} ";
+        timing[op + shape_name] += p.microseconds();
```
fengleitian: Initialization?

src/tools/nbench/nbench.cpp
```diff
@@ -33,6 +35,121 @@
...
+    return rc;
+}
+
+multimap<size_t, string> aggregate_timing(const vector<PerfShape>& perf_data)
+{
+    unordered_map<string, size_t> timing;
+    for (const PerfShape& p : perf_data)
+    {
+        string op = p.name().substr(0, p.name().find('_'));
+        timing[op] += p.microseconds();
```
fengleitian: Initialization?

src/tools/nbench/nbench.cpp
rkimballn1: unordered_map will create an entry if one does not exist and initialize it to zero or the default ctor. `timinig["some string not in map"] == 0`

fengleitian: COOL. Did know that.

PR #1526, opened 2018-08-30
Add conv add fusion

gaurides:


src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ set(SRC
     op/max_pool_with_indices.cpp
```
jbobba: Please revert all permission changes to the files.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -474,6 +475,26 @@ namespace ngraph
...
+                        ConvolutionLayout<ngraph::op::ConvolutionAdd, false, false>(
+                            node, i_mds, o_mds);
+                        // Force second input to sum to use the same layout as convolution output
+                        i_mds.push_back(o_mds[0]);
+                        node = insert_input_conversions(external_function, node, i_mds);
+                        set_output_layouts(node, o_mds);
+                    }
+                    else
+                    {
+                        set_native_layouts(external_function, node);
```
jbobba: might be better to throw an error if not mkldnn here as well.

src/ngraph/runtime/cpu/pass/cpu_fusion.hpp
```diff
@@ -72,6 +72,11 @@ class ngraph::runtime::cpu::pass::CPUFusion : public ngraph::pass::GraphRewrite
             construct_conv_bias();
             construct_sigmoid_multiply();
         }
+        if (fusions & REGULAR_FUSIONS)
+        {
+            construct_conv_add();
```
jbobba: Add to the previous conditional for REGULAR_FUSIONS

gaurides: If added to previous condition, then conv_bias was getting replaced with conv_add. 
Hence, I added it after conv_bias.

jbobba: Oh, right. We should be able to move all the differentiable_fusions before regular_fusions. Would be good to add a comment about the ordering requirement between conv_bias and conv_add.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
gaurides: You mean instead of using set_native_layouts() , throw ngraph_error() ? 
I will do it, but i don't understand why though.

src/ngraph/runtime/cpu/pass/cpu_fusion.hpp
gaurides: Yes, moving all differentiable_fusions before regular_fusions works. Will add comment too.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
jbobba: There is no non-mkldnn kernel to execute an in-place ConvolutionAdd. So we shouldn't hit this case ever.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -219,6 +220,33 @@ namespace ngraph
...
+                        data_dilated = data_dilated || (s != 1);
+                    }
+
+                    if (!data_dilated && arg0_rank == 4 && arg1_rank == 4 &&
+                        node->get_input_element_type(0) == element::f32)
+                    {
+                        auto op_annotations =
+                            std::make_shared<ngraph::runtime::cpu::CPUOpAnnotations>();
+                        op_annotations->set_mkldnn_op(true);
+                        const int ADD_INPUT = 2; // CHECK: 3 or 2?
```
jbobba: Change to 2.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
gaurides: Ok. Done

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
gaurides: Done!

gaurides:
test now

src/ngraph/runtime/cpu/op/conv_add.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include "ngraph/util.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+void op::util::validate_conv_shapes(const Shape& data_shape, const Shape& filters_shape)
+{
+    if (data_shape[1] != filters_shape[1])
+    {
+        throw ngraph_error(
```
Krovatkin: please switch to NGRAPH_TYPE_CHECK; see other ops for an example

src/ngraph/runtime/cpu/op/conv_add.cpp
```diff
@@ -0,0 +1,121 @@
...
+    auto& data_batch_et = data_batch->get_element_type();
+    auto& filters_shape = filters->get_shape();
+    auto& filters_et = filters->get_element_type();
+
+    //
+    // Make sure data batch and filter element types match.
+    //
+    if (data_batch_et != filters_et)
+    {
+        throw ngraph_error("Convolution data batch and filter element types do not match");
```
Krovatkin: NGRAPH_TYPE_CHECK

src/ngraph/runtime/cpu/op/conv_add.cpp
gaurides: Strangely, I couldn't find any examples for NGRAPH_TYPE_CHECK, even on master.
Can you please point me to a specific example?

aprocter: I think @Krovatkin is referring to NODE_VALIDATION_ASSERT

gaurides: Great! Thanks. Now there are lot of examples.
will do the change

gaurides:
TESTNOW

gaurides:
TESTNOW

PR #1533, opened 2018-08-31
Better node validation error messages

aprocter:
Update almost all node validation error messages to use the new error message infrastructure, which provides more context/details.

Main changes:

- Changed `TypeCheckError` to `NodeValidationError` and `TYPE_CHECK_ASSERT` to `NODE_VALIDATION_ASSERT`.
- Changed `copy_with_new_args` to throw `NodeValidationError` if `new_args` size is wrong, moved logic for those checks into a function.
- Changed almost (see below) all node validation error messages to use `TYPE_CHECK_ASSERT`, with richer context info.

There are some ops I haven't gotten to yet but will do in a follow-on PR:

- Convolution and its backprops
- BatchNorm and its backprop
- MaxPool and its backprop

There are some ops I did not do for this PR and may do in a follow-on PR, if we decide we need to keep them:

- Reduce
- ReduceWindow
- SelectAndScatter

Most of the changes are regular in form, so I will go through and drop comments on this PR for the cases where I've gone a bit beyond the strict mandate.


src/ngraph/op/concat.cpp
```diff
@@ -32,56 +32,59 @@ op::Concat::Concat(const NodeVector& args, size_t concatenation_axis)
 
 void op::Concat::validate_and_infer_types()
 {
```
aprocter: NB: this function got somewhat heavier surgery than most.

src/ngraph/op/relu.cpp
```diff
@@ -24,38 +24,23 @@ op::Relu::Relu(shared_ptr<Node> arg)
     : UnaryElementwiseArithmetic("Relu", {arg})
 {
     constructor_validate_and_infer_types();
-    set_output_type(0, arg->get_element_type(), arg->get_shape());
```
aprocter: NB: I dropped what I think is a redundant `set_output_type` here (`Relu` is in `UnaryElementwiseArithmetic`).

src/ngraph/op/relu.cpp
```diff
@@ -24,38 +24,23 @@ op::Relu::Relu(shared_ptr<Node> arg)
...
-    if (new_args.size() != 1)
-    {
-        throw ngraph_error("Incorrect number of new arguments");
-    }
+    check_new_args_count(this, new_args, 1);
     return make_shared<Relu>(new_args.at(0));
 }
 
 op::ReluBackprop::ReluBackprop(shared_ptr<Node> arg, shared_ptr<Node> delta)
-    : Op("ReluBackprop", check_single_output_args({arg, delta}))
```
aprocter: NB: I moved this to `BinaryElementwiseArithmetic`.

src/ngraph/op/function_call.cpp
```diff
@@ -53,6 +57,7 @@ op::FunctionCall::FunctionCall(shared_ptr<Function> function, const NodeVector&
 
 shared_ptr<Node> op::FunctionCall::copy_with_new_args(const NodeVector& new_args) const
 {
+    check_new_args_count(this, new_args, m_function->get_parameters().size());
```
aprocter: NB: This check was not here before, but I think it should be.

src/ngraph/op/result.cpp
```diff
@@ -44,15 +42,10 @@ void op::Result::validate_and_infer_types()
...
-    {
-        throw ngraph_error("Incorrect number of new arguments");
-    }
+    check_new_args_count(this, new_args, 1);
 
-    if (new_args.at(0)->get_outputs().size() != 1)
-    {
-        throw ngraph_error("Result::copy_with_new_args expected a single-output argument");
-    }
+    NODE_VALIDATION_ASSERT(this, new_args[0]->get_outputs().size() == 1)
```
aprocter: I have preserved this check, but I think it is redundant. (The condition will be checked at validation.)

src/ngraph/op/concat.cpp
```diff
@@ -32,56 +32,59 @@ op::Concat::Concat(const NodeVector& args, size_t concatenation_axis)
...
+                NODE_VALIDATION_ASSERT(this, first_input_shape[j] == get_inputs()[i].get_shape()[j])
+                    << "Dimensions of argument " << i << " do not match for axis " << j
+                    << " (expected " << first_input_shape[j] << ", got "
+                    << get_inputs()[i].get_shape()[j] << ").";
             }
-            else if (j == m_concatenation_axis)
+            else
             {
-                concatenation_axis_length += input_i_shape.at(j);
+                concatenation_axis_output_length += get_inputs()[i].get_shape()[j];
```
diyessi: `get_input_shape(i)[j]` is more direct.

src/ngraph/op/result.cpp
diyessi: We should drop it -- in the future if the input node hasn't been validated yet, the test will fail.

PR #1534, opened 2018-08-31
GPU logical reshape pass-through

csullivan:
Add pass-through support for Reshape that doesn't change the memory layout (logical reshape only).


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -516,6 +516,36 @@ namespace ngraph
                     return;
                 }
                 auto reshape = static_cast<const op::Reshape*>(node);
+                auto can_skip_reshape = [&]() {
```
fengleitian: check if the input name is same as output name.

fengleitian: if(same) 
   skip;
else 
   copy;

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -516,6 +516,36 @@ namespace ngraph
...
+                    if (!reshape->get_is_transpose())
```
Krovatkin: Please don't do this! Emitter isn't a MemoryLayout pass. The decision whether to copy or not should be as localized as possible.
Do this:
```cpp
if (args[0].get_name() == out[0].get_name())
{
   // print skipping comment
}
else
{
  // emit a copy
}
```

fengleitian: @Krovatkin 
anything to do with `op_annotations->add_in_place_oi_pair({0, 0, false});`? should we use true here?

Krovatkin: Logical `Reshape`s are non-destructive, so it should be `false`. It's `true` for `ReplaceSlice`, for example. The reason why CPU has those checks in Emitter is somewhat convoluted and it has to do with a `CPULayout` pass optimizing away unnecessary transposes. You guys shouldn't have these issues so a simple equal comparison should work.

csullivan:
@rkimballn1 @fengleitian , the problem I was seeing with the logical reshape elimination was coming from the CommonFunctionCollection pass. By caching the function emission for pass-through ops the elimination of the reshape wouldn't end up being emitted, because the version with the memcpy had already been cached. I've added a check in CommonFunctionCollection to not cache ops with `oi_pairs` and simplified the reshape emitter check.

I've also added functions to propagate in place inputs and outputs - borrowed from IA. Thanks to @Krovatkin for pointing them out.

src/ngraph/runtime/gpu/pass/gpu_layout.cpp
```diff
@@ -54,6 +55,33 @@ namespace ngraph
                         rep_slice->set_op_annotations(op_annotations);
                     }
                 }
+                template <>
+                void GPULayout::LAYOUT_DECL(ngraph::op::Reshape)
+                {
+                    auto reshape = static_cast<ngraph::op::Reshape*>(node.get());
+                    if (reshape->get_is_transpose() &&
+                        reshape->get_output_shape().size() ==
+                            reshape->get_argument(0)->get_shape().size())
```
fengleitian: does `reshape->get_output_shape().size() ==
                             reshape->get_argument(0)->get_shape().size()` required. Looks fine if we go 1^m to 1^n, or {2, 2} to {4} etc.

fengleitian: Actually we don't need to check if it's identical.

fengleitian: take it back, here you are check the case that can not do inplace.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -516,17 +516,23 @@ namespace ngraph
                     return;
                 }
                 auto reshape = static_cast<const op::Reshape*>(node);
+
+                if (out[0].get_name() == args[0].get_name() && !reshape->get_is_transpose())
```
fengleitian: since we don’t have a in place transpose.  `out[0].get_name() == args[0].get_name()` should covered everything.

csullivan: If we add one we will need to remember to put this back. Does it hurt anything having it here?

csullivan: All right, we can leave it for later.

PR #1539, opened 2018-09-01
Cmake flags update

avijit-nervana:
Added the following CMake flags:
* NGRAPH_USE_CXX_ABI to specify _GLIBCXX_USE_CXX11_ABI 
  -  When NGRAPH_USE_CXX_ABI is set 0 or 1, the value is passed to `gcc`.  
* NGRAPH_DOC_BUILD_DISABLE 
  - When set, documentation is not generated

These two flags (especially the first one) would allow building the nGraph TensorFlow bridge from pre-installed TensorFlow (as opposed to first building TensorFlow from source - which takes a while to build).


CMakeLists.txt
```diff
@@ -281,4 +288,7 @@ else()
     message(STATUS "unit tests disabled")
 endif()
 
-add_subdirectory(doc)
+if (NOT NGRAPH_DOC_BUILD_DISABLE)
```
rkimballn1: Should be called NGRAPH_DOC_BUILD_ENABLE to conform to all other flags

avijit-nervana: Done.

PR #1541, opened 2018-09-02
IntelGPU backend: helper function improvement

shssf:


src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -75,11 +77,11 @@ string
     {
         if (axis.find(var_idx) == axis.end())
         {
-            buffer += "[i" + to_string(var_idx) + "]";
+            buffer += "[" + var + to_string(var_idx) + "]";
```
rkimballn1: using an std::stringstream would make the string construction look better, no to_string needed.

PR #1546, opened 2018-09-02
[ONNXIFI] implement onnxGetBackendIDs() interface function

apwojcik:
This PR implements the `onnxGetBackendIDs` method from ONNXIFI API.
From ONNXIFI docs:

> Get stable IDs of available backends on the system.
> 
> ONNXIFI backend is a combination of software layer and hardware device used to run an ONNX graph. The same software layer may expose multiple backends (e.g. one ONNXIFI backend for each GPU in the system, or one ONNXIFI backend for GPU and another for CPU, both implemented in the same software). Backends implemented in the same software, but targeting different devices (e.g. "MyNN" for CPU and "MyNN" for GPU) have different backend IDs.
> 
> Note that some (hot-pluggable) backends can be connected and disconnected at any time, and thus subsequent calls to this function may return different number or set of backend IDs. The returned IDs, however, stay valid even if the hardware device used by the backend disconnects from the system.
> 
> To avoid resource leak, the backend ID MUST be released through a call to onnxReleaseBackendID when it is no longer needed.

Refer to https://github.com/onnx/onnx/blob/53b20f623a030d65e5b43b6ed0e994fdcc87e74b/onnx/onnxifi.h#L1013 for details


src/ngraph/frontend/onnxifi/backend_manager.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "backend.hpp"
+#include "backend_manager.hpp"
+
+namespace ngraph
+{
+    namespace onnxifi
+    {
+        BackendManager::BackendManager()
+        {
+            auto registered_backends = runtime::BackendManager::get_registered_backends();
```
rkimballn1: I'm trying but I don't understand what this class is going. registered_backends is a vector<string> and then you are casting the address to this temporary string to a uintptr_t on line 36. This looks wrong.

src/ngraph/frontend/onnxifi/backend.hpp
```diff
@@ -0,0 +1,85 @@
...
+        public:
+            Backend(const Backend&) = delete;
+            Backend& operator=(const Backend&) = delete;
+
+            Backend(Backend&&) = default;
+            Backend& operator=(Backend&&) = default;
+
+            Backend() = delete;
+
+            explicit Backend(std::string type)
```
rkimballn1: take a const string& and then skip the move statement. Exact same number of copies and more normal way to do  it.

src/ngraph/frontend/onnxifi/backend.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include <memory>  // std::shared_ptr
+#include <string>  // std::string
+#include <utility> // std::move
+#include <vector>  // std::vector
+
+#include "ngraph/function.hpp"
+#include "ngraph/runtime/backend.hpp"
+#include "ngraph/runtime/tensor_view.hpp"
+
+namespace ngraph
```
rkimballn1: Why this class? Use runtime::Backend. Is it missing something you need?

src/ngraph/frontend/onnxifi/backend_manager.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "ngraph/runtime/backend_manager.hpp"
+
+#include "backend.hpp"
+#include "backend_manager.hpp"
+
+namespace ngraph
+{
+    namespace onnxifi
+    {
+        BackendManager::BackendManager()
```
rkimballn1: If you need a special class rather than runtime::BackendManager then you also need a comment explaining why.

src/ngraph/frontend/onnxifi/backend_manager.cpp
apwojcik: `registered_backends` is used to eliminate the potential _undefined behavior_ which may occur if one uses the result of a function (which is `vector`) as `range_expression` of the range-based loop. To resolve this issue future standard (i.e. C++20) will provide `init-statement` section of the range-based loop (see https://en.cppreference.com/w/cpp/language/for for details).

apwojcik: `m_registered_backends` is the `std::map` which is mapping **handles** to `onnxifi::Backend` **objects**. The backend uses pointer (via `std::uintptr_t` because by the specification its size is equal size of pointer) to `type` as handle to `ngraph::runtime::Backend`.
So it does not matter whether `type` is temporar or permament becuaus it is used only as **handle** (a reference to actual object). The lifetime of `onnxifi::BackendManager` is determined by the lifetime of application (singleton pattern).
Accessing objects through handles is mandatory requirement of ONNXIFI specification.

src/ngraph/frontend/onnxifi/backend.hpp
apwojcik: This change was suggested by clang-tidy, to pass by value and use move semantics.

_The purpose of `clang-tidy` is to provide an extensible framework for diagnosing and fixing typical programming errors, like style violations, interface misuse, or bugs that can be deduced via static analysis._

BTW. Why do you think move semantics is not a normal way of expressing things in C++?

src/ngraph/frontend/onnxifi/backend.hpp
apwojcik: Yes, `runtime::Backend` is missing a log of from what ONNXIFI requires. This is just an early implementation of `onnxifi::Backend`. Consecutive PRs will add much more to the class.

src/ngraph/frontend/onnxifi/backend_manager.cpp
apwojcik: `runtime::BackendManager::create(const std::string& type) -> std::unique_ptr<runtime::Backend>`

The most attractive feature of `std::unique_ptr` is that it easily and effectively converts to a `std::shared_ptr`. The backward conversion (`std::shared_ptr` to `std::unique_ptr`) is not allowed by design.

90% of nGraph code shoud be `std::unique_ptr` and ownership passing.

src/ngraph/frontend/onnxifi/backend_manager.cpp
rkimballn1: So add a comment say that. You can't just write opaque code without comment.

src/ngraph/frontend/onnxifi/backend.hpp
rkimballn1: Please change it. For some reason you refuse to make your code look like all the rest of the code in the repo and seem to do things just to be different or to appease clion.

src/ngraph/frontend/onnxifi/backend.hpp
rkimballn1: Can you please add comment to the code explaining this? Adding comments to the PR don't help people in the future.

apwojcik: done

src/ngraph/frontend/onnxifi/backend.hpp
```diff
@@ -0,0 +1,86 @@
...
+{
+    namespace onnxifi
+    {
+        /// \brief ONNXIFI extensions to nGraph backend
+        class Backend
```
arogowie-intel: How one would get id for _specific_ `Backend`?  There are some ONNXIFI functions that require backend id i.e. `onnxGetBackendInfo` or  `onnxInitBackend`

apwojcik: Please use `onnxGetBackendIDs()` to obtain IDs and `onnxGetBackendInfo()` to choose the backend (based on properties you choose).

apwojcik: or to use `onnxGetBackendCompatibility()` instead of `onnxGetBackendInfo()`

PR #1559, opened 2018-09-04
Add in_place support for ReplaceSlice

gaurides:


src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -483,6 +484,30 @@ namespace ngraph
...
+                void CPUAssignment::ASSIGN_DECL(ngraph::op::ReplaceSlice)
+                {
+                    auto replace_slice = static_cast<op::ReplaceSlice*>(node);
+
+                    //auto arg0_shape = node->get_input_shape(0);
+                    //auto arg0_rank = arg0_shape.size();
+                    //auto result_shape = node->get_output_shape(0);
+
+                    //TO confirm : not checking ranks as we support ranks 0/1/2/4
+                    if (node->get_input_element_type(0) == element::f32)
```
Krovatkin: I think we can support any supported element type.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -483,6 +484,30 @@ namespace ngraph
                     }
                 }
 
+                template <>
+                void CPUAssignment::ASSIGN_DECL(ngraph::op::ReplaceSlice)
+                {
+                    auto replace_slice = static_cast<op::ReplaceSlice*>(node);
+
+                    //auto arg0_shape = node->get_input_shape(0);
```
Krovatkin: remove dead code

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -2422,16 +2422,31 @@ namespace ngraph
...
-                                           args[1].get_name(),
-                                           out[0].get_name(),
-                                           args[1].get_shape(),
-                                           out[0].get_shape(),
-                                           replace_slice->get_lower_bounds(),
-                                           replace_slice->get_upper_bounds(),
-                                           replace_slice->get_strides());
+                if (args[0].get_name() != out[0].get_name())
+                {
+                    kernel::emit_replace_slice(writer,
```
Krovatkin: why can't we use the original version of `emit_replace_slice`?

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
gaurides: I think so too. 
we would need to check if the type for arg0 and arg1 are of same type

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
gaurides: will do. just wanted to confirm we don't need to use ranks anywhere.
Again, i feel we should just make sure the ranks are same for arg0 and arg1

src/ngraph/runtime/cpu/cpu_emitter.cpp
gaurides: My main intention is to avoid a copy in emit_replace_slice(). 
So, I was self debating where to do the if check : inside `emit_replace_slice()` or even inside `emit_pointwise_copy()`
However, I thought it was easier to read and understand the intention from the very top. Hence i added there.
On second thoughts, `emit_replace_slice()` might be a good place too. I will change it that way.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -483,6 +484,23 @@ namespace ngraph
...
 
+                template <>
+                void CPUAssignment::ASSIGN_DECL(ngraph::op::ReplaceSlice)
+                {
+                    auto replace_slice = static_cast<op::ReplaceSlice*>(node);
+
+                    // ReplaceSlice is independent of data type. Hence not checking type
+                    auto op_annotations =
+                        std::make_shared<ngraph::runtime::cpu::CPUOpAnnotations>();
+                    op_annotations->set_mkldnn_op(true);
```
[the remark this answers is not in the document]
gaurides: Oh.. yes. removed the check.

gaurides:
TESTNOW

gaurides:
TESTNOW

PR #1562, opened 2018-09-05
[ONNX] Reduce* operations

arogowie-intel:


src/ngraph/frontend/onnx_import/utils/reduction.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace ngraph
+{
+    namespace onnx_import
+    {
+        namespace reduction
+        {
+            namespace detail
+            {
+                inline std::vector<std::size_t> get_reduction_axes(const Node& node)
```
tsocha: Consider use of ngraph::AxisVector as return type.

src/ngraph/frontend/onnx_import/op/reduce.hpp
```diff
@@ -0,0 +1,217 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#pragma once
+
+#include <memory>
+
+#include "ngraph/node_vector.hpp"
+#include "ngraph/op/abs.hpp"
```
postrational: Please consider moving the implementation into a `.cpp` file, so you won't have to include all of these headers in this header, which is included by the `ops_bridge`.

src/ngraph/frontend/onnx_import/utils/common.hpp
```diff
@@ -0,0 +1,106 @@
...
+                    ///
+                    template <typename ForwardIterator, typename T>
+                    void fill_monotonic_range(ForwardIterator first,
+                                              ForwardIterator last,
+                                              T init_value,
+                                              T step)
+                    {
+                        while (first != last)
+                        {
+                            *first++ = init_value;
```
apwojcik: I think regular `for()` is better, please notice post increment requires additional memory to hold previous state
```
for (;first != last; ++first, init_value += step)
{
    *first = init_value;
}
```

src/ngraph/frontend/onnx_import/utils/reshape.cpp
```diff
@@ -59,9 +60,8 @@ namespace ngraph
 
             AxisVector get_default_axis_vector(std::size_t data_shape_size, std::size_t start_value)
             {
-                AxisVector axis_vector(data_shape_size);
-                std::iota(std::begin(axis_vector), std::end(axis_vector), start_value);
-                return axis_vector;
+                return AxisVector{onnx_import::common::get_monotonic_range<std::size_t>(
```
apwojcik: `common::get_monotonic_range<>` unless this piece of code is not part of `onnx_import` namespace.

apwojcik: and `return {...}`, the `AxisVector` is not required.

src/ngraph/frontend/onnx_import/utils/reduction.hpp
```diff
@@ -0,0 +1,111 @@
...
+                    return op_node;
+                }
+
+                auto output_shape = data_shape;
+                // flatten reduced axes
+                for (const auto& idx : reduction_axes)
+                {
+                    output_shape.at(idx) = 1;
+                }
+                auto reshaped_op_node = std::make_shared<ngraph::op::Reshape>(
```
apwojcik: `return std::make_shared<ngraph::op::Reshape>(...);` the variable `reshape_op_node` is not required.

src/ngraph/frontend/onnx_import/utils/reduction.hpp
```diff
@@ -0,0 +1,111 @@
...
+                inline AxisSet get_reduction_axes(const Node& node)
+                {
+                    auto reduction_axes =
+                        node.get_attribute_value<std::vector<std::size_t>>("axes", {});
+                    if (reduction_axes.empty())
+                    {
+                        reduction_axes = onnx_import::common::get_monotonic_range<std::size_t>(
+                            node.get_ng_inputs().at(0)->get_shape().size());
+                    }
+                    return AxisSet{reduction_axes};
```
apwojcik: `AxisSet` is not required unless it is `explicit` then it is required.

src/ngraph/frontend/onnx_import/utils/reduction.hpp
postrational: The variable `reshape_op_node` makes the code more readable. Unless we're making an unnecessary copy here, let's keep it.

src/ngraph/frontend/onnx_import/utils/reshape.cpp
postrational: Does adding `AxisVector` hurt anything? If not then, it makes the code more explicit and less magical...

src/ngraph/frontend/onnx_import/utils/reduction.hpp
arogowie-intel: I think that this code is more readable.

src/ngraph/frontend/onnx_import/utils/reshape.cpp
arogowie-intel: I think that returning `AxisVector{...}` is more readable.

src/ngraph/frontend/onnx_import/utils/reduction.hpp
arogowie-intel: The `return std::make_shared<ngraph::op::Reshape>()` clearly states that the function returns node which does reshape op.

PR #1563, opened 2018-09-05
[WIP] Split Batchnorm into BatchnormTraining and BatchnormInference

Krovatkin:


test/backend_test.in.cpp
```diff
@@ -8843,7 +8843,7 @@ NGRAPH_TEST(${BACKEND_NAME}, batchnorm_fprop_globalstats_b2c2w2h1)
     auto beta = make_shared<op::Parameter>(element::f32, beta_shape);
     double eps = 0.001;
     auto shape_r = Shape{2, 2, 2, 1};
-    auto bn = make_shared<op::BatchNorm>(eps, gamma, beta, input, mean, var, true);
+    auto bn = make_shared<op::BatchNormTraining>(eps, gamma, beta, input, mean, var);
```
csullivan: Should this not also be `op::BatchNormInference` since the parameters are provided? I tested that this works on CPU. 

I have a commit that enables GPU support and changes this test, if you agree with the above I will push the commit to your branch.

csullivan: After further review, the test is fine as is. BatchNormTrainingForward always updates the mean and variance. The ability to pass the mean/var as parameters allows the normalization to have non-trivial initial conditions, but in both cases (whether they are provided as arguments or maintained internally) the forward prop should update the mean and variance. This is the behavior in both mkldnn and cudnn. 

PR #1569 provides GPU support for this and targets this branch.

PR #1564, opened 2018-09-05
Make apply-code-format set correct permissions on source files

Krovatkin:


postrational:
Good idea. We could also add something to change line endings to Unix LF.

Krovatkin:
@postrational I could definitely add that. Let's see what @diyessi and @rkimballn1 think.

Krovatkin:
@diyessi they are being set too permissively if a file is edited on windows over a samba share which is a really really really convenient way to edit files 😄

diyessi:
Then you could just set umask in the script so everything get created less permissively.

Krovatkin:
@diyessi `umask` **seems** to work to my surprise (because every new session still shows the old `umask` 🤷‍♂️ ) but it needs to be set before any files are edited or created which is a relatively big caveat. Let alone the fact that some users might **not** appreciate our scripts modifying their `umask` setting even if it is only for the current session. I find setting using `chmod 0644` is a way more reliable and direct way to achieve what we want 😄

diyessi:
umask should only affect what happens in the script since it's a separate process.

aprocter:
Should we update check-code-format as well? (Mispermissioned files would block CI in that case, not sure how we feel about that).

Krovatkin:
@aprocter I don't know either. This is kinda the best effort w/o over-imposing.

PR #1565, opened 2018-09-06
Add support for Dequantize op via mkldnn for IA backend (codegen + DEX)

nbpatel:


src/ngraph/runtime/cpu/op/dequantize.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+    if (max->get_shape().size() != 0)
+    {
+        throw ngraph_error("Max is not a scalar!");
+    }
+
+    if (!(std::dynamic_pointer_cast<op::Constant>(min) &&
+          std::dynamic_pointer_cast<op::Constant>(max)))
+    {
+        throw ngraph_error("Min and max have to be constants!");
```
aprocter: If these are required to be constants, they should be attributes rather than data flow inputs. Not necessarily a blocker if this is somehow more convenient for the CPU backend ops, but for core ops we can't have restrictions like this.

PR #1569, opened 2018-09-06
Add GPU support for the new BN ops

csullivan:


src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -1229,14 +1229,16 @@ size_t runtime::gpu::CUDNNEmitter::build_batchnorm(const cudnnBatchNormMode_t& b
                                                    const Prop& direction,
                                                    const Shape& tensor_shape,
                                                    const Shape& param_shape,
-                                                   double epsilon)
+                                                   double epsilon,
+                                                   bool global_stats)
```
csullivan: @fengleitian FYI: This and lines 1305 (and their use in the cudnn call) are the only meaningful additions to the BN kernel. Now if the mean and variance are passed to BNForward, they will be correctly used and updated by the cudnn kernel.

fengleitian: OK.

PR #1570, opened 2018-09-06
add ISO C++ Guideline Support Library (gsl-lite)

apwojcik:
Currently, only ONNXIFI is using it, but in the future, nGraph may benefit, too.

The Guideline Support Library (GSL) contains functions and types that are suggested for use by the _C++ Core Guidelines_ maintained by the _Standard C++ Foundation_. The library includes types like `owner<>`, `not_null<>`, `span<>`, `string_span` and much more.

The `gsl-lite` recognizes when it is compiled for the CUDA platform and decorates functions (methods) with `__host__` and `__device__`.

Refer to https://github.com/isocpp/CppCoreGuidelines for more details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -244,6 +244,8 @@ if(NOT DEFINED EXTERNAL_PROJECTS_ROOT)
     set(EXTERNAL_PROJECTS_ROOT ${CMAKE_CURRENT_BINARY_DIR})
 endif()
 
+include(cmake/external_gsl-lite.cmake)
+
 if (NGRAPH_ONNX_IMPORT_ENABLE)
     if (NOT NGRAPH_USE_SYSTEM_PROTOBUF)
         include(cmake/external_protobuf.cmake)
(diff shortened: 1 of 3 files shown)


apwojcik:
For starter, I would like to use `gsl::span<>`, it is described here https://en.cppreference.com/w/cpp/container/span. Trained ONNX models may have a gigabyte or more in size. I want to view the memory as ONNX model in form of a set of objects without copying the data.

TL;DR:
The _Cpp Core Guidelines_ is a set of core guidelines for modern C++, the recommendations from the standard committee signed by Bjarne Stroustrup and Herb Sutter, taking likely future enhancements and ISO Technical Specifications (TSs) into account. The aim is to help C++ programmers to write simpler, more efficient, more maintainable code. The guidelines are not mandatory. We may adopt only those we think we benefit the most.

The _Guidelines Support Library_ is the implementation of the guidelines in form of class templates. `gsl-lite` is single header file implementation of the library based on Microsoft's GSL.

apwojcik:
Here's the link to the header file:
https://github.com/martinmoene/gsl-lite/blob/master/include/gsl/gsl-lite.hpp

apwojcik:
I asked Jim John to start the approval process.

diyessi:
The Microsoft version wants C++14 support, and we have requirements for older compilers.

diyessi:
What is the status of this (getting approval, etc.)?

PR #1571, opened 2018-09-07
Add support for Quantized Pooling(Max + Avg) op via mkldnn for IA backend (codegen + DEX)

nbpatel:


src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -3046,6 +3048,85 @@ namespace ngraph
                 }
             }
 
+            template <>
+            void CPU_Emitter::EMITTER_DECL(ngraph::op::QuantizedMaxPool)
+            {
+                auto qmax_pool = static_cast<const ngraph::op::QuantizedMaxPool*>(node);
```
jbobba: Instead of duplicating this code between builder and codegen, please create a builder method in mkldnn_emitter

src/ngraph/runtime/cpu/op/quantized_avg_pool.cpp
```diff
@@ -0,0 +1,232 @@
...
+    Shape result_shape(1 + 1 + spatial_dimension_count);
+    result_shape[0] = batch_size;
+    result_shape[1] = channel_count;
+    copy(output_item_shape.begin(), output_item_shape.end(), result_shape.begin() + 2);
+
+    set_output_size(3);
+    set_output_type(0, get_input_element_type(0), result_shape);
+    //TODO(nbpatel): Change to Shape{} once the mkldnn version is updated
+    set_output_type(1, element::f32, Shape{1});
+    set_output_type(2, element::f32, Shape{1});
```
jbobba: seems like a lot of code duplication with the core op. Can we refactor and reuse from there? Perhaps @aprocter has some thoughts.

aprocter: Heartily agree we need to refactor (probably moving duplicated logic into utility functions but maybe into subclasses). I don’t think it’s necessary for this PR, but should be done Very Soon.

aprocter: (Ideally the scope of that refactor would cover the core ops as well, which themselves have an embarrassing amount of duplication.)

nbpatel: Most of the Cpu ops also have a lot of duplication from the core ops. So i guess it is a different task in itself

src/ngraph/runtime/cpu/op/quantized_avg_pool.cpp
```diff
@@ -0,0 +1,232 @@
...
+    : Op("QuantizedAvgPool", check_single_output_args({arg, min, max}))
+    , m_window_shape(window_shape)
+    , m_window_movement_strides(window_movement_strides)
+    , m_padding_below(padding_below)
+    , m_padding_above(padding_above)
+    , m_include_padding_in_avg_computation(include_padding_in_avg_computation)
+{
+    constructor_validate_and_infer_types();
+
+    if (arg->get_element_type() != element::u8 && arg->get_element_type() != element::i8)
```
Krovatkin: why aren't these checks in `validate_and_infer`? I'm a bit confused by this split.

PR #1575, opened 2018-09-07
Use MKLDNN for non-strided slice ops.

ayzhuang:


src/ngraph/runtime/cpu/builder/slice.cpp
```diff
@@ -48,52 +48,74 @@ namespace ngraph
...
-                                   lower_bounds,
-                                   upper_bounds,
-                                   strides);
-                        };
                     functors.emplace_back(functor);
                 }
                 else
                 {
-                    std::function<decltype(runtime::cpu::kernel::slice<float, 2>)> kernel;
+                    bool strided = false;
```
Krovatkin: I'd suggest adding this as a utility function to `ngraph_util.cpp`. You could also use `std::all_of` or `std::none_of` if it's any clearer.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -1431,6 +1432,34 @@ namespace ngraph
...
+                    if (mkldnn_utils::use_mkldnn_kernel(node.get()))
+                    {
+                        // pass input format to output
+                        auto input_md = mkldnn_utils::get_input_mkldnn_md(node.get(), 0);
+                        NGRAPH_DEBUG << "input memory format: " << input_md.data.format << "\n";
+                        auto result_format =
+                            static_cast<mkldnn::memory::format>(input_md.data.format);
+                        if (result_format == mkldnn::memory::blocked)
+                        {
+                            //TODO get named blocked format for output?
```
Krovatkin: could we try `get_named_md` which might be able to canonicalize a blocked layout into `nChw8c` or `nChw16c` simple_reorder seems to support both.

jbobba: Can we create an MKLDNN pd and query for output format?

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -642,6 +643,30 @@ namespace ngraph
...
+                    for (auto stride : strides)
+                    {
+                        if (stride != 1)
+                        {
+                            strided = true;
+                            break;
+                        }
+                    }
+
+                    if (!strided)
```
jbobba: We'll also need to check for datatypes. MKLDNN cant handle all the datatypes.

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
ayzhuang: Output gets its format from input. I have modified the code to call create_blocked_mkldnn_md when input format is blocked.

src/ngraph/runtime/cpu/builder/slice.cpp
```diff
@@ -48,52 +48,64 @@ namespace ngraph
...
-                    }
-                }
+                    auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
+                    auto input_desc = mkldnn_utils::get_input_mkldnn_md(node, 0);
+                    auto result_desc = mkldnn_utils::get_output_mkldnn_md(node, 0);
 
-                if (strided)
-                {
-                    std::function<decltype(runtime::cpu::kernel::strided_slice<float, 2>)> kernel;
+                    auto slice_index = mkldnn_emitter->build_slice(
```
jmenon: Is the mkldnn primitive actually faster than the existing kernel? Or is the primary motivation to eliminate layout conversions?

ayzhuang: to eliminate layout conversions.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -690,6 +691,19 @@ namespace ngraph
...
+                    auto strides = slice->get_strides();
+                    if (!is_strided(strides) && node->get_input_element_type(0) == element::f32)
+                    {
+                        auto op_annotations =
+                            std::make_shared<ngraph::runtime::cpu::CPUOpAnnotations>();
+                        op_annotations->set_mkldnn_op(true);
+                        slice->set_op_annotations(op_annotations);
+                    }
+                }
+
```
jbobba: Add a template<>

PR #1576, opened 2018-09-07
Add support for Quantize op via mkldnn for IA backend (codegen + DEX)

nbpatel:


src/ngraph/runtime/cpu/quantization_util.hpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    namespace runtime
+    {
+        namespace cpu
+        {
+            namespace quantization_util
+            {
+                static inline void get_min_max_range(const ngraph::Node* node,
+                                                     std::vector<float>& quant_util)
+                {
```
jbobba: Make this independent of quantize node. You can take input_min_range and input_max_range as inputs and return adjusted min/max and scale.

test/quantize_cpu.cpp
```diff
@@ -72,3 +73,81 @@ TEST(quantize_cpu, dequantize_from_int8)
...
+    auto f = make_shared<Function>(NodeVector{output_data, output_min, output_max},
+                                   op::ParameterVector{A});
+    auto backend = runtime::Backend::create("CPU");
+    // Create some tensors for input/output
+    auto a = backend->create_tensor(element::f32, shape_a);
+    copy_data(a, a_data);
+    auto result = backend->create_tensor(element::u8, shape_a);
+    auto result_min = backend->create_tensor(element::f32, Shape{});
+    auto result_max = backend->create_tensor(element::f32, Shape{});
+    backend->call(f, {result, result_min, result_max}, {a});
```
jbobba: use call_with_validate()

test/quantize_cpu.cpp
```diff
@@ -72,3 +73,81 @@ TEST(quantize_cpu, dequantize_from_int8)
...
```
jbobba: call_with_validate()

src/ngraph/runtime/cpu/builder/quantize.cpp
```diff
@@ -0,0 +1,73 @@
...
+                    auto input_desc = mkldnn_utils::get_input_mkldnn_md(node, 0);
+                    auto result_desc = mkldnn_utils::get_output_mkldnn_md(node, 0);
+
+                    vector<float> quant_util; // min_range, max_range & scale.
+                    quantization_util::get_min_max_range(node, quant_util);
+                    std::vector<float> scales;
+                    scales.push_back(quant_util[2]);
+                    mkldnn::primitive_attr attr;
+                    attr.set_output_scales(0, scales);
+                    attr.set_int_output_round_mode(mkldnn::round_mode::round_nearest);
```
jbobba: set attr inside build_quantize_reorder.

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -2774,6 +2775,42 @@ namespace ngraph
...
+                    auto input_data_desc = mkldnn_utils::get_input_mkldnn_md(node, 0);
+                    auto result_desc = mkldnn_utils::get_output_mkldnn_md(node, 0);
+
+                    vector<float> quant_util; // min_range, max_range & scale.
+                    quantization_util::get_min_max_range(node, quant_util);
+                    std::vector<float> scales;
+                    scales.push_back(quant_util[2]);
+                    mkldnn::primitive_attr attr;
+                    attr.set_output_scales(0, scales);
+                    attr.set_int_output_round_mode(mkldnn::round_mode::round_nearest);
```
jbobba: set attr inside build_quantize_reorder

src/ngraph/runtime/cpu/builder/quantize.cpp
```diff
@@ -0,0 +1,73 @@
...
+                    auto& arg_tensor = external_function->get_tensor_data(args[0].get_name());
+                    auto& out_tensor = external_function->get_tensor_data(out[0].get_name());
+                    auto& out1_tensor = external_function->get_tensor_data(out[1].get_name());
+                    auto& out2_tensor = external_function->get_tensor_data(out[2].get_name());
+                    auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
+                    auto input_desc = mkldnn_utils::get_input_mkldnn_md(node, 0);
+                    auto result_desc = mkldnn_utils::get_output_mkldnn_md(node, 0);
+
+                    vector<float> quant_util; // min_range, max_range & scale.
+                    quantization_util::get_min_max_range(node, quant_util);
```
jbobba: Can we do this at op definition time inside the constructor instead of in emitter and builder? We can compute and save the adjusted ranges and scales inside the quantize op.

src/ngraph/runtime/cpu/quantization_util.hpp
```diff
@@ -0,0 +1,81 @@
...
+                    max_range = std::max(0.0f, max_range);
+                    // end code copied and pasted from
+                    // github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/quantize_op.cc
+                    const float max_abs = std::max(std::abs(min_range), std::abs(max_range));
+                    bool is_signed = (quantize->get_quantize_et()).is_signed();
+                    const float target_range =
+                        static_cast<float>((is_signed ? std::pow(2, 7) : std::pow(2, 8)) - 1);
+                    max_range = max_abs;
+                    min_range = is_signed ? -max_abs : 0;
+                    const float scale = target_range / max_abs;
```
adstraw: I am not convinced that a quantize API consisting of min/max/type is sufficient.  Take a look at the TF quantize op here:

https://www.tensorflow.org/api_docs/python/tf/quantize

There are three ways to calculate 'scale' from min_range and max_range corresponding to the three 'mode' attributes in tf.quantize (MIN_COMBINED, MIN_FIRST, SCALED).

How can this single method of computing scale map to three methods of computing scale at the framework level?  NOTE:  There are three modes *just* within TF, not to mention other frameworks.

I am thinking about the design of the nGraph core quantize (and dequantize) op.  My initial guess was to force the framework to provide the 'scale' in the quantize API.

nbpatel: We'd have to use Eigen for the other modes.

diyessi: Add scale and zero-point args. I think with all four, a bridge can deal with all the modes.
So Q = max(R-min, max-min)/scale+zero_point

nbpatel: @diyessi does it have to go in this PR or we'd be adding it for the ngraph quantize op?

diyessi: We can add it later.

PR #1579, opened 2018-09-09
IntelGPU backend: BatchNorm operation optimization

shssf:
It provide ~2x performance gain with inference run with ResNet50.
Currently the kernel looks like:
```
__attribute__((reqd_work_group_size(1,1,1)))
__kernel void batch_norm_BatchNorm_461_0(const __global float *input0, const __global float *input1, const __global float *input2, const __global float *input3, const __global float *input4, __global float *output)
{
    // input array dims: input0[16][2048][7][7]
    const uint i1 = get_global_id(1); /* channel_axis trip count 2048*/
    const float    gamma = input1[i1];
    const float     beta = input2[i1];
    const float     mean = input3[i1];
    const float variance = input4[i1];
    const float var_sqrt = (gamma / sqrt(variance + 1.001e-05));
    const uint i0 = get_global_id(0); /* batch axis trip count 16*/
    const uint idx0 = (i0 * 100352) + (i1 * 49);
    for (uint i3 = 0; i3 < 49; ++i3)
    {
        const uint idx = idx0 + i3;
        output[idx] = (input0[idx] - mean) * var_sqrt + beta;
    }
}
```

### DIFF
--- a/src/ngraph/runtime/intelgpu/intelgpu_op_batchnorm.cpp
+++ b/src/ngraph/runtime/intelgpu/intelgpu_op_batchnorm.cpp
@@ -48,6 +48,11 @@ static Shape get_channel_shape(const Shape& shape, const string& function_name)
     return {shape.at(channel_axis)};
 }
 
+static size_t get_idx_size(const Shape& shape, size_t pos)
+{
+    return accumulate(shape.cbegin() + pos, shape.cend(), 1, multiplies<size_t>());
+}
+
 void runtime::intelgpu::do_create_mean(cldnn::topology& topology,
                                        const string& output_name,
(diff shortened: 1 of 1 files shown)


src/ngraph/runtime/intelgpu/intelgpu_op_batchnorm.cpp
```diff
@@ -197,6 +197,11 @@ void runtime::intelgpu::do_create_variance(cldnn::topology& topology,
     topology.add(op_variance);
 }
 
+static size_t get_idx_size(const Shape& shape, size_t pos)
```
dmyershov: Could you please move this function into static functions block?

PR #1583, opened 2018-09-10
Remove heap allocations during execution

rkimballn1:
Using std::vector to create a temporary array of points for invoke_primitive requires heap allocations. Changed to create simple temporary arrays. I believe it is easier to read too.
What was
```
    // Add_3(Parameter_0_0, Parameter_1_0, Add_3_0)
    {
        gpu::invoke_primitive(ctx, 0, std::vector<void*>{((float*)(inputs[0])), ((float*)(inputs[1]))}.data(), std::vector<void*>{((float*)((char *)pool_base_ptr + 0))}.data());
    }
```
becomes
```
    // Add_3(Parameter_0_0, Parameter_1_0, Add_3_0)
    {
        void* input[] = {((float*)(inputs[0])), ((float*)(inputs[1]))};
        void* output[] = {((float*)(pool_base_ptr + 0))};
        gpu::invoke_primitive(ctx, 0, input, output);
    }
```


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -125,11 +125,10 @@ namespace ngraph
...
 
-                    writer << "gpu::invoke_primitive(ctx, " << index << ", ";
-                    writer << "std::vector<void*>{" << args[0].get_name() << ","
-                           << args[1].get_name() << "}.data(), ";
-                    writer << "std::vector<void*>{" << out[0].get_name() << "}.data()";
-                    writer << ");\n";
+                    writer << "void* input[] = {" << node_names(args) << "};\n";
+                    writer << "void* output[] = {" << node_names(out) << "};\n";
+                    writer << "gpu::invoke_primitive(ctx, " << index << ", input, output);\n";
+                    cout << writer.get_code();
```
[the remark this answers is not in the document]
rkimballn1: Oops!

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -458,17 +458,18 @@ void runtime::gpu::GPU_ExternalFunction::emit_temp_mem_pool_allocation(
     {
         m_writer << "// Allocate the memory pool\n";
         // TODO memory pool malloc.
-        m_writer << "void* pool_base_ptr = ngraph::runtime::gpu::invoke_memory_primitive(ctx, "
-                 << m_tensor_memory_buffers->at(current_function->get_name()) << ");\n";
+        m_writer
+            << "char* pool_base_ptr = (char*)ngraph::runtime::gpu::invoke_memory_primitive(ctx, "
```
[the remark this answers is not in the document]
rkimballn1: Because we cast it to char* everywhere because you can't do pointer math on a void*. If it is defined as char* then we can remove the cast from everywhere else.

src/ngraph/runtime/gpu/gpu_emitter.hpp
```diff
@@ -87,18 +87,24 @@ namespace ngraph
...
+                        writer << "void* input[] = {" << node_names(args) << "};\n";
+                        writer << "void* output[] = {" << node_names(out) << "};\n";
+                        writer << "gpu::invoke_primitive(ctx, " << ew_index
+                               << ", input, output);\n";
                     }
                     writer.block_end();
                 }
+
+            private:
+                static std::string node_names(const std::vector<GPU_TensorViewWrapper>& args)
```
fengleitian: cool

src/ngraph/runtime/gpu/gpu_emitter.cpp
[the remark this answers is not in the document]
csullivan: Yeah, it's the cuDNN impl. of op::Add. Right now we are using our CUDA impl. instead `GPU_Emitter::emit_elementwise<op::Add>` as @fengleitian found it had better performance for the LSTM graphs.

src/ngraph/runtime/gpu/gpu_emitter.hpp
csullivan: Being able to use node_names when the order is different would be nice.

Can we add an additional argument that specifies the order of the args? It should default to the order of args in the vector of tensor views, but some of our ops rearrange the tensor order in the invocation. 

e.g.
```
writer << "void* input[] = {" << node_names(args, {2,0,1}) << "};\n";

```

rkimballn1: Do you also want to skip certain values? like `{2, -1, 1}` would only yield `{arg[2], arg[1]}`

rkimballn1: or `{2, 1}'

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1512,30 +1470,29 @@ namespace ngraph
...
+                                                             out[0].get_type(),
+                                                             CUDNNEmitter::Prop::Forward,
+                                                             args[0].get_shape());
                     }
 
-                    writer << "gpu::invoke_primitive(ctx, " << softmax_index << ", ";
-                    writer << "std::vector<void*>{" << args[0].get_name() << "}.data(), ";
-                    writer << "std::vector<void*>{" << out[0].get_name() << "}.data()";
-                    writer << ");\n";
+                    writer << "void* input[] = {" << node_names(args, {0, 0}) << "};\n";
```
fengleitian: should be just `node_names(args)` here?

src/ngraph/runtime/gpu/gpu_emitter.hpp
fengleitian: I think this is good, but I don't think it's necessary. It's easy to handle in the function you pass.

fengleitian: I see what you mean. looks like easier to handle in gnu_emitter in this case.

src/ngraph/runtime/gpu/gpu_emitter.cpp
csullivan: Yeah, for softmax, I agree this should just be `node_names(args)`

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1387,15 +1350,15 @@ namespace ngraph
                         }
                     }
```
csullivan: Not necessary, but convenient: could you delete lines 1344 - 1351, which are no longer used?


```
                     int num_nontrivial_dims = 0;
                     for (int64_t i = input_shape.size() - 1; i > 1; i--)
                     {
                         if (input_shape[i] > 1)
                         {
                             num_nontrivial_dims++;
                         }
                     }
```

PR #1586, opened 2018-09-11
[ONNX] Shape operator

tsocha:


src/ngraph/frontend/onnx_import/op/shape.hpp
```diff
@@ -0,0 +1,49 @@
...
+// limitations under the License.
+//*****************************************************************************
+
+#pragma once
+
+#include <memory>
+
+#include "ngraph/node.hpp"
+#include "ngraph/node_vector.hpp"
+#include "ngraph/shape.hpp"
```
postrational: Please split the implementation into a `.cpp` file, so we don't add too many `includes`

tsocha: There is no need to split file only because of includes because they have `#pragma once` inside.

postrational: There is a reason to split the `#includes` out to avoid unnecessarily polluting the `ops_bridge` compilation unit's space with all includes.

PR #1590, opened 2018-09-11
Use validated call in the quantization unit test and the correct shape

jbobba:


test/quantize_cpu.cpp
```diff
@@ -46,10 +46,10 @@ void DequantizeTest(int input, float min, float max, float expected_output)
     auto f = make_shared<Function>(r, op::ParameterVector{A});
     auto backend = runtime::Backend::create("CPU");
     // Create some tensors for input/output
-    auto a = backend->create_tensor(element::from<T>(), Shape{});
+    auto a = backend->create_tensor(element::from<T>(), Shape{1});
```
nbpatel: why shape{1}?

diyessi: `shape_a` has shape {1}. Could also make `shape_a` have shape {} and leave the others go I think.

nbpatel: yes, we can make shape_a have shape{} and use shape_a everywhere.  my bad that the test still has shape_a as Shape{1}

diyessi: There's an upcoming PR for making sure layouts assigned to a tensor agree with the shape of the tensor, and it caught this test.

jbobba: Shape{1} is needed to get things into MKLDNN. Currently MKLDNN doesn't like zero-dim tensors.

nbpatel: ah ok ya, I hit the same issue in pooling as well, they do support it in the newer release though

PR #1600, opened 2018-09-12
add distributed class

sfvaroglu:


src/ngraph/distributed.hpp
```diff
@@ -0,0 +1,42 @@
...
+* limitations under the License.
+*******************************************************************************/
+
+#pragma once
+
+#ifdef IN_NGRAPH_LIBRARY
+#error("distributed.hpp is for external use only")
+#endif
+
+#include <mpi.h>
```
wenzhe-nrv: do we need distributed flag here during compile?

sfvaroglu: Not really. This file is used externally.

wenzhe-nrv: Sorry I don't understand. Could you please elaborate how it's used, e.g. for MXNet?

sfvaroglu: The bridge will `#include` this header file, not nGraph. So, it depends on what distributed flag the bridge uses. If MXNet uses `MXNET_NGRAPH_DISTRIBUTED_ENABLE` and TF uses `NGRAPH_TF_DISTRIBUTED`, then having `#ifdef NGRAPH_DISTRIBUTED #include <mpi.h> #endif` in this file will not work, since the bridge doesn't know about `NGRAPH_DISTRIBUTED`. Every bridge has its own style for defining these flags, so it's hard to unify.

wenzhe-nrv:
just realized that we need to modify the dist mnist example and update doc accordingly: https://github.com/NervanaSystems/ngraph/blob/master/doc/examples/mnist_mlp/dist_mnist_mlp.cpp
https://github.com/NervanaSystems/ngraph/blob/master/doc/sphinx/source/howto/distribute-train.rst

sfvaroglu:
Closing. Opened #1619.

PR #1606, opened 2018-09-13
[WIP] GenerateMask

Krovatkin:


src/ngraph/op/activate.hpp
```diff
@@ -0,0 +1,63 @@
...
+        public:
+            Activate(const std::shared_ptr<State>& state)
+                : Op("Activate", {})
+                , m_state(state)
+            {
+                constructor_validate_and_infer_types();
+            }
+
+            Activate(const std::shared_ptr<Node> arg, const std::shared_ptr<State>& state)
+                : Op("Activate", {check_single_output_args({arg})})
```
diyessi: Why {} around `check_single_output_args`?

src/ngraph/op/activate.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <memory>
+#include "ngraph/op/constant.hpp"
+#include "ngraph/op/op.hpp"
+#include "ngraph/state/rng_state.hpp"
+namespace ngraph
+{
+    namespace op
+    {
+        /// \brief Activate
```
diyessi: This is a little too brief.

src/ngraph/state/rng_state.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "except.hpp"
+#include "rng_state.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+void ngraph::RNGState::activate()
+{
+    //reentrant?
+    if (!is_active())
```
diyessi: Either throw an error if it's already active, or keep a counter for how active it is so that deactivate doesn't happen to early. Probably throwing an error is correct for now.

src/ngraph/op/activate.hpp
```diff
@@ -0,0 +1,63 @@
...
+        ///
+        class Activate : public op::Op
```
diyessi: I think `ActivateState` would be better -- `Activate` is too generic a name.

src/ngraph/op/generate_mask.hpp
```diff
@@ -0,0 +1,81 @@
...
+        /// \brief GenerateMask
+        ///
+        class GenerateMask : public op::Op
+        {
+        public:
+            GenerateMask(const std::shared_ptr<Node>& training,
+                         const std::shared_ptr<Node>& activate,
+                         const Shape& shape,
+                         const element::Type& element_type,
+                         double prob,
```
diyessi: Frameworks are very inconsistent about whether this is the keep or the drop probability, so we should call it `keep_probability` or `dropout_probability` to be clear.

src/ngraph/runtime/reference/generate_mask.hpp
```diff
@@ -0,0 +1,41 @@
...
+            template <typename T>
+            void generate_mask(
+                T* out, size_t count, unsigned int seed, double probability, bool training)
+            {
+                std::mt19937 gen(seed);
+                std::bernoulli_distribution bd(probability);
+
+                for (size_t i = 0; i < count; i++)
+                {
+                    out[i] = training ? static_cast<T>(bd(gen)) : static_cast<T>(1);
```
diyessi: You can pre-multiply the mask by 1/prob.

Krovatkin:
Closing in favour of https://github.com/NervanaSystems/ngraph/pull/1942

PR #1609, opened 2018-09-14
- fixed performance_counter for CPU DEX

pruthviIntel:


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1489,12 +1489,16 @@ void runtime::cpu::CPU_ExternalFunction::build()
                 {
                     for (size_t j = 0; j < p.second; j++)
                     {
-                        if (runtime::cpu::IsTracingEnabled())
+                        // we will only update the profile_counter for the first functor within
```
jmenon: This is incorrect (thought I documented it but guess not). There needs to be _exactly_ one functor per op. If you have multiple computation stages then you'll need to build one functor with those stages. The matmul + bias builder has an example. The maxpoolbackprop builder should be fixed.

jbobba:
@pruthviIntel please make the changes in the TBB executor as well.

pruthviIntel:
@jbobba done

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1396,45 +1399,33 @@ void runtime::cpu::CPU_ExternalFunction::build()
...
-                    }
-
                     tbb::flow::continue_node<tbb::flow::continue_msg, tbb::flow::lightweight>*
                         flowgraph_node = new tbb::flow::continue_node<tbb::flow::continue_msg,
                                                                       tbb::flow::lightweight>(
-                            *(ctx->G), [&, ftrs](const tbb::flow::continue_msg& msg) {
+                            *(ctx->G), [&](const tbb::flow::continue_msg& msg) {
                                 if (p.first(ctx) || ctx->first_iteration)
                                 {
-                                    for (size_t j = 0; j < p.second; j++)
```
jmenon: Is the mapped value needed elsewhere? If not, it doesn't need to be calculated and stored

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1352,10 +1351,13 @@ void runtime::cpu::CPU_ExternalFunction::build()
             };
         }
 
-        enables.emplace_back(make_pair(enable, functors.size() - functor_count));
+        enables.emplace_back(make_pair(enable, 1));
```
jmenon: I was more implying that we could remove it altogether (or use some other container) since it's not very useful to store a '1' there, right?

pruthviIntel:
@jmenon address the feedback

PR #1618, opened 2018-09-14
Change GPU backend to use op_tbl

rkimballn1:
Yes, this PR looks really scary, but it is not all that bad. It may be easier to just look at the source files rather than attempting to wade through the diffs since so many lines changed but here is the TL;DR
The OpMap dispatcher table is replaces with a switch statement. Why a switch statement? Because with the switch we can check that *every* op is handled. If a new op is added to ngraph core (ngraph/op/op_tbl.hpp) then the build will fail if that new op is not addressed in the GPU backend. Addressing the new op can be as simple as throwing an exception, but you must do that or it won't compile. This new check found 5 ops that were nowhere to be found in the GPU backend.
Another benefit is that the templated functions of the emitter are no longer templated. The templates were an attempt to address the missing op issue but did not work. With no templated functions the cpp file is indented 12 less spaces for everything.


src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -158,91 +164,461 @@ static string emit_string_array(const vector<string>& s, size_t max_line_length)
...
+            external_function, writer, node, args, out);
+        break;
+    }
+    case OP_TYPEID::Log:
+    {
+        runtime::gpu::GPU_Emitter::emit_elementwise<ngraph::op::Log>(
+            external_function, writer, node, args, out);
+        break;
+    }
+    case OP_TYPEID::LRN: { throw unsupported_op("Unsupported op '" + node->description() + "'");
```
fengleitian: is this by clang, or should we have a new line for `{`

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -158,91 +164,461 @@ static string emit_string_array(const vector<string>& s, size_t max_line_length)
...
+            external_function, writer, node, args, out);
+        break;
+    }
+    case OP_TYPEID::Tanh:
+    {
+        runtime::gpu::GPU_Emitter::emit_elementwise<ngraph::op::Tanh>(
+            external_function, writer, node, args, out);
+        break;
+    }
+    case OP_TYPEID::TopK: { throw unsupported_op("Unsupported op '" + node->description() + "'");
```
fengleitian: same here.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -36,6 +36,8 @@
 #include "ngraph/op/add.hpp"
 #include "ngraph/op/allreduce.hpp"
 #include "ngraph/op/and.hpp"
+#include "ngraph/op/argmax.hpp"
```
shssf: Maybe it is better to use the same auto generation procedure for all headers related to particular operation?

rkimballn1: I did think about that but the class names are CamelCase and the header files are snake_case and conversion inside a macro is difficult/impossible and I was not that interested in trying.

rkimballn1:
I agree that there is no explicit contract and I can't think of a way to add one, but were someone to leave a new op out of op_tbl then they can't serialize/deserialize the op and the interpreter won't support the op. The lack interpreter support can be hidden in the test manifest or the op can be added with no unit tests. 
The more I think about it, we can add a check so that if someone tries to *use* an op that is not in op_tbl then we could throw an exception. Not great but any unit test for the new op would fail.
I don't see why it is only a half-measure. For the GPU you *must* implement a function `emit_NewOp` or compile fails. Yes, the new op can be empty, but we can't stop people from doing that short of a PR review, which we should be doing.
The macro use is light weight in my opinion.

PR #1619, opened 2018-09-14
refactor distributed code

sfvaroglu:


test/main.cpp
```diff
@@ -23,24 +23,16 @@
 using namespace std;
 
 #ifdef NGRAPH_DISTRIBUTED
-#include <mpi.h>
-
+#include "ngraph/distributed.hpp"
 class MpiEnvironment : public ::testing::Environment
```
wenzhe-nrv: is this class still needed? I wonder if ln52 could be replaced by `ngraph::Distributed dist;`?
if we need to keep this, please rename the class to `DistributedEnvironment`.

PR #1620, opened 2018-09-14
Add support for Quantized Convolution op via mkldnn for IA backend (codegen + DEX)

nbpatel:


src/ngraph/runtime/cpu/quantization_util.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace cpu
+        {
+            namespace quantization_util
+            {
+                template <class T1, class T2, class T3>
+                void QuantizationRangeForMultiplication(
```
jbobba: quantization_range_for_multiplication

src/ngraph/runtime/cpu/quantization_util.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace cpu
+        {
+            namespace quantization_util
+            {
+                template <class T1, class T2, class T3>
```
jbobba: Add TF copyright

nbpatel: The quantize PR will go in before this, so its a merging thing.

PR #1622, opened 2018-09-14
Add info for cpu ops in VisualizeTree

Krovatkin:


src/ngraph/pass/manager.hpp
```diff
@@ -54,6 +54,14 @@ class ngraph::pass::Manager
         }
     }
 
+    void set_ops_to_details(
+        std::unordered_map<std::type_index,
+                           std::function<void(std::shared_ptr<Node>, std::stringstream& ss)>>*
```
rkimballn1: If you just made this `Node*` instead of `std::shared_ptr<Node>` it would be smaller and clearer. Or maybe a Node&
Also use and ostream& just to be a little more generic.

rkimballn1: Would this look better from a user point of view if it were an initializer list?

src/ngraph/pass/visualize_tree.cpp
```diff
@@ -101,14 +103,25 @@ std::string pass::VisualizeTree::get_attributes(shared_ptr<Node> node)
...
                                                       : vector_to_string(node->get_shape()));
     }
 
+    if (m_ops_to_details)
+    {
+        const Node& n = *node;
+        auto eh = m_ops_to_details->find(TI(n));
+        if (eh != m_ops_to_details->end())
+        {
+            eh->second(node, ss);
```
rkimballn1: Here you can pass in n instead of node if you use const Node& in signature.

src/ngraph/pass/manager.hpp
```diff
@@ -65,4 +73,7 @@ class ngraph::pass::Manager
     ManagerState m_state;
     bool m_visualize = false;
     bool m_serialize = false;
+    std::unordered_map<std::type_index,
```
rkimballn1: I wonder if I should just get rid of ManagerState if nobody is going to use it. You could put the ops_to_detail map in there.

src/ngraph/pass/visualize_tree.hpp
```diff
@@ -37,6 +41,13 @@ class ngraph::pass::VisualizeTree : public ModulePass
     bool run_on_module(std::vector<std::shared_ptr<ngraph::Function>>&) override;
 
     static std::string get_file_ext();
+    void set_ops_to_details(
+        std::unordered_map<std::type_index,
```
rkimballn1: As much as I hate `using ops_to_detail_t = std::unordered_map<stuff>` I think I might hate seeing this line repeated over and over.

PR #1626, opened 2018-09-17
IntelGPU backend. Change Intel GPU backend to use op_tbl

shssf:


shssf:
In case of the question about std::map vs std::unordered_map used here.
I measured performance of the *IntelGPUBackend::compile()* function with *mxnet-ngraph-Function_6-CompileForward-fprop.json* graph with 488407 operations. I measured only the backend function, not all SW stack that belongs to it.
```
std::map:          16.93s | 16.79s | 16.75s
std::unordered_map 16.99s | 16.89s | 16.95s
```
It looks *std::map* slightly faster.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -230,7 +260,9 @@ bool runtime::intelgpu::IntelGPUBackend::compile(shared_ptr<Function> func)
 
     for (shared_ptr<Node> op : func->get_ops())
     {
-        if ("Parameter" == op->description())
+        switch (get_typeid(op->description()))
```
rkimballn1: You should add the gcc pragmas
```
#pragma GCC diagnostic push
#pragma GCC diagnostic error "-Wswitch"
#pragma GCC diagnostic error "-Wswitch-enum"
```
So that all cases are required. For unsupported ops you would simply throw an unsupported_op exception inside the case. No default: is need for this as all cases are covered.

PR #1641, opened 2018-09-19
Add Quantized conv+relu

nbpatel:


src/ngraph/runtime/cpu/builder/quantized_conv.cpp
```diff
@@ -19,6 +19,7 @@
 #include "ngraph/runtime/cpu/cpu_builder.hpp"
 #include "ngraph/runtime/cpu/mkldnn_invoke.hpp"
 #include "ngraph/runtime/cpu/mkldnn_utils.hpp"
+#include "ngraph/runtime/cpu/op/quantized_conv_relu.hpp"
```
aprocter: I think this header is out of order.

PR #1645, opened 2018-09-20
[ONNX] LRN op

tsocha:


src/ngraph/frontend/onnx_import/op/lrn.cpp
```diff
@@ -0,0 +1,46 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include <memory>
+
+#include "ngraph/op/lrn.hpp"
+
+#include "core/node.hpp"
+
```
arogowie-intel: We've agreed to keep _onnx importer_ header files in one group. Please conform.

PR #1652, opened 2018-09-20
Update clang/llvm to 5.0.2 and add macOS prebuild binary support.

silee2:
macOS with prebuilt binary passes all unit tests.


silee2:
I checked the LLVM code on LLVM github repo and the only code change on the LLVM side is just changing the version string to 5.0.2 from 5.0.1. clang and openmp is still the same SHA.

silee2:
Shouldn't have any regression as the functional code for LLVM/Clang is exactly the same as 5.0.1

PR #1661, opened 2018-09-21
add nGraph quantize op

adstraw:


adstraw:
This should be ready to review.  Last change was to use a quantized offset value which I know @diyessi may not agree with.  I don't plan to make any other changes except to fix build errors and address code feedback.

src/ngraph/coordinate.hpp
```diff
@@ -75,13 +75,19 @@ namespace ngraph
 
     // Removes some values from a vector of axis values
     template <typename AXIS_VALUES>
-    AXIS_VALUES project(const AXIS_VALUES& axis_values, const AxisSet& deleted_axes)
+    AXIS_VALUES project(const AXIS_VALUES& axis_values,
```
diyessi: I'd rather this be two methods, `project` (with the new semantics) and `remove` or as a renaming of the current one. `remove` could just call `project` with the correct arguments.

src/ngraph/op/quantize.cpp
```diff
@@ -0,0 +1,100 @@
...
+        OFFSET
+    };
+
+    set_output_size(1);
+    set_output_type(0, m_type, get_input_shape(INPUT));
+
+    // TODO: longer term we probably want quantized types
+    // 1) for type safety - so quantized types are not passed to non-quantized ops
+    // 2) to reflect quantized type min/max which can vary e.g. [-127, 127] for "scaled" int8
+    NODE_VALIDATION_ASSERT(this, (m_type == element::u8 || m_type == element::i8))
```
diyessi: Let's instead add an `is_quant_type` method to element types.

src/ngraph/op/quantize.cpp
```diff
@@ -0,0 +1,100 @@
...
+        << "Output element type (" << m_type << ") must be an 8-bit integer";
+
+    NODE_VALIDATION_ASSERT(this, get_input_element_type(INPUT).is_real())
+        << "Input element type (" << get_input_element_type(INPUT)
+        << ") must be a floating point number";
+
+    NODE_VALIDATION_ASSERT(this, get_input_element_type(SCALE) == get_input_element_type(INPUT))
```
diyessi: I think real might be better here. fp16/bloat16 inputs might need the extra precision for scale/offset.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -831,7 +831,8 @@ static const runtime::cpu::pass::AssignOpMap s_dispatcher{
      &runtime::cpu::pass::CPUAssignment::assign<ngraph::op::QuantizedAvgPool>},
     {TI(ngraph::op::Softmax), &runtime::cpu::pass::CPUAssignment::assign<ngraph::op::Softmax>},
     {TI(ngraph::op::Slice), &runtime::cpu::pass::CPUAssignment::assign<ngraph::op::Slice>},
-    {TI(ngraph::op::Quantize), &runtime::cpu::pass::CPUAssignment::assign<ngraph::op::Quantize>},
+    {TI(ngraph::op::QuantizeCPU),
```
diyessi: `CPUQuantize` ? I think the prefix convention is more common. I defer to @jbobba

src/ngraph/runtime/reference/quantize.hpp
```diff
@@ -0,0 +1,64 @@
...
+                {
+                    Coordinate scale_offset_coord = project(input_coord, axes, false);
+
+                    // apply scale and offset
+                    REAL qvalue =
+                        std::round(input[input_transform.index(input_coord)] /
+                                   scale[scale_offset_transform.index(scale_offset_coord)]) +
+                        offset[scale_offset_transform.index(scale_offset_coord)];
+
+                    // clamp
```
diyessi: Clamp t he reals-- you can get overflow this way.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
adstraw: I think the idea is that we would delete this operator and @nbpatel would rework his kernels to use the nGraph (vs. CPU) Quantize op.  So, given this is temporary maybe it's not such a big deal.  However, I can rename to what @jbobba and @nbpatel decide.

src/ngraph/op/quantize.cpp
```diff
@@ -0,0 +1,100 @@
...
+        << "Scale element type (" << get_input_element_type(SCALE)
+        << ") must match input element type (" << get_input_element_type(INPUT) << ")";
+
+    NODE_VALIDATION_ASSERT(this, get_input_element_type(OFFSET) == m_type)
+        << "Offset element type (" << get_input_element_type(OFFSET)
+        << ") must match input element type (" << m_type << ")";
```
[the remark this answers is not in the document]
diyessi: I don't think that will work, although I think it might work to use double the value, assuming we allow each of min/max to fall either on the edge or center of scale region and nothing else.

src/ngraph/op/quantize.hpp
```diff
@@ -0,0 +1,60 @@
...
+        {
+        public:
+            enum class RoundMode
+            {
+                HALF_AWAY_FROM_ZERO,
+                HALF_TO_EVEN
+            };
+
+            Quantize(std::shared_ptr<Node> input,
+                     std::shared_ptr<Node> scale,
```
nbpatel: With the scale, offset design is there a way the backend would know which MODE is used to calculate the scale. The reason I am asking is because we should be able to distinguish between SCALED/MIN_COMBINED/MIN_FIRST mode for the transformer to make a decision on calling mkldnn or eigen

src/ngraph/op/quantize.hpp
```diff
@@ -0,0 +1,60 @@
...
+namespace ngraph
+{
+    namespace op
+    {
+        class Quantize : public ngraph::op::Op
+        {
+        public:
+            enum class RoundMode
+            {
+                HALF_AWAY_FROM_ZERO,
```
nbpatel: The mkldnn supports round nearest mode, not sure which one of these corresponds to that

src/ngraph/runtime/cpu/builder/quantize.cpp
```diff
@@ -29,11 +29,11 @@ namespace ngraph
         namespace cpu
         {
             template <>
-            void Builder::BUILDER_DECL(ngraph::op::Quantize)
+            void Builder::BUILDER_DECL(ngraph::op::QuantizeCPU)
             {
                 if (runtime::cpu::mkldnn_utils::use_mkldnn_kernel(node))
                 {
-                    auto quantize = static_cast<const ngraph::op::Quantize*>(node);
+                    auto quantize = static_cast<const ngraph::op::QuantizeCPU*>(node);
```
nbpatel: I dont see a reason why will be needing this op once we have ngraph ops ready. But IMO for time being we can have them like this. Maybe @jbobba can give us some more details on how to go about with this.

src/ngraph/op/quantize.hpp
adstraw: I designed this API specifically so that nGraph would not have to know about MODE.  The modes you mentioned come from TensorFlow, but there may me more as we expand to other frameworks.  We need to have a discussion on how to integrate MKLDNN kernels into this schema.  I talked with @jbobba and he indicated it was possible though I am short on the exact details.

src/ngraph/runtime/cpu/builder/quantize.cpp
adstraw: Yes, let's take a future action to remove QuantizeCPU and use Quantize for the CPU backend.

src/ngraph/op/quantize.hpp
adstraw: Offline discussion:  MKLDNN cares about having the offset=0.  Hence we are covered.

PR #1663, opened 2018-09-21
zero dim elem fix

Krovatkin:


src/ngraph/pass/zero_dim_tensor_elimination.cpp
```diff
@@ -37,7 +37,16 @@ static bool has_zero_dim(std::shared_ptr<Node> node)
     {
         throw ngraph_error("has_zero_dim is called on multi-output op");
     }
-    return shape_size(node->get_shape()) == 0;
+
```
diyessi: `find` would be shorter.

PR #1664, opened 2018-09-21
Add Quantized conv+relu

nbpatel:


src/ngraph/runtime/cpu/quantization_util.hpp
```diff
@@ -89,7 +89,32 @@ namespace ngraph
                     quant_util.push_back(scale);
                 }
 
-                float get_scale(const ngraph::Node* node);
```
pruthviIntel: can you add some comments or refer to link regarding this scale/offset min/max calculation?. It will be useful to understand the difference with TF/Mxnet choice of doing v/s the ngraph.

src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -198,7 +205,24 @@ namespace ngraph
                     }
                     else if (std::is_same<OP, ngraph::op::QuantizedConvolution>())
```
gaurides: You can use 1 if statement to check for both OP types? and pass `OP` to get_scale()?

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
```diff
@@ -67,7 +68,49 @@ namespace ngraph
...
+                    auto& out1_tensor = external_function->get_tensor_data(out[1].get_name());
+                    auto& out2_tensor = external_function->get_tensor_data(out[2].get_name());
+
+                    auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
+
+                    auto conv_index =
+                        mkldnn_emitter->build_convolution<ngraph::op::QuantizedConvolutionRelu>(
+                            node, args, out);
+                    auto& deps = mkldnn_emitter->get_primitive_deps(conv_index);
+                    float min_freezed_output = qconvolution_relu->get_freezed_output_min();
```
gaurides: trying to understand, why is it called `freezed_output_min()` and `freezed_output_max()` and not just output_min/max?

test/quantize_cpu.cpp
```diff
@@ -331,3 +332,100 @@ TEST(quantize_cpu, quantize_to_int8)
     EXPECT_EQ((vector<float>{-127}), read_vector<float>(result_min));
     EXPECT_EQ((vector<float>{127}), read_vector<float>(result_max));
 }
+
+TEST(quantize_cpu, quantizedConv2D_with_relu)
```
gaurides: trying to understand, what's the difference with `quantizedConv2D_fused_relu` apart from Shape and values?

nbpatel: None. I was just out of names :)

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
nbpatel: Just because we do some graph transforms including freezing the graph and extracting values logged during training.

src/ngraph/runtime/cpu/mkldnn_emitter.hpp
nbpatel: I tried that. But for some reason the templates try to use all possible values of OP and hence it crashes. Maybe a template within template thing.

gaurides: ok. no worries.

PR #1671, opened 2018-09-23
Reshape with no permutation optimized out

shssf:
This optimization pass should remove case where no data changes needed in `Reshape` operation.
For example, `Shape{}` to `Shape{1}`

### DIFF
--- a/src/ngraph/pass/nop_elimination.cpp
+++ b/src/ngraph/pass/nop_elimination.cpp
@@ -24,6 +24,7 @@
 #include "ngraph/op/constant.hpp"
 #include "ngraph/op/convert.hpp"
 #include "ngraph/op/pad.hpp"
+#include "ngraph/op/reshape.hpp"
 #include "ngraph/op/slice.hpp"
 #include "ngraph/op/stop_gradient.hpp"
 #include "ngraph/op/sum.hpp"
@@ -101,6 +102,17 @@ HANDLER_DECL(eliminate_broadcast)
     return false;
 }
 
+HANDLER_DECL(eliminate_reshape)
+{
+    auto reshape = std::dynamic_pointer_cast<ngraph::op::Reshape>(node);
+    if (!reshape->get_is_transpose())
+    {
+        ngraph::replace_node(node, node->get_argument(0));
+        return true;
+    }
+    return false;
+}
+
 HANDLER_DECL(eliminate_stop_gradient)
 {
     ngraph::replace_node(node, node->get_argument(0));
@@ -113,6 +125,7 @@ static const std::unordered_map<std::type_index,
                {TI(ngraph::op::Sum), &eliminate_sum},
                {TI(ngraph::op::Convert), &eliminate_convert},
                {TI(ngraph::op::Slice), &eliminate_slice},
+               {TI(ngraph::op::Reshape), &eliminate_reshape},
                {TI(ngraph::op::StopGradient), &eliminate_stop_gradient},
                {TI(ngraph::op::BroadcastLike), &replace_broadcast_like},
                {TI(ngraph::op::Broadcast), &eliminate_broadcast}};


shssf:
I see strange things in test system on my machine.
If I ran all CPU tests by `./test/unit-test --gtest_filter=CPU.*` it goes well for some time. But from 50 or 60 tests it shows all other tests failed with:
```
[ RUN      ] CPU.pad_interior_exterior_2d
unknown file: Failure
C++ exception with description "function failed to compile" thrown in the test body.
[  FAILED  ] CPU.pad_interior_exterior_2d (112 ms)
[ RUN      ] CPU.pad_exterior_2d_0x0
unknown file: Failure
C++ exception with description "function failed to compile" thrown in the test body.
[  FAILED  ] CPU.pad_exterior_2d_0x0 (109 ms)
[ RUN      ] CPU.pad_exterior_2d_0x3
unknown file: Failure
C++ exception with description "function failed to compile" thrown in the test body.
[  FAILED  ] CPU.pad_exterior_2d_0x3 (122 ms)
[ RUN      ] CPU.pad_exterior_2d_3x0
^C
```
if execute particular test from failed tests set:
```
shalnov@nnwsshalnov11:/home/work/nGraph/0.build_gpu$ ./test/unit-test --gtest_filter=CPU.pad_exterior_2d_0x3
Note: Google Test filter = CPU.pad_exterior_2d_0x3
[==========] Running 1 test from 1 test case.
[----------] 1 test from CPU
[ RUN      ] CPU.pad_exterior_2d_0x3
[       OK ] CPU.pad_exterior_2d_0x3 (1905 ms)
[----------] 1 test from CPU (1905 ms total)

[==========] 1 test from 1 test case ran. (1905 ms total)
[  PASSED  ] 1 test.
```

aprocter:
I don't think this will work in its current form.

Consider something like this (apologies for pseudocode):

```
p0 is a parameter of shape [16]
p1 is a parameter of shape [4,2]
t0 := non_transposing_reshape(p0,new_shape=[4,4])
t1 := concat([t0,p1],axis=1)
```

If we eliminate the non-transposing reshape here, `t1` is now trying to concatenate `p0` and `p1` together on axis 1, which doesn't make sense. There's nothing stopping a backend from aliasing p0 and t0's storage here, but even then we still need to keep track of the shape info somehow.

shssf:
@Krovatkin Yes, I changed this PR and now only `Reshape` with the same Shapes eliminated. It is also helps at least in ResNet50 to avoid such operations. Maybe it is a next step to eliminate situation where scalar(`Shape{}`) reshaped to `Shape{1}` because in this case we lost information about Shape changing.

aprocter:
LGTM with that change.

Krovatkin:
@shssf I wonder if we could use this one to optimize this special case? : https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/pass/reshape_elimination.cpp#L40

shssf:
@Krovatkin Hm... I didn't find it in any backend as optimization pass. Perhaps I missed it by this reason. Need time to understand impact of this pass to the models we use. I saw cases where "optimization" brings worse performance in inference mode.

Krovatkin:
@shssf Heh, I do think you missed it exactly for the reason you mentioned: this pass *isn't* being used by any of our backends.  However, it's invoked explicitly by MXNet. We haven't noticed any perf regression due to this pass so far, so, hopefully, it will work for your case, as well. Also, I could add more switches so you could run any one particular transformation you need from this pass (similar to https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/pass/constant_folding.hpp#L49)

shssf:
@Krovatkin Sorry but I see some misunderstanding. This PR add small optimization for all backend because I added this into common pass that used in other backends. I didn't implement some special case optimization for INTELGPU backend.
Thank you for pointing me an optimization I could use for INTELGPU. I will try, and perhaps use it, for the backend.
Thank you very much again.

Krovatkin:
@shssf I think we should make all backends run `ReshapeElimination`. I don't see any immediate downsides to it. It's a very low overhead. CPU and GPU have other optimizations that subsume **some** (but not all) transformations `RE` does, so `RE` is still somewhat relevant. The other option would be to fold `RE` into `NopElimination`. That way we might get nice cascading effects from interactions of several transformations.

shssf:
@Krovatkin Yes, I completely agree. I'm also see that optimizations of the Function needs to be redeveloped in nGraph. It should be splitted, at least, on "common" (which should run in common place) and "backend specific". Also, quality of the optimizations might be improved. But all these things is another story.

PR #1672, opened 2018-09-23
Add lots more fun output to nbench

rkimballn1:
Directory processing now supports all options, so you can dump statistics on a directory.
Added types used to statistics output.
You can now run multiple options, like --statistics and -b at the same time and get output for all.


src/tools/nbench/nbench.cpp
```diff
@@ -269,117 +287,123 @@ OPTIONS
...
+    for (const string& model : models)
     {
-        shared_ptr<Function> f = deserialize(model);
+        cout << "\n============================================================================"
+                "\n";
+        cout << "---- Processing '" << model << "'\n";
+        cout << "============================================================================\n";
+        if (visualize)
+        {
+            shared_ptr<Function> f = deserialize(model);
```
csullivan: Is it desired behavior for nbench to not catch this exception:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Error parsing json at node 'ScalarConstantLike_7958'
Aborted
```

Possible alternative: catch and log the throw, move to the next model and report the error at-exit.

PR #1675, opened 2018-09-24
add DropOut op and stub to Documentation ToC

indie:


indie:
Also adding @crlishka since I think we can debug broken doc search by adding the `python3-sphinx` package to apt-get commands as described in the latest Sphinx docs: http://www.sphinx-doc.org/en/master/usage/installation.html#linux

This is a fairly recent change to their process, might be what is breaking our search.

contrib/docker/Dockerfile.ngraph.centos74
```diff
@@ -23,14 +23,15 @@
 FROM centos:7
 
 # Added install for perl Data::Dumper to avoid a compile error
+# Sphinx docs specify python-sphinx package
```
indie: for CentOS, that is.

doc/sphinx/source/ops/drop_out.rst
```diff
@@ -0,0 +1,70 @@
...
+   DropOut  // DropOut operations
+
+Description
+===========
+
+.. TODO
+
+Inputs
+------
+
```
diyessi: Let's remove the content of these boxes and the mathematical definition so we're not defining.
Maybe we could have a default stub page so we don't accidentally pull in other op doc.

indie: Good idea.  I was thinking it might be easier to write LaTEX kind of math when you have an example there already, but might just be confusing

indie:
Adding @DawnStone

indie:
Adding @aprocter in case changes to checklist are needed

PR #1676, opened 2018-09-24
Move GPU Call Frame into GPU Backend class

rkimballn1:
Move necessary functions from CallFrame to Backend class. Move more info to the FunctionInstance class.


src/ngraph/runtime/gpu/gpu_backend.cpp
```diff
@@ -143,7 +137,26 @@ bool runtime::gpu::GPU_Backend::call(shared_ptr<Function> func,
 
     // ensure the GPURuntimeContext primitive pointers are valid
     m_context->prepare_runtime_context();
-    instance.m_call_frame->call(outputs, inputs, m_context->m_runtime_context.get());
+
+    // Device tensors
+    vector<void*> inputs;
+    vector<void*> outputs;
+
+    for (size_t i = 0; i < input_tvs.size(); i++)
```
csullivan: Since `TensorView` -> `Tensor` soon, perhaps `inputs/outputs_tvs`  should remain `inputs/outputs`?

Now that this is in the backend, can we refactor these below two loops into a static helper that returns a void** of device pointers? Prefer to not have l-value references for the device pointers hanging around.

Optionally: 
```
auto ctx = m_context->m_runtime_context.get();
instance.m_compiled_function(device_buffers(inputs), device_buffers(outputs), ctx);
```

Either that or naming the device pointers as inputs/outputs_d.

rkimballn1: this PR was done as a "minimal change" merge of CallFrame into Backend. I wanted to change as few lines of code possible. When I copied this block of code from CallFrame to Backend there was a naming conflict with inputs/outputs. The smallest change I could make was to change to *_tvs. Lame, but it fit with the minimal change ethic.
I agree with you regarding the refactor, but it is a separate PR. Diff had a hard time detecting changes and marked way too much as changed when it was only copied. With that many changes and substantive changes would have been hidden in the noise.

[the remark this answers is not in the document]
rkimballn1: This code was copied directly from CallFrame with minimal modification. We can change in a separate PR.

PR #1679, opened 2018-09-24
Don't reuse memory for non-destructive ops

csullivan:


src/ngraph/pass/memory_layout.cpp
```diff
@@ -53,10 +53,8 @@ bool pass::MemoryLayout::run_on_function(shared_ptr<ngraph::Function> function)
 
                     // an input tensor can be reused if this is the last use or
                     // an op isn't destructive (i.e. Reshape(DimShuffle))
```
fengleitian: change comments accordingly, why not for non-destructive ops?

Krovatkin: One of indirect users of a reshape(a user of a user) might be an in-place destructive op which destroys the original value before it's consumed by all the direct users of reshape.

fengleitian: if non-destructive passthrough followed by a destructive passthrough, that means we will change the buffer which should be non-destructive

fengleitian: but if it's destructive, it will change if it's last used.

PR #1680, opened 2018-09-25
[WIP][ONNX] Add ENV Flag to compile onnx import

tsocha:


tsocha:
#1687 was chosen as better solution.

PR #1684, opened 2018-09-25
IAT: Skip reshapes that are removing or adding size-1 dimensions

jbobba:


src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -1276,84 +1276,182 @@ namespace ngraph
...
+                            skip_reshape = true;
+                            skip_input_reorder = true;
                         }
                         else
                         {
-                            set_native_layouts(external_function, node);
-                            return;
+                            if (!reshape->get_is_transpose())
+                                skip_reshape = true;
                         }
```
csullivan: Could combine this if with the else.

PR #1686, opened 2018-09-25
Fix DEX_ONLY compile error in cpu_external_function.cpp

aprocter:
Seeing this error with `-DNGRAPH_DEX_ONLY=TRUE`:

```
/Users/amprocte/Work/ngraph/src/ngraph/runtime/cpu/cpu_external_function.cpp:1139:12: error: unknown type name 'StaticInitializers'
    static StaticInitializers s_static_initializers(s_debug_dir);
```

Looks like the definition of this class should be outside the big `#if !defined(NGRAPH_DEX_ONLY)`.

### DIFF
--- a/src/ngraph/runtime/cpu/cpu_external_function.cpp
+++ b/src/ngraph/runtime/cpu/cpu_external_function.cpp
@@ -373,32 +373,6 @@ static void
     writer << "}\n";
 }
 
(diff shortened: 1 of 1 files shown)


aprocter:
Whoops, there's another link error way down at the end I might as well fix while I'm here...

PR #1688, opened 2018-09-26
IntelGPU backend: Switch Select operation to use implementation from …

dmyershov:
…clDNN


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -346,6 +349,8 @@ bool runtime::intelgpu::IntelGPUBackend::compile(shared_ptr<Function> func)
         {
             arguments_check(op, 3, 1);
 
+// Leave it here for some time
```
shssf: If we don't set this line with comment, the style formatter do quite wrong things with sources (see first commit in this PR).

PR #1689, opened 2018-09-26
Address deferred comments from PR 1676

rkimballn1:


src/ngraph/runtime/gpu/gpu_backend.cpp
```diff
@@ -139,24 +159,11 @@ bool runtime::gpu::GPU_Backend::call(shared_ptr<Function> func,
...
-            static_pointer_cast<runtime::gpu::GPU_TensorView>(input_tvs[i]);
-        inputs.push_back(tv->m_allocated_buffer_pool);
-    }
-    for (size_t i = 0; i < output_tvs.size(); i++)
-    {
-        shared_ptr<runtime::gpu::GPU_TensorView> tv =
-            static_pointer_cast<runtime::gpu::GPU_TensorView>(output_tvs[i]);
-        outputs.push_back(tv->m_allocated_buffer_pool);
-    }
+    initialize_io(instance.m_inputs.data(), inputs);
```
csullivan: I guess this just overwrites the cached inputs/outputs every time, correct?

src/ngraph/runtime/gpu/gpu_backend.cpp
```diff
@@ -139,24 +159,11 @@ bool runtime::gpu::GPU_Backend::call(shared_ptr<Function> func,
...
+    initialize_io(instance.m_outputs.data(), outputs);
```
csullivan: Any reason to not pass in the vector (or make this helper a member of FunctionInstance), and just push_back on the vector instead of pre-initializing the vector with resize and passing in the data pointer? The ownership conundrum of having an external method touch the data pointer of a member vector worries me as I'm less certain that something won't happen to the vector that invalidates the pointer.

src/ngraph/runtime/gpu/gpu_backend.cpp
fengleitian: should be in this?
```
if (instance.m_external_function == nullptr)
```

rkimballn1: Yes. The whole point of this is to not have to create new vectors with each call. The vectors are really just used as safe ways to deal with raw memory buffers.

src/ngraph/runtime/gpu/gpu_backend.cpp
rkimballn1: push_back requires that we clear the vector. Don't think of it as a vector, think of it as a `void**` or an array of `void*`. I only use a vector as a holder for memory. I could use malloc/free but that is so old-school. Would you be uncomfortable if I just used raw memory?

csullivan: Admittedly, no. After seeing these changes requested from the previous PR I realize that my hesitation is mostly with the fact that FunctionInstance is behaving as a struct without encapsulation. If it were closed down a bit more, then something like

```
    instance.initialize_io(inputs);
    instance.initialize_io(outputs);
```

would be better. However as many changes are being made to its members (e.g. external_function) soon, it makes sense to wait and decide on this particular point once the larger refactor has been done.

PR #1692, opened 2018-09-26
Partial Shapes, Part 1: Classes for partially known shapes, possibly unknown dimensions

aprocter:


src/ngraph/dimension.hpp
```diff
@@ -0,0 +1,84 @@
...
+        ///
+        ///        Requires that size_t != std::numeric_limits<size_t>::max(). If that condition
+        ///        does not hold, throws std::invalid_argument.
+        Dimension(size_t dimension)
+            : m_dimension(dimension)
+        {
+            if (dimension == s_undetermined_val)
+            {
+                throw std::invalid_argument(
+                    "Cannot convert std::numeric_limits<size_t>::max() to Dimension");
```
diyessi: Use that value of `s_undetermined_val` instead of `std::numeric_limits<size_t>::max()` and then it will always be the correct value. And the message should have a brief why in it.

aprocter:
@rkimballn1 @silee2 FWIW/FYI: it looks like you guys were tagged because you are the code owners for `CMakeLists.txt`.

aprocter:
TESTNOW

aprocter:
TESTNOW

PR #1693, opened 2018-09-26
Check output shape when setting memory layout for slice op.

ayzhuang:


jbobba:
We can close this. More generic way in #1712

PR #1696, opened 2018-09-27
[ONNX] Header files for ONNX_ML

srinivasputta:


src/ngraph/frontend/onnx_import/core/attribute.hpp
```diff
@@ -16,7 +16,11 @@
 
 #pragma once
 
+#ifdef ONNX_ML
```
postrational: * We would prefer to avoid using `#ifdef` if possible. Perhaps we should just switch to using ONNX-ML wholesale?
* `ONNX_ML` does not seem specific enough as an option name.
* We should add the option to a list of available options in `CMakeLists`

srinivasputta: 1. Customers having prebuilt/custom onnx will encounter issues, if we hard-code either of onnx or onnx_ml. They have to use whatever the header filer they have available with.
2 & 3. I was thinking about it and have patch to create ONNX_ML as an option. However my skills on cmake are limited, so I may need your help if it doesn't follow guidelines.

CMakeLists.txt
```diff
@@ -100,6 +100,7 @@ option(NGRAPH_DEX_ONLY "Build CPU DEX without codegen" FALSE)
 if (NGRAPH_ONNX_IMPORT_ENABLE)
     option(NGRAPH_USE_SYSTEM_PROTOBUF "Use system provided Protobuf shared object" FALSE)
     option(NGRAPH_ONNXIFI_ENABLE "Enable ONNX Interface for Framework Integration" TRUE)
+    option(NGRAPH_ONNX_ML "Build ONNX with ONNX_ML enabled" FALSE)
```
rkimballn1: Should be `NGRAPH_ONNX_ML_ENABLE` to match other enable flags

srinivasputta: Done.

arogowie-intel:
This https://github.com/NervanaSystems/ngraph/pull/1738 should resolve this.

srinivasputta:
Closing this as PR#1738 meets my requirement.

PR #1698, opened 2018-09-27
GPU External Function cleanup

rkimballn1:
Lots of header cleanup. This knocks about 1 minute off of the unit test time, from ~295 seconds to something like 220 seconds, so ok, more than a minute.
Also some general cleanup/simplification. Lots more to come but this PR is bad enough.


src/ngraph/runtime/gpu/cuda_error_check.hpp
```diff
@@ -16,17 +16,16 @@
 
 #pragma once
 
-#include <iostream>
-#include <sstream>
-#include <stdexcept>
-#include <stdint.h>
-#include <string>
-
```
csullivan: nit: Better to separate the std lib headers from the cuda-related headers?

rkimballn1: I was 50:50 on this.

PR #1701, opened 2018-09-27
Reshape Sinking

Krovatkin:


src/ngraph/runtime/cpu/pass/cpu_reshape_sinking.cpp
```diff
@@ -0,0 +1,260 @@
...
+                    delete_reshape(reorders.at(arg));
+                    insert_reshape(n, reorders.at(arg), i);
+                    //no swimming up
+                }
+            }
+            reorders[n] = create_default_reshape(n);
+        }
+    }
+
+    //fix wrong shape info wholesale
```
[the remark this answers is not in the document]
Krovatkin: why we are fixing shapes? it will allow us to serialize, copy a graph after `ReshapeSinking`. Also, some optimizations (`ReshapeElimination` and most likely `CPULayout`) rely on correct shape info.

src/ngraph/runtime/cpu/pass/cpu_reshape_sinking.cpp
```diff
@@ -0,0 +1,316 @@
+/*******************************************************************************
```
rkimballn1: Wrong license header

test/cpu_reshape_sinking.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "util/all_close.hpp"
+#include "util/autodiff/backprop_function.hpp"
+#include "util/autodiff/numeric_compare.hpp"
+#include "util/ndarray.hpp"
+#include "util/random.hpp"
+#include "util/test_tools.hpp"
+
+using namespace ngraph;
+using namespace std;
+
```
diyessi: Add a test for a output that gets used as an input to two ops, one of which is eligible for a reshape.

PR #1691, opened 2018-09-26
Gauri/static initialization problem r08

gaurides:

PR #1702, opened 2018-09-27
Gauri/static initialization problem r08 cherry-pick from #1691

gaurides:


gaurides:
TESTNOW

gaurides:
just cherry-picking from r0.8 to master

PR #1704, opened 2018-09-27
Sigmoid and tanh  doc, edit for abs

diyessi:


doc/sphinx/source/ops/sigmoid.rst
```diff
@@ -8,6 +8,15 @@ Sigmoid
 
    Sigmoid  // Elementwise sigmoid operation
 
+
+Description
+===========
+
+Produces a single output tensor of the same element type and shape as
+``arg,`` where the value at each coordinate of ``output`` is the
+sigmoid of ``arg`` at the same coordinate.
```
aprocter: I think we should define what we mean by "the sigmoid" here. Wikipedia, which is likely the first thing that people are going to land on when they Google, defines "a sigmoid function" as being a function with an "S-like" curve, with f(x) = 1/(1+exp(-x)) only being one example.

doc/sphinx/source/ops/tanh.rst
```diff
@@ -8,22 +8,51 @@ Tanh
...
-        /// | ---------------------- | ------------------------------------------------------------------------------------- |
-        /// | \f$N[d_1,\dots,d_n]\f$ | The tensor \f$T\f$, where \f$T[i_1,\dots,i_n] = \tanh(\texttt{arg}[i_1,\dots,i_n])\f$ |
-
-..       /// \brief Constructs a hyperbolic tangent operation.
-            ///
-            /// \param arg Node that produces the input tensor.
+Description
+===========
 
+Produce a tensor with the same shape and element typye as ``arg,``
```
indie: type

doc/sphinx/source/ops/sigmoid.rst
diyessi: I added this to the math definition.

PR #1712, opened 2018-09-28
IAT: More convolution folding optimizations

jbobba:


src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -1403,3 +1403,215 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_bounded_relu()
...
+                                                      Strides{1, 1},
+                                                      CoordinateDiff{0, 0},
+                                                      CoordinateDiff{0, 0},
+                                                      Strides{1, 1});
+    auto conv_label = std::make_shared<pattern::op::Label>(conv, nullptr, NodeVector{conv});
+
+    auto Ac = std::make_shared<pattern::op::Label>(element::f32, Shape{2});
+    auto A = std::make_shared<op::Broadcast>(Ac, Shape{2, 2, 1, 1}, AxisSet{0, 2, 3});
+    auto A_label = std::make_shared<pattern::op::Label>(A, nullptr, NodeVector{A});
+    auto multiply = std::make_shared<op::Multiply>(conv_label, A_label);
```
ayzhuang: Why do we need conv_label and A_label?

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -1403,3 +1403,215 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_bounded_relu()
...
+            auto bn_eps = op::Constant::create(element::f32, Shape{}, {m_bn->get_eps_value()});
+            auto var_eps = std::make_shared<op::Add>(
+                pattern_map[var],
+                std::make_shared<op::Broadcast>(bn_eps, pattern_map[var]->get_shape(), AxisSet{0}));
+            auto sqrt_var_eps = std::make_shared<op::Sqrt>(var_eps);
+
+            auto mean_gamma = std::make_shared<op::Multiply>(
+                std::make_shared<op::Subtract>(pattern_map[bias], pattern_map[mean]),
+                pattern_map[gamma]);
+            auto new_biases = std::make_shared<op::Subtract>(
```
gaurides: Should this be Subtract or Add? the comment above says  ` +beta`

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -1403,3 +1403,215 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_bounded_relu()
...
+
+    auto pconv = std::make_shared<op::ConvolutionBias>(input,
+                                                       filters,
+                                                       bias,
+                                                       Strides{1, 1},
+                                                       Strides{1, 1},
+                                                       CoordinateDiff{0, 0},
+                                                       CoordinateDiff{0, 0},
+                                                       Strides{1, 1});
+    auto mean_shape = Shape{2};
```
gaurides: do we need these shape variables if they are just used here?

src/ngraph/runtime/cpu/pass/cpu_fusion.hpp
```diff
@@ -64,6 +64,8 @@ class ngraph::runtime::cpu::pass::CPUFusion : public ngraph::pass::GraphRewrite
             construct_zero_padded_conv();
             construct_zero_padded_conv_backprop_filters();
             construct_conv_bias_bprop();
+            construct_conv_bias_folded_batch_norm();
```
gaurides: 1. Should `construct_conv_bias_folded_batch_norm()` be before `construct_batch_norm_relu()`? If yes, can you please add a comment?

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -1403,3 +1403,215 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_bounded_relu()
...
+    auto mean = std::make_shared<pattern::op::Label>(element::f32, mean_shape);
+    auto var_shape = Shape{2};
+    auto var = std::make_shared<pattern::op::Label>(element::f32, var_shape);
+    auto gamma_shape = Shape{2};
+    auto gamma = std::make_shared<pattern::op::Label>(element::f32, gamma_shape);
```
gaurides: in review `gamma` shows up as a keyword in red. Hope that doesn't cause problem.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
ayzhuang: Answer: With labels, we can get nodes matching conv and A through pattern_map in callback.

src/ngraph/runtime/cpu/pass/cpu_fusion.hpp
jbobba: Its not required since folded_batch_norm() matches against batchnorm which will come earlier to a relu in the batch_norm_relu() pattern.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -1403,3 +1401,211 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_bounded_relu()
     auto m = std::make_shared<pattern::Matcher>(min, callback);
     this->add_matcher(m);
 }
+
+void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_folded_batch_norm()
+{
+    Shape shape{2, 2, 1, 1};
+    auto input = std::make_shared<pattern::op::Label>(element::f32, Shape{2, 2, 1, 1});
```
diyessi: It looks like input, filters were meant to use `shape`

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -1403,3 +1401,211 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_bounded_relu()
...
+        }
+
+        if (conv_m->with_relu())
+        {
+            return false;
+        }
+
+        auto A_m = std::static_pointer_cast<op::Broadcast>(pattern_map[A_label]);
+
+        // Check if values are being broadcast along channel (2nd) dimension
```
pruthviIntel: @jbobba why its important to check if the` scaling_vector` is broadcasted across channel? is this because we are scaling  it with filter and if we have `C` filters , we need to have `scaling_vector` across `C`  filters?

jbobba: Right.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
jbobba: Removed shape.

test/cpu_fusion.cpp
```diff
@@ -1670,32 +1670,32 @@ TEST(cpu_fusion, conv_batch_norm_folding)
...
+        auto f = make_shared<Function>(
+            NodeVector{bn}, op::ParameterVector{input, weights, bias, gamma, beta, mean, var});
         return f;
     };
 
     auto int_f = make_function();
     auto cpu_f = make_function();
 
-    test::Uniform<float> rng(-100.0f, 100.0f);
+    test::Uniform<float> rng(1.0f, 100.0f);
```
gaurides: How do you select the range of values? this is all positive values whereas you had + / - earlier. Do we need to test for -ve values?

mchrusci:
TEST NOW

PR #1718, opened 2018-09-28
Partial Shapes, Part 2: Adapt Tensor class to have partial shapes

aprocter:


src/ngraph/descriptor/tensor.cpp
```diff
@@ -22,20 +22,43 @@ using namespace ngraph;
...
-    , m_shape(shape)
+    , m_shape(pshape.is_complete() ? pshape.to_shape() : Shape{})
+    , m_partial_shape(pshape)
     , m_name(name)
 {
 }
 
-void descriptor::Tensor::set_tensor_type(const element::Type& element_type, const Shape& shape)
+void descriptor::Tensor::set_tensor_type(const element::Type& element_type,
+                                         const PartialShape& pshape)
```
diyessi: I think it might be better to just set partial shape and have validate do that, and have a pass that switches over to m_shape when shapes are fully known.

aprocter: I agree that maintaining `m_shape` in parallel like this is awkward. The motivation for the moment is that I want to maintain backwards compatibility during the transition with certain APIs that are returning a `const Shape&`.

test/partial_shape.cpp
```diff
@@ -200,3 +200,66 @@ TEST(partial_shape, shapes_equal_both_complete_same_rank_different_dims)
 
     ASSERT_FALSE(ps1.compatible(ps2));
 }
+
+TEST(partial_shape, from_shape)
+{
+    Shape s{2, 4, 6, 8};
+    PartialShape ps1{s};
+
+    // TODO(amprocte): No way to examine contents of ps1 yet.
```
aprocter: Regarding this TODO: the next PR will be adding some things like `operator[]` on `PartialShape`, etc. I can update these tests in that PR.

aprocter:
TESTNOW

PR #1719, opened 2018-09-28
Fold some unary and binary ops

Krovatkin:


rkimballn1:
Is it normal/acceptable to have a PR like this with no unit tests? Is it covered by other tests?

Krovatkin:
@rkimballn1 added tests.

PR #1721, opened 2018-09-29
cudnnFind/Get interoperability

csullivan:
Update to algo. search so that it is no longer binary on/off. Now it is either off, a heuristic search (cudnnGet*) or an explicit search (cudnnFind*). cudnnFind gives ~10-20% speed up when used on resnet50 (bs128) relative to using cudnnGet. The extra compile time is minimal but for now I leave the explicit search turned off and default to the heuristic search. Open to suggestions.

### DIFF
--- a/src/ngraph/runtime/gpu/cudnn_emitter.cpp
+++ b/src/ngraph/runtime/gpu/cudnn_emitter.cpp
@@ -946,7 +946,8 @@ size_t runtime::gpu::CUDNNEmitter::build_convolution(const std::string& dtype,
     int max_algos = 0;
     CUDNN_SAFE_CALL(cudnnGetConvolutionForwardAlgorithmMaxCount(*m_ctx->cudnn_handle, &max_algos));
     std::vector<cudnnConvolutionFwdAlgoPerf_t> results(max_algos);
-    auto cudnn_algo_search = find_algo ? cudnnFindConvolutionForwardAlgorithm : cudnnGetConvolutionForwardAlgorithm_v7;
+    auto cudnn_algo_search =
+        find_algo ? cudnnFindConvolutionForwardAlgorithm : cudnnGetConvolutionForwardAlgorithm_v7;
     CUDNN_SAFE_CALL((*cudnn_algo_search)(*m_ctx->cudnn_handle,
                                          tensor_desc_0,
                                          filter_desc,
@@ -956,7 +957,8 @@ size_t runtime::gpu::CUDNNEmitter::build_convolution(const std::string& dtype,
                                          &num_algos,
                                          results.data()));
     results.resize(num_algos);
-    auto conv_fwd_algo = select_cudnn_algo<cudnnConvolutionFwdAlgoPerf_t, cudnnConvolutionFwdAlgo_t>(results);
+    auto conv_fwd_algo =
+        select_cudnn_algo<cudnnConvolutionFwdAlgoPerf_t, cudnnConvolutionFwdAlgo_t>(results);
 
     void* alpha = m_host_parameters.allocate_by_datatype(data_type, 1.0);
     void* beta = m_host_parameters.allocate_by_datatype(data_type, 0);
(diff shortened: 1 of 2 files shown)


rkimballn1:
Would it be possible to run the first few iterations running the sub-optimial kernel while compiling the optimal one? Then when the optimal kernel is ready use it.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -943,16 +943,27 @@ size_t runtime::gpu::CUDNNEmitter::build_convolution(const std::string& dtype,
...
-        CUDNN_SAFE_CALL(cudnnGetConvolutionForwardAlgorithm(*m_ctx->cudnn_handle,
-                                                            tensor_desc_0,
-                                                            filter_desc,
-                                                            conv_desc,
-                                                            tensor_desc_1,
-                                                            CUDNN_CONVOLUTION_FWD_PREFER_FASTEST,
-                                                            /*memoryLimitInBytes=*/0,
-                                                            &conv_fwd_algo));
+        int num_algos;
+        int max_algos = 0;
```
fengleitian: we only need 1, so can we set number just to 1?

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -943,16 +943,27 @@ size_t runtime::gpu::CUDNNEmitter::build_convolution(const std::string& dtype,
...
+        CUDNN_SAFE_CALL(
+            cudnnGetConvolutionForwardAlgorithmMaxCount(*m_ctx->cudnn_handle, &max_algos));
+        std::vector<cudnnConvolutionFwdAlgoPerf_t> results(max_algos);
+        auto cudnn_algo_search = (find_algo == algo_search::EXPLICIT)
```
fengleitian: just use `cudnnFindConvolutionForwardAlgorithm`

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -1091,19 +1114,32 @@ size_t runtime::gpu::CUDNNEmitter::build_convolution_backward_filter(
...
+        CUDNN_SAFE_CALL((*cudnn_algo_search)(*m_ctx->cudnn_handle,
+                                             tensor_desc_0,
+                                             tensor_desc_1,
+                                             conv_desc,
+                                             filter_desc,
+                                             static_cast<int>(results.size()),
+                                             &num_algos,
+                                             results.data()));
+        results.resize(num_algos);
+        conv_bwd_filter_algo =
```
fengleitian: `select_cudnn_algo` is not required if result size is 1.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: same for others.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: same for others.

csullivan: This max_algos is asking cudnn to try all the algorithms it knows, if we reduce it to 1 I believe it will reduce the number of tested algorithms to 1.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
csullivan: In case we want to switch back, I left the algo_search enum, but set it to do the explicit search via `cudnnFindConvolutionForwardAlgorithm` by default.

src/ngraph/runtime/gpu/cudnn_emitter.hpp
```diff
@@ -170,6 +179,24 @@ namespace ngraph
...
+                    for (auto i = 0; i != perf_results.size(); ++i)
+                    {
+                        auto const& result = perf_results[i];
+                        if (result.status == CUDNN_STATUS_SUCCESS &&
+                            result.memory <= workspace_byte)
+                        {
+                            return result.algo;
+                        }
+                    }
+                    return perf_results.at(0).algo;
```
fengleitian: should print error msg if we get here. That means all algorithm can't be used.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: yes, keep it.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: I did some test, It does search but only return one result. But I'm not sure if it usable, since I notice there is `status` in perf result, is that means it could be `false`?

PR #1723, opened 2018-10-01
print the entire path of the file printing relative to the project root

rkimballn1:
Change the way that log prints out the file path. It used to just print the file name so if there were two files with the same name in different paths you would never know the difference. Now it prints out the file's path relative to the project root, so you get output like this:
```
[INFO] 2018-10-01T15:25:18z src/ngraph/coordinate_transform.cpp 57      Strides{1, 1}
[INFO] 2018-10-01T15:25:18z test/coordinate.cpp 130
```


rkimballn1:
@aprocter The full path provides no useful information beyond the trimmed path. The information you do gain is where the repo was cloned to on the machine that built the software. I guess from that you could perhaps know who built the software.
This is a change to add more of the path to what exists now. Now we have only the filename with no path. This can be confusing when you get a file like `coordinate.cpp` which could be `test/coordinate.cpp` or `ngraph/coordinate.cpp`. Adding the path relative to the build directory tells you which source file we are talking about. If you would like to find out where bazel sticks ngraph for building then I suggest we add something that does that, but prepending the path to where the ngraph repo was cloned makes the name too long.

aprocter:
> This is a change to add more of the path to what exists now. Now we have only the filename with no path.

Whoops... I misunderstood that. Thought we had full filenames now. I think I'm mixing it up with `NGRAPH_ASSERT` and friends, which do provide full paths.

> The full path provides no useful information beyond the trimmed path. The information you do gain is where the repo was cloned to on the machine that built the software. I guess from that you could perhaps know who built the software.

There's a particular use case where I find it useful with the NGRAPH_ASSERT messages is as follows:

* TensorFlow's build process has built nGraph in `/foo/bar/bazel/god/knows/where`
* An `NGRAPH_ASSERT` fails in /foo/bar/bazel/god/knows/where/ngraph/src/ngraph/op/avg_pool.cpp:23
* I can just copy and paste that whole string from the error message, and paste it as an argument to `vim +23`

aprocter:
Objection dropped, after some offline chatter.

PR #1733, opened 2018-10-02
Show types in VisualizeTree

Krovatkin:


src/ngraph/pass/visualize_tree.cpp
```diff
@@ -112,6 +112,15 @@ std::string pass::VisualizeTree::get_attributes(shared_ptr<Node> node)
                                                       : vector_to_string(node->get_shape()));
     }
 
+    static const auto nvtot = std::getenv("NGRAPH_VISUALIZE_TREE_OUTPUT_TYPES");
```
rkimballn1: I call "auto abuse" on this. You are comparing it to a nullptr which means you know what it is. Put a real type here.

PR #1735, opened 2018-10-02
add doctools js from basic theme sphinx repo

indie:
Trying something else


indie:
Tracked the bug down this is as the source: https://github.com/rtfd/sphinx_rtd_theme/pull/672

PR #1739, opened 2018-10-03
Changes to make Klocwork a little happier

rkimballn1:
The changes to examples in the doc directory are only to the mnist_loader so it should not effect the doc line numbers.


doc/examples/mnist_mlp/mnist_loader.hpp
```diff
@@ -42,18 +42,18 @@ class MNistLoader
...
     }
 
-    std::uint32_t get_items() { return m_items; }
+    uint32_t get_items() { return m_items; }
 protected:
     std::string m_filename;
-    FILE* m_file{nullptr};
-    std::uint32_t m_magic;
-    std::uint32_t m_items;
+    FILE* m_file;
```
diyessi: Initializing with `{nullptr}` means every constructor gets the correct initialization if it's not explicit. Initialiing in the constructor means every constructor needs to explicitly initialize.

doc/examples/mnist_mlp/mnist_loader.hpp
```diff
@@ -42,18 +42,18 @@ class MNistLoader
...
+    uint32_t m_magic;
```
diyessi: These should be coming from `<cstdint>` and be `std` according to https://stackoverflow.com/questions/14883896/why-is-stduint32-t-different-from-uint32-t
Do you have a source that says otherwise?

src/ngraph/cpio.cpp
```diff
@@ -76,6 +76,7 @@ cpio::Header cpio::Header::read(istream& stream)
         {
             throw runtime_error("CPIO magic error");
         }
+        rc.magic = 0x71C7;
```
diyessi: Comments about the source of these magic numbers?

doc/examples/mnist_mlp/mnist_loader.hpp
Krovatkin: I think `std::uint32_t` is an *alias* of `uint32_t` *IF* both come from `cstdint`

rkimballn1: This is from <cstdint>. It is as @Krovatkin says, `std::uint32_t` is just an alias to `uint32_t`
```
namespace std
{
  using ::int8_t;
  using ::int16_t;
  using ::int32_t;
  using ::int64_t;

  using ::int_fast8_t;
  using ::int_fast16_t;
  using ::int_fast32_t;
  using ::int_fast64_t;

  using ::int_least8_t;
  using ::int_least16_t;
  using ::int_least32_t;
  using ::int_least64_t;

  using ::intmax_t;
  using ::intptr_t;

  using ::uint8_t;
  using ::uint16_t;
  using ::uint32_t;
  using ::uint64_t;

  using ::uint_fast8_t;
  using ::uint_fast16_t;
  using ::uint_fast32_t;
  using ::uint_fast64_t;

  using ::uint_least8_t;
  using ::uint_least16_t;
  using ::uint_least32_t;
  using ::uint_least64_t;

  using ::uintmax_t;
  using ::uintptr_t;
} // namespace std
```

doc/examples/mnist_mlp/mnist_loader.hpp
rkimballn1: I like consistency. There are some values that must be initialized in the ctor while others could be defined here in the header. If you want I can move all values that are not from parameters in the ctor to be initialized as they are defined.

diyessi: Let's start doing it that way; then the constructors only contain what is not the default.

rkimballn1: done

PR #1740, opened 2018-10-03
Support LRN for NVGPU Backend

gcwenger:


src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -1403,6 +1403,69 @@ size_t runtime::gpu::CUDNNEmitter::build_batchnorm(const cudnnBatchNormMode_t& b
...
+                                                            &output_desc,
+                                                            this,
+                                                            lrn_op,
+                                                            lrn_alpha,
+                                                            lrn_beta,
+                                                            lrn_bias,
+                                                            lrn_size,
+                                                            alpha,
+                                                            beta](void** inputs, void** outputs) {
+        CUDNN_SAFE_CALL(cudnnSetLRNDescriptor(
```
fengleitian:
```
CUDNN_SAFE_CALL(cudnnSetLRNDescriptor(
                 lrn_descriptor, static_cast<unsigned int>(lrn_size), lrn_alpha, lrn_beta, lrn_bias));
```
possible to move this outside? I think this only need to be done one time, right?

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -1403,6 +1403,69 @@ size_t runtime::gpu::CUDNNEmitter::build_batchnorm(const cudnnBatchNormMode_t& b
...
+                                             const Shape& output_shape,
+                                             const double lrn_alpha,
+                                             const double lrn_beta,
+                                             const double lrn_bias,
+                                             const size_t lrn_size)
+{
+    // construct hash to determine if kernel needs to be emitted
+    // or if it already exists in the primitive list
+    std::stringstream ss;
+    ss << "lrn_op_" << lrn_op << "_dtype_" << dtype << "_dir" << static_cast<int>(direction) << "_i"
```
fengleitian:
```
                                              const double lrn_alpha,
                                              const double lrn_beta,
                                              const double lrn_bias,
                                              const size_t lrn_size
```
Those should also in hash. Since here there is not runtime input for those numbers. So different number can not share the same primitive. 

Another way is put those number through parameter like `void** inputs, void** outputs`, so you don't need those number in hash. This require some framework change.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
fengleitian: If you do this, need to check if we have cache for `lrn_descriptor`, otherwise might need a copy instead of reference.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -1403,6 +1403,69 @@ size_t runtime::gpu::CUDNNEmitter::build_batchnorm(const cudnnBatchNormMode_t& b
     return this->m_primitive_emitter->register_primitive(batchnorm, hash);
 }
 
+size_t runtime::gpu::CUDNNEmitter::build_lrn(const cudnnLRNMode_t& lrn_op,
+                                             const std::string& dtype,
+                                             const Prop& direction,
```
fengleitian: Is `direction` been used here. do we have an `backward op` need to implement later?

src/ngraph/runtime/gpu/cudnn_emitter.cpp
gcwenger: I had noted mixed use of setting up inside lambda vs. not. But upon further analysis I agree I can do this ahead of time. Caching already supported by Chris's CUDNNDescriptors.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
gcwenger: I'll add these to the hash (at least for now). That will be consistent with doing cudnnSetLRNDescriptor once ahead of time (previous comment above). (If passed as param, I would not do cudnnSetLRNDescriptor ahead of time). But agreed that it would likely be a good future direction to pass additional info as parameter (generating less unique ops) to favor faster compilation (because runtime would likely not be adversely impacted).

src/ngraph/runtime/gpu/cudnn_emitter.cpp
gcwenger: Not used yet. Plumbed in for future support. Couldn't find any existing support of backwards direction in CPU backend.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -1403,6 +1403,62 @@ size_t runtime::gpu::CUDNNEmitter::build_batchnorm(const cudnnBatchNormMode_t& b
...
+                                             const double lrn_alpha,
+                                             const double lrn_beta,
+                                             const double lrn_bias,
+                                             const size_t lrn_size)
+{
+    // construct hash to determine if kernel needs to be emitted
+    // or if it already exists in the primitive list
+    std::stringstream ss;
+    ss << "lrn_op_" << lrn_op << "_dtype_" << dtype << "_dir" << static_cast<int>(direction) << "_i"
+       << join(input_shape, "_") << "_o" << join(output_shape, "_") << "_alpha_" << lrn_alpha
```
csullivan: Is the output_shape necessary for the hash?

gcwenger: No - it's in lockstep with input_shape. Reworked code to reflect this.

PR #1741, opened 2018-10-03
nvgpu maxpool bug fix

fengleitian:
the build_pad used in maxpool might have some bugs.
1. add a test to catch bug
2. fix bug with another padding.


test/backend_test.in.cpp
```diff
@@ -5207,6 +5207,63 @@ NGRAPH_TEST(${BACKEND_NAME}, max_pool_2d_2channel_2image)
               read_vector<float>(result));
 }
```
diyessi: Please add a short description to this test that describes the problem it is trying to detect.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -712,46 +712,45 @@ size_t runtime::gpu::CUDNNEmitter::build_primitive(const op::MaxPool* node)
...
-    if (pad_required)
+    if (padding_below != padding_above)
     {
-        auto temp_size = shape_size(shape_to_pool) * args[0].get_element_type().size();
+        Shape padding_interior(padding_below.size(), 1);
+        input_shape_padded =
+            runtime::gpu::get_padded_shape(input_shape, padding_below, padding_above, {});
+        padded_size = shape_size(input_shape_padded);
+        //currntly we set this to float point only, need to add other datatype support later
+        float pad_value = std::numeric_limits<float>::lowest();
```
csullivan: We can easily not restrict to float if we add support for `pad_value` to build_pad_dynamic directly.
```
        auto pad_value = TypeInfo::Get(args[0].get_element_type())->lowest();		
```
Can you add an issue for supporting this in pad_dynamic.

fengleitian: https://github.com/NervanaSystems/ngraph/issues/1752

PR #1742, opened 2018-10-03
fix hash value in convolution 

fengleitian:
add padding_above to hash to avoid reuse the wrong primitive.


fengleitian:
ok, not a bug, it's redundant.

PR #1754, opened 2018-10-04
IntelGPU backend: Broadcast bug fix: (output_shape.at(0) == 1) doesn'…

dmyershov:
…t mean that it is scalar

### DIFF
--- a/src/ngraph/runtime/intelgpu/intelgpu_op_broadcast.cpp
+++ b/src/ngraph/runtime/intelgpu/intelgpu_op_broadcast.cpp
@@ -138,7 +138,7 @@ void runtime::intelgpu::do_bcast_sum_operation(cldnn::topology& topology,
         else
         {
             // corner case with scalar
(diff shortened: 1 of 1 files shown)


src/ngraph/runtime/intelgpu/intelgpu_op_broadcast.cpp
```diff
@@ -138,7 +138,7 @@ void runtime::intelgpu::do_bcast_sum_operation(cldnn::topology& topology,
         else
         {
             // corner case with scalar
-            if (output_shape.empty() || (!output_shape.empty() && (output_shape.at(0) == 1)))
+            if (output_shape.empty() || (!output_shape.empty() && (shape_size(output_shape) == 1)))
```
shssf: `!output_shape.empty()` not needed in this case.
I would propose:
`if (output_shape.empty() || (shape_size(output_shape) == 1))`

PR #1756, opened 2018-10-05
[WIP] Partial Shapes and Types, Part 4: Implement partial shape/type validation for existing ops

aprocter:


aprocter:
WIP because #1728 is still pending

src/ngraph/op/allreduce.cpp
```diff
@@ -27,13 +27,9 @@ op::AllReduce::AllReduce(const shared_ptr<Node>& arg)
...
 void op::AllReduce::validate_and_infer_types()
 {
-    if (validate_punt_if_dynamic())
-    {
-        return;
-    }
-
     NODE_VALIDATION_ASSERT(this,
-                           get_input_element_type(0) == element::f32 ||
+                           get_input_element_type(0).is_dynamic() ||
```
diyessi: Shouldn't this be `is_real()` ?

aprocter: The `is_dynamic()` is there to cover the case where the element type is not determined yet, in which case type checking "provisionally" passes. At runtime, `AllReduce` only allows `f32` and `f64`.

aprocter:
@rkimballn1 @silee2 Looks like you guys got tagged for CMake.

PR #1758, opened 2018-10-05
Cyphers/doc1

diyessi:
Documentation for a few more ops.
Clean up punctuation formatting.


doc/sphinx/source/ops/abs.rst
```diff
@@ -12,7 +12,7 @@ Abs
 Description
 ===========
 
-Produces a single output tensor of the same element type and shape as ``arg``,
+Produces a single output tensor of the same element type and shape as ``arg,``
 where the value at each coordinate of ``output`` is the absolute value of the
```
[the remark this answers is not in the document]
aprocter: This is probably the best option for now, but is there any CSS hackery we could do to make it look right with the punctuation outside the backticks?

doc/sphinx/source/ops/sqrt.rst
```diff
@@ -0,0 +1,57 @@
...
+| ``output``      | Same as ``arg``         | Same as ``arg``                |
++-----------------+-------------------------+--------------------------------+
+
+
+Mathematical Definition
+=======================
+
+.. math::
+
+   \mathtt{output}_{i_0, \ldots, i_{n-1}} = \sqrt(\mathtt{arg}_{i_0, \ldots, i_{n-1}})
```
aprocter: When we render this, the radical sign isn't extending over the whole radicand. I think if we change it from

```
\sqrt(foo)
```

to

```
\sqrt{foo}
```

it should give us what we want.

doc/sphinx/source/ops/sqrt.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Backprop
+========
+
+.. math::
+
+   \overline{\mathtt{arg}} \leftarrow \frac{\Delta}{2 \mathtt{output}}
```
aprocter: The rendering of `2output` is a bit hard to read. Can we use `\dot` (or whatever) to denote the multiplication?

PR #1763, opened 2018-10-05
IAT: Collapse dimensions around arithmetic reduction operations

jbobba:


src/ngraph/pass/memory_layout.cpp
```diff
@@ -51,8 +51,10 @@ bool pass::MemoryLayout::run_on_function(shared_ptr<ngraph::Function> function)
                     auto input = &node->get_inputs().at(oi_pair.input).get_tensor();
                     auto input_node = node->get_inputs().at(oi_pair.input).get_output().get_node();
 
-                    // an input tensor can be reused if this is the last use
-                    if (node->liveness_free_list.count(input) != 0 &&
+                    // For destructive kernel, this should be the last use
+                    // Non-destructive kernels can pass through if memory sharing is disabled
+                    if ((node->liveness_free_list.count(input) != 0 ||
```
Krovatkin: i'm not sure we can do that; i'll bug you offline about this one.

src/ngraph/runtime/cpu/pass/cpu_collapse_dims.cpp
```diff
@@ -56,88 +63,155 @@ static void collapse_dims(std::vector<size_t>& output_shape,
...
+
+    collapse_dims(input_shape, operated_axes, cshape, true);
+
+    if (cshape.axis_set.size() == 0)
+    {
+        // Null reduction operation
+        ngraph::replace_node(n, node->get_argument(0));
+        replaced = true;
+    }
+    else if (input_shape.size() != cshape.rshape.size())
```
ayzhuang: should it be cshape.fshape.size()?

ayzhuang: please change it.

PR #1764, opened 2018-10-05
Handle zero sized buffer requests

jbobba:
We have some unit test graphs that dont require any intermediate buffers and hit this case.


src/ngraph/runtime/aligned_buffer.cpp
```diff
@@ -48,6 +48,11 @@ void runtime::AlignedBuffer::initialize(size_t byte_size, size_t alignment)
             m_aligned_buffer += (alignment - mod);
         }
     }
+    else
+    {
+        m_allocated_buffer = nullptr;
```
Krovatkin: shouldn't we do this in a c-tor? or even better in a field initializer?

jbobba: @rkimballn1 are you ok with removing initialize. we can do this whole thing inside the constructor. This will also prevent cases where initialize is called multiple times on the same AlignedBuffer

rkimballn1: If you remove these lines then you *must* make initialize a private member function.
Were a user to call initialize() it should work. If you move the nullptr setting to the ctor then user can't call initialize()

rkimballn1: I think initialize() exists because there were multiple ctors that did the same thing. I would remove initialize() and fold it into the ctor.

PR #1770, opened 2018-10-08
[ONNX] Assert all op types supported

postrational:
Verify that import of all op types is supported and display a list of unsupported ops found in model if any.


src/ngraph/frontend/onnx_import/core/model.cpp
```diff
@@ -0,0 +1,59 @@
...
+        Model::Model(const onnx::ModelProto& model_proto)
+            : m_model_proto{&model_proto}
+        {
+            // Verify that the ONNX graph contains only nodes of supported op_type
+            assert_all_op_types_supported();
+        }
+
+        void Model::assert_all_op_types_supported()
+        {
+            std::set<std::string> unsupported_ops;
```
arogowie-intel: Please add `<set>` header file.

src/ngraph/frontend/onnx_import/ops_bridge.cpp
```diff
@@ -109,6 +109,11 @@ namespace ngraph
                     return ops_bridge::get()(node);
                 }
 
+                static bool is_op_type_supported(const std::string& op_type)
```
arogowie-intel: Please add `<string>` header file.

arogowie-intel: This function may also be `const` qualified.

postrational: Static member function cannot have `const` qualifier.

arogowie-intel: Good point! :)

PR #1780, opened 2018-10-09
run Quantize / Dequantize reference kernels on CPU backend

adstraw:


adstraw:
@Krovatkin to add codegen support

PR #1786, opened 2018-10-09
reset buffer size, use original input size for memcpy

fengleitian:


src/ngraph/runtime/gpu/gpu_memory_manager.cpp
```diff
@@ -80,6 +80,7 @@ void runtime::gpu::GPUMemoryManager::allocate()
             m_argspace_mem.back().ptr, m_buffered_mem.data(), m_buffer_offset);
         // add an empty node to the end of the list and zero offset
         m_argspace_mem.push_back({nullptr, 0});
+        m_buffered_mem.resize(initial_buffer_size);
```
csullivan: I don't think it's required since we maintain our own buffer offset, but we could be safe and clear the space before resize. We could also zero-initialize the space here and in the constructor.

```
        m_buffered_mem.clear();
        m_buffered_mem.resize(initial_buffer_size, 0);
```

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
```diff
@@ -97,7 +98,8 @@ void runtime::gpu::GPUMemoryManager::allocate()
 size_t runtime::gpu::GPUMemoryManager::queue_for_transfer(const void* data, size_t size)
 {
     // if the current allocation will overflow the host buffer
-    size_t new_size = m_buffer_offset + size;
+    size_t aligned_size = ngraph::pass::MemoryManager::align(size, runtime::gpu::GPUMemoryManager::alignment);
```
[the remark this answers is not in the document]
fengleitian: This does not resolve that issue.

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
fengleitian: Yes, it's not required but could release some resource. 
Right, it will be safe to  zero-initialize.

fengleitian: let me try add a test.

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
```diff
@@ -26,7 +26,7 @@ constexpr const uint32_t initial_buffer_size = 10 * 1024 * 1024;
 
 runtime::gpu::GPUMemoryManager::GPUMemoryManager(GPUPrimitiveEmitter* emitter)
     : m_buffer_offset(0)
-    , m_buffered_mem(initial_buffer_size)
+    , m_buffered_mem(initial_buffer_size, 0)
```
gcwenger: Why zero initialize?

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
```diff
@@ -80,6 +80,8 @@ void runtime::gpu::GPUMemoryManager::allocate()
             m_argspace_mem.back().ptr, m_buffered_mem.data(), m_buffer_offset);
         // add an empty node to the end of the list and zero offset
         m_argspace_mem.push_back({nullptr, 0});
+        m_buffered_mem.clear();
```
gcwenger: Why reallocate & zero initialize?

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
```diff
@@ -109,12 +113,12 @@ size_t runtime::gpu::GPUMemoryManager::queue_for_transfer(const void* data, size
 
     if (need_resize)
     {
-        m_buffered_mem.resize(buffer_size);
+        m_buffered_mem.resize(buffer_size, 0);
```
gcwenger: Why zero initialize?

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
fengleitian: good question, it's not necessary to zero initialize it. but it's a good practice to zero initialize. It's by default for array in c++.

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
fengleitian: Here will release resource after use. for example, you might used 10GB buffer, you can release it and reallocate to initial size, which is much smaller, like 10MB.

fengleitian: same as above for zero initialize.

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
fengleitian: same as above for zero initialize.

src/ngraph/runtime/gpu/gpu_memory_manager.cpp
fengleitian: I should not say by default, if you do like this, it will be zero initialized.
```
T () ;
T t = {} ; 
T {} ;
```

PR #1789, opened 2018-10-10
Switch to scale and offset design from min and max for Quantization

nbpatel:
This commit uses scale and offset approach instead of min and max. Also, removes QuantizeCPU and DequantizeCPU.


adstraw:
Adding @ashokei to comment on the constant-ness of min/max ----> scale/offset in MXNet

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
```diff
@@ -50,16 +50,15 @@ namespace ngraph
                         mkldnn_emitter->build_convolution<ngraph::op::QuantizedConvolution>(
                             node, args, out);
                     auto& deps = mkldnn_emitter->get_primitive_deps(conv_index);
-                    float min_freezed_output = qconvolution->get_freezed_output_min();
-                    float max_freezed_output = qconvolution->get_freezed_output_max();
+                    float scale = qconvolution->get_scale();
+                    float offset = qconvolution->get_offset();
```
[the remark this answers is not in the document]
diyessi: We're supporting invoking the mkl-dnn ops here; we should take the parameters they do, which I think means just a float scale and no offset.

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -2825,69 +2823,6 @@ namespace ngraph
                 }
             }
 
-            template <>
-            void CPU_Emitter::EMITTER_DECL(ngraph::op::QuantizeCPU)
```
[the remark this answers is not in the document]
nbpatel: this can be another PR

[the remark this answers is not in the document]
adstraw: Agree.

nbpatel:
@ashokei mkldnn does not support runtime scale.

ashokei:
@nbpatel can we pass scale/offset (or min/max) at op invoke time (like inputs/data). Currently it seems min/max are constants (quantize op has to be recreated if we want to use different min/max for different data)

nbpatel:
@ashokei with the current design no. But something we'd be looking into after things stabilize

src/ngraph/builder/quantization.cpp
```diff
@@ -0,0 +1,162 @@
...
+            // TODO: make a builder to calculate
+            auto scale = op::Constant::create(element::f32, Shape{}, {1});
+
+            return make_shared<op::QuantizedAvgPool>(arg,
+                                                     window_shape,
+                                                     window_movement_strides,
+                                                     padding_below,
+                                                     padding_above,
+                                                     include_padding_in_avg_computation,
+                                                     scale);
```
jbobba: We dont need scale for avg pool since we dont requantize.

src/ngraph/builder/quantization.cpp
```diff
@@ -0,0 +1,162 @@
...
+                                           const std::shared_ptr<Node> min_input,
+                                           const std::shared_ptr<Node> max_input,
+                                           const std::shared_ptr<Node> min_filter,
+                                           const std::shared_ptr<Node> max_filter,
+                                           const std::shared_ptr<Node> min_freezed_output,
+                                           const std::shared_ptr<Node> max_freezed_output,
+                                           const bool with_relu)
+        {
+            // TODO: make a builder to calculate
+            auto scale = op::Constant::create(element::f32, Shape{}, {1});
```
jbobba: probably better to call this requantization_scale instead of scale.

src/ngraph/builder/quantization.cpp
```diff
@@ -0,0 +1,162 @@
...
+                                           const Strides& data_dilation_strides,
+                                           const std::shared_ptr<Node> min_input,
+                                           const std::shared_ptr<Node> max_input,
+                                           const std::shared_ptr<Node> min_filter,
+                                           const std::shared_ptr<Node> max_filter,
+                                           const std::shared_ptr<Node> min_freezed_output,
+                                           const std::shared_ptr<Node> max_freezed_output)
+        {
+            // TODO: make a builder to calculate
+            auto scale = op::Constant::create(element::f32, Shape{}, {1});
```
jbobba: requantization_scale

src/ngraph/builder/quantization.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+            return make_shared<op::QuantizedMaxPool>(arg,
+                                                     window_shape,
+                                                     window_movement_strides,
+                                                     padding_below,
+                                                     padding_above,
+                                                     scale);
```
jbobba: dont need scale.

src/ngraph/builder/quantization.cpp
nbpatel: Yes, correct. Adam and me had this discussion. Will make the changes.

postrational:
TEST NOW

PR #1793, opened 2018-10-10
[ONNX] Tool to serialize ONNX models

postrational:
Simple tool to use for debugging problems with ONNX models.


src/tools/serialize_onnx/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ******************************************************************************
+
+add_executable(serialize_onnx serialize_onnx.cpp)
+add_dependencies(serialize_onnx ngraph)
+add_dependencies(serialize_onnx onnx_import_interface)
```
apwojcik: This should not be required. If the build does not work without this target then the dependencies for ngraph are incorrectly set by ONNX.

src/tools/serialize_onnx/serialize_onnx.cpp
```diff
@@ -0,0 +1,83 @@
...
+        }
+        else if (arg == "-h" || arg == "--help")
+        {
+            help();
+            return 0;
+        }
+    }
+
+    ifstream f(input);
+    if (f)
```
apwojcik: Instead, use `f.is_open()` to check if the file stream has an associated file.

src/tools/CMakeLists.txt
```diff
@@ -16,3 +16,4 @@
 
 add_subdirectory(nbench)
 add_subdirectory(reserialize)
+add_subdirectory(serialize_onnx)
```
apwojcik: Add the subdirectory ONLY if ONNX is enabled.

src/tools/serialize_onnx/CMakeLists.txt
apwojcik: Remove dependency on `onnx_import_interface`. By adding this you force build not having ONNX enabled to build the incomplete ONNX support in nGraph.

src/tools/serialize_onnx/serialize_onnx.cpp
```diff
@@ -0,0 +1,84 @@
...
+        }
+        else if (arg == "-h" || arg == "--help")
+        {
+            help();
+            return 0;
+        }
+    }
+
+    ifstream f(input);
+    if (f.is_open())
```
rkimballn1: Better to just use `if (f)` which also tests for errors, not just open.

arogowie-intel:
TEST NOW

arogowie-intel:
TEST NOW

PR #1797, opened 2018-10-10
Handle workspace reservation when workspace size is zero

AyanmoI:
Return nullptr when workspace size is zero+modify insert method to accept const lvalue ref


AyanmoI:
Added memory manager unit test

PR #1798, opened 2018-10-10
DEX Debugger

Krovatkin:


test/debugger.cpp
```diff
@@ -0,0 +1,225 @@
...
+// limitations under the License.
+//*****************************************************************************
+
+#include <algorithm>
+#include <cstdio>
+#include <iostream>
+#include <list>
+#include <memory>
+
+#include "gtest/gtest.h"
```
rkimballn1: This #include list seems excessive. Everything here is needed?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1524,17 +1524,28 @@ void runtime::cpu::CPU_ExternalFunction::build()
         }
         else
         {
-            for (const auto& p : enables)
+            std::advance(functor, ctx->pc);
```
Krovatkin: @jbobba can I change `enables` and `functors` to use `std::vector`? We might take a negligible perf hit when those get resized at a *build* step (which could also be avoided if we pass the size in `CPU_ExternalFunction` c-tor) but then we will be able to just use regular `for` loop with an index w/o this iterator voodoo?

src/ngraph/runtime/cpu/cpu_call_frame.cpp
```diff
@@ -79,6 +82,78 @@ void runtime::cpu::CPU_CallFrame::call(
...
+        return false;
+    }
+
+    ctx->breakpoints.insert(ctx->pc + 1);
+    inner_call(output_tvs, input_tvs);
+    ctx->breakpoints.erase(ctx->pc);
+    return true;
+}
+
+void runtime::cpu::CPU_CallFrame::kontinue(
```
ransford2011: Is `kontinue` due to naming clash? 😆

Krovatkin: @ransford2011 but @diyessi gave me a great idea to rename it to `resume`

src/ngraph/runtime/cpu/cpu_call_frame.cpp
```diff
@@ -79,6 +82,78 @@ void runtime::cpu::CPU_CallFrame::call(
...
+    const std::vector<std::shared_ptr<runtime::Tensor>>& input_tvs)
+{
+    if (ctx->pc >= m_external_function->op_names.size())
+    {
+        return false;
+    }
+
+    ctx->breakpoints.insert(ctx->pc + 1);
+    inner_call(output_tvs, input_tvs);
+    ctx->breakpoints.erase(ctx->pc);
```
diyessi: If you step over a breakpoint, you end up erasing it, probably not what the user intended.

src/ngraph/runtime/cpu/cpu_call_frame.hpp
```diff
@@ -52,6 +53,21 @@ namespace ngraph
...
+                bool step(const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
+                          const std::vector<std::shared_ptr<runtime::Tensor>>& inputs);
+
+                /// \brief Continue to execute from the current PC
+                void resume(const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
+                            const std::vector<std::shared_ptr<runtime::Tensor>>& inputs);
+
+                /// \brief Add a breakpoint to a node
+                bool add_breakpoint(std::shared_ptr<Node> op);
+                /// \brief Remove a breakpoint to a node
```
diyessi: `from` a node

src/ngraph/runtime/cpu/cpu_call_frame.cpp
Krovatkin: ah a good catch.. I want to *erase* it only if it wasn't there before!

src/ngraph/runtime/cpu/cpu_call_frame.cpp
```diff
@@ -79,6 +82,86 @@ void runtime::cpu::CPU_CallFrame::call(
...
+    inner_call(output_tvs, input_tvs);
+    return;
+}
+*/
+
+void runtime::cpu::CPU_CallFrame::call(
+    const std::vector<std::shared_ptr<runtime::Tensor>>& output_tvs,
+    const std::vector<std::shared_ptr<runtime::Tensor>>& input_tvs)
+{
+    ctx->pc = 0;
```
jbobba: propagate_layouts here instead of in inner_call.

src/ngraph/runtime/cpu/cpu_call_frame.cpp
```diff
@@ -79,6 +82,86 @@ void runtime::cpu::CPU_CallFrame::call(
     }
 }
 
+/*
```
jbobba: remove commented out code.

src/ngraph/runtime/cpu/cpu_call_frame.hpp
```diff
@@ -52,6 +56,21 @@ namespace ngraph
...
+                /// \brief Continue to execute from the current PC
+                //void resume(const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
+                //            const std::vector<std::shared_ptr<runtime::Tensor>>& inputs);
+
+                /// \brief Add a breakpoint to a node
+                //bool add_breakpoint(std::shared_ptr<Node> op);
+                /// \brief Remove a breakpoint from a node
+                //bool delete_breakpoint(std::shared_ptr<Node> op);
+
+                //void* inspect(std::shared_ptr<Node> op, size_t output_index = 0);
```
jbobba: remove.

src/ngraph/runtime/cpu/cpu_call_frame.hpp
```diff
@@ -63,6 +82,11 @@ namespace ngraph
                 CPU_CallFrame(CPU_CallFrame&&) = delete;
                 CPU_CallFrame& operator=(const CPU_CallFrame&) = delete;
 
+                //void* inspect(const std::string& tensor_descriptor_name);
```
jbobba: remove.

test/debugger.cpp
```diff
@@ -0,0 +1,226 @@
+//*****************************************************************************
```
jbobba: Move to cpu_debugger.cpp

test/debugger.cpp
```diff
@@ -0,0 +1,226 @@
...
+#include "ngraph/runtime/cpu/cpu_layout_descriptor.hpp"
+#include "ngraph/runtime/cpu/cpu_tensor_view.hpp"
+#include "ngraph/runtime/cpu/op/sigmoid_mul.hpp"
+#include "ngraph/util.hpp"
+#include "util/test_tools.hpp"
+
+using namespace ngraph;
+using namespace std;
+
+TEST(debugger, add_breakpoint)
```
jbobba: cpu_debugger

PR #1806, opened 2018-10-11
Leona/10112018

indie:
Update landpage graphic and make Release Notes page import from `changes.md` automagically


README.md
```diff
@@ -24,18 +25,29 @@ whatever scenario you need.
...
-![nGraph ecosystem][ngraph-ecosystem]
+See the [Release Notes] for recent changes.
+
+
+|  Framework     | bridge available? | ONNX support?  |
+|----------------|-------------------|----------------|
+| TensorFlow*    | yes               | yes            |
+| MXNet*         | yes               | yes            |
+| PaddlePaddle   | yes               | yes            |   
+| PyTorch*       | not required      | yes            |
```
diyessi: Since "not required" is because of ONNX support, and they all have ONNX support, they should then all say "not required" so I think "no" is better, and more accurate since ONNX only provides partial support for a framework -- inference only, and only if your model can be expressed in ONNX.

PR #1807, opened 2018-10-11
Type prop tests for Quantize/Dequantize

aprocter:
Added some type prop tests for Quantize/Dequantize, so I'll have a reliable baseline for implementing the partial type/shape validation for these.

I did make one change to the constraints: while both the input and scale types for Quantize (and respectively, the output and scale types for Dequantize) were validated to be floating-point type, there was no check that they were the _same_ floating-point type, which I think was the intent. I can change it back if this is not correct.


src/ngraph/op/dequantize.cpp
```diff
@@ -50,9 +50,9 @@ void op::Dequantize::validate_and_infer_types()
     NODE_VALIDATION_ASSERT(this, m_type.is_real()) << "Output element type (" << m_type
                                                    << ") must be a floating point number";
 
-    NODE_VALIDATION_ASSERT(this, get_input_element_type(SCALE).is_real())
+    NODE_VALIDATION_ASSERT(this, get_input_element_type(SCALE) == m_type)
```
aprocter: This is the change I alluded to in the PR description.

aprocter:
Adding @Krovatkin to cover for @adstraw, who is off work.

src/ngraph/op/dequantize.cpp
adstraw: My original PR has == m_type (instead of is_real) and @diyessi asked me to change.  I like it better this way, TBH.

adstraw:
I see @diyessi has already approved so I am assuming he is OK with the change to m_type vs. is_real

PR #1817, opened 2018-10-12
Refactor convolution type prop

aprocter:
This is an overhaul of type propagation for both fprop and (especially) bprop, which will help greatly with dynamic types/shapes.

Old algorithm for bprop type propagation:

```
- Do all the fancy transformations to translate bprop into fprop with axes switched around, filters flipped, etc.
- Pass parameters for the above to a common infer_convolution_shape function.
```

This was error prone, and generated error messages that were hard to interpret since the original parameters from the graph were transformed internally.

New algorithm for bprop type propagation:

```
- Validate shapes as if we were doing forward prop, using the same function we would use for fprop.
- Make sure that the incoming output-delta matches the inferred fprop shape.
- Set output shape to the shape of the input that we are backpropping onto.
```

The new validation function (which I may refactor in a later PR into something that's also usable for pooling ops and batch norm) lives in the newly-created `validation_util.hpp`. CPU backend still uses the old function, so I have kept it for legacy usage for now; I can submit a separate PR to update the fused CPU conv ops to use the new function.

A heads-up: as I was working on this I found that some cases were slipping by which I don't think we intended to support. (Thus this technically `Changes API`, or at least the "de facto" API.) Specifically, data batches with dim-zero spatial axes were being allowed, and there is a test for this in ZeroDimElimination that no longer works as a result (because you can't construct it in the first place). I have removed that test, but am open to discussing a different solution.


test/zero_dim_tensor_elimination.cpp
```diff
@@ -55,26 +55,6 @@ TEST(zero_dim_tensor_elimination, zero_sum)
     ASSERT_EQ(count_ops_of_type<op::Sum>(f), 0);
 }
 
-TEST(zero_dim_tensor_elimination, zero_const_conv)
-{
-    Shape zero_shape{0};
-    auto A = std::make_shared<op::Parameter>(element::f32, Shape{1, 1, 0});
-    auto weights = std::make_shared<op::Parameter>(element::f32, Shape{1, 1, 4});
```
aprocter: Attn: @Krovatkin. This is actually invalid (weights are bigger than the data batch along spatial dims, and dims has a spatial dimension of zero, neither of which are "supposed to be" allowed), so this test is relying on a bug in validation. We could relax the spec of the convolution op to allow dim-zero axes on both data and filters.

test/zero_dim_tensor_elimination.cpp
```diff
@@ -55,45 +55,6 @@ TEST(zero_dim_tensor_elimination, zero_sum)
...
-    pass_manager.register_pass<ngraph::pass::ZeroDimTensorElimination>();
-    pass_manager.register_pass<pass::VisualizeTree>("after.pdf");
-    pass_manager.run_passes(f);
-    ASSERT_EQ(count_ops_of_type<op::Convolution>(f), 0);
-}
-
-TEST(zero_dim_tensor_elimination, zero_const_avg_pool)
-{
-    Shape zero_shape{0};
-    auto A = std::make_shared<op::Parameter>(element::f32, Shape{1, 1, 0});
```
aprocter: Attn @Krovatkin: Same thing here.

postrational:
Please note the failing test in ONNX CI:
```
        return AvgPool(data_batch, Shape(window_shape), Strides(window_strides), Shape(padding_below),

>                      Shape(padding_above), include_padding)

E       RuntimeError: While validating node 'AvgPool[AvgPool_102809](Concat_102808)' of type 'AvgPool':

E       Assertion 'window_shape[i] <= data_shape[i]' failed at /root/ngraph/src/ngraph/validation_util.cpp:71:

E       Window shape (Shape{7, 7}) is larger than data shape (Shape{6, 6}) at axis 0.
```

Let us know if you need our assistance with addressing it.

aprocter:
Thanks @postrational. That test should have passed, and I'm a bit annoyed with myself for not including an equivalent test in the C++ tests. I'll need to add some. I'm actually going to move this back to `[WIP]` for now.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -533,10 +533,10 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_zero_padded_conv_backprop_
         pad_input, pad_value, Shape{0, 0, 0, 0}, Shape{0, 0, 0, 0}, Shape{0, 0, 0, 0});
     auto pad_label = std::make_shared<pattern::op::Label>(pad, nullptr, NodeVector{pad});
 
-    auto output_delta = std::make_shared<pattern::op::Label>(element::f32, Shape{1, 1, 1, 1});
+    auto output_delta = std::make_shared<pattern::op::Label>(element::f32, Shape{1, 1, 3, 3});
```
jbobba: @aprocter Shape{1, 1, 1, 1} seems more accurate?

aprocter: I think you're right, and this test was getting caught up in the same error that @postrational identified. I'll give it a try and likely revert.

aprocter: Confirmed- this has been reverted

aprocter:
Just realized that the changes I made to @Krovatkin's zero-dim stuff are unnecessary as well. Back to `[WIP]`! :3

aprocter:
Okay, *NOW* I think we're good. I removed the "Changes API" tag because it turns out that all the "bad" cases I thought I had identified were legitimate cases of dimension-zero data batches with padding.

PR #1821, opened 2018-10-14
hybrid at core

sasadep:
* bare minimum hybrid infrastructure. 
* currently using backend unit test for testing.


src/ngraph/runtime/hybrid/hybrid_backend.cpp
```diff
@@ -0,0 +1,226 @@
...
+shared_ptr<runtime::Backend> runtime::hybrid::HYBRIDBackend::get_cached_backend(Placement placement)
+{
+    if (m_cached_backends.find(placement) == m_cached_backends.end())
+    {
+        m_cached_backends[placement] = runtime::Backend::create(placement_to_string(placement));
+    }
+    return m_cached_backends.at(placement);
+}
+
+shared_ptr<runtime::Tensor> runtime::hybrid::HYBRIDBackend::create_tensor(const element::Type& type,
```
rkimballn1: This function worries me. What if this is a GPU tensor? You need to create the tensor on the proper device but I'm not sure how right now.

src/ngraph/runtime/hybrid/hybrid_backend.cpp
```diff
@@ -0,0 +1,226 @@
...
+            Placement placement = get_colocated_function_placement(sub_function);
+            auto backend = get_cached_backend(placement);
+            backend->compile(sub_function);
+        }
+    }
+    NGRAPH_INFO << "hybrid compile -End ";
+    return true;
+}
+
+bool runtime::hybrid::HYBRIDBackend::call(shared_ptr<Function> function,
```
rkimballn1: This entire function is too inefficient. Too many new objects created. We have to make this work with minimal new objects and copies.

src/ngraph/runtime/hybrid/hybrid_backend.cpp
diyessi: What if there is no proper device, i.e. the tensor is used on more than one device?

rkimballn1:
This does not pass tests if you don't build the CPU backend. I normally don't build the CPU backend.

src/ngraph/runtime/backend.cpp
```diff
@@ -107,3 +107,10 @@ void runtime::Backend::validate_call(shared_ptr<const Function> function,
         }
     }
 }
+
+bool runtime::Backend::is_supported(const Node& node) const
+{
```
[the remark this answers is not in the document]
rkimballn1: I was only making it behave they way it does today. If we make it false then it operates differently than current backends.

sasadep:
@rkimballn1 hybrid needs at least two backends to work in its current state 
CPU and Interpreter is the first two that I started with.

PR #1826, opened 2018-10-15
Minor GPU code cleanup

rkimballn1:


src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -241,9 +241,9 @@ void runtime::gpu::GPU_ExternalFunction::emit_timer_functions()
         m_writer << "// Declare debug timers\n";
         vector<string> names;
         size_t index = 0;
-        for (shared_ptr<Function> current_function : m_pass_manager.get_state().get_functions())
+        for (const auto& p : m_function_ordered_ops)
         {
-            for (shared_ptr<Node> node : m_function_ordered_ops.at(current_function))
+            for (shared_ptr<Node> node : p.second)
```
csullivan: nit: Use of auto& above, and shared_ptr<Node> here? Maybe both `auto&`, but at least both by reference? even better `auto& const`.

src/ngraph/runtime/gpu/gpu_external_function.cpp
```diff
@@ -296,9 +296,9 @@ void runtime::gpu::GPU_ExternalFunction::emit_timer_functions()
 void runtime::gpu::GPU_ExternalFunction::emit_constant_declarations()
 {
     m_writer << "// Declare all constants\n";
-    for (shared_ptr<Function> current_function : m_pass_manager.get_state().get_functions())
+    for (const auto& p : m_function_ordered_ops)
```
csullivan: same as above, and for the rest of the PR

PR #1831, opened 2018-10-16
Add support for perf counter collection in DEX

jbobba:
Supports --timing_detail in nbench.


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -414,6 +412,12 @@ void runtime::cpu::CPU_ExternalFunction::compile()
     writer << "// Generated by the nGraph CPU backend\n";
     if (m_use_tbb)
     {
+        if (runtime::cpu::IsTracingEnabled() || m_emit_timing)
+        {
+            NGRAPH_WARN
```
rkimballn1: Libraries should not print things. Remove the warning or find some way to get this information to an app like nbench.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1143,6 +1147,13 @@ void runtime::cpu::CPU_ExternalFunction::build()
     {
         return;
     }
+
+    if (m_use_tbb && (runtime::cpu::IsTracingEnabled() || m_emit_timing))
+    {
+        NGRAPH_WARN << "CPU Backend: Tracing and performance breakdowns might not be accurate with "
```
rkimballn1: Remove warning

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1615,6 +1663,49 @@ const runtime::cpu::LayoutDescriptorPtrs&
     return result_layout_descriptors;
 }
 
+const vector<runtime::PerformanceCounter>& runtime::cpu::CPU_ExternalFunction::get_perf_counters()
+{
+    if (m_direct_execution)
+    {
+        return m_perf_counters;
+    }
+#if !defined(NGRAPH_DEX_ONLY)
```
rkimballn1: I don't understand why perf counters can't be used with dex.

jbobba: Not sure if I follow your concern. They are being added to DEX with this PR.

src/ngraph/runtime/cpu/cpu_external_function.cpp
jbobba: This is only going to kick in for ngraph developers. Dont want to spend the time implementing the plumbing for this. Will make this an assert for now.

PR #1839, opened 2018-10-16
Add builder for {de}quantize to make API's consistent and support {de}quantize with mkldnn

nbpatel:


ashokei:
seems missing unit-tests...you may want to reuse previous `quantizeCPU` unit-tests that used to take min/max values as inputs?

src/ngraph/builder/quantization.cpp
```diff
@@ -33,6 +34,33 @@ namespace ngraph
 {
     namespace builder
     {
+        std::shared_ptr<Node> ScaledQuantize(std::shared_ptr<Node> input,
```
ashokei: declaration in header?

src/ngraph/builder/quantization_util.hpp
```diff
@@ -99,6 +99,59 @@ namespace ngraph
...
+                                                   const std::shared_ptr<Node> max_input,
+                                                   bool is_signed)
+            {
+                auto min_input_const_op = std::static_pointer_cast<ngraph::op::Constant>(min_input);
+                auto max_input_const_op = std::static_pointer_cast<ngraph::op::Constant>(max_input);
+                float input_min_range =
+                    *(static_cast<float const*>(min_input_const_op->get_data_ptr()));
+                float input_max_range =
+                    *(static_cast<float const*>(max_input_const_op->get_data_ptr()));
+
```
sayantan-nervana: Can we use `get_vector` here (instead of get_data_ptr), assert that it is of size 1 and then read the first element into input_min_range

src/ngraph/builder/quantization_util.hpp
```diff
@@ -99,6 +99,59 @@ namespace ngraph
...
+                    std::max(1.0f, std::max(fabsf(input_min_range), fabsf(input_max_range))) /
+                    100.0f;
+                max_range = std::max(input_max_range, min_range + epsilon);
+                max_range = std::max(0.0f, max_range);
+                // end code copied and pasted from
+                // github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/quantize_op.cc
+                const float max_abs = std::max(std::abs(min_range), std::abs(max_range));
+                const float target_range =
+                    static_cast<float>((is_signed ? std::pow(2, 7) : std::pow(2, 8)) - 1);
+                max_range = max_abs;
```
sayantan-nervana: this is hard coded to num_bits = 8, which works for now because we only support qi8 and qu8

We could have num_bits passed to this function, in case there are plans for qi16, qu16, q132 etc

src/ngraph/builder/quantization_util.hpp
```diff
@@ -99,6 +99,59 @@ namespace ngraph
...
+                float input_min_range =
+                    *(static_cast<float const*>(min_input_const_op->get_data_ptr()));
+                float input_max_range =
+                    *(static_cast<float const*>(max_input_const_op->get_data_ptr()));
+
+                const float max_abs =
+                    std::max(std::abs(input_min_range), std::abs(input_max_range));
+                const float target_range =
+                    static_cast<float>((is_signed ? std::pow(2, 7) : std::pow(2, 8)) - 1);
+                const float scale_factor = max_abs / target_range;
```
sayantan-nervana: same comment as above. assumes num_bits = 8

src/ngraph/builder/quantization_util.hpp
nbpatel: trying to understand the advantage of it. Also, the plan in future is to have channel specific scales, so the mins and maxes will we vectors of length > 1

src/ngraph/builder/quantization.cpp
nbpatel: my bad

src/ngraph/builder/quantization.cpp
```diff
@@ -33,6 +34,33 @@ namespace ngraph
...
+
+        std::shared_ptr<Node> ScaledDequantize(std::shared_ptr<Node> input,
+                                               std::shared_ptr<Node> min,
+                                               std::shared_ptr<Node> max,
+                                               const ngraph::element::Type& type,
+                                               const ngraph::AxisSet& axes)
+        {
+            float scale =
+                builder::quantization_util::get_dequantize_scale(min, max, type.is_signed());
+            auto offset = op::Constant::create(element::f32, Shape{1}, {0});
```
ashokei: offset type should be same as output type ?

src/ngraph/builder/quantization_util.hpp
ashokei: atleast in mxnet, these are passed as vectors. I would prefer vector too. (we currently have to use extra reshape)

sayantan-nervana: Ah, I see. In TF min/max is a single number, hence was thinking that the min/max const is of size 1. If per-channel info is to be supported then no sense assert size==1

nbpatel:
Added the mkldnn support for {de}quantize with mkldnn as well in this PR.

src/ngraph/builder/quantization_util.hpp
ashokei: @nbpatel can we make min/max be vectors ? above static_cast will need to change accordingly.

src/ngraph/builder/quantization_util.hpp
```diff
@@ -99,6 +99,59 @@ namespace ngraph
                     (std::pow(2, -24) * static_cast<double>(max_abs32 / max_abs8)));
                 return scale;
             }
+
+            static inline float get_quantize_scale(const std::shared_ptr<Node> min_input,
+                                                   const std::shared_ptr<Node> max_input,
+                                                   bool is_signed)
+            {
+                auto min_input_const_op = std::static_pointer_cast<ngraph::op::Constant>(min_input);
+                auto max_input_const_op = std::static_pointer_cast<ngraph::op::Constant>(max_input);
```
adstraw: would it be safer to dynamic cast and check for nullptr?

src/ngraph/builder/quantization_util.hpp
adstraw: my intention is to add type traits (min, max) to each element type.  this way, you could have, for example, a int8 with range [-128, 127] and also qint8 (or some such name) with range [-127, 127].  so, why don't we pass `type` to this function and then call `is_signed` on that type?  you could also call `bitwidth` for now to do this calculation.

src/ngraph/builder/quantization_util.hpp
```diff
@@ -99,6 +99,59 @@ namespace ngraph
                     (std::pow(2, -24) * static_cast<double>(max_abs32 / max_abs8)));
                 return scale;
             }
+
+            static inline float get_quantize_scale(const std::shared_ptr<Node> min_input,
```
adstraw: We might need a template function here given that `scale` may either be `float` or `double` and the type is tightly controlled by Quantize / Dequantize.  If Quantize input is float, scale is float.  If Quantize input is double, scale is double.  Same goes for Dequantize ouput matching scale.

src/ngraph/builder/quantization_util.hpp
```diff
@@ -99,6 +99,59 @@ namespace ngraph
...
+                const float max_abs = std::max(std::abs(min_range), std::abs(max_range));
+                const float target_range =
+                    static_cast<float>((is_signed ? std::pow(2, 7) : std::pow(2, 8)) - 1);
+                max_range = max_abs;
+                min_range = is_signed ? -max_abs : 0;
+                const float scale = target_range / max_abs;
+                return scale;
+            }
+
+            static inline float get_dequantize_scale(const std::shared_ptr<Node> min_input,
```
adstraw: does it make sense to parameterize `get_quantize_scale` so we can use a common function?

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -775,6 +778,43 @@ namespace ngraph
...
+                template <>
+                void CPUAssignment::ASSIGN_DECL(ngraph::op::Quantize)
+                {
+                    auto quantize = static_cast<op::Quantize*>(node);
+                    auto offset_const_op =
+                        std::static_pointer_cast<ngraph::op::Constant>(quantize->get_argument(2));
+                    float offset = *(static_cast<float const*>(offset_const_op->get_data_ptr()));
+                    op::Quantize::RoundMode round_mode = quantize->get_round_mode();
+                    if (node->get_input_element_type(0) == element::f32 && offset == 0 &&
+                        round_mode == op::Quantize::RoundMode::HALF_TO_EVEN)
```
adstraw: seems like there should be a "TODO" here at least.  need to hook up MKLDNN directly to Quantize round mode once HALF_TO_EVEN is supported.  need to fall back to reference kernel for HALF_AWAY_FROM_ZERO.

src/ngraph/builder/quantization_util.hpp
nbpatel: so are you guys suggesting having the min and max as vectors in the builder and then converting to a ngraph constant op before calling the ngraph quantize?

src/ngraph/builder/quantization.cpp
```diff
@@ -33,6 +34,34 @@ namespace ngraph
...
+        std::shared_ptr<Node> ScaledQuantize(std::shared_ptr<Node> input,
+                                             std::shared_ptr<Node> min,
+                                             std::shared_ptr<Node> max,
+                                             const ngraph::element::Type& type,
+                                             const ngraph::AxisSet& axes,
+                                             op::Quantize::RoundMode round_mode)
+        {
+            float scale =
+                builder::quantization_util::get_quantize_scale(min, max, type.is_signed());
+            auto offset = op::Constant::create(element::f32, Shape{1}, {0});
```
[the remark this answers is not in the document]
sayantan-nervana: get_quantize_scale accepts 2 nodes, typecasts them to const node, reads data from them, and computes float scaleoffset. That is a slightly mixed api because it accepts nodes and outputs float.

a) get_quantize_scale should accept vectors or floats and spit out floats (so its a function that does not accept ngraph nodes)
or
b) get_quantize_scale accepts min-max ngraph nodes and returns scale-offset ngraph nodes.

I prefer approach (b) because: for now we just accept min-max as consts and compute the scale offset in c++ code. the problem is that the quantize ops must declare min-max as static inputs (must be const or inputs to encapsulates, that they cannot be in the middle of the encapsulate or in the middle of an ngraph graph)
later we could have ngraph nodes compute scale-offset from min-max nodes. In that case option (b) would create the subgraph that converts min-max to scale-offset nodes.

src/ngraph/builder/quantization_util.hpp
diyessi: Yes. We shouldn't even be assuming they are constant.

src/ngraph/builder/quantization_util.hpp
diyessi: We'll need per-channel. Add a broadcast if they are scalars.

src/ngraph/builder/quantization.cpp
diyessi: Needs to go in cpp file.

diyessi: yes

src/ngraph/builder/quantization_util.hpp
```diff
@@ -74,23 +74,23 @@ namespace ngraph
...
-                    input_min, input_max, filter_min, filter_max, &min_out_value, &max_out_value);
+                quantization_range_for_multiplication<uint8_t, int8_t, int32_t>(input_min[0],
+                                                                                input_max[0],
+                                                                                filter_min[0],
+                                                                                filter_max[0],
+                                                                                &min_out_value,
+                                                                                &max_out_value);
                 const float max_abs32 = std::max(std::abs(min_out_value), std::abs(max_out_value));
-                const float max_abs8 = std::max(std::abs(output_min), std::abs(output_max));
+                const float max_abs8 = std::max(std::abs(output_min[0]), std::abs(output_max[0]));
```
adstraw: Should there be an error message if these values are not scalar?

src/ngraph/builder/quantization_util.hpp
```diff
@@ -99,6 +99,55 @@ namespace ngraph
...
+                    throw ngraph_error("min_input shouldn't be nullptr!");
+                else if (max_input_const_op == nullptr)
+                    throw ngraph_error("max_input shouldn't be nullptr!");
+
+                auto input_min_range = min_input_const_op->get_vector<T>();
+                auto input_max_range = max_input_const_op->get_vector<T>();
+
+                T min_range;
+                T max_range;
+                if (is_quantize)
```
adstraw: There may come a time, some day, where we want to adjust min/max for Dequantize as well.  Hence, this flag could take a more generic name based on what this code is doing:  adjusting a too-close min/max by delta.

src/ngraph/builder/quantization.cpp
```diff
@@ -83,7 +144,8 @@ namespace ngraph
                                                              padding_below,
                                                              padding_above,
                                                              data_dilation_strides,
-                                                             requantization_scale);
+                                                             requantization_scale,
+                                                             with_relu);
```
adstraw: what is this?

test/backend_test.in.cpp
```diff
@@ -5518,6 +5518,38 @@ NGRAPH_TEST(${BACKEND_NAME}, dequantize)
               read_vector<output_c_type>(y));
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, dequantize_cpu)
```
adstraw: why just a test for Dequantize and not Quantize?  Maybe existing tests for Quantize round mode, etc. covered it?

src/ngraph/builder/quantization_util.hpp
```diff
@@ -99,6 +99,55 @@ namespace ngraph
...
+            {
+                auto min_input_const_op =
+                    std::dynamic_pointer_cast<ngraph::op::Constant>(min_input);
+                auto max_input_const_op =
+                    std::dynamic_pointer_cast<ngraph::op::Constant>(max_input);
+
+                if (min_input_const_op == nullptr)
+                    throw ngraph_error("min input must be constant");
+                else if (max_input_const_op == nullptr)
+                    throw ngraph_error("max input must be constant");
```
mbrookhart: The const assumption here breaks the functionality of the general quantize/dequantize ops in both TF and MXNet. I cannot implement these ops in the bridge using this builder as it stands and still pass unit tests. Thus, I approve under strong objection mostly to move the other projects forward.

src/ngraph/builder/quantization_util.hpp
```diff
@@ -99,6 +99,55 @@ namespace ngraph
...
+                                                   const std::shared_ptr<Node> max_input,
+                                                   const ngraph::element::Type& type,
+                                                   bool bump_by_eps = false)
+            {
+                auto min_input_const_op =
+                    std::dynamic_pointer_cast<ngraph::op::Constant>(min_input);
+                auto max_input_const_op =
+                    std::dynamic_pointer_cast<ngraph::op::Constant>(max_input);
+
+                if (min_input_const_op == nullptr)
```
rkimballn1: All code blocks need curly braces

src/ngraph/runtime/cpu/builder/quantization.cpp
```diff
@@ -34,122 +37,171 @@ namespace ngraph
...
+                    auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
+                    auto input_desc = mkldnn_utils::get_input_mkldnn_md(node, 0);
+                    auto result_desc = mkldnn_utils::get_output_mkldnn_md(node, 0);
 
-                const ngraph::op::Quantize* quantize =
-                    static_cast<const ngraph::op::Quantize*>(node);
-                function<void(CPURuntimeContext*)> functor;
+                    auto scale_const_op =
+                        std::dynamic_pointer_cast<ngraph::op::Constant>(quantize->get_argument(1));
+                    if (scale_const_op == nullptr)
```
rkimballn1: curly braces

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -4670,31 +4670,79 @@ namespace ngraph
...
-                       << static_cast<int>(quantize->get_round_mode()) << "));\n";
+                if (runtime::cpu::mkldnn_utils::use_mkldnn_kernel(node))
+                {
+                    auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
+                    auto input_data_desc = mkldnn_utils::get_input_mkldnn_md(node, 0);
+                    auto result_desc = mkldnn_utils::get_output_mkldnn_md(node, 0);
+
+                    auto scale_const_op =
+                        std::dynamic_pointer_cast<ngraph::op::Constant>(quantize->get_argument(1));
+                    if (scale_const_op == nullptr)
```
rkimballn1: curly braces

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -797,6 +799,57 @@ namespace ngraph
...
+                }
+
+                template <>
+                void CPUAssignment::ASSIGN_DECL(ngraph::op::Quantize)
+                {
+                    auto quantize = static_cast<op::Quantize*>(node);
+                    auto offset_const_op =
+                        std::static_pointer_cast<ngraph::op::Constant>(quantize->get_argument(2));
+                    op::Quantize::RoundMode round_mode = quantize->get_round_mode();
+                    if (round_mode != op::Quantize::RoundMode::ROUND_NEAREST_TOWARD_EVEN)
```
rkimballn1: curly braces

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -797,6 +799,57 @@ namespace ngraph
...
+                        return;
+                    if (node->get_output_element_type(0) == element::u8)
+                    {
+                        auto offset = offset_const_op->get_vector<uint8_t>();
+                        if (offset[0] != 0)
```
rkimballn1: curly braces

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -797,6 +799,57 @@ namespace ngraph
...
+                            return;
+                    }
+                    if (node->get_output_element_type(0) == element::i8)
+                    {
+                        auto offset = offset_const_op->get_vector<int8_t>();
+                        if (offset[0] != 0)
```
rkimballn1: curly braces

src/ngraph/runtime/gpu/unit_test.manifest
```diff
@@ -53,6 +53,7 @@ quantize_int8
 quantize_clamp
 dequantize
 dequantize_axes
+dequantize_cpu
```
rkimballn1: This test is not a generalized backend unit test (not NGRAPH_TEST) so it does not need to be in the manifest

src/ngraph/runtime/intelgpu/unit_test.manifest
```diff
@@ -19,6 +19,7 @@ batch_norm_one_output
 batch_norm_three_outputs
 dequantize
 dequantize_axes
+dequantize_cpu
```
rkimballn1: This test is not a generalized backend unit test (not NGRAPH_TEST) so it does not need to be in the manifest

PR #1840, opened 2018-10-16
Add some fixes to speed up RN50 training

mbrookhart:
make_zeros was adding an extra OutputElement to Batchnorm that broke fprop_cache performance a little, and the Relu change enables BN+relu and Conv+relu fusion for training


mbrookhart:
Props to @Krovatkin for figuring out why I was getting 5 output elements from Batchnorm.

PR #1841, opened 2018-10-16
Cache and use fprop stats in cudnn batchnorm bprop

csullivan:
This adds a new CUDNNBatchNorm op which has two extra outputs, batch mean and batch inverse variance. These are then passed directly into BatchNormBackprop as the mean and variance parameters. The mean output is equivalent to the batch mean output when the cumulative moving average factor is 1.0. 

`gpu::pass::BatchNormCache` replaces any BatchNorm arguments to BatchNormBackprop with CUDNNBatchNorm.

This needs to be updated to target new batchnorm spec.


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -335,17 +336,28 @@ void runtime::gpu::GPU_Emitter::emit_BatchNorm(EMIT_ARGS)
...
 }
 
+void runtime::gpu::GPU_Emitter::emit_BatchNorm(EMIT_ARGS)
+{
+    ::emit_BatchNorm(external_function, writer, node, args, out, false);
+}
+
+void runtime::gpu::GPU_Emitter::emit_CUDNNBatchNorm(EMIT_ARGS)
+{
+    ::emit_BatchNorm(external_function, writer, node, args, out, true);
```
fengleitian: name it something like 'emit_BatchNormWithSavedMeanVar'?

csullivan: How about `emit_CUDNNBatchNorm -> emit_BatchNormWithStats`. We change the op name too? `CUDNNBatchNorm` -> `BatchNormWithStats`? 

Note that when the new API for batch norm goes in, it would be `BatchNormTraining` and `BatchNormTrainingWithStats`

test/backend_test.in.cpp
```diff
@@ -18,11 +18,13 @@
 #include <cinttypes>
 #include <cmath>
 #include <cstdlib>
+#include <fstream>
```
gcwenger: Is this new include needed?

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: Good.

test/backend_test.in.cpp
```diff
@@ -5865,3 +5866,34 @@ NGRAPH_TEST(${BACKEND_NAME}, quantize_ROUND_DOWN)
...
+        {1.0f},       // gamma
+        {1.0f},       // beta
+        {1.1f, 1.0f}, // x
+        {1.0f, 1.0f}, // dy
+    };
+
+    auto func = std::make_shared<Function>(dx, op::ParameterVector{g, b, input, delta});
+    auto results = execute(func, args, "${BACKEND_NAME}");
+    EXPECT_TRUE(test::all_close_f(std::vector<float>{350.957, -388.67}, results.at(0)));
+}
```
fengleitian: no new line at the end of file.

PR #1842, opened 2018-10-16
Partial Shapes and Types, Part 4j: Quantize/Dequantize

aprocter:
(Branches from #1756)

~Work in progress~ All done:
- [x] Implement Quantize
- [x] Tests for Quantize
- [x] Implement Dequantize
- [x] Tests for Dequantize


src/ngraph/op/dequantize.cpp
```diff
@@ -35,53 +35,89 @@ op::Dequantize::Dequantize(shared_ptr<Node> input,
...
         SCALE,
         OFFSET
     };
 
-    set_output_size(1);
-    set_output_type(0, m_type, get_input_shape(INPUT));
-
-    NODE_VALIDATION_ASSERT(this, get_input_element_type(INPUT).is_quantized())
-        << "Input element type (" << get_input_element_type(INPUT) << ") must be a quantized type";
+    NODE_VALIDATION_ASSERT(this, m_type.is_static()) << "Output element type must not be dynamic";
```
aprocter: Something to think about: do we want this restriction? We could also allow `element::dynamic` for the output type, which would have this behavior:

```
m_type is dynamic, scale input type is dynamic -> output is dynamic
m_type is dynamic, scale input type is foo -> output is foo
m_type is foo, scale input type is dynamic -> output is foo
m_type is foo, scale input type is foo -> output is foo
m_type is foo, scale input type is bar -> error
```

aprocter: I think it can. In the current (fully static) system, `m_type` has to be the same as the `scale` input.

aprocter: (This means `m_type` is technically redundant, FWIW.)

PR #1828, opened 2018-10-15
BatchNorm splitting into ops (2nd try)

Krovatkin:

PR #1843, opened 2018-10-17
Cache and use fprop stats in cudnn batchnorm bprop [Pt. 2]

csullivan:
Update to the new batch norm API introduced in #1828.


csullivan:
Updated #1841 to match newly merged batch norm API and so this PR is defunct.

PR #1848, opened 2018-10-17
IntelGPU backend: Enable nBench statistic

shssf:


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -278,7 +278,20 @@ extern "C" void delete_backend(runtime::Backend* backend)
 
 runtime::intelgpu::IntelGPUBackend::IntelGPUBackend()
 {
-    ocl_engine = make_shared<cldnn::engine>();
+    bool profiling = false;
+
+    if (getenv("NGRAPH_INTELGPU_STAT") != nullptr)
```
rkimballn1: I don't understand what this env var is for? Performance monitor is enabled via enable_performance_data()

shssf: In case of INTELGPU backend we have to create `cldnn::engine` with parameter "`profiling == true`". I can't use `enable_performance_data()` for this because it is too late (`cldnn::engine` created already and I can't change this due to clDNN design).

postrational:
TEST NOW

postrational:
There was a build error in ONNX CI, rerunning.

PR #1851, opened 2018-10-17
cast to ArgMin/ArgMax based on reduce_op

fengleitian:
The result was correct, but it could be wrong if ArgMin/ArgMax has difference implementation of 'get_reduction_axis'.


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -176,23 +176,33 @@ void runtime::gpu::GPU_Emitter::emit_ArgMin(EMIT_ARGS)
...
     if (out[0].get_size() == 0)
     {
         return;
     }
-    auto argmax = static_cast<const ngraph::op::ArgMax*>(node);
-    std::vector<size_t> axes{argmax->get_reduction_axis()};
-    auto axis_set = AxisSet(axes);
+
+    size_t axis;
+    if (reduce_op == CUDNN_REDUCE_TENSOR_MIN)
```
rkimballn1: There are many more values than CUDNN_REDUCE_TENSOR_MIN and CUDNN_REDUCE_TENSOR_MAX. You need to handle all of the other values either explicitly or throw an exception. I would vote for exception.

fengleitian: updated.

PR #1853, opened 2018-10-17
Cuda < 9 compilation fix suggested by Chris Sullivan

gcwenger:


gcwenger:
This is known to be a quick fix because there are other combinations of CUDA and compilers that should work.

src/ngraph/runtime/gpu/CMakeLists.txt
```diff
@@ -72,8 +72,32 @@ if (NGRAPH_GPU_ENABLE)
 	    -gencode=arch=compute_70,code=compute_70)
     endif()
 
-    set_source_files_properties( ${CUDA_SRC} PROPERTIES CUDA_SOURCE_PROPERTY_FORMAT OBJ)
-    cuda_compile(CUDA_OBJ ${CUDA_SRC} STATIC)
+    set (DO_CUDA_COMPILE FALSE)
+    if (CUDA9_FOUND)
+        set (DO_CUDA_COMPILE TRUE)
```
fengleitian: no need this one?

src/ngraph/runtime/gpu/CMakeLists.txt
```diff
@@ -72,8 +72,32 @@ if (NGRAPH_GPU_ENABLE)
...
+        if (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU"))
+            # CUDA 9 supports up to gcc 6.x
+            if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0)
+                set (DO_CUDA_COMPILE TRUE)
+            endif()
+        # CUDA 9 supports up to clang 3.9
+        elseif (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.0)
+            set (DO_CUDA_COMPILE TRUE)
+        endif()
+    elseif (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU"))
```
fengleitian: need to check cuda8 found?

src/ngraph/runtime/gpu/CMakeLists.txt
rkimballn1: Not just no need, it is incorrect. Remove line 77

src/ngraph/runtime/gpu/CMakeLists.txt
```diff
@@ -72,8 +72,32 @@ if (NGRAPH_GPU_ENABLE)
...
+        # Current release of CUDA 8 supports up to gcc 5.4
+        if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.5)
+            set (DO_CUDA_COMPILE TRUE)
+        endif()
+    # CUDA 8 supports up to clang 3.8
+    elseif(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.9)
```
rkimballn1: Should check for clang, not just assume it is clang. There are other compilers that we currently don't use like windows and icc.

src/ngraph/runtime/gpu/CMakeLists.txt
fengleitian: oh, right.

fengleitian: what if you have cuda9 + cuda8 + clang3.8

src/ngraph/runtime/gpu/CMakeLists.txt
gcwenger: Ok - will rework. Thought maybe I was safe given this warning from another CMakeLists.txt:
message(WARNING "You are using an unsupported compiler. Compilation has only been tested with Clang (${CLANG_MIN_VERSION} and up), Apple Clang (${APPLE_CLANG_MIN_VERSION} and up), and GCC (${GCC_MIN_VERSION} and up).")

src/ngraph/runtime/gpu/CMakeLists.txt
gcwenger: Will fix.

src/ngraph/runtime/gpu/CMakeLists.txt
fengleitian: Is it possible have cuda9 + cuda8 + clang3.8 on system?

gcwenger: I'm not sure. We look for cuda9 first. Both cuda9 & cuda8 should be good with clang3.8.

csullivan: It is. To start the CMake searches for an optional CUDA 9 install. If it is not found, CUDA 8 is required. Thus, if both 8 and 9 are present CUDA 9 will be used. If clang-3.8 is used in that case, then it will be CUDA 9 + 3.8, which should work without issue.

fengleitian: I see, cuda 9 support up to 3.9.

fengleitian:
LGTM

PR #1855, opened 2018-10-18
Support for parameterized testing combined with backend driven manifests

gcwenger:


test/backend_test.in.cpp
```diff
@@ -7422,38 +7437,8 @@ NGRAPH_TEST(${BACKEND_NAME}, avg_pool_3d_strided_uneven_padded_do_not_include_in
...
-    }
-    auto int_results = execute(int_f, args, "INTERPRETER");
-    auto backend_results = execute(cpu_f, args, "${BACKEND_NAME}");
-    for (size_t i = 0; i < backend_results.size(); i++)
-    {
-        EXPECT_TRUE(test::all_close(backend_results.at(i), int_results.at(i), 1.0e-4f, 1.0e-4f));
-    }
-}
+// avg_pool_3d case generation
+NGRAPH_INSTANTIATE_TEST_CASE_P(${BACKEND_NAME}, include_pad, avg_pool_3d_params, testing::Bool(), );
```
rkimballn1: Is the comma at the end of the line required or is it just a stray? If it is required then there needs to be some comment.

gcwenger:
Overarching goal is to add support for an NGRAPH_TEST_F (equivalent to TEST_F, but with per backend support), NGRAPH_TEST_P (equivalent to TEST_P), and NGRAPH_INSTANTIATE_TEST_CASE_P (equivalent to INSTANTIATE_TEST_CASE_P).

Fixture and parametrized testing are powerful ways to increase backend testing code coverage with significantly less need to copy & paste unit test code (as we've been doing to date).
In this article:
https://blog.jetbrains.com/rscpp/unit-testing-google-test/
The "Test Fixtures" section is a useful description of TEST_F (stopping when he starts discussing how fixtures integrate with ReSharper). Similarly "How Do I Test Lots of Data?" section is a good description of TEST_P.
A useful technique that he doesn't cover is that you can do parameterized testing similar to what he describes, but use ValuesIn(method_that_generates_a_vector()).

Original documentation on TEST_F is found here:
https://github.com/google/googletest/blob/master/googletest/docs/primer.md
under "Test Fixtures: Using the Same Data Configuration for Multiple Tests".

Original documentation on TEST_P is found here:
https://github.com/google/googletest/blob/master/googletest/docs/advanced.md
under "How to Write Value-Parameterized Tests".

For backend test generation NGRAPH_TEST re-uses the test_case_name for the backend_name. But this doesn't work for NGRAPH_TEST_F and NGRAPH_TEST_P. The original TEST_F has a first parameter of the fixture class. Similarly the original TEST_P has a first parameter of the parameterized fixture class (TestWithParam<T> subclass).
So NGRAPH_TEST_F, NGRAPH_TEST_P, NGRAPH_INSTANTIATE_TEST_CASE_P introduce a 0th parameter for the backend_name and then maintain the same subsequent parameters as the original macros. Similarly NGRAPH_GTEST_TEST_ was modified to support backend_name and test_case_name (rather than assuming that backend_name and test_case_name are the same). Other changes percolated through to combine backend_name and test_case_name for unique naming and registration.

PR #1857, opened 2018-10-18
Flush denormals to zero in nbench

jbobba:
Random data used in nbench could lead to denormals during computations leading to unpredictable performance fluctuations. In real workloads, these denormals hopefully dont show up. 

http://carlh.net/plugins/denormals.php


jbobba:
@ashokei Yes, its been popping up. Mobilenet first layer, one of the last layers in inception v3 and if you run multiple models through nbench.

PR #1858, opened 2018-10-18
Primitive CPU Executor for managing thread pools and executing kernels across partitioned resources

jbobba:
Creates a single CPUExecutor which contains multiple thread pools. For now, all execution happens on thread pool '0' until we figure out a way to schedule graphs to more thread pools. 

KernelFunctors get an extra argument CPUExecutionContext which carries execution-time information.


src/ngraph/runtime/cpu/cpu_executor.cpp
```diff
@@ -0,0 +1,108 @@
...
+#else
+                        num_threads_per_pool = GetNumCores();
+#endif
+                        m_thread_pools.push_back(std::unique_ptr<Eigen::ThreadPool>(
+                            new Eigen::ThreadPool(num_threads_per_pool)));
+                        m_thread_pool_devices.push_back(std::unique_ptr<Eigen::ThreadPoolDevice>(
+                            new Eigen::ThreadPoolDevice(m_thread_pools[i].get(), GetNumCores())));
+                        m_tbb_arenas.emplace_back(1);
+                    }
+                }
```
pruthviIntel: @jbobba , please let me know if my understanding is correct.

With this PR changes we will be using  `Eigen::ThreadPool` for the eigen computation and `TBB::threadpool` for TBB on the arena define by CPUExecutionContext which is `0` for now. for each arena we can define number of threads by `ngraph_intra_op_parallelism` correct?.

 Should we also have method specific for TBB to define/bind the threads to the specific set of cores?. for example if we have two arena's 

arena_0 -> {0, 1,2, 3...13}
arena_1 -> {14, 15,....27}.

As per my understanding and discussion with Pablo, we can only restrict number of threads that participate in arena. but cannot defined `which threads` should participate in arena .

src/ngraph/runtime/cpu/builder/one_hot.cpp
```diff
@@ -47,7 +47,8 @@ namespace ngraph
                     std::function<decltype(runtime::cpu::kernel::one_hot_rank_0<float>)> kernel;
                     SELECT_KERNEL(
                         kernel, out[0].get_element_type(), runtime::cpu::kernel::one_hot_rank_0);
-                    auto functor = [&, kernel, out_shape, one_hot_axis](CPURuntimeContext* ctx) {
+                    auto functor = [&, kernel, out_shape, one_hot_axis](CPURuntimeContext* ctx,
+                                                                        CPUExecutionContext* ectx) {
                         kernel(arg_tensor, out_tensor, out_shape, one_hot_axis);
```
gaurides: @jbobba : I am trying to understand: when do you pass `ectx->arena` to a kernel and when not? Eg: the kernel calls in functor of dot.cpp have a new argument `ectx->arena` .

src/ngraph/runtime/cpu/builder/convolution.cpp
```diff
@@ -341,7 +349,8 @@ namespace ngraph
                                 node, args, out);
```
gaurides: These kernels if (!mkldnn), also don't have arena. do they need an arena?

src/ngraph/runtime/cpu/builder/dot.cpp
```diff
@@ -251,7 +256,7 @@ namespace ngraph
 
                 auto functor =
                     [&, kernel, arg0_shape, arg1_shape, result_shape, reduction_axes_count](
-                        CPURuntimeContext* ctx) {
+                        CPURuntimeContext* ctx, CPUExecutionContext* ectx) {
                         kernel(arg0_tensor,
```
gaurides: This kernel() call doesn't have arena. whereas other kernel calls in this file do.

src/ngraph/runtime/cpu/builder/convolution.cpp
jbobba: reference kernels dont run in arenas for now.

PR #1862, opened 2018-10-18
ArgReduce 64 bit indices

csullivan:


src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -236,44 +236,80 @@ size_t runtime::gpu::CUDNNEmitter::build_reduce_forward(const cudnnReduceTensorO
...
+                        debug_sync();
+                    }});
+            }
+        }
+        else
         {
             std::stringstream ss_er;
-            ss_er
-                << "Unsupported Type: Only uint32 currently supported for indices in op ArgReduce ";
+            ss_er << "Unsupported Type: Only uint32 & uint64 currently supported for indices in op "
```
fengleitian: nit: add a `output_type.c_type_string()` in error message, so we know what's is required.

fengleitian:
@gcwenger very good point, but if the array is supported by cuDNN, it's guarantee the dimension size < 32bit.

PR #1883, opened 2018-10-22
Build LLVM only when needed.

silee2:
I'm trying to control LLVM build as a first step to creating a smaller distribution of nGraph.
The current logic I have only builds LLVM if GPU is enabled or CPU is enabled and not DEX only.
Not sure how that would impact NNP and Intel IGP.
LLVM build can be controlled manually with a new option.


silee2:
"Currently llvm is only downloaded or built only when needed." Did not realize that it already works that way. Closing the PR.

PR #1888, opened 2018-10-22
Add PlaidML backend

earhart:


src/ngraph/runtime/plaidml/plaidml_backend.hpp
```diff
@@ -0,0 +1,65 @@
...
+              const std::vector<std::shared_ptr<runtime::Tensor>>& inputs) final;
+
+    void remove_compiled_function(std::shared_ptr<Function> func) final;
+
+    void save(std::shared_ptr<Function> func,
+              const std::string& filename,
+              plaidml_file_format format);
+
+private:
+    Config config_;
```
rkimballn1: ngraph coding standard, prepend `m_` to all class member variables. m_config, m_compiler, etc..

src/ngraph/runtime/plaidml/plaidml_backend.cpp
```diff
@@ -0,0 +1,96 @@
+/*******************************************************************************
```
rkimballn1: Our license headers have changed to use single line comments `//`. Please update all license headers

src/ngraph/runtime/plaidml/plaidml_builder.cpp
```diff
@@ -0,0 +1,850 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+#include "ngraph/runtime/plaidml/plaidml_builder.hpp"
+
```
rkimballn1: remove this line. Also `<>` headers go before `""`

src/ngraph/runtime/plaidml/plaidml_builder.hpp
```diff
@@ -0,0 +1,391 @@
...
+    Function& operator=(Function&&) = default;
+
+    // Builds the final string form of the function.
+    std::string to_string() const;
+
+    // Finalizes a function, transforming it into a PlaidML function application object.
+    vertexai::plaidml::application finalize() const;
+
+    // Adds an input to the function.
+    Function& add(Input input) &;
```
rkimballn1: What is the problem being solved with returning `&` and `&&`? I haven't used this before.

src/ngraph/runtime/plaidml/plaidml_compilation_cache.hpp
```diff
@@ -0,0 +1,54 @@
...
+        {
+            class CompilationCache;
+        }
+    }
+}
+// A compilation cacher.
+class ngraph::runtime::plaidml::CompilationCache final
+{
+public:
+    // Looks up the supplied function in the compilation cache.  If the function is not in the cache, returns an empty pointer.
```
rkimballn1: I didn't count but these comments look longer than 100 chars. Can we wrap them?

src/ngraph/runtime/plaidml/plaidml_builder.hpp
```diff
@@ -0,0 +1,391 @@
...
+    TernaryContraction&& add_constraints(L lambda) &&
+    {
+        lambda(std::back_inserter(constraints_));
+        return std::move(*this);
+    }
+
+private:
+    friend class Function;
+
+    std::string agg_op_;
```
rkimballn1: fix names to conform to coding standard

src/ngraph/runtime/plaidml/plaidml_convpool_formatter.cpp
```diff
@@ -0,0 +1,785 @@
...
+        .set_first(input)
+        .set_second(forward_output)
+        .set_third(incoming_deriv);
+}
+
+std::string ngraph::runtime::plaidml::ConvPoolFormatter::c()
+{
+    return "c";
+}
+std::string ngraph::runtime::plaidml::ConvPoolFormatter::ci()
```
rkimballn1: newline between functions.

src/ngraph/runtime/plaidml/plaidml_ops_batch_norm.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace plaidml
+        {
+            // BatchNorm implements batch normalization.
+            template <>
+            void Impl<op::BatchNorm>::operator()()
```
rkimballn1: A lot of these files would probably look better if we removed the nested namespaces and prepended ngraph::runtime::plaidml:: to the implementations.

aprocter:
Should probably designate `CODEOWNERS`(s) for `src/ngraph/runtime/plaidml`

src/ngraph/runtime/plaidml/plaidml_backend.hpp
earhart: Done

src/ngraph/runtime/plaidml/plaidml_backend.cpp
earhart: Done

src/ngraph/runtime/plaidml/plaidml_builder.cpp
earhart: Done.

(FWIW: This pattern comes from the Google coding standards; the idea is that each header should be the first header included by some .cpp file, to ensure that each header is properly self-contained.)

src/ngraph/runtime/plaidml/plaidml_builder.hpp
earhart: The idea was to make it easy to use `this` as either an lvalue and rvalue-reference, so that you could do either `MyObject foo; foo.bar().baz(); return foo.finalize()` or `return MyObject{}.bar().baz().finalize()`, depending on what reads best at the call site.

src/ngraph/runtime/plaidml/plaidml_compilation_cache.hpp
earhart: Done

src/ngraph/runtime/plaidml/plaidml_convpool_formatter.cpp
earhart: Done

src/ngraph/runtime/plaidml/plaidml_ops_batch_norm.cpp
earhart: Done.  (For some reason, I thought you couldn't specialize a template from another namespace; looks like I should've tried it before assuming.  :-)

src/ngraph/runtime/plaidml/plaidml_builder.hpp
earhart: Done

postrational:
Other external dependencies are automatically downloaded and built though cmake configurations. 
Do we want to do the same with PlaidML?

earhart:
We haven't released an SDK for PlaidML yet; it's currently just a Bazel rule that builds a tar file with the right contents.  I think we'd like to have a non-Python binary release of some sort, but it's not our highest priority, and I think we'd need to think a bit more about what it'd look like.  (E.g. on MacOS, it seems like we might want it to be a Framework.)

It might be possible to download PlaidML and compile it via shelling out to Bazel.  I suspect it'd take me a while to get that working, but it might be simple for someone with more expertise with cmake.  It'll probably be easier after our next open-source release (hopefully soon); that should include a stable source tarball.

(I very briefly thought about trying to talk nGraph devs into adding a Bazel build system, which would make it trivial to compile PlaidML as part of the build.  That seems like a much longer conversation, though, and doesn't really help unless it becomes the only build system.)

earhart:
FWIW -- I don't have access to the Jenkins server; anyone want to let me know what failed?  :-/

aprocter:
@earhart I put in the correct (I think :/) access requests for you. Can take a look at the CI for you later this afternoon if the approval hasn't come through by then.

earhart:
Awesome; thanks!  I'll keep checking it periodically.

aprocter:
Looks like @rkimballn1 already got it. The problem was the source code formatting check. You can run `maint/check-code-format.sh` to check the formatting, and `maint/apply-code-format.sh` to automatically reformat.

PR #1889, opened 2018-10-22
GraphRewrite until fixed point

Krovatkin:


src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -60,6 +60,38 @@
...
+// In this case, you should be able to request another pass of CPUFusion.
+// To request another pass, you will need to register fusions in a callback:
+// i.e. you will need to pass `this` into a callback and then call `this->construct_X`
+// This will schedule another pass of CPUFusion with the following fusion.
+// This approach should only be used if you are either:
+// a) need more than one fusion occur on the same node
+// b) you are modifying nodes after the current node in the topological order
+// c) there's no linear order of fusions which will give
+//    the correct final fusion. i.e. the same fusion needs to occur before and after some other fusion
+
```
jbobba: shouldn't this comment be in graph_rewrite.hpp or somewhere non-cpu specific, may be in core_fusion.cpp?

diyessi: Agree.

src/ngraph/pass/graph_rewrite.cpp
```diff
@@ -22,36 +22,72 @@
...
-                rewritten = true;
-                if (matcher->process_match())
+                NGRAPH_DEBUG << "Running matcher " << matcher->get_name() << "("
+                             << matcher->get_pattern()->get_name() << ") on " << node->get_name();
+                if (matcher->match(node))
                 {
-                    break;
+                    NGRAPH_DEBUG << "Matcher " << matcher << matcher->get_name() << " matched "
+                                 << node->get_name();
+                    rewritten = true;
```
diyessi: should be set inside the conditional below.

PR #1891, opened 2018-10-23
fix compiler warnings

rkimballn1:


cmake/clang_4_0_flags.cmake
```diff
@@ -43,3 +43,6 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-padded")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-sign-compare")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-parameter")
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-conversion")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-double-promotion")
```
aprocter: Do we want these flags permanently or are they just a quick workaround? If the latter, let's remember to open a JIRA to track.

PR #1899, opened 2018-10-23
[CI fix] Roll back #906 and #1853

csullivan:


fengleitian:
is there CI machine with only cuda9 on it.

crlishka:
@fengleitian The three servers we use in the Jenkins "algo" pool, for nGraph CUDA testing, all have CUDA 9 drivers on them.  The CUDA libraries are determined by the Dockerfile used, and @csullivan and I checked that the Dockerfile derives from an NVida docker image with CUDA 9.1 libraries.

fengleitian:
@crlishka Thank you. I was worried if we require cuda8, it will fail on cuda9. But @csullivan just tested and showed that it's compatible(cuda 9 is OK if you required cuda 8, which means at least 8).

PR #1901, opened 2018-10-23
remove 5arg BNTraining

Krovatkin:


aprocter:
Looks like `op::gpu::BatchNormTrainingWithStats` has a 5-arg ctor that is delegating to this. I'm guessing it can be cut out too.

mbrookhart:
Why remove it? We'll need this back in when we start training SSD/RCNN type models.

Krovatkin:
> Why remove it? We'll need this back in when we start training SSD/RCNN type models. 
@aprocter wants to split it into 2 ops. Might as well start from scratch

aprocter:
> @aprocter wants to split it into 2 ops. Might as well start from scratch

It doesn't follow that we want to eliminate functionality. @Krovatkin I think your reasoning for cutting out 5-arg BN was that it was broken for now anyway, right?

Krovatkin:
> It doesn't follow that we want to eliminate functionality. @Krovatkin I think your reasoning for cutting out 5-arg BN training was that it was broken for now anyway, right? (Meaning that we're not actually removing any existing functionality.)

Yes,  Matt told me that the op was broken and he wasn't using it yet.

PR #1903, opened 2018-10-24
Reenabled Chris's nvcc building

gcwenger:
Reenabled Chris's nvcc building. Improved support for nvcc include path and variations of cuda 8/9 + clang/gcc. On nervana-titanxp41 tested building the following combinations:
* cuda 8 + clang 3.9
* cuda 9 + clang 3.9
* cuda 8 + gcc
* cuda 9 + gcc


src/ngraph/runtime/gpu/CMakeLists.txt
```diff
@@ -44,15 +44,103 @@ set(SRC
...
+                --compiler-options -Wno-unused-macros;
+                --compiler-options -Wno-used-but-marked-unused)
+        endif()
+
+        message(STATUS "Doing cuda compile...")
+        set_source_files_properties( ${CUDA_SRC} PROPERTIES CUDA_SOURCE_PROPERTY_FORMAT OBJ)
+        cuda_include_directories(${CUDA_INC})
+        cuda_compile(CUDA_OBJ ${CUDA_SRC} STATIC)
+    else()
+        message(STATUS "Not doing cuda compile ")
```
csullivan: > Not doing cuda compile 
I think that these messages might be confusing considering most of our cuda kernels are built at runtime. If I saw cmake say that it was not doing cuda compilation, I might think there is not CUDA support via ngraph. So I propose to change all the message references to CUDA to something about NVCC. We could possibly have a message about kernels being compiled via NVRTC at runtime, e.g.
> "Not precompiling static CUDA kernels via NVCC, runtime compilation via NVRTC will be used." or something similar.

fengleitian: maybe print info like "CUDA x and XXX is required for XXX, static CUDA kernels will not .."

PR #1908, opened 2018-10-24
Support TopK for NvidiaGPU backend

AyanmoI:
Design Philosophy:
- Modify graph by inserting reshape ops to reshape layout such that the topk op is performed for a 2nd order Tensor (except for 1D cases) and on the last dimension.
- Perform a heap sort upto K to get the topk values and indices

There are two kernels for two modes: shared mem and global mem modes. The global mem mode kernel is invoked when the shared mem required per block is greater than sharedMemPerBlock


src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.hpp
```diff
@@ -85,6 +85,12 @@ namespace ngraph
                                                 size_t out_rank,
                                                 size_t reduce_rank);
 
+                static void get_topk(codegen::CodeWriter& writer,
+                                     const std::string& name,
+                                     std::vector<std::string>& dtypes,
```
rkimballn1: This should be `const vector<string>&`

src/ngraph/runtime/gpu/cuda_emitter.hpp
```diff
@@ -50,6 +50,13 @@ namespace ngraph
                 size_t build_primitive(const op::ReplaceSlice* node, bool in_place_op);
 
             public:
+                size_t build_topk(const std::vector<element::Type>& dtypes,
+                                  NVShape input_shape,
```
rkimballn1: const NVShape&

fengleitian:
Need more time to review this one.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -202,6 +202,141 @@ void runtime::gpu::CudaKernelBuilder::get_ew_collective_op(
...
+        writer << "out_id = out_id + blockIdx.x * topk_k;\n";
+        writer << "out_val = out_val + blockIdx.x * topk_k;\n";
+        if (!use_malloc)
+        {
+            writer << "extern __shared__ Entry entry[];\n";
+        }
+
+        writer << "for (size_t i = threadIdx.x; i < num_cols; i += blockDim.x)\n";
+        writer.block_begin();
+        {
```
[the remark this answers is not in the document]
fengleitian: click the wrong button, re-submit.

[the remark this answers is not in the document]
AyanmoI: That is correct..I can do that while loading into shared memory (atleast when shared mem is used). I can do that in future while doing further optimizations

src/ngraph/runtime/gpu/pass/gpu_layout.cpp
```diff
@@ -79,6 +81,97 @@ namespace ngraph
                         reshape->set_op_annotations(op_annotations);
                     }
                 }
+                template <>
+                void GPULayout::LAYOUT_DECL(ngraph::op::TopK)
```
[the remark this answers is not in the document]
AyanmoI: I guess doing it through a graph transform would be cleaner and more modular design.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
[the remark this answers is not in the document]
AyanmoI: I can start working on this as a separate PR. That'll be easier for me to keep track. Is it ok to merge this since this is a working version already?

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -167,6 +167,104 @@ size_t runtime::gpu::CUDAEmitter::build_concat(const std::vector<std::string>& d
...
+    size_t primitive_index = m_primitive_emitter->lookup(hash);
+    if (primitive_index != std::numeric_limits<size_t>::max())
+    {
+        return primitive_index;
+    }
+    uint32_t block_size_x = 32;
+    uint32_t aligned_grid_size_x = num_rows;
+    size_t shared_data_bytes = num_cols * (dtypes[0].size() + index_elem_type.size());
+    // if the kernel has not been compiled, build it
+    bool use_malloc = ((shared_data_bytes > (48 << 10)) ? true : false);
```
fengleitian: share memory size can be query by 'cudaGetDeviceProperties'.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -202,6 +202,141 @@ void runtime::gpu::CudaKernelBuilder::get_ew_collective_op(
...
+            writer << "entry[i].set_value(in[i]);\n";
+            writer << "entry[i].set_index(i);\n";
+        }
+        writer.block_end();
+
+        writer << "__syncthreads();\n";
```
fengleitian: sync is not required if you only use one warp.

fengleitian: But let's keep it here in case someone accidentally change the block size.

src/ngraph/runtime/gpu/pass/gpu_layout.cpp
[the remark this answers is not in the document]
AyanmoI: Will do through cuda_emitter in separate PR

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
AyanmoI: Ok..keeping the syncthreads

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -202,6 +202,141 @@ void runtime::gpu::CudaKernelBuilder::get_ew_collective_op(
...
+
+        writer << "if (threadIdx.x == 0)\n";
```
csullivan: Since this is roughly O(nlogn), the transpose (O(n^2)) will dominate the kernel and so doing a strided load and write will likely be better. tensorflow is ok with this because they rely mainly on a radix sort from cub (for n <=1000, when k=n or when k > 100). It would be good to get a serialized graph with topk in it to compare. but this is lower priority right now.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1585,7 +1585,32 @@ void runtime::gpu::GPU_Emitter::emit_Tanh(EMIT_ARGS)
...
+    {
+        return;
+    }
+    auto topk = static_cast<const ngraph::op::TopK*>(node);
+    size_t topk_axis = topk->get_top_k_axis();
+    size_t topk_k = topk->get_k();
+    auto index_elem_type = topk->get_index_element_type();
+    bool compute_max = topk->get_compute_max();
+    std::vector<element::Type> dtypes{args[0].get_element_type()};
+    NGRAPH_ASSERT(out.size() <= 2) << "TopK can only have 2 outputs";
```
csullivan: == 2

test/backend_topk_argReduce.in.cpp
```diff
@@ -0,0 +1,858 @@
+//*****************************************************************************
```
csullivan: Why a separate file for topk+argmin/max? seems like a separate file for topk and argreduce each would be better if you want to refactor. also filenames should be all lower snake case.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -167,6 +167,104 @@ size_t runtime::gpu::CUDAEmitter::build_concat(const std::vector<std::string>& d
...
+    auto compiled_kernel = m_ctx->compiled_kernel_pool->get(kernel_name.str());
```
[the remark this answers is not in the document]
AyanmoI: I tried doing this in a previous commit. The issue that comes is that the arg list need to be changed within the primitive, it did not do it this way. But I guess I can just use placeholders

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
AyanmoI: Agreed!

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -167,6 +167,138 @@ size_t runtime::gpu::CUDAEmitter::build_concat(const std::vector<std::string>& d
...
+
+    size_t shared_struct_bytes = (((dtypes[0].size() + index_elem_type.size()) <= 8) ? 8 : 16);
+    size_t shared_data_bytes = num_cols * shared_struct_bytes;
+
+    // Use global memory when each row size exceeds shared mem allowed per block
+    int device_num = 0;
+    CUDA_RT_SAFE_CALL(cudaGetDevice(&device_num));
+    cudaDeviceProp prop;
+    CUDA_RT_SAFE_CALL(cudaGetDeviceProperties(&prop, device_num));
+    bool use_malloc = ((shared_data_bytes > prop.sharedMemPerBlock) ? true : false);
```
csullivan: Shared mem consumption here is a little complicated - blocks on the same SM share shared memory. As it is, if the number of columns is large, this could easily serialize down to one block per SM. Just FYI since we will likely eventually do a different sort when num_col is large.

PR #1909, opened 2018-10-24
Change Backend::create to return std::unique_ptr<Backend>

rkimballn1:
Change from shared_ptr to unique_ptr for Backend::create. There was no need for a shared_ptr in this case.

While this does change the API it is not nearly as bad as one might think. Most calls are of the type:
```
auto backend = runtime::Backend::create("");
```
which just works. Also
```
shared_ptr<runtime::Backend> backend = runtime::Backend::create("");
```
works too because the right-hand side unique_ptr can be converted to a shared_ptr.


aprocter:
@mbrookhart Can you just convert the `unique_ptr` to a `shared_ptr` on the bridge side?

src/ngraph/runtime/backend_manager.cpp
```diff
@@ -111,21 +111,9 @@ shared_ptr<runtime::Backend> runtime::BackendManager::create_backend(const std::
...
-
-        runtime::Backend* backend = new_backend(config.c_str());
-        rc = shared_ptr<runtime::Backend>(backend, [=](runtime::Backend* b) {
-            delete_backend(b);
-            // CLOSE_LIBRARY(handle);
-        });
+        backend = new_backend(config.c_str());
     }
-    return rc;
+    return unique_ptr<runtime::Backend>(backend);
```
earhart: One minor note -- if we ever want to support unloading a backend, we won't be able to do it via adding a deleter to the shared pointer.  But it's probably better to use a wrapper for that anyway, so we can defer unload as long as there are any outstanding objects.  (And it might be best to never unload backends; unloading dlls is tricky.)

src/ngraph/runtime/backend_manager.cpp
```diff
@@ -83,7 +83,7 @@ shared_ptr<runtime::Backend> runtime::BackendManager::create_backend(const std::
     if (it != registry.end())
     {
         new_backend_t new_backend = it->second;
-        rc = shared_ptr<runtime::Backend>(new_backend(config.c_str()));
+        backend = new_backend(config.c_str());
```
earhart: Is there any reason not to return a `std::unique_ptr<runtime::Backend>` directly from the `new_backend()` call?  (We could pass in `config` as a string, too.)

test/util/autodiff/numeric_derivative.hpp
```diff
@@ -34,7 +34,7 @@ namespace ngraph
         /// \returns vector of dy/dvar, where each dy/dvar's shape is concat(y.shape(), var.shape())
         template <typename T>
         std::vector<std::shared_ptr<runtime::Tensor>>
-            numeric_derivative(const std::shared_ptr<runtime::Backend>& backend,
+            numeric_derivative(runtime::Backend& backend,
```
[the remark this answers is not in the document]
shssf: @rkimballn1 Is it necessary to eliminate `const` here?

test/cpu_fusion.cpp
```diff
@@ -143,7 +143,7 @@ TEST(cpu_fusion, gemm_cpu_broadcast_row)
 
     auto f = make_shared<Function>(cg, op::ParameterVector{A, B});
 
-    auto backend = runtime::Backend::create("CPU");
+    shared_ptr<runtime::Backend> backend = runtime::Backend::create("CPU");
```
shssf: why `shared_ptr` used here instead `uniq_ptr`?

postrational:
Unfortunately this currently produces a segfault (code -11) in the pybind code of the Python API.
Further changes may be needed in `python/pyngraph/runtime/backend.cpp` and related files.

src/ngraph/runtime/backend_manager.cpp
rkimballn1: Since new_backend is a C function it shouldn't return c++ objects. Pointers are ok.

test/cpu_fusion.cpp
rkimballn1: I got lazy. Some of the tests depend on backend being a shared_ptr so I just did a global search/replace.

test/util/autodiff/numeric_derivative.hpp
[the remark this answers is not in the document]
rkimballn1: `const shared_ptr<X>` is a `const shared_ptr`, not a `const X`. I did not get rid of const on X as it was never there. I did try making this const but the functions compile() and call() on backend are not const so backend can't be const.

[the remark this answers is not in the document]
rkimballn1: @earhart I like your rule. Right now we pretty much pass everything around everywhere as shared_ptr which makes me crazy. I will talk this idea over with a few folks here to see if we can come to a conclusion.

src/ngraph/runtime/backend_manager.cpp
rkimballn1: I was disappointed to find that unique_ptr does not handle deleters the same way that shared_ptr does and I could not find a way to add a deleter that I could pass without having to add another template arg to every reference of unique_ptr.

earhart: There's always the hack of specializing `std::default_delete<>`, but I don't think it'd make sense for something like this.  :-)

test/util/autodiff/numeric_derivative.hpp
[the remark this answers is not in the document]
rkimballn1: @earhart I talked with Scott and we are both meh about the pointer vs reference. I *did* change it to pointer because it was a pointer before and it reduces the changes on other lines of code.

PR #1917, opened 2018-10-25
Move to TBB2019 and bug fix to capture functor

jbobba:


cmake/external_tbb.cmake
```diff
@@ -20,7 +20,7 @@
 
 if(NGRAPH_TBB_ENABLE)
     set(TBB_GIT_REPO_URL https://github.com/01org/tbb)
-    set(TBB_GIT_TAG "tbb_2018")
+    set(TBB_GIT_TAG "tbb_2019")
```
silee2: tbb_2019 is a branch that can change over time. Is better to use an actual tag. 2019_U1 is the latest tag for tbb 2019.

jbobba: good point. will change.

cmake/external_tbb.cmake
```diff
@@ -20,7 +20,7 @@
 
 if(NGRAPH_TBB_ENABLE)
     set(TBB_GIT_REPO_URL https://github.com/01org/tbb)
-    set(TBB_GIT_TAG "tbb_2018")
+    set(TBB_GIT_TAG "2019_U1")
```
pruthviIntel: should we also set the flag in the parent `CMakeLists.txt` to include the profiling tools for TBB. I build ngraph with this in my branch to generate graph.xml/trace.xml file for FGA.

`set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -DTBB_USE_THREADING_TOOLS")`

PR #1928, opened 2018-10-26
Do not fuse nodes if one node is predecessor of another node in horiz…

ayzhuang:
…ontal fusion.


src/ngraph/runtime/cpu/pass/cpu_horizontal_fusion.cpp
```diff
@@ -96,6 +98,20 @@ void ngraph::runtime::cpu::pass::CPUHorizontalFusion::cpu_conv_horizontal_fusion
             bias_nodes.push_back(u->get_argument(2));
             conv_bias_nodes.push_back(u);
         }
+
+        for (auto node_start : conv_bias_nodes)
```
Krovatkin: this will be pretty expensive :(

diyessi: Couldn't you instead check to see if `node_start` was a predecessor of any of the `conv_bias_nodes` simultaneously, instead of doing each individually?

PR #1930, opened 2018-10-26
always using create_gpu_buffer and free_gpu_buffer instead of cudamalloc and cudafree

fengleitian:


fengleitian:
@csullivan  exactly.

PR #1932, opened 2018-10-27
IntelGPU backend: Profile data improved

shssf:


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -1662,17 +1662,17 @@ vector<runtime::PerformanceCounter>
     return rc;
 }
 
-static Shape get_shape_by_name(const shared_ptr<Function> func, const string& name)
+static shared_ptr<Node> get_shape_by_name(const shared_ptr<Function> func, const string& name)
```
rkimballn1: Shouldn't this function be renamed to `get_node_by_name`?

rkimballn1: This could return a Node* to make this a little less shared_ptr dependent. Or perhaps const Node*

PR #1933, opened 2018-10-28
Quantize(reorder) bias to int32 

nbpatel:


src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -357,16 +417,78 @@ size_t MKLDNNEmitter::build_quantized_convolution(const mkldnn::memory::desc& in
...
+
+    auto result_desc = mkldnn_utils::get_output_mkldnn_md(node, 0);
+
+    mkldnn::post_ops ops;
+
+    auto add_relu = [&]() {
+        return (dynamic_cast<const ngraph::op::QuantizedConvolutionBias*>(node))->with_relu();
+    };
+
+    if (add_relu())
```
sayantan-nervana: Unless the lambda `add_relu` is used elsewhere, why define it.

instead this would work
```
if ((dynamic_cast<const ngraph::op::QuantizedConvolutionBias*>(node))->with_relu())
```

or:
```
bool add_relu = (dynamic_cast<const ngraph::op::QuantizedConvolutionBias*>(node))->with_relu();
if (add_relu)
```

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -357,16 +417,78 @@ size_t MKLDNNEmitter::build_quantized_convolution(const mkldnn::memory::desc& in
...
+    auto convolution = static_cast<const ngraph::op::QuantizedConvolutionBias*>(node);
+
+    // For dilation, MKLDNN wants to know how many elements to insert between, not how far
+    // apart to space the elements like nGraph. So we have to subtract 1 from each pos.
+    ngraph::Strides window_dilation_strides_adjusted;
+
+    for (size_t s : convolution->get_window_dilation_strides())
+    {
+        window_dilation_strides_adjusted.push_back(s - 1);
+    }
```
sayantan-nervana: Instead of `push_back`ing the vector, can we do this?
```
ngraph::Strides window_dilation_strides_adjusted = convolution->get_window_dilation_strides();
```

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
```diff
@@ -106,20 +106,61 @@ namespace ngraph
                     auto& out0_tensor = external_function->get_tensor_data(out[0].get_name());
 
                     auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
+                    bool is_bias_i32 = (args[2].get_element_type() == element::i32);
+                    if (!is_bias_i32)
+                    {
+                        auto reorder_bias_desc = mkldnn_emitter->build_memory_descriptor(
+                            args[2].get_shape(), ngraph::element::i32, mkldnn::memory::format::x);
+                        auto conv_index =
+                            mkldnn_emitter->build_bias_reorder(node, args, out, reorder_bias_desc);
```
jbobba: Please modify build_convolution<>() for this instead of duplicating code in another function. You can do the args[2].get_element_type() check inside build_convolution as well, create reorder descriptors there.

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
```diff
@@ -106,20 +106,61 @@ namespace ngraph
                     auto& out0_tensor = external_function->get_tensor_data(out[0].get_name());
 
                     auto& mkldnn_emitter = external_function->get_mkldnn_emitter();
+                    bool is_bias_i32 = (args[2].get_element_type() == element::i32);
```
jbobba: Cant we define QuantizedConvolutionBias as taking int32 bias and insert a reorder in the builder itself instead of doing it here?

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -312,21 +314,79 @@ size_t MKLDNNEmitter::build_quantized_convolution(const mkldnn::memory::desc& in
...
+         mkldnn_utils::global_cpu_engine},
+        *m_mkldnn_primitives[input_data_index],
+        *m_mkldnn_primitives[weights_index],
+        *m_mkldnn_primitives[bias_index],
+        *m_mkldnn_primitives[result_index]));
+    m_primitive_deps[conv_index] = {input_data_index, weights_index, bias_index, result_index};
+    return conv_index;
+}
+
+std::vector<size_t> MKLDNNEmitter::build_quantized_convolution_bias_reorder(
```
jbobba: overload build_quantized_convolution_forward

src/ngraph/builder/quantization.cpp
```diff
@@ -121,13 +121,20 @@ namespace ngraph
...
                                                                 min_freezed_output,
-                                                                max_freezed_output);
+                                                                max_freezed_output,
+                                                                is_output_signed);
             auto requantization_scale = op::Constant::create(element::f32, Shape{1}, {scale});
+            float scale_bias = (bias->get_element_type() == element::i32)
+                                   ? 1.0
+                                   : builder::quantization_util::get_bias_scale(
+                                         min_input, max_input, min_filter, max_filter);
+            auto bias_scale = op::Constant::create(element::f32, Shape{1}, {scale_bias});
```
jbobba: Can we use a quantize op here to convert bias to int32?

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
jbobba: you dont need to return a vector. See this: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/builder/max_pool.cpp#L129

src/ngraph/builder/quantization_util.hpp
```diff
@@ -95,11 +96,33 @@ namespace ngraph
                 // s32 = f32 * std::pow(2, 31)/ max_abs32;
                 // s8 = f32 * std::pow(2, 7)/ max_abs8;
                 // s8 = s32 * std::pow(2, -24) * max_abs32 / max_abs8;
-                const float scale = static_cast<float>(
-                    (std::pow(2, -24) * static_cast<double>(max_abs32 / max_abs8)));
+                float scale = static_cast<float>((std::pow(2, is_output_signed ? -24 : -23) *
```
adstraw: comments not updated to reflect signed case.  could you add some comments as to why this math makes sense?  I don't have any idea why this calculation does what it does other than "that's what TensorFlow does".

src/ngraph/builder/quantization.cpp
```diff
@@ -121,13 +121,20 @@ namespace ngraph
...
         {
+            bool is_output_signed = !with_relu;
             float scale = builder::quantization_util::get_scale(min_input,
                                                                 max_input,
                                                                 min_filter,
                                                                 max_filter,
                                                                 min_freezed_output,
-                                                                max_freezed_output);
+                                                                max_freezed_output,
+                                                                is_output_signed);
```
adstraw: Instead of booleans, we should be passing element::Type and then getting the information we want from that Type in the function.  Booleans do not make great APIs.

src/ngraph/builder/quantization.cpp
adstraw: @jbobba are you talking about an nGraph::op::Quantize?

src/ngraph/builder/quantization_util.hpp
```diff
@@ -95,11 +96,33 @@ namespace ngraph
...
+                    std::static_pointer_cast<ngraph::op::Constant>(min_filter);
+                auto max_filter_const_op =
+                    std::static_pointer_cast<ngraph::op::Constant>(max_filter);
+                auto input_min = min_input_const_op->get_vector<float>();
+                auto input_max = max_input_const_op->get_vector<float>();
+                auto filter_min = min_filter_const_op->get_vector<float>();
+                auto filter_max = max_filter_const_op->get_vector<float>();
+
+                float bias_scale =
+                    255.0 * 127.0 / (std::max(std::abs(input_min[0]), std::abs(input_max[0])) *
```
adstraw: 255.0 and 127.0 are the std::numeric_limits<uint8_t>::max() and std::numeric_limits<uint8_t>::min(), right?  we should use that here instead of these magic numbers.  Note that soon I want to add `min` and `max` to element::Type so that these values can be defined in the nGraph type system.  not asking you to make that change.  just remove the magic numbers.

src/ngraph/op/experimental/quantized_conv_bias.cpp
```diff
@@ -53,6 +55,10 @@ op::QuantizedConvolutionBias::QuantizedConvolutionBias(const shared_ptr<Node>& d
     auto scale_val = scale_const_op->get_vector<float>();
     this->m_scale = scale_val[0];
 
+    auto bias_scale_const_op = std::static_pointer_cast<ngraph::op::Constant>(bias_scale);
```
adstraw: this is not a safe cast.  at a minimum:  use dynamic cast and check for nullptr.  better yet, keep this as a `Node` and wait to get the `float` scale value in the CPU backend before calling MKLDNN.  Same goes for `scale_val` above.  Same goes for other experimental ops.  Can we get this cleaned up?

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
adstraw: tailoring the op to a specific type seems overly optimized to the current graphs we are working on unless I am missing something here.  can we always count on bias being a certain type?

src/ngraph/builder/quantization_util.hpp
```diff
@@ -125,7 +126,37 @@ namespace ngraph
...
+                auto shape = min_input->get_shape();
+                if (shape != max_input->get_shape() || shape != min_filter->get_shape() ||
+                    shape != max_filter->get_shape())
+                {
+                    throw ngraph_error("get_scale: min and max must have same shape");
+                }
+
+                auto max_abs_input_range = max_abs(min_input, max_input);
+                auto max_abs_filter_range = max_abs(min_filter, max_filter);
+                auto range = make_constant(type, shape, 255.0 * 127.0);
```
adstraw: I would still prefer these not be "magic numbers".  This is std::numeric_limits<uint8_t>::max() and std::numeric_limits<int8_t>::min()?

src/ngraph/builder/quantization_util.hpp
```diff
@@ -125,7 +126,37 @@ namespace ngraph
                 // s32 = f32 * std::pow(2, 31)/ max_abs32;
                 // s8 = f32 * std::pow(2, 7)/ max_abs8;
                 // s8 = s32 * std::pow(2, -24) * max_abs32 / max_abs8;
-                return make_constant(type, shape, std::pow(2, -24)) * (max_abs32 / max_abs8);
+
+                return make_constant(
```
adstraw: Needs more documentation.  Why does this math make sense?

src/ngraph/builder/quantization_util.hpp
```diff
@@ -125,7 +126,37 @@ namespace ngraph
...
+                                                 std::shared_ptr<Node> max_input,
+                                                 std::shared_ptr<Node> min_filter,
+                                                 std::shared_ptr<Node> max_filter)
+            {
+                auto type = min_input->get_element_type();
+                if (type != max_input->get_element_type() ||
+                    type != min_filter->get_element_type() ||
+                    type != max_filter->get_element_type())
+                {
+                    throw ngraph_error("get_scale: min and max must have same type");
```
adstraw: "get_bias_scale"

src/ngraph/builder/quantization_util.hpp
```diff
@@ -125,7 +126,37 @@ namespace ngraph
...
+                }
+
+                auto shape = min_input->get_shape();
+                if (shape != max_input->get_shape() || shape != min_filter->get_shape() ||
+                    shape != max_filter->get_shape())
+                {
+                    throw ngraph_error("get_scale: min and max must have same shape");
```
adstraw: "get_bias_scale"

src/ngraph/builder/quantization.cpp
```diff
@@ -121,23 +121,50 @@ namespace ngraph
...
+            {
+                auto zero = make_constant(element::i32, min_input->get_shape(), 0);
+                AxisSet quantization_axes;
+                auto bias_scale =
+                    quantization_util::get_bias_scale(min_input, max_input, min_filter, max_filter);
+                op::Quantize::RoundMode round_mode =
+                    op::Quantize::RoundMode::ROUND_NEAREST_TOWARD_EVEN;
+
+                auto reordered_bias = make_shared<op::Quantize>(
+                    bias, bias_scale, zero, element::i32, quantization_axes, round_mode);
```
adstraw: Is there testing for this?  I am pretty sure this line will throw when called as element::i32 is not considered to be a quantized type.

src/ngraph/builder/quantization.cpp
```diff
@@ -121,23 +121,50 @@ namespace ngraph
...
-                                                             requantization_scale,
-                                                             with_relu);
+            if (bias->get_element_type() != element::i32)
+            {
+                auto zero = make_constant(element::i32, min_input->get_shape(), 0);
+                AxisSet quantization_axes;
+                auto bias_scale =
+                    quantization_util::get_bias_scale(min_input, max_input, min_filter, max_filter);
+                op::Quantize::RoundMode round_mode =
+                    op::Quantize::RoundMode::ROUND_NEAREST_TOWARD_EVEN;
```
adstraw: How do you know the rounding mode?  Shouldn't this match the rounding mode used for the input?

src/ngraph/builder/quantization.cpp
```diff
@@ -121,23 +121,50 @@ namespace ngraph
...
-                                                             filters,
-                                                             bias,
-                                                             window_movement_strides,
-                                                             window_dilation_strides,
-                                                             padding_below,
-                                                             padding_above,
-                                                             data_dilation_strides,
-                                                             requantization_scale,
-                                                             with_relu);
+            if (bias->get_element_type() != element::i32)
```
adstraw: First off, the QuantizedConvolutionBias experimental op should enforce that bias is of type element::i32.  The fact that those experimental ops omit type checking needs to change.  Second, I am not sure it should be the responsibility of this builder to meet those type requirements.  There isn't enough information (see rounding mode comment below) to make this conversion.  My feeling is that it should be the responsibility of the bridge to call these builders with properly formatted inputs.  If not the bridge, we likely need a graph pass to conincalize inputs.  What is the situation with TF DO graphs?  What is the type of the bias input we are converting from?

src/ngraph/builder/quantization.cpp
```diff
@@ -121,23 +121,50 @@ namespace ngraph
...
+                                                                 window_dilation_strides,
+                                                                 padding_below,
+                                                                 padding_above,
+                                                                 data_dilation_strides,
+                                                                 requantization_scale,
+                                                                 with_relu);
+            }
+            else
+            {
+                return make_shared<op::QuantizedConvolutionBias>(input,
```
adstraw: There's no need to have multiple return statements here.  Instead, just set bias = new bias.

src/ngraph/builder/quantization_util.hpp
```diff
@@ -125,7 +126,37 @@ namespace ngraph
...
                 // s8 = f32 * std::pow(2, 7)/ max_abs8;
                 // s8 = s32 * std::pow(2, -24) * max_abs32 / max_abs8;
-                return make_constant(type, shape, std::pow(2, -24)) * (max_abs32 / max_abs8);
+
+                return make_constant(
+                           type, shape, std::pow(2, (output_type == element::i8) ? -24 : -23)) *
+                       (max_abs32 / max_abs8);
+            }
+
+            std::shared_ptr<Node> get_bias_scale(std::shared_ptr<Node> min_input,
```
adstraw: There are already three ways to calculate a `scale` value just for TensorFlow and just for scaled mode.  This isn't surprising to me, I guess... it's the reason why I opted for scale/bias implementation for Quantize and Dequantize.  However, is there some way to combine or group these calculations into some more common theme?  It feels very "wild wild west" right now.  How will one know which scale builders to use?  We need more documentation at a minimum.

src/ngraph/builder/quantization.cpp
adstraw: It should be the responsibility of the bridge to quantize bias / make bias input conform to the builder/op type checking.  That said, I won't hold up the PR for it provided that... 1) type checking for experimental quantized ops and related testing (test/type_prop.cpp) is included in a separate PR and 2) that this code is moved to the bridge in separate PR.  NOTE:  a separate builder can be used for #2 if the code to do the conversion is too complex for the bridge create on its own.

src/ngraph/builder/quantization_util.hpp
adstraw: Still feels a little "wild west" to me, but not sure how to fix right now.  Let's look to consolidate and simplify our builders which calculate scale at some later date.

PR #1941, opened 2018-10-29
[ONNX] Support for external weights to enable Caffe2 models

apwojcik:
PyTorch and Caffe2 ONNX models are not 100% ONNX compliant. The difference is in how they pass weights to model. Compliant ONNX models have weights embedded into a model as `constants` or as `initializers`. However, due to the limitation of `protobuf` serializer (2GB limit, i.e. trained VGG16 cannot be serialized!), PyTorch and Caffe2 are separating model and weights, and pass both independently to a framework.

### DIFF
--- a/src/ngraph/frontend/onnx_import/core/value_info.hpp
+++ b/src/ngraph/frontend/onnx_import/core/value_info.hpp
@@ -25,6 +25,7 @@
 
 #include "node.hpp"
 #include "tensor.hpp"
+#include "weight.hpp"
 
 namespace ngraph
 {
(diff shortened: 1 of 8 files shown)


src/ngraph/frontend/onnx_import/core/value_info.hpp
```diff
@@ -103,7 +104,8 @@ namespace ngraph
             }
 
             std::shared_ptr<ngraph::Node>
-                get_ng_node(op::ParameterVector& parameters,
+                get_ng_node(const Weights& weights,
```
postrational: Please make `weights` an optional parameter here.

src/ngraph/frontend/onnx_import/onnx.hpp
```diff
@@ -34,16 +35,19 @@ namespace ngraph
                                Operator fn);
 
         // Convert on ONNX model to a vector of nGraph Functions (input stream)
```
postrational: Please add documentation for the `weights` params.

PR #1942, opened 2018-10-29
Dropout for CPU

Krovatkin:


src/ngraph/runtime/gpu/unit_test.manifest
```diff
@@ -10,6 +10,7 @@ concat_matrix_int64
 divide_by_zero_int32
```
jbobba: Please revert mode changes.

Krovatkin: will do!

src/ngraph/runtime/intelgpu/unit_test.manifest
```diff
@@ -109,3 +109,24 @@ zero_sized_sqrt
 zero_sized_subtract
 zero_sized_tan
 zero_sized_tanh
+argmin_trivial
```
shssf: It is strange. These two tests works well on my side.
```
shalnov@nnwsshalnov11:/home/work/nGraph/0.build$ ./test/unit-test --gtest_filter=INTELGPU.argmin*
Note: Google Test filter = INTELGPU.argmin*
[==========] Running 3 tests from 1 test case.
[----------] 3 tests from INTELGPU
[ RUN      ] INTELGPU.argmin_trivial
[       OK ] INTELGPU.argmin_trivial (502 ms)
[ RUN      ] INTELGPU.argmin_4D_axis_3
[       OK ] INTELGPU.argmin_4D_axis_3 (404 ms)
[ RUN      ] INTELGPU.argmin_4D_axis_3_i64
[       OK ] INTELGPU.argmin_4D_axis_3_i64 (400 ms)
[----------] 3 tests from INTELGPU (1306 ms total)

[==========] 3 tests from 1 test case ran. (1306 ms total)
[  PASSED  ] 3 tests.
```

src/ngraph/state/rng_state.hpp
```diff
@@ -0,0 +1,52 @@
...
+        RNGState(unsigned int seed, double probability)
+            : State()
+            , m_generator(seed)
+            , m_distribution(probability)
+        {
+        }
+        virtual void activate() override;
+        virtual void deactivate() override;
+        virtual ~RNGState() {}
+        std::mt19937& get_generator() { return m_generator; }
```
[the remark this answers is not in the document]
Krovatkin: As far as I understand that type is just a typedef of another concrete type : `typedef mersenne_twister_engine<uint_fast32_t, 32,624,397,31,0x9908b0df,11,0xffffffff,7,0x9d2c5680,15,0xefc60000,18,1812433253> mt19937;`; Random device is another **concrete** type of a random number generator

src/ngraph/op/generate_mask.hpp
```diff
@@ -0,0 +1,65 @@
...
+            {
+                constructor_validate_and_infer_types();
+            }
+
+            virtual std::shared_ptr<Node>
+                copy_with_new_args(const NodeVector& new_args) const override;
+
+            std::shared_ptr<Node> get_training_node() const { return get_argument(0); }
+            double get_probability() const { return m_probability; }
+            unsigned int get_seed() const { return m_seed; }
```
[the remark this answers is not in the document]
Krovatkin: I think they might be useful for serializing at the very least. Which reminds me that I need to add the serialization logic. I believe these need to be inspectable attributes. I'll add docstrings.

src/ngraph/state/rng_state.hpp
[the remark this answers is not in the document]
diyessi: I think this is okay for now; multinomial is probably next, and will give us a better picture of the lay of the land.

src/ngraph/op/generate_mask.cpp
```diff
@@ -0,0 +1,34 @@
...
+shared_ptr<Node> op::GenerateMask::copy_with_new_args(const NodeVector& new_args) const
+{
+    check_new_args_count(this, new_args);
+    return make_shared<GenerateMask>(
+        new_args.at(0), m_shape, m_element_type, m_seed, m_probability);
+}
+
+void ngraph::op::GenerateMask::validate_and_infer_types()
+{
+    NGRAPH_ASSERT(shape_size(get_training_node()->get_shape()) == 1)
```
aprocter:
```suggestion
    NODE_VALIDATION_ASSERT(this, get_input_partial_type(0).compatible(PartialShape{}))
```

Any restrictions on the input element type?

src/ngraph/op/generate_mask.hpp
```diff
@@ -0,0 +1,70 @@
...
+        {
+        public:
+            /// \brief Constructs a GenerateMask node with a given shape, sed,
+            /// probability and training/inference mode
+            GenerateMask(const std::shared_ptr<Node>& training,
+                         const Shape& shape,
+                         const element::Type& element_type,
+                         unsigned int seed,
+                         double prob)
+                : Op("GenerateMask", {check_single_output_args({training})})
```
aprocter:
```suggestion
                : Op("GenerateMask", check_single_output_args({training}))
```

aprocter: Move the constructor implementation to `.cpp`

src/ngraph/op/generate_mask.hpp
```diff
@@ -0,0 +1,70 @@
...
+                , m_probability(prob)
+            {
+                constructor_validate_and_infer_types();
+            }
+
+            virtual std::shared_ptr<Node>
+                copy_with_new_args(const NodeVector& new_args) const override;
+
+            /// \brief Returns the argument specifying if this mask is used in training or inference
+            std::shared_ptr<Node> get_training_node() const { return get_argument(0); }
```
aprocter: I'd remove this.

src/ngraph/op/generate_mask.hpp
```diff
@@ -0,0 +1,70 @@
...
+            /// \brief Returns the probability of a trial generating 1 (i.e. an element being kept)
+            double get_probability() const { return m_probability; }
+            /// \brief Returns the seed value supplied to by a random generator
```
aprocter:
```suggestion
            /// \brief Returns the seed value supplied to a random generator
```

src/ngraph/op/generate_mask.hpp
```diff
@@ -0,0 +1,70 @@
...
+        protected:
+            virtual void generate_adjoints(autodiff::Adjoints& adjoints,
+                                           const NodeVector& deltas) override
+            {
+            }
+
+            void validate_and_infer_types() override;
+            Shape m_shape;
+            element::Type m_element_type;
+            unsigned int m_seed = 0;
```
aprocter: Remove the initialization here (constructor takes care of it)

src/ngraph/op/generate_mask.hpp
```diff
@@ -0,0 +1,70 @@
...
+            double m_probability = 1.;
```
aprocter: Remove the initialization here (constructor takes care of it)

src/ngraph/op/generate_mask.cpp
aprocter: We should also check something like:

```
    NODE_VALIDATION_ASSERT(this, m_element_type.is_static())
        << "Output element type must not be dynamic.";
```

aprocter: I'm not sure why we want the "training flag" to be an input at all. Can't a bridge just omit `GenerateMask` if the mode is known statically, and replace it with something like `Select(is_training,GenerateMask,Constant(1))` if the mode is not known until runtime?

src/ngraph/runtime/intelgpu/unit_test.manifest
```diff
@@ -25,6 +25,7 @@ divide_by_zero_int32
 dot_3d_multi_axis
 dot_4d_5d_multi_axis
 dot_4d_5d_multi_axis_more
+generate_mask
```
aprocter: Revert mode changes

src/ngraph/runtime/interpreter/int_backend.hpp
```diff
@@ -312,6 +314,11 @@ class ngraph::runtime::interpreter::INTBackend : public Backend
                                    avg_pool->get_include_padding_in_avg_computation());
             break;
         }
+        case OP_TYPEID::GenerateMask:
+        {
+            throw ngraph_error(
+                "GenerateMask is an experimental op that's only supported on CPU backend");
```
aprocter: I would've preferred to put it in the `experimental` directory if this is the case, but I shan't fight it.

test/backend_test.in.cpp
```diff
@@ -4764,6 +4766,30 @@ NGRAPH_TEST(${BACKEND_NAME}, reverse_sequence_n4d2c3h2w2)
...
+    auto f = make_shared<Function>(NodeVector{gen_mask, gen_mask2}, op::ParameterVector{});
+
+    auto backend = runtime::Backend::create("${BACKEND_NAME}");
+
+    auto result_tv1 = backend->create_tensor<float>(result_shape);
+    auto result_tv2 = backend->create_tensor<float>(result_shape);
+    backend->call_with_validate(f, {result_tv1, result_tv2}, {});
+    auto result1 = read_vector<float>(result_tv1);
+    auto result2 = read_vector<float>(result_tv2);
+    ASSERT_EQ(result1, result2);
```
aprocter: Should also check that `result1` and `result2` (and `result1_2`, `result2_2` below) consist entirely of zeros and ones.

test/backend_test.in.cpp
```diff
@@ -4764,6 +4766,30 @@ NGRAPH_TEST(${BACKEND_NAME}, reverse_sequence_n4d2c3h2w2)
     EXPECT_EQ(read_vector<int>(result), expected);
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, generate_mask)
+{
+    Shape scalar{};
+    Shape result_shape{1, 20};
```
aprocter: I'm being a dork here but can we increase this to, say, `{1, 128}`? With `{1, 20}` (and if the seed weren't fixed), the probability of one of the `ASSERT_NE` tests below failing on a compliant implementation is only about 1/500000  :)

src/ngraph/op/generate_mask.cpp
mbrookhart: This always generates the mask, even for inference, I belive? That gives us a slowdown. We're compiling separate inference and training graphs anyway, the selection happens at the graph level, not the op level.

src/ngraph/state/rng_state.hpp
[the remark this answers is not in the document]
Krovatkin: @diyessi I think the next step is to provide an interface to abstract random number generator and then probability objects because other backends might have different implementations and in stdlib all those are concrete types, so we need some kind of a wrapper with well defined API

src/ngraph/op/generate_mask.cpp
mbrookhart: Then again, since I'm already compiling two graphs, I probably just won't insert the generate mask op into the graph if it's inference without randomization.

aprocter: Right, that’s what I was thinking (two graphs). If it really is a run-time decision I understand that it’s a problem though. I wish that we had a way to do something like if_else(cond,a,b) without forcing evaluation of both a and b, but we don’t.

It’s fine like it is in this respect, no need to block on this.

test/backend_test.in.cpp
aprocter: I should have said: the probability is about 1/500000, which is a bit high :)

src/ngraph/runtime/interpreter/int_backend.hpp
aprocter: Thanks for doing this.

aprocter:
@Krovatkin Just two last things:

1. I think you still need to fix the mode change on `hybrid/unit_test.manifest`. (If I'm wrong just tell me, GitHub's UI is confusing to me here.)

2. Please delete `state/state.cpp`, or add a copyright header (even though the file is otherwise empty) so we don't forget it if code is added to the file later.

test/backend_test.in.cpp
```diff
@@ -14,6 +14,7 @@
 // limitations under the License.
 //*****************************************************************************
 
+#include <algorithm>
```
aprocter: This is already included on the following line.

PR #1946, opened 2018-10-30
TopK additional tests for nvGPU backend

AyanmoI:
Added tests for malloc mode and graph transform


test/gpu_test.cpp
```diff
@@ -160,3 +164,105 @@ TEST(gpu_test, memory_manager_seperate_workspaces_allocsize)
...
+                                         op::ParameterVector{A_gpu});
+    auto gpu_f_1 = make_shared<Function>(make_shared<op::GetOutputElement>(B_gpu, 1),
+                                         op::ParameterVector{A_gpu});
+
+    test::Uniform<float> rng(0.0f, 1.0f);
+    vector<vector<float>> args_index;
+
+    for (shared_ptr<op::Parameter> param : cpu_f_0->get_parameters())
+    {
+        vector<float> tensor_val(shape_size(param->get_shape()));
```
diyessi: You really don't want to use random values because you end up with tests that sometimes fail, and then you can never get them to fail when you run them by hand.

test/gpu_test.cpp
```diff
@@ -160,3 +164,105 @@ TEST(gpu_test, memory_manager_seperate_workspaces_allocsize)
...
+        rng.initialize(tensor_val);
+        args_index.push_back(tensor_val);
+    }
+
+    auto cpu_results_0 = execute<float, int64_t>(cpu_f_0, args_index, "CPU");
+    auto gpu_results_0 = execute<float, int64_t>(gpu_f_0, args_index, "GPU");
+
+    for (size_t i = 0; i < gpu_results_0.size(); i++)
+    {
+        EXPECT_TRUE(test::all_close(gpu_results_0.at(i), cpu_results_0.at(i)));
```
diyessi: This is one of the few cases where `all_close` is wrong. When you rearrange a list of floats, they won't change value. If they do, something is very wrong.
I just checked -- ONNX topk, which this is implementing, specifies you order by (-value, index), so all backends should agree. Here you should be checking against INTERPRETER (reference implementation) rather than CPU. Then we can move the test into the backend-independent tests. We might need to fix the reference implementation first.
https://github.com/onnx/onnx/blob/master/docs/Operators.md#TopK

diyessi: For reference, neither TensorFlow nor MXNet specify what to do on a tie.

test/gpu_test.cpp
```diff
@@ -1,3 +1,4 @@
+
```
rkimballn1: Remove line before header

test/gpu_test.cpp
```diff
@@ -160,3 +164,105 @@ TEST(gpu_test, memory_manager_seperate_workspaces_allocsize)
...
+        vector<float> tensor_val(shape_size(param->get_shape()));
+        rng.initialize(tensor_val);
+        args.push_back(tensor_val);
+    }
+    auto cpu_results_1 = execute(cpu_f_1, args, "CPU");
+    auto gpu_results_1 = execute(gpu_f_1, args, "GPU");
+
+    for (size_t i = 0; i < gpu_results_1.size(); i++)
+    {
+        EXPECT_TRUE(test::all_close(gpu_results_1.at(i), cpu_results_1.at(i), 1.0e-4f, 1.0e-4f));
```
rkimballn1: use all_close_f instead

test/gpu_test.cpp
```diff
@@ -160,3 +164,105 @@ TEST(gpu_test, memory_manager_seperate_workspaces_allocsize)
...
+    }
+
+    vector<vector<float>> args;
+    for (shared_ptr<op::Parameter> param : cpu_f_1->get_parameters())
+    {
+        vector<float> tensor_val(shape_size(param->get_shape()));
+        rng.initialize(tensor_val);
+        args.push_back(tensor_val);
+    }
+    auto cpu_results_1 = execute(cpu_f_1, args, "CPU");
```
rkimballn1: The CPU is an optional backend so you need to either conditionally run this test or use interpreter or construct your input such that you get a known output.

test/gpu_test.cpp
```diff
@@ -160,3 +164,105 @@ TEST(gpu_test, memory_manager_seperate_workspaces_allocsize)
...
+    auto r0 = backend->create_tensor(element::i32, out_shape);
+    auto r1 = backend->create_tensor(element::i32, out_shape);
+    auto r2 = backend->create_tensor(element::f32, out_shape);
+    auto r3 = backend->create_tensor(element::f32, out_shape);
+
+    backend->call_with_validate(gpu_f, {r0, r1, r2, r3}, {a, b, c, d, e});
+
+    EXPECT_EQ((vector<int32_t>{2, 1, 1, 2, 1, 2, 0, 1}), read_vector<int32_t>(r0));
+    EXPECT_EQ((vector<int32_t>{2, 1, 1, 2, 1, 2, 0, 1}), read_vector<int32_t>(r1));
+    EXPECT_EQ((vector<float>{4, 4, 3, 3, 3, 4, 2, 3}), read_vector<float>(r2));
```
rkimballn1: In general it is bad to compare floats, that is what all_close_f is for. In this case it is probably ok, but I would use all_close_f with very tight tolerances.

test/gpu_test.cpp
```diff
@@ -160,3 +163,105 @@ TEST(gpu_test, memory_manager_seperate_workspaces_allocsize)
...
+    vector<vector<float>> args_index;
+    for (shared_ptr<op::Parameter> param : interp_f_0->get_parameters())
+    {
+        vector<float> tensor_val(shape_size(param->get_shape()));
+        iota(tensor_val.begin(), tensor_val.end(), 0.0f);
+        args_index.push_back(tensor_val);
+    }
+
+    auto interp_results_0 = execute<float, int32_t>(interp_f_0, args_index, "INTERPRETER");
+    auto gpu_results_0 = execute<float, int32_t>(gpu_f_0, args_index, "GPU");
```
shssf: Sorry, I can't see any GPU specific things in these tests. Why don't add them to a common place for all backends?

AyanmoI: topk_3d_global_mem test is a case specific to nvGPUs wherein memory is allocated on the gpu global memory as compared to the shared memory on the GPU when the size of a row exceeds a particular limit. topk_fanout_graph_transform test locks the behavior wherein reshape ops are added before and after topk. This is also specific to nvGPU backend. These tests would definitely pass for all backends but I added these to gpu_test.cpp since these tests are intended to test changes specific to nvGPU backends. I can move them to backend_topk.in.cpp if you think that is appropriate.

test/gpu_test.cpp
```diff
@@ -160,3 +163,105 @@ TEST(gpu_test, memory_manager_seperate_workspaces_allocsize)
...
+        vector<float> tensor_val(shape_size(param->get_shape()));
+        iota(tensor_val.begin(), tensor_val.end(), 0.0f);
+        args.push_back(tensor_val);
+    }
+    auto interp_results_1 = execute(interp_f_1, args, "INTERPRETER");
+    auto gpu_results_1 = execute(gpu_f_1, args, "GPU");
+
+    for (size_t i = 0; i < gpu_results_1.size(); i++)
+    {
+        EXPECT_TRUE(test::all_close_f(gpu_results_1.at(i), interp_results_1.at(i), 0.0f, 0.0f));
```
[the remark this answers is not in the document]
rkimballn1: You need to read the docs for all_close_f if you are going to fiddle with mantissa and tolerance. For float the mantissa is 23 bits? and the tolerance is how many bits must match.

test/gpu_test.cpp
AyanmoI: moved to general tests

PR #1949, opened 2018-10-31
[PlaidML] Add PlaidML documentation

earhart:


doc/sphinx/source/programmable/index.rst
```diff
@@ -5,33 +5,57 @@
 Interact with Backends
```
indie: The changes we made to this file awhile ago aren't relevant to current nGraph, and this might confuse people since they weren't added at the same time as the code was... probably better to build docs in this section on the latest nGraph.

indie:
Most of these changes are getting added to other PRs so probably okay to close this

PR #1955, opened 2018-10-31
Update "JSON for Modern C++" to a version that officially supports gc…

silee2:
…c 4.8


silee2:
JSON for modern c++ not requires CMAKE > 3.8 which causes CI failure.

aprocter:
This is a header-only library, right? We might not need to run its cmake.

PR #1904, opened 2018-10-24
Partial Shapes and Types, Part 4m: BatchNorm and backprops

aprocter:
Reasons for the `Changes API` tag:

(1) I removed the `BatchNormBase` class, and updated CPU/nvGPU/IntelGPU backends to deal with this.
(2) I removed the restriction on rank for `BatchNormTrainingBackprop`, because this restriction is inconsistent with the forward ops.

PR #1963, opened 2018-11-01
Update BatchNorm to set_output_size at construction time

aprocter:
(Will have to redo this one if #1904 goes in first, but whatever.)

~I'm actually not 110% sure this is needed. If it seems wrong, please kick it back to me and I'll give it more thought.~

Actually I think I remembered why I want to do this. At `Node` construction time we need to be able to do `get_outputs` on the arguments.


aprocter:
I will need to rework this since #1904 won the race.

PR #1975, opened 2018-11-02
add dtype-generic load definitions and clean up nvrtc helpers

csullivan:
This extends data type support to the `load` functions previously hard coded in `include_helpers` and refactors them into `nvrtc/helpers.hpp`.

Fixes bug found in tensorflow test: `math_ops_test.py`.

### DIFF
--- a/src/ngraph/runtime/gpu/cuda_emitter.cpp
+++ b/src/ngraph/runtime/gpu/cuda_emitter.cpp
@@ -1245,7 +1245,6 @@ size_t runtime::gpu::CUDAEmitter::build_avg_pool(const std::array<std::string, 2
     if (compiled_kernel == nullptr)
     {
         codegen::CodeWriter writer;
-        writer << include_helpers();
         CudaKernelBuilder::get_avg_pool(writer, kernel_name, dtypes, include_pad);
         compiled_kernel = m_ctx->compiled_kernel_pool->set(kernel_name, writer.get_code());
     }
@@ -1597,7 +1596,6 @@ size_t runtime::gpu::CUDAEmitter::build_softmax_divide(const std::vector<std::st
     {
         codegen::CodeWriter writer;
         CudaKernelBuilder::add_pod_typedefs(writer);
-        writer << include_helpers();
         CudaKernelBuilder::get_softmax_divide_op(
             writer, kernel_name, dtypes, axes_flag, input_shape.size());
         compiled_kernel = m_ctx->compiled_kernel_pool->set(kernel_name, writer.get_code());
@@ -1713,7 +1711,6 @@ size_t runtime::gpu::CUDAEmitter::build_reduce_to_nd(const std::vector<std::stri
     {
         codegen::CodeWriter writer;
         CudaKernelBuilder::add_pod_typedefs(writer);
-        writer << include_helpers();
         if (kernel)
         {
             CudaKernelBuilder::get_device_helper(
@@ -1789,7 +1786,6 @@ size_t runtime::gpu::CUDAEmitter::build_reduce_to_scalar(const std::vector<std::
     {
         codegen::CodeWriter writer;
         CudaKernelBuilder::add_pod_typedefs(writer);
-        writer << include_helpers();
         if (kernel)
         {
             CudaKernelBuilder::get_device_helper(
@@ -1858,7 +1854,6 @@ size_t runtime::gpu::CUDAEmitter::build_reduce_to_scalar_acc(const std::vector<s
     {
         codegen::CodeWriter writer;
         CudaKernelBuilder::add_pod_typedefs(writer);
-        writer << include_helpers();
         if (kernel)
         {
             CudaKernelBuilder::get_device_helper(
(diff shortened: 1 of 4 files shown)


fengleitian:
what's the bug?

src/ngraph/runtime/gpu/nvrtc/helpers.hpp
```diff
@@ -0,0 +1,208 @@
...
+    ss << R"(
+#define WARP_SIZE 32
+#define __ballot_sync(mask, predicate) __ballot(predicate)
+#define __shfl_down_sync(mask, val, delta, width) __shfl_down(val, delta, width)
+#define __shfl_xor_sync(mask, val, laneMask, width) __shfl_xor(val, laneMask, width)
+)";
+#endif
+
+    // add modern type definitions
+    ss << "typedef signed char int8_t;\n";
```
rkimballn1: Don't know if it matters but `using` statements are more modern.
`using int8_t = signed char`

csullivan: Yeah, I would prefer using as well, but it is unfortunately not supported by nvrtc.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -2843,109 +2835,6 @@ void runtime::gpu::CUDAEmitter::print_tensor_from_gpu(codegen::CodeWriter& write
...
-    asm("vmad.s32.u32.u32 %0, %1.h0, %2.h0, %3;" : "=r"(res) : "r"(a), "r"(b), "r"(c));
-    return res;
-}
-__device__ __forceinline__ int msub16(int a, int b, int c)
-{
-    int res;
-    asm("vmad.s32.u32.u32 %0, %1.h0, %2.h0, -%3;" : "=r"(res) : "r"(a), "r"(b), "r"(c));
-    return res;
-}
-__device__ __forceinline__ float  load(const float*  __restrict__ in, int i=0, bool b=true)
```
csullivan: > what's the bug?

@fengleitian, there was no `double load(const double* ...)` overload.

src/ngraph/runtime/gpu/nvrtc/helpers.hpp
```diff
@@ -0,0 +1,208 @@
...
+}
+)";
+    return ss.str();
+}
+
+std::string ngraph::runtime::gpu::nvrtc::define_zero(const std::string& dtype,
+                                                     const std::string& name)
+{
+    std::stringstream ss;
+    std::string zero = "static_cast<" + dtype + ">(0)";
```
fengleitian: 0 does not need to cast.

src/ngraph/runtime/gpu/nvrtc/helpers.hpp
```diff
@@ -0,0 +1,208 @@
...
+}
+
+std::string ngraph::runtime::gpu::nvrtc::define_vzero(std::string dtype,
+                                                      const uint32_t& n,
+                                                      const std::string& name)
+{
+    std::stringstream ss;
+    if (n == 1 || n == 2 || n == 4)
+    {
+        std::string zero = "static_cast<" + dtype + ">(0)";
```
fengleitian: same here.

src/ngraph/runtime/gpu/nvrtc/helpers.hpp
```diff
@@ -0,0 +1,208 @@
...
+        throw ngraph_error("Invalid request for vector zero of " + dtype + std::to_string(n));
+    }
+    return ss.str();
+}
+#define LOAD_C                                                                                     \
+    "__device__ __forceinline__ " << dtype << " " << name << "(const " << dtype                    \
+                                  << "*  __restrict__ in, int i=0, bool b=true) { " << dtype       \
+                                  << "  v; zero_(v); if (b) v = in[i]; return v; }\n"
+
+std::string ngraph::runtime::gpu::nvrtc::define_coherent_load(const std::string& dtype,
```
fengleitian: do you mean `aligned/unaligned` here with `coherent/noncoherent`?

csullivan: Nope, the `__ldg()` intrinsic is a read only load that is not coherent with global memory (updates are not propagated back to DRAM). Recently it has been implemented as a pull into constant memory for fast access that can help performance for memory bound kernels.

csullivan: `define_coherent_load` creates a load function that uses `ptr[offset]` for doing a coherent load from global memory.

`define_non_coherent_load` defines a load that uses `__ldg()`.

src/ngraph/runtime/gpu/nvrtc/helpers.hpp
csullivan: fixed.

src/ngraph/runtime/gpu/nvrtc/helpers.hpp
csullivan: fixed.

src/ngraph/runtime/gpu/nvrtc/helpers.hpp
csullivan: We can change the name of these templates if that's preferred, I was just trying to be explicit. Maybe there is something better.

fengleitian: ok, got, name is good if that's the case.

PR #1979, opened 2018-11-04
NOP backend

rkimballn1:
This is a simple NOP backend which does nothing. It will be used to benchmark backend call overhead.
There were a couple of unit tests that actually passed with this backend so I "fixed" them so they now fail with NOP as they should. Backend tests can't be run with this backend because everything will fail.


test/build_graph.cpp
```diff
@@ -25,6 +25,16 @@
 using namespace std;
 using namespace ngraph;
 
+TEST(build_graph, broadcast_to_non_existent_axis)
```
aprocter: FWIW this test can probably be dropped altogether. It seems to be a less precise version of `type_prop.broadcast_axes_oob`.

aprocter:
@shssf I believe @rkimballn1's original motivation for moving it out of `backend_test.in.cpp` is that it's not actually a backend test. The expected failure occurs in type propagation, and we never even hit the backend.

The reason I'd favor scrapping it is that it seems to be redundant due to the existence of https://github.com/NervanaSystems/ngraph/blob/d9f615b7a9ca5b4d2611744eade4687f7e01cc29/test/type_prop.cpp#L37 .

It is true that these are not _technically_ testing _quite_ the same thing: the shapes/axes are different, the removed test has the op construction inside the constructor call for `Function`, and the removed test is a bit less specific about the exception. I don't think the removed test represents any particularly important boundary values, but if we're not sure then we could move it over to `type_prop`.

PR #1980, opened 2018-11-04
address issues which surface with clang 6.0

rkimballn1:


src/ngraph/runtime/cpu/builder/dot.cpp
```diff
@@ -175,56 +175,55 @@ namespace ngraph
...
+                                              &n,
+                                              &k,
+                                              &alpha,
+                                              a_array,
+                                              &lda_array,
+                                              b_array,
+                                              &ldb_array,
+                                              &beta,
+                                              c_array,
+                                              &ldc_array,
```
jbobba: group_count is being used here. So we need to pass it into the lambda. Clang complains since it does constant propagation through the lambda. Other compilers are not required to do this which will lead to incorrect execution. i think this has come up once before. we might want to turn off clang checks around this function instead.

rkimballn1: If this warning can lead to programming errors then I think it best we just disable it. I have reverted the changes to dot.cpp and pass/cpu_fusion.cpp and added the lambda warning to our whitelist for the entire project.

PR #1982, opened 2018-11-05
Ayzhuang/propagate cacheability

ayzhuang:


src/ngraph/runtime/cpu/cpu_op_annotations.cpp
```diff
@@ -0,0 +1,38 @@
...
+    namespace runtime
+    {
+        namespace cpu
+        {
+            std::function<std::shared_ptr<ngraph::op::util::OpAnnotations>(void)>
+                get_op_annotations_func()
+            {
+                std::function<std::shared_ptr<ngraph::op::util::OpAnnotations>(void)> func =
+                    []() -> std::shared_ptr<ngraph::op::util::OpAnnotations> {
+                    auto op_annotations = std::make_shared<ngraph::op::util::OpAnnotations>();
```
jbobba: Create an return CPUOpAnnotations here.

src/ngraph/runtime/cpu/mkldnn_utils.cpp
```diff
@@ -639,7 +639,13 @@ bool runtime::cpu::mkldnn_utils::is_mkldnn_padded_layout(const mkldnn::memory::d
 bool runtime::cpu::mkldnn_utils::use_mkldnn_kernel(const ngraph::Node* node)
 {
     auto op_annotations = static_cast<const ngraph::op::Op*>(node)->get_op_annotations();
-    return (op_annotations &&
-            static_pointer_cast<ngraph::runtime::cpu::CPUOpAnnotations>(op_annotations)
-                ->is_mkldnn_op());
+    if (op_annotations)
```
jbobba: Since all op annotations will be created in the backend, we should be able to statically cast to CPUOpAnnotations.

src/ngraph/op/util/op_annotations.hpp
```diff
@@ -54,9 +54,13 @@ namespace ngraph
                     return m_in_place_oi_pairs;
                 }
 
+                bool is_cacheable() { return m_cacheable; }
```
rkimballn1: make function const

src/ngraph/pass/manager_state.hpp
```diff
@@ -60,7 +60,19 @@ class ngraph::pass::ManagerState
         return m_visualize_tree_ops_map;
     }
 
+    void set_op_annotations_func(
+        std::function<std::shared_ptr<ngraph::op::util::OpAnnotations>(void)> func)
+    {
+        m_op_annotations_func = func;
+    }
+
+    std::function<std::shared_ptr<ngraph::op::util::OpAnnotations>(void)> get_op_annotations_func()
```
rkimballn1: make function const

src/ngraph/pass/propagate_cacheability.cpp
```diff
@@ -0,0 +1,71 @@
...
+            {
+                bool cacheable = true;
+                for (auto arg : node->get_arguments())
+                {
+                    NGRAPH_DEBUG << "propagate cacheability: arg is " << arg->get_name();
+                    if (auto arg_op = std::dynamic_pointer_cast<op::Op>(arg))
+                    {
+                        auto arg_op_annotations = arg_op->get_op_annotations();
+                        NGRAPH_ASSERT(arg_op_annotations);
+                        cacheable = cacheable && arg_op_annotations->is_cacheable();
```
Krovatkin: nitpick: we could probably have an early break here: as soon as we see a non-cacheable argument, there should be no point in iterating over the rest of args?

src/ngraph/pass/manager.cpp
```diff
@@ -98,6 +99,10 @@ void ngraph::pass::Manager::run_passes(shared_ptr<Function> func, bool transitiv
         }
         else if (function_pass)
         {
+            if (auto pc_pass = dynamic_pointer_cast<pass::PropagateCacheability>(function_pass))
+            {
+                pc_pass->set_op_annotations_func(get_state().get_op_annotations_func());
```
Krovatkin: can we come up with a better name? `set_annotation_factory`, `set_annotation_creator`, `set_annotation_producer`? @jbobba

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1054,6 +1057,7 @@ void runtime::cpu::CPU_ExternalFunction::register_common_passes(ngraph::pass::Ma
     pass_manager.register_pass<runtime::cpu::pass::CPUMemoryOptimization>();
     pass_manager.register_pass<ngraph::pass::GetOutputElementElimination>();
     pass_manager.get_state().set_visualize_tree_ops_map(runtime::cpu::get_visualize_tree_ops_map());
+    pass_manager.get_state().set_op_annotations_func(runtime::cpu::get_op_annotations_func());
```
Krovatkin: @ayzhuang @jbobba @diyessi  I was wondering if we could directly inject it into a pass instead of going through `State`like we did for CSE : https://github.com/NervanaSystems/ngraph/pull/1972/files#diff-a8b86c1cce292e0e2e89b028c5518ad6R1053 .There's a very good reason why `VisualizeTree` needs to go through state because we create more than one VisualizeTree pass dynamically in PassManager.

src/ngraph/op/util/op_annotations.hpp
```diff
@@ -54,9 +54,13 @@ namespace ngraph
                     return m_in_place_oi_pairs;
                 }
 
+                bool is_cacheable() const { return m_cacheable; }
+                void set_cacheable(bool val) { m_cacheable = val; }
             private:
                 // map of output-input pairs for which in-place computation is valid
                 std::vector<struct oi_pair> m_in_place_oi_pairs;
+
+                bool m_cacheable = false;
```
aprocter: Not specific to this PR, but I really don't think we should have member variables in this class at all---including the ones that are already there. Originally, it was intended to be a completely abstract base class(though the absence of a pure virtual, or something like that, may have made that technically not the case), with specific backends subclassing it (e.g. `CPUOpAnnotations` and `GPUOpAnnotations`). I recognize that there's commonality between `CPUOpAnnotations` and `GPUOpAnnotations`, but I think we need to think a better way to handle that commonality than just jamming more and more fields into this class.

src/ngraph/pass/manager.cpp
jbobba: set_annotation_factory

src/ngraph/op/util/op_annotations.hpp
Krovatkin: any reason not to? I thought it was actually conceived as a kitchen sink for ad-hoc node attributes?

aprocter: I think of the `OpAnnotations` class itself as being an abstract base class for whatever particular brand of kitchen sink a backend wants. Putting a field all the way up in `OpAnnotations` means that every backend that attaches `OpAnnotations` has this field whether it wants it or not. If we think all backends need this stuff, then I don't think `OpAnnotations` is where it belongs.

jbobba: Agree with @aprocter The only issue is that when we want to use some OpAnnotation info in core passes (like what we have in MemoryLayout), we will end up with ugly dynamic_casts

aprocter: Ah, that's a tricky one! Anyway, no need to stop the presses over this.

PR #1983, opened 2018-11-05
cse for convert layout

pruthviIntel:


src/ngraph/runtime/cpu/cpu_cse.cpp
```diff
@@ -29,7 +29,25 @@ using namespace std;
 
 static bool cse_convertlayout(std::shared_ptr<Node> a, std::shared_ptr<Node> b)
 {
-    return false;
+    NGRAPH_DEBUG << "In cse_convertlayout for " << a->get_name() << " and " << b->get_name();
+
+    auto ar_a = std::static_pointer_cast<runtime::cpu::op::ConvertLayout>(a);
+    auto ar_b = std::static_pointer_cast<runtime::cpu::op::ConvertLayout>(b);
+
+    auto get_tvl = [](std::shared_ptr<Node> node) {
```
diyessi: Since we don't have tensor views any more, we should call these something other than tv. Tensor?

pruthviIntel: sure make sense.

pruthviIntel: @diyessi done.

src/ngraph/runtime/cpu/cpu_cse.cpp
```diff
@@ -29,7 +29,26 @@ using namespace std;
...
-    return false;
+    NGRAPH_DEBUG << "In cse_convertlayout for " << a->get_name() << " and " << b->get_name();
+
+    auto ar_a = std::static_pointer_cast<runtime::cpu::op::ConvertLayout>(a);
+    auto ar_b = std::static_pointer_cast<runtime::cpu::op::ConvertLayout>(b);
+
+    // gets the tensor layout from the given node
+    auto get_tl = [](std::shared_ptr<Node> node) {
+        auto tensor = node->get_output_tensor_ptr();
+        auto cpu_tl = dynamic_cast<ngraph::runtime::cpu::LayoutDescriptor*>(
```
jbobba: static_cast<> since we know that ConvertLayout can only come from the CPU backend.

src/ngraph/runtime/cpu/cpu_cse.cpp
```diff
@@ -29,7 +29,26 @@ using namespace std;
...
+            tensor->get_tensor_layout().get());
+        return cpu_tl;
+    };
+
+    auto a_layout_desc = dynamic_cast<runtime::cpu::LayoutDescriptor*>(get_tl(a));
```
jbobba: do we need to cast again?

src/ngraph/runtime/cpu/cpu_cse.cpp
```diff
@@ -29,7 +29,26 @@ using namespace std;
...
 static bool cse_convertlayout(std::shared_ptr<Node> a, std::shared_ptr<Node> b)
 {
-    return false;
+    NGRAPH_DEBUG << "In cse_convertlayout for " << a->get_name() << " and " << b->get_name();
+
+    auto ar_a = std::static_pointer_cast<runtime::cpu::op::ConvertLayout>(a);
+    auto ar_b = std::static_pointer_cast<runtime::cpu::op::ConvertLayout>(b);
+
+    // gets the tensor layout from the given node
+    auto get_tl = [](std::shared_ptr<Node> node) {
```
Krovatkin: if you call this lambda `get_tensor_layout` you won't probably need that comment :-)

src/ngraph/runtime/cpu/cpu_cse.cpp
Krovatkin: @jbobba does the cast (i.e. `LayoutDescriptor`) always succeed?

pruthviIntel: @Krovatkin . it should, we make sure all the tensor will be populated with `LayoutDescriptor` in `cpu_layout` pass.

src/ngraph/runtime/cpu/cpu_cse.cpp
jbobba: @Krovatkin https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/op/convert_layout.cpp#L63

PR #1988, opened 2018-11-05
nvgpu cuda reduce

fengleitian:
Use cuda instead of cudnn, add support for `And/Or`.


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1021,12 +1021,22 @@ void runtime::gpu::GPU_Emitter::emit_Reduce(EMIT_ARGS)
...
-                // 2. the op should be in the op_map
-                // otherwise, throw an error message
-                cudnnReduceTensorOp_t reduce_tensor_op;
+                auto axes_set = reduce_op->get_reduction_axes();
+                ngraph::AxisVector axes_vec;
+                for (auto a : axes_set)
+                {
+                    axes_vec.push_back(a);
+                }
+                vector<string> dtypes;
```
csullivan: nit: std::vector, to be consistent with the rest of your standard lib use.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1038,35 +1048,53 @@ void runtime::gpu::GPU_Emitter::emit_Reduce(EMIT_ARGS)
...
-                    external_function->get_primitive_emitter()->get_cudnn_emitter();
-                auto reduce_index =
-                    cudnn_emitter->build_reduce_forward(reduce_tensor_op,
-                                                        dtypes,
-                                                        args[0].get_shape(),
-                                                        reduce_op->get_reduction_axes(),
-                                                        CUDNNEmitter::ReductionMode::Reduce);
 
+                auto& fn = *reduce_op;
+                if (TI(fn) == TI(ngraph::op::Add))
```
rkimballn1: We really need a better way to test if a node is a particular type.

csullivan: We should down cast here `if (std::dynamic_pointer_cast<ngraph::op::Add>(fn))`

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: fixed.

src/ngraph/runtime/gpu/gpu_emitter.cpp
fengleitian: maybe node should have a enum of unique name.

fengleitian:
@csullivan https://github.com/NervanaSystems/ngraph/pull/1988/files#diff-036693cdc95f05c01e4fdbbc1c22e1aaR2029.  will be resolved in PR2014.

src/ngraph/runtime/gpu/gpu_emitter.cpp
csullivan: Can we change these conditionals to use dynamic_pointer_casts? This is the proper way of using RTTI in this case. type_index works, but it is a bit more expensive

PR #1994, opened 2018-11-06
[ONNXIFI] 2. Add exceptions for statuses

apwojcik:
This PR adds several exception types to easily distinguish between different ONNXIFI statuses. Internally ONNXIFI-nGraph backed is implemented in C++ and exceptions are used to handle error conditions. However, ONNXIFI is C API and for each error condition, the function needs to return appropriate status code instead.

### DIFF
--- a/src/ngraph/frontend/onnxifi/exceptions.hpp
+++ b/src/ngraph/frontend/onnxifi/exceptions.hpp
@@ -1,0 +1,323 @@
+//*****************************************************************************
+// Copyright 2017-2018 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


src/ngraph/frontend/onnxifi/exceptions.hpp
```diff
@@ -0,0 +1,323 @@
...
+
+            struct backend_unavailable : runtime
+            {
+                constexpr backend_unavailable()
+                    : runtime{ONNXIFI_STATUS_BACKEND_UNAVAILABLE}
+                {
+                }
+            };
+
+        } // namespace error
```
arogowie-intel: Typo - should be `namespace status`

PR #1996, opened 2018-11-06
[ONNXIFI] 3. Add class Span

apwojcik:
Template class implementing spanning class with dynamic extent. The class replaces `(pointer, length)` pair to refer a sequence of contiguous objects. It can be thought of as a pointer to an array or a vector, but it knows its bounds. However, a `Span` does not own the elements it points to. It is not a container like an `std::array` or an `std::vector`, it is a view of the contents of such a container.

By dynamic extent, it means the length is known at runtime, in contrast to a static extent where length is deduced from a template parameter.

### DIFF
--- a/src/ngraph/frontend/onnxifi/span.hpp
+++ b/src/ngraph/frontend/onnxifi/span.hpp
@@ -1,0 +1,110 @@
+//*****************************************************************************
+// Copyright 2017-2018 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


src/ngraph/frontend/onnxifi/span.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include <cstddef>
+#include <iterator>
+#include <stdexcept>
+
+namespace ngraph
+{
+    namespace onnxifi
+    {
+        template <typename T>
+        class Span
```
rkimballn1: Do we really need this? I looks like a standard container. Can't we just use a standard container?
There are no comments describing what this is.
There are no unit tests for this.

postrational: Please add a note about this as documentation.

rkimballn1: @apwojcik you can move your comment you just wrote into the header file as a comment. This is information which should have been in the header in the first place.
Does Span function identically to std::span? If so then you can put a c++ version check around it and put it in the std namespace under the class name `std::span`. That way when we switch to c++20 all of the code works the same.

rkimballn1:
There are still no unit tests for Span. Please add them

rkimballn1:
Now if we can just get some unit tests for this we will be good to go

apwojcik:
Unit test added

PR #1998, opened 2018-11-06
Add a real HybridBackend

rkimballn1:
The GPU backend was used as a test case for hybridizing but its `is_supported` function is in no way complete, just a proof of concept.
The proof of concept hybrid backend contains only the GPU with no backup backend. Anything not implemented by the GPU throws an exception which is what I wanted for testing that this check works. It is trivial to add a backup/fallback backend to make everything work but that is another PR.


src/ngraph/runtime/interpreter/int_backend.hpp
```diff
@@ -163,6 +163,7 @@ class ngraph::runtime::interpreter::INTBackend : public Backend
     std::vector<PerformanceCounter>
         get_performance_data(std::shared_ptr<Function> func) const override;
 
+    bool is_supported(const Node& node) const override { return true; }
```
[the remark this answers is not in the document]
rkimballn1: I think the code speaks for itself. The sad fact is that the interpreter does have a few unit tests in it's manifest file meaning that there are things it does not support.

src/ngraph/runtime/tensor.hpp
```diff
@@ -20,6 +20,7 @@
 #include <vector>
 
 #include "ngraph/descriptor/tensor.hpp"
+#include "ngraph/descriptor/layout/tensor_layout.hpp"
```
sasadep: is this needed ?

rkimballn1: Yes. Without this there is a compile error.

src/ngraph/runtime/hybrid/hybrid_backend.cpp
```diff
@@ -74,49 +44,149 @@ std::vector<T> read_vector(std::shared_ptr<ngraph::runtime::Tensor> tv)
...
     return true;
 }
 
-bool runtime::hybrid::HYBRIDBackend::call(shared_ptr<Function> function,
+bool runtime::hybrid::HybridBackend::call(shared_ptr<Function> func,
                                           const vector<shared_ptr<runtime::Tensor>>& outputs,
                                           const vector<shared_ptr<runtime::Tensor>>& inputs)
 {
-    validate_call(function, outputs, inputs);
+    // auto it = m_backend_list.begin();
```
diyessi: debug code?

src/ngraph/runtime/hybrid/hybrid_backend.cpp
```diff
@@ -74,49 +44,149 @@ std::vector<T> read_vector(std::shared_ptr<ngraph::runtime::Tensor> tv)
...
+    if (it == m_function_map.end())
+    {
+        compile(func);
+        it = m_function_map.find(func);
+    }
 
-    compile(function);
+   if (it == m_function_map.end())
+    {
+        throw runtime_error("Error constructing backend.");
```
diyessi: This needs to be a little more informative,

rkimballn1: This will never happen. The only way that the function map won't contain an instance is if the compile() function throws an exception in which case we won't even execute this code. I added a little more info to the message but there is no much to say since it really will never happen.

src/ngraph/runtime/hybrid/hybrid_backend.cpp
```diff
@@ -74,49 +44,149 @@ std::vector<T> read_vector(std::shared_ptr<ngraph::runtime::Tensor> tv)
...
-    return make_shared<runtime::HostTensor>(type, shape, memory_pointer, "external");
+    auto it = m_backend_list.begin();
+    return it->second->create_tensor(element_type, shape, memory_pointer);
 }
 
-bool runtime::hybrid::HYBRIDBackend::compile(shared_ptr<Function> function)
+bool runtime::hybrid::HybridBackend::compile(shared_ptr<Function> func)
 {
-    if (m_function_map.find(function) == m_function_map.end())
+    // auto it = m_backend_list.begin();
```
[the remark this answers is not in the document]
rkimballn1: removed

PR #2017, opened 2018-11-07
[ONNX] return single nGraph function with multiple outputs

apwojcik:
**This PR changes ONNX importer API.**

Previously multiple nGraph functions were returned each representing single output (result) from a model (in case of multiple outputs). This PR is changing this ONNX importer behavior and multiple outputs for a single nGraph function representing ONNX model (graph) is returned instead.


diyessi:
Looks like there is a variable name mismatch in a unit test.

PR #2018, opened 2018-11-07
Hybrid groundwork

rkimballn1:
A bit of shared_ptr cleanup. Trying to make functions that don't need a shared_ptr not pass a shared_ptr.
Also added make_unique because I can.


src/ngraph/util.hpp
```diff
@@ -217,3 +217,43 @@ namespace ngraph
 } // end namespace ngraph
 
 std::ostream& operator<<(std::ostream& os, const ngraph::NodeVector& nv);
+
+#if (__cplusplus < 201402)
+// Copied from https://isocpp.org/files/papers/N3656.txt
```
diyessi: This needs to go into the BOM first.

PR #2019, opened 2018-11-07
Fix gtest and LLVM builds with ABI flag setting

rkimballn1:


ransford2011:
@en-em .  Please see this PR FYI

PR #2022, opened 2018-11-08
[WIP][test]Compare backend using serialized graph.

fengleitian:


src/tools/nbench/benchmark.hpp
```diff
@@ -34,3 +39,118 @@ std::vector<ngraph::runtime::PerformanceCounter> run_benchmark(std::shared_ptr<n
...
+
+bool all_close_f(const std::vector<float>& a,
+                 const std::vector<float>& b,
+                 int mantissa_bits,
+                 int tolerance_bits);
+
+/// \brief A predictable pseudo-random number generator
+/// The seed is initialized so that we get repeatable pseudo-random numbers for tests
+template <typename T>
+class Uniform
```
rkimballn1: Many of these functions are available in test utils. delete the copies

test/backend_test.in.cpp
```diff
@@ -5431,3 +5431,61 @@ NGRAPH_TEST(${BACKEND_NAME}, batchnorm_fprop_bprop_2step)
...
+    for (shared_ptr<op::Parameter> param : new_func->get_parameters())
+    {
+        vector<float> tensor_val(shape_size(param->get_shape()));
+        rng.initialize(tensor_val);
+        args.push_back(tensor_val);
+    }
+
+    auto cpu_func = clone_function(*new_func);
+    auto bk_func = clone_function(*new_func);
+    auto cpu_results = execute(cpu_func, args, "CPU");
```
rkimballn1: This test is only valid if the optional CPU backend is built. You need to remove this test otherwise.

src/tools/nbench/benchmark.cpp
```diff
@@ -253,3 +254,125 @@ vector<runtime::PerformanceCounter> run_benchmark(shared_ptr<Function> f,
...
+    {
+        vector<float> tensor_val(shape_size(param->get_shape()));
+        rng.initialize(tensor_val);
+        args.push_back(tensor_val);
+    }
+
+    auto cpu_func = ngraph::clone_function(*new_func);
+    auto bk_func = ngraph::clone_function(*new_func);
+
+    auto cpu_results = execute(cpu_func, args, "CPU");
```
rkimballn1: This is only valid if CPU backend is built. You need to make this code conditional. Move code to separate source file and only add it if the CPU is built.

src/tools/nbench/nbench.cpp
```diff
@@ -225,6 +226,10 @@ int main(int argc, char** argv)
         {
             timing_detail = true;
         }
+        else if (arg == "--validation")
```
rkimballn1: This option is only valid if the CPU backend is built. Add condition around it.

PR #2030, opened 2018-11-08
Interpreter rework

rkimballn1:
I would like to change from passing vector<shared_ptr<Tensor>> to just passing vector<Tensor*> for the call() method on the backend, eliminating the creation of shared_ptr to put into the array.
I was playing with this change and found that the Interpreter backend depended heavily on shared_ptr internally so I decided to first address this issue on the Interpreter implementation then address the API change in a separate PR.

This PR changes the Interpreter internals to use Liveness/MemoryLayout for all temporaries and then pass around void* for buffers.


src/ngraph/runtime/interpreter/int_backend.cpp
```diff
@@ -134,35 +145,42 @@ bool runtime::interpreter::INTBackend::call(shared_ptr<Function> function,
...
 
         // get op outputs from map or create
-        vector<shared_ptr<runtime::HostTensor>> op_outputs;
+        vector<void*> op_outputs;
+        vector<shared_ptr<runtime::HostTensor>> htv_outputs;
         for (size_t i = 0; i < op->get_output_size(); ++i)
         {
             descriptor::Tensor* tv = op->get_output_tensor_ptr(i).get();
-            shared_ptr<runtime::HostTensor> htv;
+            void* htv = nullptr;
```
aprocter: Might rename this variable since it is no longer a  `HostTensor` (but then there are vestigial `v`s throughout anyway, so maybe that's another PR).

src/ngraph/runtime/interpreter/int_backend.cpp
```diff
@@ -202,20 +220,7 @@ bool runtime::interpreter::INTBackend::call(shared_ptr<Function> function,
         }
         if (instance.m_nan_check_enabled)
         {
-            perform_nan_check(op_outputs, op);
-        }
-
-        // delete any obsolete tensors
-        for (const descriptor::Tensor* t : op->liveness_free_list)
```
aprocter: Should we not still do this cleanup one way or another?

src/ngraph/runtime/interpreter/int_backend.cpp
```diff
@@ -355,3 +360,29 @@ void runtime::interpreter::INTBackend::perform_nan_check(const vector<shared_ptr
         arg_number++;
     }
 }
+
+void runtime::interpreter::INTBackend::create_tensor_array(
```
aprocter: Maybe `arrays` (shrug)

src/ngraph/runtime/interpreter/int_backend.hpp
```diff
@@ -408,115 +427,104 @@ class ngraph::runtime::interpreter::INTBackend : public Backend
...
                                  in_shapes,
-                                 out[0]->get_shape(),
+                                 node.get_output_shape(0),
                                  concat->get_concatenation_axis());
             break;
         }
         case OP_TYPEID::Constant:
         {
-            const op::Constant* c = static_cast<const op::Constant*>(&node);
-            reference::constant<T>(
```
aprocter: Maybe we should remove `reference::constant` at some point, if it's not used elsewhere.

src/ngraph/runtime/interpreter/int_backend.cpp
rkimballn1: We never allocate temporaries except for the single AlignedBuffer which is deleted on backend destruction

src/ngraph/runtime/interpreter/int_backend.hpp
rkimballn1: Sounds like a good idea. I think Constant is handled uniquely for each backend and none use reference::constant

src/ngraph/runtime/interpreter/int_backend.cpp
rkimballn1: I figured since I am here already I will rename them, so I did.

Issue #1989, opened 2018-11-05
tensorflow subgraph : nn_test failure 2


csullivan:
In `SoftmaxTest.testGradient` and `MomentsTest.doOutput4DInput123`.
TF script, serialized graph, test log, and nbench result for one of the offending serialized graphs attached. 
[nn_test2.zip](https://github.com/NervanaSystems/ngraph/files/2550954/nn_test2.zip)

PR #2033, opened 2018-11-09
Add cudnn sum and test that exposes tf test failure for nvGPU

AyanmoI:
This change fixes tensorflow test failures


src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -1563,6 +1563,11 @@ void runtime::gpu::GPU_Emitter::emit_Subtract(EMIT_ARGS)
 }
 
 void runtime::gpu::GPU_Emitter::emit_Sum(EMIT_ARGS)
+{
+    runtime::gpu::GPU_Emitter::emit_Sum_1(external_function, writer, node, args, out);
+}
+
+void runtime::gpu::GPU_Emitter::emit_Sum_0(EMIT_ARGS)
```
rkimballn1: Comments for something ambiguously named are mandatory. What does this function do? Same comment for Sum_1
Also in comment, why this is better or worse or whatever than other functions. What does this function bring to the table?

test/backend_sum.in.cpp
```diff
@@ -427,3 +427,30 @@ NGRAPH_TEST(${BACKEND_NAME}, sum_5d_to_scalar)
...
+    {
+        vector<float> tensor_val(shape_size(param->get_shape()));
+        rng.initialize(tensor_val);
+        args.push_back(tensor_val);
+    }
+
+    auto cpu_func = clone_function(*f);
+    auto bk_func = clone_function(*f);
+
+    auto cpu_results = execute(cpu_func, args, "CPU");
```
rkimballn1: This test is only valid if the CPU backend is built. Please make it conditional on that.

jbobba: why are we comparing to CPU and not interpreter?

src/ngraph/runtime/cpu/unit_test.manifest
```diff
@@ -12,3 +12,4 @@ one_hot_vector_1_fp
 one_hot_vector_1_fp_nonint
 backwards_batch_norm_three_outputs
 backwards_maxpool_n2_c1_hw5_3x3_str2_max_pad1x2_2x3
+sum_stable_acc
```
jbobba: Would prefer to have only failing tests here.

test/backend_sum.in.cpp
AyanmoI: INTERPRETER is also producing incorrect results. The purpose of this test is address the issue of failures in tensorflow tests. Both GPU and INTERPRETER produced similar results which differs from tensorflow. CPU  passes the tf tests.

csullivan: @AyanmoI, if it is easy, we could quickly change the INTERPRETER to use the kahan sum and try comparing against it. Can you take a look?

aprocter: The current situation with respect to `Sum` is a bit messy. `INTERPRETER` is a naive sum, while `CPU` was switched to a Kahan sum some time ago. The behavior should not be inconsistent like this; if it's part of the spec that `Sum` needs to be (or for that matter is _allowed_ to be) a stable sum, we should write that into the spec.

@rkimballn1 has raised this issue a number of times, but we haven't gotten around to resolving it. I'd actually favor separate `Sum` and `StableSum` ops, and leaving it up to the bridge to select the correct one.

aprocter: Also plausible: the current `Sum`'s spec could be updated with some kind of stability guarantee, and we could add a `NaiveSum` if anybody actually wants it.

src/ngraph/runtime/gpu/gpu_emitter.cpp
AyanmoI: Done

AyanmoI: Done

test/backend_sum.in.cpp
AyanmoI: Done

AyanmoI: I just pushed a kahan sum implementation for INTERPRETER `Sum`

AyanmoI: Also @jbobba , changed to comparing with INTERPRETER rather than CPU

AyanmoI: Please suggest if there needs to be two separate `sum`s for INTERPRETER

test/backend_sum.in.cpp
```diff
@@ -427,3 +427,38 @@ NGRAPH_TEST(${BACKEND_NAME}, sum_5d_to_scalar)
...
+    test::Uniform<float> rng(1000.0f, 1000.1f, 2112);
+    vector<vector<float>> args;
+    for (shared_ptr<op::Parameter> param : f->get_parameters())
+    {
+        vector<float> tensor_val(shape_size(param->get_shape()));
+        rng.initialize(tensor_val);
+        args.push_back(tensor_val);
+    }
+
+    auto cpu_func = clone_function(*f);
```
jbobba: minor nit. Change to ref_func and ref_results.

AyanmoI:
Reference Issue: https://github.com/NervanaSystems/ngraph/issues/1989

PR #2040, opened 2018-11-09
Update l2_norm and std_dev builders to use op::Sqrt

aprocter:


src/ngraph/builder/reduce_ops.cpp
```diff
@@ -24,6 +24,7 @@
 #include "ngraph/op/multiply.hpp"
 #include "ngraph/op/power.hpp"
```
mbrookhart: Can you get rid of the power include?

aprocter: Just noticed that  :)

PR #2042, opened 2018-11-10
Support dynamic scales for quantization in the CPU backend

jbobba:


src/ngraph/runtime/cpu/builder/quantization.cpp
```diff
@@ -165,24 +166,58 @@ namespace ngraph
...
+                        // execution
+                        scales.push_back(1.0f);
+                        size_t quantize_index =
+                            mkldnn_emitter->build_quantize_reorder(input_desc, result_desc, scales);
+                        auto& deps = mkldnn_emitter->get_primitive_deps(quantize_index);
+                        auto functor = [&, input_desc, result_desc, scales_size](
+                            CPURuntimeContext* ctx, CPUExecutionContext* ectx) {
+                            // Create MKLDNN reorder primitive during the first iteration.
+                            // Assumes the scales dont change for the duration of the graph
+                            static mkldnn::reorder r_prim = [&]() {
```
ashokei: can we make sure this static gets recreated for a different instance of quantize op.

diyessi: It shouldn't matter -- if it references something in the functor and/or node, it will be recreated, if not, it doesn't matter.

nbpatel: Won't we need to add the same stuff for Codegen as well?

ashokei: @jbobba i think scott addressed it, i wasnt sure if this static variable in this lambda behaves just like regular function static variable (where it gets instantiated once per func definition). for eg:
https://stackoverflow.com/questions/16750933/static-variables-in-c-lambdas

src/ngraph/runtime/cpu/mkldnn_invoke.cpp
```diff
@@ -44,3 +44,16 @@ extern "C" void ngraph::runtime::cpu::mkldnn_utils::mkldnn_invoke_primitive(CPUR
         throw ngraph_error("Could not run mkdnn primitive " + e.message);
     }
 }
+
+extern "C" void ngraph::runtime::cpu::mkldnn_utils::mkldnn_submit(mkldnn::primitive prim)
```
ashokei: LGTM. may want to exclude this as it is not used anywhere.

PR #2020, opened 2018-11-07
Heterogenous serialized graph testing across backends

gcwenger:
Based on work started by @csullivan. Compares serialized graph for a given backend with CPU. Currently only enabled to test GPU backend, although that can be tweaked in future.
Currently compares all intermediate data and so disables any fusions (will likely be extended in future PR).

PR #2044, opened 2018-11-10
Refactor heterogeneous graph compare into a test utility

csullivan:
Minor refactor of @gcwenger's powerful new type-heterogeneous graph comparison test framework. This branch removes the BACKEND_NAME cmake variable dependence and opts for using template params to pick the desired backends to compare. With this change the comparison methods introduced in #2020 can be used as a header utility in individual backend source files.


test/CMakeLists.txt
```diff
@@ -92,7 +92,10 @@ if (NGRAPH_CPU_ENABLE)
 endif()
 
 if(NGRAPH_GPU_ENABLE)
-    list(APPEND SRC cudnn.cpp gpu_test.cpp gpu_fusion.cpp)
+  list(APPEND SRC cudnn.cpp gpu_test.cpp gpu_fusion.cpp)
```
rkimballn1: Indenting looks wrong here. Should be 4 spaces.

csullivan:
For backend comparisons of graphs, there are three identified use cases:
1) Backend specific comparison tests. Tests a specific backend is making to ensure specific regressions
2) Graph specific comparison tests. Tests the results of all backends to ensure each produce equivalent results.
3) A combinations of 1 and 2 in a separate utility binary.

Current `master`(28002287d410d14a610fb4808c038af2ee97c3a6) has partial support for 1 and 2, whereas this PR adds full support for (1) but only partial supports (2) as it would require multiple comparator and test instantiations for each backend. 

Thus, the proposal is to close this PR until all three points above are met.

PR #2046, opened 2018-11-12
Doc for Batchnorm, fix argument order, cleanup some comments

diyessi:
Initial documentation for BatchNorm
Put the arguments in an order more in agreement with other ops. I left the args order as it was since it is hard to change that safely since matchers and emitters are positional.
Clean up some old tensor view comments.

There will be a follow-up PR with the math for the backprop and a bug fix for the reference implementation.


src/ngraph/op/batch_norm.cpp
```diff
@@ -22,99 +22,163 @@
...
-{
-    check_new_args_count(this, new_args);
-    return std::make_shared<BatchNormTraining>(
-        m_epsilon, new_args.at(0), new_args.at(1), new_args.at(2));
+        new_args.at(2), new_args.at(0), new_args.at(1), new_args.at(3), new_args.at(4), m_epsilon);
 }
 
 ngraph::op::BatchNormTrainingBackprop::BatchNormTrainingBackprop(
-    double eps,
+    std::shared_ptr<ngraph::Node> input,
```
aprocter: I agree that it's better to have `input` first, but shouldn't we also reorder the `Input`s, i.e. change the order in which we pass things to `check_single_output_args`? "Too much work for this PR" is an acceptable answer.

doc/sphinx/source/ops/batch_norm_training_backprop.rst
```diff
@@ -0,0 +1,71 @@
...
++----------------------+-------------------------+------------------------------+
+| ``gamma``            | same as ``input``       | :math:`(C)`                  |
++----------------------+-------------------------+------------------------------+
+| ``beta``             | same as ``input``       | :math:`(C)`                  |
++----------------------+-------------------------+------------------------------+
+| ``mean``             | same as ``input``       | :math:`(C)`                  |
++----------------------+-------------------------+------------------------------+
+| ``variance``         | same as ``input``       | :math:`(C)`                  |
++----------------------+-------------------------+------------------------------+
+| ``normalized_delta`` | same as ``input``       | :math:`input`                |
```
aprocter:
```suggestion
| ``normalized_delta`` | same as ``input``       | same as ``input``                |
```

src/ngraph/op/batch_norm.cpp
aprocter: I'm seeing now that you addressed this in your initial comment. This works for me.

diyessi:
Will be incorporated in a new PR.

PR #2049, opened 2018-11-12
Support to selectively enable/disable passes through env variable

jbobba:


src/ngraph/util.cpp
```diff
@@ -498,3 +498,25 @@ AxisVector ngraph::get_permutation_to_default_order(const AxisVector& axis_order
     }
     return out;
 }
+
+void ngraph::get_pass_enables(std::unordered_map<std::string, bool>& enables)
```
diyessi: Instead of doing it this way, have an enables initialized from the environment and functions for setting/clearing values, and for testing if something is set. That way, unit tests can easily change a value. It might be good to have an RAII save/restore mechanism so a unit test didn't accidentally leave something changed and mess up the remaining tests.

jbobba: @diyessi If a unit-test creates a PassConfig object, how would it pass it to the backend transformer. Do you want me to modify the compile() API to take an additional argument?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -179,6 +181,28 @@
 using namespace std;
 using namespace ngraph;
 
+#define REGISTER_KNOBBED_PASS(name, enable_by_default, pass)                                       \
+    if (pass_map.find(name) != pass_map.end())                                                     \
+    {                                                                                              \
+        if (pass_map[name])                                                                        \
+            pass_manager.register_pass<pass>();                                                    \
```
rkimballn1: curly braces around all code blocks

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -179,6 +181,28 @@
...
+    else if (enable_by_default)                                                                    \
+    {                                                                                              \
+        pass_manager.register_pass<pass>();                                                        \
+    }
+
+#define REGISTER_KNOBBED_PASS_WITH_ARGS(name, enable_by_default, pass, ...)                        \
+    if (pass_map.find(name) != pass_map.end())                                                     \
+    {                                                                                              \
+        if (pass_map[name])                                                                        \
+            pass_manager.register_pass<pass>(__VA_ARGS__);                                         \
```
rkimballn1: curly braces here too

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -179,6 +181,28 @@
 using namespace std;
 using namespace ngraph;
 
+#define REGISTER_KNOBBED_PASS(name, enable_by_default, pass)                                       \
```
rkimballn1: You could change this macro so you don't have to pass in the string name and the class name since they are both the same.
```
REGISTER_KNOBBED_PASS(LikeReplacement, true, ngraph::pass)
```
The use the first argument to emit a string for the name and append `::first_arg` to the namespace. Redundancy with strings is difficult to debug and with this change a misspelled class name would be a compile-time error.

If this makes no sense let me know and I can code something up for this.

PR #2052, opened 2018-11-13
runtime::Backend API change

rkimballn1:
Change the Backend API by adding the use of an opaque Handle instead of using shared_ptr<Function> as a handle. This allows us to add a `save()` and `load()` function to the Backend. It also means that the caller does not need to maintain a copy of shared_ptr<Function> for the lifetime of a Backend.
You must now call `compile()` or `load()` prior to calling `call()`
This is a WIP. I still need to get validate() working.

### DIFF
--- a/python/pyngraph/runtime/backend.cpp
+++ b/python/pyngraph/runtime/backend.cpp
@@ -35,22 +35,22 @@ void regclass_pyngraph_runtime_Backend(py::module m)
                     const ngraph::element::Type&, const ngraph::Shape&)) &
                     ngraph::runtime::Backend::create_tensor);
     backend.def("compile",
-                (void (ngraph::runtime::Backend::*)(std::shared_ptr<ngraph::Function>)) &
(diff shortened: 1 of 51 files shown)


shssf:
What if the same `Function` goes into `Backend` twice? Should it be compiled again? As far as I can see you removed `Function` cache at least in `INTELGPU` backend.

rkimballn1:
@avijit-nervana I don't see how I can make an API change only PR without also changing the implementation.
You only need to review the part of the PR that affects you.

rkimballn1:
@shssf I thought if you called compile twice with the same Function you should get two separate handles. However that is handled internally is up to the backend.

avijit-nervana:
> @avijit-nervana I don't see how I can make an API change only PR without also changing the implementation.
> You only need to review the part of the PR that affects you.
You need to refactor the code and submit a new set or PRs

rkimballn1:
I'm going to divide this up into multiple PRs so closing this

PR #2053, opened 2018-11-13
Embedding fprop

Krovatkin:


src/ngraph/op/embedding.cpp
```diff
@@ -0,0 +1,55 @@
...
+    element::Type result_et = get_input_element_type(1);
+
+    const PartialShape& arg0_shape = get_input_partial_shape(0);
+    const PartialShape& arg1_shape = get_input_partial_shape(1);
+
+    PartialShape result_shape;
+
+    if (arg0_shape.rank().is_static() && arg1_shape.rank().is_static())
+    {
+        NODE_VALIDATION_ASSERT(this, static_cast<size_t>(arg1_shape.rank()) == 2)
```
aprocter: We should move this check up, outside the `if`, and make it something like this:

```
NODE_VALIDATION_ASSERT(this, arg1_shape.rank().is_dynamic() || static_cast<size_t>(arg1_shape.rank()) == 2)
    << "blah";
```

That way we'll fail if arg1's rank is bad, even if we don't know arg0's rank statically.

src/ngraph/op/embedding.cpp
```diff
@@ -0,0 +1,55 @@
...
+            << "weights are expected to be a matrix";
+        std::vector<Dimension> result_dims(static_cast<size_t>(arg0_shape.rank()) + 1);
+        for (size_t i = 0; i < static_cast<size_t>(arg0_shape.rank()); i++)
+        {
+            result_dims[i] = arg0_shape[i];
+        }
+        result_dims[result_dims.size() - 1] = arg1_shape[1];
```
aprocter: (In conjunction with the above change), replace `= arg1_shape[1]` with `= (arg1_shape.rank().is_dynamic() ? Dimension::dynamic() : arg1_shape[1])`.

That way, even if we we don't know arg1's rank, but we know that arg0's shape is "(d1,...,dn)" we can still "provisionally" fill the output type in with "(d1,...,dn,?)".

shssf:
I think `case OP_TYPEID::Embedding:` needs to be added at the and of the large switch in `intelgpu_backend.cpp` because `error: enumeration value 'Embedding' not handled in switch` build error in INTELGPU backend.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -441,6 +442,10 @@ bool runtime::intelgpu::IntelGPUBackend::compile(shared_ptr<Function> func)
             function_output_names.push_back(get_input_name(op));
             break;
         }
+        case OP_TYPEID::EmbeddingLookup:
```
shssf: Please move this line to the end of the `switch` into line 1515.

Krovatkin: done

src/ngraph/op/embedding_lookup.cpp
```diff
@@ -0,0 +1,58 @@
...
+    element::Type result_et = get_input_element_type(1);
+
+    const PartialShape& arg0_shape = get_input_partial_shape(0);
+    const PartialShape& arg1_shape = get_input_partial_shape(1);
+
+    PartialShape result_shape;
+
+    if (arg0_shape.rank().is_static())
+    {
+        NODE_VALIDATION_ASSERT(
```
aprocter: Still need to move this up, outside the `if`. We should fail in this case even if arg0's rank is dynamic.

test/type_prop.cpp
```diff
@@ -2385,6 +2386,44 @@ TEST(type_prop, or_bad_arguments)
         });
 }
 
+TEST(type_prop, embedding_lookup_static_shapes)
```
aprocter: These tests are good but we need to implement some tests for failing cases as well.

PR #2054, opened 2018-11-13
Move ParameterVector and ResultVector to the ngraph namespace

rkimballn1:
while there are a lot of changes they all fit into one of these categories
1) move src/ngraph/op/*_vector.hpp to the src/ngraph directory
2) move ngraph::op::ParameterVector to ngraph::ParameterVector. ResultVector was already in the correct namespace
3) change #include statements


indie:
Seems like this could break doc builds...Is this why the other CI tests fail?

PR #2059, opened 2018-11-13
Add test revealing issue in current cudnn pooling implementation.

csullivan:
To fix we will need a core change: `op::MaxPoolBackprop` will need to consume the pooling output as an input argument.

See specifically line 1246:
```
    const size_t num_elements = 3 * 1024 + 1; // change 1 to 0 and GPU will pass on Titan XP
```

Speculation as to the cause: Even though cudnnPoolingBackward takes as an input the output `y` of cudnnPoolingForward, when the tensor fits within L2 cache of the device, it will just recompute the forward pass. Using 256 threads would produce:

(3* 1024 * 4 bytes / thread) * 256 threads = 3145728 bytes = the size of the L2 cache on TitanXP. 

Thus, a tensor of greater size will have global memory spills that reduce performance. To mitigate that, they use the output of the forward pass to simply do a search for the max arg in the input. The search is slower than the MaxPool when the pooling occurs from L2 cache, but faster than pooling when done from global memory.

This would be why our NVGPU implementation of op::MaxPoolBackprop works for all our unit tests even though we haven't provided `y` previously. The tensors are small enough that cuDNN implicitly recalculates it. Can't be sure without inspecting the cudnn binaries, but this is my best guess.


csullivan:
Adding @aprocter and @diyessi, for review as this is the motivation for the change to the MaxPoolBackprop core op.

aprocter:
Also, I think `EXPECT_EQ` instead of some sort of `EXPECT_NEAR` at the end is okay here, because we are taking dy and always multiplying by exactly 0 or exactly 1. But you might want to (1) double-check me on that and (2) explain in a comment.

test/backend_pool.in.cpp
```diff
@@ -1235,3 +1235,84 @@ NGRAPH_TEST_P(${BACKEND_NAME}, avg_pool_3d_params, avg_pool_3d_uneven_strided_pa
 
 // avg_pool_3d case generation
 NGRAPH_INSTANTIATE_TEST_CASE_P(${BACKEND_NAME}, include_pad, avg_pool_3d_params, testing::Bool());
+
```
csullivan: @aprocter, I added a description of the test as suggested and some comments through out. Feel free to poke at it if it isn't quite clear.

test/backend_pool.in.cpp
```diff
@@ -1235,3 +1235,84 @@ NGRAPH_TEST_P(${BACKEND_NAME}, avg_pool_3d_params, avg_pool_3d_uneven_strided_pa
...
+//
+NGRAPH_TEST(${BACKEND_NAME}, maxpool_bprop_larger_than_cache)
+{
+    Shape window_shape{1, 2};
+    Strides move_strides{1, 2};
+    Shape padding_below{0, 0};
+    Shape padding_above{0, 0};
+
+    // 200 MB tensor to exceed cache
+    const size_t num_elements = 50 * 1024 * 1024;
```
csullivan: I increased the tensor size well past the identified failure condition so that if the failure is related to cache sizes, the test won't accidentally pass on future hardware (with larger caches).

csullivan: @shssf I had to disable this test on INTELGPU because the above chosen tensor size is too large. For the Titan XP used in CI, it needs to be at leas 3*1024 + 1 elements to spill into global DRAM. Ideally the number of elements chosen here is larger than any future GPU cache sizes. If you have a suggestion for something that meets this need but also can run on INTELGPU, I will update and re-enable the test.

PR #2062, opened 2018-11-14
QCBiasAdd and QCBiasSignedAdd for mkldnn

nbpatel:


aprocter:
@diyessi So far I've been assuming that it's not required to add `experimental` ops into the op_tbl.

nbpatel:
Yes, none of the experimental ops are added in the op table

src/ngraph/op/experimental/quantized_conv_bias.cpp
```diff
@@ -90,3 +90,146 @@ shared_ptr<Node> op::QuantizedConvolutionBias::copy_with_new_args(const NodeVect
                                                          new_args.at(3),
                                                          m_with_relu));
 }
+
+op::QuantizedConvolutionBiasAdd::QuantizedConvolutionBiasAdd(const shared_ptr<Node>& data_batch,
```
adstraw: It feels like overkill having a new op here.  @nbpatel tells me we support the same op on the CPU backend with the aid of pattern matching for f32.  The same should be true for int8.  nGraph has a reduced instruction set architecture.  We've made some accommodations in the experimental directory given our immediate goals but (so far) this is just for TensorFlow, just for CPU and just for the few models we are targeting in the near term.  How many models does this op set enable?  Two?  I'm concerned that if we keep expanding the experimental ops the solution will be untenable - not for  TensorFlow and the CPU backend, of course, but as we expand to other frameworks and backends.  I'm not arguing the builder for the same op... we need that.  I'm saying the builder should build a subgraph in this case, not return an op and the CPU backend should pattern match in the right MKL-DNN kernel.

adstraw: It feels like overkill having a new op here.  @nbpatel tells me we support the f32 version of this op in the CPU backend via pattern matching.  Ideally, the same would be true for the int8 version.

Assume you get to keep the builder, but not the op.  What would the builder do to enable pattern matching in the CPU backend?  Would we need new (more elemental) quantized ops to make this work?

I'm concerned about feature creep.  So far, we are working just on TensorFlow, just on the CPU backend, and just for the few models we are targeting for our near-term quantization goals.  Worried that this sort or "complex" instruction set will break as we add more frameworks and backends later on.  Need to align as close as possible to nGraph "reduced" instruction set architecture.

rkimballn1:
```
[ RUN      ] builder.scaled_QC_with_bias_add_and_relu
unknown file: Failure
C++ exception with description "Scale must be a Constant" thrown in the test body.
[  FAILED  ] builder.scaled_QC_with_bias_add_and_relu (10 ms)
[ RUN      ] builder.scaled_QC_with_bias_signed_add_and_relu
unknown file: Failure
C++ exception with description "Scale must be a Constant" thrown in the test body.
[  FAILED  ] builder.scaled_QC_with_bias_signed_add_and_relu (8 ms)
```

rkimballn1:
```
[ RUN      ] builder.scaled_QC_with_bias_add_and_relu
unknown file: Failure
C++ exception with description "Scale must be a Constant" thrown in the test body.
[  FAILED  ] builder.scaled_QC_with_bias_add_and_relu (10 ms)
[ RUN      ] builder.scaled_QC_with_bias_signed_add_and_relu
unknown file: Failure
C++ exception with description "Scale must be a Constant" thrown in the test body.
[  FAILED  ] builder.scaled_QC_with_bias_signed_add_and_relu (8 ms)
```

PR #2063, opened 2018-11-14
When processing in place slice, do not change the offset of the slice…

ayzhuang:
… node if the argument pointer comes from function input.


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1336,12 +1336,18 @@ void runtime::cpu::CPU_ExternalFunction::process_in_place_slice(
...
                 {
-                    auto arg = slice->get_argument(0);
-                    auto input = slice->get_input_from(arg);
+                    auto input = &slice->get_inputs().at(0);
+                    auto arg = input->get_output().get_node();
                     auto index = input->get_output().get_index();
                     auto input_node = std::dynamic_pointer_cast<ngraph::op::Op>(arg);
                     auto input_tensor = &input_node->get_output_tensor(index);
                     auto offset = input_tensor->get_pool_offset();
+                    if (offset == 0)
```
jbobba: can we run process_in_place_slice after propagate_in_place_input and propagate_in_place_output? If yes, then you can use tensor_roles. Offset == 0 seems to be a fragile check and might not cover for the case where offset is non-zero but the input is set to a parameter input through in_place_input propagation

PR #2064, opened 2018-11-14
Backend API change pre-work

rkimballn1:
This is pre-work for the Backend API update.
The Backend::compile() call is changed from
```
bool compile(shared_ptr<Function> func)
```
to
```
Handle compile(shared_ptr<Function> func)
```
where Handle will ultimately be an opaque handle but currently is just a `shared_ptr<Function>`

The big change in this PR is that you now must call `compile()` before you call `call()`

The unit test changes are just to call `compile` before calling `call`. It is a lot of changes in a lot of files, but it really is just the same change over and over.


src/ngraph/runtime/backend.hpp
```diff
@@ -79,8 +80,8 @@ class ngraph::runtime::Backend
 
     /// \brief Compiles a Function.
     /// \param func The function to compile
-    /// \returns true if compile is successful, false otherwise
-    virtual bool compile(std::shared_ptr<Function> func) = 0;
+    /// \returns compiled function or nullptr on failure
+    virtual Handle compile(std::shared_ptr<Function> func) = 0;
```
avijit-nervana: How is the compilation error conveyed to the caller? Exceptions I suppose. Please provide some information about it in the comment for this function.

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -82,22 +82,20 @@ bool runtime::cpu::CPU_Backend::compile(shared_ptr<Function> func)
...
-        rc = compile(func);
-    }
-
-    if (!rc)
-    {
-        throw ngraph_error("couldn't compile a function");
+        auto rc = compile(func);
+        if (!rc)
+        {
+            throw ngraph_error("couldn't compile a function");
```
avijit-nervana: Need better error messages than this. This is **extremely** important for end users.

src/ngraph/runtime/backend.hpp
avijit-nervana: There are two things I would like to suggest.
1. We keep the `call(...)` API as is and internally that function calls `compile()` (instead of the error thrown). 
2. We change the signature to `call()` so that it now takes a `Handle` as opposed to a function. That way - caller will call compile and then call. Can we rename `call()` to `execute()`? We are really executing the function - no?
I prefer the second option as it enforced the caller to call `compile()` first.

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -112,7 +110,7 @@ bool runtime::cpu::CPU_Backend::call(shared_ptr<Function> func,
     FunctionInstance& instance = m_function_map[func];
     if (instance.m_external_function == nullptr)
     {
-        rc = compile(func);
+        throw runtime_error("compile() must be called before call().");
```
avijit-nervana: is there any reason `call(...)` cannot internally call `compile(...)`? But a better option is to change the signature - as I proposed above.

src/ngraph/runtime/backend.hpp
rkimballn1: Your second option is what I did in the last giant PR that you wanted split into smaller PRs. This is part 1 of several PRs that will ultimately change call to be `call(Handle)`
The point of this PR is to enforce that the unit tests all call compile prior to calling call.
I agree that execute() is a much better name than call() but that should be a separate PR.

src/ngraph/runtime/cpu/cpu_backend.cpp
rkimballn1: We have gone around and around on this. If there is an error here there is pretty much nothing that a user can do except report the error to us developers. I suppose I could change it to something like "Internal compile error, please file a bug report"
When we enabled dumping out of the compile error that was causing problems with the unit tests long ago.

src/ngraph/runtime/cpu/cpu_backend.cpp
rkimballn1: Yes, because call will ultimately take a Handle, not a shared_ptr<Function> so there will be no way it can compile. If you load a saved compiled model you will not get a shared_ptr<Function> when you load it, you will get a handle.  This PR is just one of several.

src/ngraph/runtime/backend.hpp
avijit-nervana: Sounds great!

src/ngraph/runtime/cpu/cpu_backend.cpp
avijit-nervana: The way I am looking at is the following:
1. If the compilation error is related to nGraph language (i.e., mismatch of op/parameters etc.) they the user needs to know that so tat they can fix the framework level code to resolve.
2. If the error is internal then of course also need to know so that they can file a bug report. 

In this case the cpu backend `compile(...)` needs to propagate that error message.

src/ngraph/runtime/cpu/cpu_backend.cpp
avijit-nervana: Ok - that makes sense.

PR #2065, opened 2018-11-15
Add MaxPool as an explicit argument to MaxPoolBackprop

csullivan:
Core changes:
* Removes the optional weak reference to the forward op in the existing MaxPoolBackprop constructor * Adds an additional constructor which takes the forward MaxPool op as an explicit argument.
* Updates `op::MaxPool::generate_adjoints` to use the new 3-arg backward constructor (thus ngraph-mxnet should require no changes)
* Add serializer support

GPU changes: 
* Support backward pooling without the pooled tensor by recomputing the forward pass (2-arg c-tor)
* Use cudnnBackwardPooling when the pooled result is available (3-arg c-tor)

CPU changes:
* Add layout for MaxPool arg in layout for backprop
* Update workspace insertion pattern matcher to support the new c-tor


postrational:
This would be a good opportunity to make sure that `MaxPool` and `AvgPool` have the same interface (both in forward and backprop).

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -1521,22 +1521,24 @@ size_t runtime::gpu::CUDNNEmitter::build_convolution_backward_filter(
...
     // or if it already exists in the primitive list
     std::stringstream ss;
-    ss << "pool_op" << pool_op << "dtype_" << dtype << "_dir" << static_cast<int>(direction) << "_i"
-       << join(input_shape, "_") << "_o" << join(output_shape, "_") << "_ws"
-       << join(window_shape, "_") << "_wst" << join(window_strides, "_") << "_pb"
-       << join(padding_below, "_") << "_pb" << join(padding_above, "_");
+    ss << "pool_op" << pool_op << "dtype_" << dtype.c_type_string() << "_dir"
+       << static_cast<int>(direction) << "_i" << join(input_shape, "_") << "_o"
+       << join(output_shape, "_") << "_ws" << join(window_shape, "_") << "_wst"
+       << join(window_strides, "_") << "_pb" << join(padding_below, "_") << "_pb"
```
rkimballn1: The second "_pb" really should be "_pa"
Do you need to include bprop_needs_pooling in the hash?

csullivan: Yes, nice catch - fixed.

PR #2069, opened 2018-11-16
 check if the nodes are reachable before returning the users of a node

pruthviIntel:


src/ngraph/node.cpp
```diff
@@ -390,7 +391,10 @@ NodeVector Node::get_users() const
     {
         for (auto input : get_output_inputs(i))
         {
-            result.push_back(input->get_node());
+            if (is_used(input->get_node().get()))
```
jbobba: i think @Krovatkin was concerned about the perf impact here. You can possibly make this optional with an additional argument that is set to false by default.

pruthviIntel:
@Krovatkin can u approve this, if it looks okay? :)

Issue #2007, opened 2018-11-06
tensorflow : math_grad_test failure


csullivan:
[math_grad_test.zip](https://github.com/NervanaSystems/ngraph/files/2554761/math_grad_test.zip)

Issue #2006, opened 2018-11-06
tensorflow : image_ops_test failure


csullivan:
[image_ops_test.zip](https://github.com/NervanaSystems/ngraph/files/2554758/image_ops_test.zip)

PR #2070, opened 2018-11-16
Use cuda_reduce for int32 & int8 input_type instead of cudnn_reduce

AyanmoI:
This PR fixes the following issues:
https://github.com/NervanaSystems/ngraph/issues/2007
https://github.com/NervanaSystems/ngraph/issues/2006


test/backend_test.in.cpp
```diff
@@ -3358,6 +3443,33 @@ NGRAPH_TEST(${BACKEND_NAME}, max_matrix_rows_zero)
...
+    auto backend = runtime::Backend::create("${BACKEND_NAME}");
+
+    // Create some tensors for input/output
+    auto a = backend->create_tensor(element::i32, shape_a);
+    copy_data(a, vector<int32_t>{});
+    auto result = backend->create_tensor(element::i32, shape_rt);
+    copy_data(result, vector<int32_t>({3, 3, 3}));
+
+    backend->call_with_validate(f, {result}, {a});
+    EXPECT_EQ((vector<int32_t>{-std::numeric_limits<int32_t>::infinity(),
```
fengleitian: int does not have infinity.

AyanmoI: Thnks for fixing this

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -732,8 +732,48 @@ void runtime::gpu::GPU_Emitter::emit_Max(EMIT_ARGS)
...
+            return;
+        }
+        else if (args[0].get_size() == out[0].get_size())
+        {
+            kernel::emit_memcpyDtD(writer, out[0], args[0]);
+            return;
+        }
+        else
+        {
+            auto axes_set = max->get_reduction_axes();
```
csullivan: Rather than manually configuring from AxisSet -> AxisVector, let's just add a constructor to nvshape.hpp:
```
NVShape(const AxisVector& vec)
```
and then pass the reduction AxisSet directly to build_reduce.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -732,8 +732,48 @@ void runtime::gpu::GPU_Emitter::emit_Max(EMIT_ARGS)
...
+            ngraph::AxisVector axes_vec;
+            for (auto a : axes_set)
+            {
+                axes_vec.push_back(a);
+            }
+            vector<string> dtypes;
```
csullivan: nit: can we add a small helper that takes args/out and returns a vector<string> and another one that returns vector<element_type> for each tensor in args and out? 

doesn't have to be in this PR, just thinking about it since we do this a lot

AyanmoI: Sure I can do that

AyanmoI: @csullivan I will address this in the next PR.

src/ngraph/runtime/gpu/gpu_emitter.cpp
AyanmoI: Done

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -732,8 +730,28 @@ void runtime::gpu::GPU_Emitter::emit_Max(EMIT_ARGS)
...
+        vector<string> dtypes;
+        dtypes.push_back(args[0].get_type());
+        dtypes.push_back(out[0].get_type());
+        auto& cuda_emitter = external_function->get_primitive_emitter()->get_cuda_emitter();
+        index = cuda_emitter->build_reduce<ngraph::op::Max>(
+            dtypes, out[0].get_element_type().size(), args[0].get_shape(), axes_vec);
+    }
+    else
+    {
+        auto& cudnn_emitter = external_function->get_primitive_emitter()->get_cudnn_emitter();
```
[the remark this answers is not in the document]
AyanmoI: In addition to `int32`, `int8`, `double` and `float` are the data types cudnn is supposed to handle. `double` works with cudnn. Hence added `int8` to check.

src/ngraph/runtime/gpu/gpu_emitter.cpp
```diff
@@ -827,8 +845,28 @@ void runtime::gpu::GPU_Emitter::emit_Min(EMIT_ARGS)
...
+        vector<string> dtypes;
+        dtypes.push_back(args[0].get_type());
+        dtypes.push_back(out[0].get_type());
+        auto& cuda_emitter = external_function->get_primitive_emitter()->get_cuda_emitter();
+        index = cuda_emitter->build_reduce<ngraph::op::Min>(
+            dtypes, out[0].get_element_type().size(), args[0].get_shape(), axes_vec);
+    }
+    else
+    {
+        auto& cudnn_emitter = external_function->get_primitive_emitter()->get_cudnn_emitter();
```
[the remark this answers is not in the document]
AyanmoI: Same as above

PR #2073, opened 2018-11-16
Update SDL

rkimballn1:


cmake/sdl.cmake
```diff
@@ -0,0 +1,59 @@
...
+# Manage secure Development Lifecycle-related compiler flags
+#===============================================================================
+
+if(SDL_cmake_included)
+    return()
+endif()
+set(SDL_cmake_included true)
+
+if(UNIX)
+    set(CMAKE_CCXX_FLAGS "-fPIC -Wformat -Wformat-security")
```
aprocter: What's the difference between `CMAKE_CCXX_FLAGS` and `CMAKE_CXX_FLAGS`?

rkimballn1: I stole this from mkldnn. It is just an internal, poorly named variable. I will fix it and other "custom" variables. Took me a minute to see what they were doing too.

PR #2074, opened 2018-11-16
Build clDNN when building IntelGPU

rkimballn1:


cmake/external_cldnn.cmake
```diff
@@ -0,0 +1,59 @@
...
+    GIT_REPOSITORY ${CLDNN_GIT_REPO_URL}
+    GIT_TAG ${CLDNN_GIT_LABEL}
+    # Disable install step
+    INSTALL_COMMAND ""
+    UPDATE_COMMAND ""
+    CMAKE_ARGS 
+                # -DCLDNN__OUTPUT_DIR=out/Debug
+                -DCMAKE_BUILD_TYPE=Release
+                -DCLDNN__BOOST_VERSION=${BOOST_VERSION}
+                -DCLDNN__INCLUDE_TESTS=FALSE
```
shssf: It is good to add `-DCLDNN__INCLUDE_TUTORIAL=FALSE` to avoid building unnecessary tutorial binary

PR #2076, opened 2018-11-17
nvgpu cuda reduce with stable sum

fengleitian:
- add stable sum for `float` and `double` type
- add 0 size input support
- add `init_value` for different reduce op and data type for 0 size input
- add support for input size is same as output
- remove the need for initial value in kernel.
- add helper to simplify reduce shape
- add helper to get reduce strides
- add helper to get magic and shift for a vector
- add unit test for `double` type stable sum
- add skip for CPU and intelGPU for `double` type stable sum.


src/ngraph/runtime/gpu/gpu_host_parameters.hpp
```diff
@@ -86,6 +87,220 @@ namespace ngraph
...
+                    {
+                        return getVal<uint32_t>(val);
+                    }
+                    else if (type == "uint64_t")
+                    {
+                        return getVal<uint64_t>(val);
+                    }
+                }
+
+                void* val_by_datatype(std::string type, int64_t val)
```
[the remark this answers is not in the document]
rkimballn1: I agree with codacy on this

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -2805,3 +2952,167 @@ void runtime::gpu::CUDAEmitter::debug_sync()
...
+        magic.push_back(_magic);
+        shift.push_back(_shift);
+    }
+}
+
+void* runtime::gpu::CUDAEmitter::get_init_reduce_val(std::string reduce_op, std::string data_type)
+{
+    if (reduce_op == "max")
+    {
+        return m_host_parameters->min_by_datatype(data_type);
```
csullivan: TypeInfo would work well for min/max here. How about something like this, 
https://gist.github.com/c55d8276ff9af5d2d28f375855a8791a

This is just an example, you can modify it to do the check for `has_infinity` if needed.

On that note, in your current implementation you use numeric_limits<T>::min(), where you likely want ::lowest() instead (numeric_limits<float>::min = 1.17549e-38 / ::lowest() = -3.40282e+38). Your check on has_infinity probably makes it ok, but better to be safe.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -1856,19 +1946,53 @@ void runtime::gpu::CudaKernelBuilder::coordinate_transform_to_multi_d(codegen::C
...
         }
         writer << "int " << o_coordinates << i << " = division_by_invariant_multiplication("
-               << "coordinate_product, " << i_stride_magic << brace_open << i << brace_close << ", "
-               << i_stride_shift << brace_open << i << brace_close << ");\n";
+               << o_coordinates << "product, " << i_stride_magic << brace_open << i << brace_close
+               << ", " << i_stride_shift << brace_open << i << brace_close << ");\n";
     }
 }
+
+void runtime::gpu::CudaKernelBuilder::collective_coordinate_transform_helper(
```
csullivan: Can we just modify the existing coordinate transform to support a `std::string reduced_idx` param? It can be a default parameter if you prefer not to change other uses of the coordinate transform.

src/ngraph/runtime/gpu/gpu_host_parameters.hpp
```diff
@@ -86,6 +87,220 @@ namespace ngraph
                     return &m_uint64_t_params.back();
                 }
 
+                template <typename T>
+                void* getMin()
```
csullivan: See comments about using TypeInfo here.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -483,9 +505,25 @@ void runtime::gpu::CudaKernelBuilder::get_reduce_to_nd_op(
     runtime::gpu::GPUKernelArgs& args,
     const std::vector<std::string>& data_types,
     const std::string& reduce_op,
-    size_t out_rank,
+    size_t non_reduce_rank,
     size_t reduce_rank)
 {
+    bool stable_sum =
+        ((reduce_op == "add") && (data_types[1] == "float" || data_types[1] == "double"));
+    auto stable_sum_lambda = [&]() {
```
csullivan: Nice. Will this stable_sum snippet be used elsewhere (e.g. Softmax)? If so, maybe make it a helper function instead of an in context lambda.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1790,17 +1873,16 @@ size_t runtime::gpu::CUDAEmitter::build_reduce_to_scalar(const std::vector<std::
     uint32_t nthreads = static_cast<uint32_t>(shape_size(input_shape));
     uint32_t n = nthreads;
     uint32_t block_size_x = 1;
-    while (n > 1)
+    while ((block_size_x << 1) <= n)
```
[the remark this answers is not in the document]
fengleitian: good point.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1691,41 +1766,44 @@ size_t runtime::gpu::CUDAEmitter::build_reduce_to_nd(const std::vector<std::stri
...
+                       simplified_reduce_axis,
+                       non_reduce_shape,
+                       non_reduce_strides,
+                       non_reduce_strides_in_input,
+                       reduce_shape,
+                       reduce_strides,
+                       reduce_strides_in_input);
+
+    // std::vector<int> reduce_strides_magic;
+    // std::vector<int> reduce_strides_shift;
```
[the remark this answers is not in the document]
fengleitian: Was going to implement another kernel that need this. but this PR is already big. Remove and will add back in another PR.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1908,27 +1996,41 @@ size_t runtime::gpu::CUDAEmitter::build_reduce_to_scalar_acc(const std::vector<s
     return this->m_primitive_emitter->register_primitive(reduce_acc, hash);
 }
 
-size_t runtime::gpu::CUDAEmitter::build_reduce(const std::vector<std::string>& dtypes,
-                                               const size_t data_bytes,
-                                               const NVShape& input_shape,
-                                               const NVShape& reduce_axis,
+size_t runtime::gpu::CUDAEmitter::build_reduce(const std::vector<element::Type>& dtypes,
+                                               NVShape input_shape,
```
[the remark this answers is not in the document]
fengleitian: right.

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: the infinity will cover it. but you are right, lowest should be the better one here.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1767,14 +1845,19 @@ size_t runtime::gpu::CUDAEmitter::build_reduce_to_nd(const std::vector<std::stri
...
 }
 
-size_t runtime::gpu::CUDAEmitter::build_reduce_to_scalar(const std::vector<std::string>& dtypes,
-                                                         const size_t data_bytes,
+size_t runtime::gpu::CUDAEmitter::build_reduce_to_scalar(const std::vector<element::Type>& dtypes,
                                                          NVShape input_shape,
                                                          const char* op,
                                                          const char* kernel)
 {
+    std::vector<std::string> dtypes_str;
```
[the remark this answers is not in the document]
fengleitian: good point

csullivan:
@fengleitian, see the CollapseDims pass in CPU https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/pass/cpu_collapse_dims.cpp. We could add this pass and simplify the dimension reduction in the kernel

src/ngraph/runtime/gpu/cuda_emitter.cpp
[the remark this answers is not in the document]
fengleitian: done.

src/ngraph/runtime/gpu/cuda_emitter.cpp
fengleitian: change min to lowest, but still keep infinity, actually that what unit test requested.

fengleitian:
@csullivan CollapseDims can be done in another PR.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1691,41 +1766,44 @@ size_t runtime::gpu::CUDAEmitter::build_reduce_to_nd(const std::vector<std::stri
...
-        else
-        {
-            non_reduce_strides.push_back(input_strides[i]);
-            output_shape.push_back(input_shape[i]);
-        }
-    }
-    NVShape output_strides = row_major_strides(output_shape);
-    uint32_t nthreads = static_cast<uint32_t>(shape_size(output_shape));
+    NVShape reduce_strides_in_input;
+    get_reduce_strides(simplified_input_shape,
```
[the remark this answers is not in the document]
fengleitian: agree.
But in another PR.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
fengleitian: combined.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
csullivan: nit: Maybe the codegen itself could be in the helper?

src/ngraph/runtime/gpu/type_info.hpp
```diff
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <limits>
+#include <list>
```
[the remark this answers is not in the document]
fengleitian: nice catch, was use a list, but changed.

src/ngraph/runtime/gpu/gpu_host_parameters.hpp
```diff
@@ -18,6 +18,7 @@
 
 #include <cinttypes>
 #include <list>
+#include "ngraph/except.hpp"
```
diyessi: line between these groups of includes.

test/backend_sum.in.cpp
```diff
@@ -535,4 +535,35 @@ NGRAPH_TEST(${BACKEND_NAME}, sum_stable_acc)
...
+    if (backend_name == "INTERPRETER")
+    {
+        return;
+    }
+    Shape shape_a{10, 10, 20, 300};
+    auto A = make_shared<op::Parameter>(element::f64, shape_a);
+
+    Shape shape_rt{10};
+    auto f = make_shared<Function>(make_shared<op::Sum>(A, AxisSet{1, 2, 3}), ParameterVector{A});
+
```
diyessi: Just use a sequence known to fail if added sequentially and known to pass if added Kahnaentially.

fengleitian: Add two simple test for float and double.

PR #2079, opened 2018-11-19
Klocwork issues addressed

rkimballn1:


src/ngraph/runtime/cpu/cpu_executor.cpp
```diff
@@ -24,32 +24,32 @@ static int GetNumCores()
...
-        return count;
+        count = std::atoi(ngraph_intra_op_parallelism);
     }
     else
     {
-        count = std::thread::hardware_concurrency() >> 1;
+        count = std::thread::hardware_concurrency() / 2;
     }
-    return count ? count : 1;
+    return count < 1 ? 1 : count;
```
Krovatkin: i wonder if it makes sense to set `int count = 1`. Then we don't really have to check if it was set or not? We already have a `catch-all` else; so `count` should not be equal to `0`.

src/ngraph/log.cpp
```diff
@@ -69,9 +69,12 @@ LogHelper::LogHelper(LOG_TYPE type,
 
     time_t tt = chrono::system_clock::to_time_t(chrono::system_clock::now());
     auto tm = gmtime(&tt);
-    char buffer[256];
-    strftime(buffer, sizeof(buffer), "%Y-%m-%dT%H:%M:%Sz", tm);
-    m_stream << buffer << " ";
+    if (tm)
```
Krovatkin: `if (auto tm = gmtime(&tt))` ?

src/ngraph/runtime/cpu/cpu_executor.cpp
```diff
@@ -24,32 +24,32 @@ static int GetNumCores()
...
 
-    if (ngraph_inter_op_parallelism && (count = std::atoi(ngraph_inter_op_parallelism)))
+    if (ngraph_inter_op_parallelism)
     {
-        return count;
+        count = std::atoi(ngraph_inter_op_parallelism);
     }
 
-    return 1;
+    return count < 1 ? 1 : count;
```
Krovatkin: ditto

PR #2080, opened 2018-11-19
[PRIVATE BRANCH] Style fixes

jbobba:


src/ngraph/runtime/cpu/pass/cpu_rnn_fusion.cpp
```diff
@@ -113,213 +121,195 @@ void ngraph::runtime::cpu::pass::LSTMFusion::construct_lstm_fprop()
...
+    //   i_t = sigmoid(W_{ii} x_t + b_{ii} + W_{hi} h_{(t-1)} + b_{hi});
+    //   g_t = tanh   (W_{ig} x_t + b_{ig} + W_{hg} h_{(t-1)} + b_{hg});
+    //   o_t = sigmoid(W_{io} x_t + b_{io} + W_{ho} h_{(t-1)} + b_{ho});
+    //   c_t = f_t * c_{(t-1)} + i_t * g_t;
+    //   h_t = o_t * \ tanh(c_t);
+    //
+
+    // Inputs to the sub-graph
+    // Assumes weights for all the 4 gates are fused in the order -
+    //                      forget (f), input (i), block (g) and output (0)
```
pruthviIntel: here the input ordering the of the gates should be` input (i),  forget (f), , block (g) and output (0)`.

src/ngraph/runtime/cpu/pass/cpu_rnn_fusion.cpp
```diff
@@ -113,213 +121,195 @@ void ngraph::runtime::cpu::pass::LSTMFusion::construct_lstm_fprop()
...
+    auto w_i2h = std::make_shared<pattern::op::Label>(element::f32, Shape{100, 400});
```
pruthviIntel: we can also a comment saying we assume the weights to be in the order `ldigo` (after reshape) which MKLDNN expects

PR #2081, opened 2018-11-20
Sigmoid layout bugfix

louisfeng:
Allow sigmoid mkldnn op to query input and output layouts. Added unit test to verify the before (failure) and after (correct) results. Shape channel size is a non multiple of 8 or 16 to trigger mkldnn layout convert to pad with zeros.


louisfeng:
TESTNOW

louisfeng:
I'm not sure if these new errors are false positives or not.

PR #2086, opened 2018-11-20
IntelGPU backend: Use clDNN BatchNorm in operations

shssf:


shssf:
@rkimballn1 It looks like the build procedure still uses `DCLDNN_ROOT_DIR=/home/clDNN` for the backend. This PR requires latest `clDNN` as it already implemented in automatic `clDNN` build.
I think it is better to remove usage of `CLDNN_ROOT_DIR` from the test procedure to allow automatic library build

rkimballn1:
We can't remove the use of CLDNN_ROOT_DIR until after we merge this PR and we can't merge this PR until after we remove the use of CLDNN_ROOT_DIR. A workaround is to
1) remove the support for CLDNN_ROOT_DIR from this PR and merge that to master
2) remove the use of CLDNN_ROOT_DIR in CI
3) add support for CLDNN_ROOT_DIR in new PR
An alternative is to rename the CLDNN_ROOT_DIR to something like NGRAPH_USE_PREBUILT_CLDNN for this pr, then everything works. You will need to change your use to be NGRAPH_USE_PREBUILT_CLDNN. For LLVM we use an option NGRAPH_USE_PREBUILT_LLVM

shssf:
clDNN version in CI should be updated for this PR

PR #2090, opened 2018-11-20
re-enable quantize_clamp_int32 test on CPU

adstraw:


diyessi:
Blocking while checking for constant folding interactions.

PR #2093, opened 2018-11-21
Re-enable disabled INTERPRETER tests

aprocter:
Re-enables some unit tests for `INTERPRETER`.

I can't find any reason that these were disabled, and they seem to work fine:

```
argmin_4D_axis_3_i64
computation_reuse
```

I presume these tests were disabled for `INTERPRETER` because they compare against `INTERPRETER` as their baseline. They are still valid tests, however:

```
topk_3d_large_input_max
topk_3d_large_input_min
```

This one was getting tripped up because `INTERPRETER` was unaware of its unusual choice of output element types (fixed that):

```
topk_int64
```


AyanmoI:
May be we can add something like     
   std::string backend_name = "${BACKEND_NAME}";
   if (backend_name == "INTERPRETER")
    {
        return;
    }
to 
```topk_3d_large_input_max
topk_3d_large_input_min```

**Comment by AyanmoI** [comment]
May be we can add something like
std::string backend_name = "${BACKEND_NAME}";
```
if (backend_name == "INTERPRETER")
{
return;
}
```
to `topk_3d_large_input_max` and `topk_3d_large_input_min` tests

aprocter:
@AyanmoI We could do that, but the observable effect should be the same and it's not like we're wasting any real time in the test.

aprocter:
Looking a bit more closely at `computation_reuse`, I am guessing it was disabled because it is meant to test the `set_stale` flag which is only "really" implemented on `CPU`. However, I think it is still valid and useful to test that `set_stale` on `INTERPRETER` has no ill effect, i.e., it doesn't change the output.

aprocter:
Regarding `argmin_4D_axis_3_i64`, the test was originally disabled in #1862, and it appears that the likely underlying interpreter bug was fixed in #1922.

PR #2094, opened 2018-11-21
remove version tagging from shared libraries

rkimballn1:


ransford2011:
@rkimballn1. Can we please guard this by a variable as this is needed for my release efforts. Thanks!

avijit-nervana:
@ransford2011 Yes - I agree. This could be guarded by a cmake flag. The requested behavior would be: When this cmake flag is specified, the libraries will have the version numbers.

rkimballn1:
TESTNOW

diyessi:
Blocked while constant folding issue is investigated.

PR #2096, opened 2018-11-21
Adding leaky relu

jbobba:
LeakyRelu - Max(input, input*alpha), alpha >0


src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -723,6 +724,25 @@ namespace ngraph
...
+                    auto arg0_rank = arg0_shape.size();
+                    auto result_shape = node->get_output_shape(0);
+
+                    if ((arg0_rank == 4 || arg0_rank == 2) &&
+                        node->get_input_element_type(0) == element::f32)
+                    {
+                        auto op_annotations =
+                            std::make_shared<ngraph::runtime::cpu::CPUOpAnnotations>();
+                        op_annotations->set_mkldnn_op(true);
+                        leaky_relu->set_op_annotations(op_annotations);
```
gaurides: Can this also be `in_place` op like `Relu`?

PR #2101, opened 2018-11-21
nvgpu cuda softmax optimization

fengleitian:
performance 6.8x:
23,851 us -> 3,503 us

- when reduce size is 1, set all output to 1.
- choose between two different kernel for better cache

before:
```
============================================================================
---- Processing 'tfl/1024/mxnet-ngraph-Function_1-MakeForwardFunction-pre-optimized-fprop.json'
============================================================================

---- Benchmark ----
compile time: 5,059ms
93.1ms per iteration

Softmax              23,851us
Rnn                  11,150us
Dot                   4,088us
Add                   3,592us
Broadcast             2,678us
Reshape                 547us
GetOutputElement        234us
Concat                  103us
Result                    1us

---- Aggregate times per op type/shape/count ----
Softmax {61440, 2000} 1             23,851us
Rnn {61440, 200} 1                  11,150us
Dot {61440, 2000} 1                  4,088us
Add {61440, 2000} 1                  3,592us
Broadcast {61440, 2000} 1            2,598us
Reshape {61440, 200} 2                 259us
Reshape {1024, 60, 200} 1              258us
GetOutputElement {61440, 200} 1        234us
Broadcast {1024, 200} 4                 80us
Concat {2048, 200} 2                    52us
Concat {643200} 1                       51us
Reshape {200, 2000} 1                   20us
Reshape {160000} 4                       4us
Reshape {800} 4                          4us
Reshape {60, 1024, 200} 1                2us
Result {61440, 2000} 1                   1us
```

after:
```
============================================================================
---- Processing 'tfl/1024/mxnet-ngraph-Function_1-MakeForwardFunction-pre-optimized-fprop.json'
============================================================================

---- Benchmark ----
compile time: 5,456ms
70.96ms per iteration

Rnn                  11,302us
Dot                   4,187us
Add                   3,584us
Softmax               3,503us
Broadcast             2,737us
Reshape                 548us
GetOutputElement        234us
Concat                  107us
Result                    1us

---- Aggregate times per op type/shape/count ----
Rnn {61440, 200} 1                  11,302us
Dot {61440, 2000} 1                  4,187us
Add {61440, 2000} 1                  3,584us
Softmax {61440, 2000} 1              3,503us
Broadcast {61440, 2000} 1            2,657us
Reshape {61440, 200} 2                 261us
Reshape {1024, 60, 200} 1              258us
GetOutputElement {61440, 200} 1        234us
Broadcast {1024, 200} 4                 80us
Concat {2048, 200} 2                    56us
Concat {643200} 1                       51us
Reshape {200, 2000} 1                   20us
Reshape {160000} 4                       4us
Reshape {800} 4                          4us
Result {61440, 2000} 1                   1us
Reshape {60, 1024, 200} 1                1us
```


src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -498,6 +501,286 @@ void runtime::gpu::CudaKernelBuilder::get_softmax_op(codegen::CodeWriter& writer
...
+    };
+
+    auto max_lambda = [&]() { writer << "r_max = r_max > input_i ? r_max : input_i;\n"; };
+
+    auto divide_lambda = [&]() {
+        writer << "input_i = expf(input_i - r_max) / r_sum;\n";
+        writer << "out[input_idx] = input_i;\n";
+    };
+
+    // auto unroll_loop_lambda = [&](void (*func)()) -> void {
```
rkimballn1: Can we remove this commented out code or is it to be used later?

fengleitian: yes, removed for now. Might add later.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -498,6 +501,262 @@ void runtime::gpu::CudaKernelBuilder::get_softmax_op(codegen::CodeWriter& writer
...
+        {
+            if (block_size_x > i)
+            {
+                writer << "input_i = __shfl_down_sync(0xffffffff, r_max, " << i << ", 32);\n";
+                max_lambda();
+            }
+        }
+        if (block_size_x > 32)
+        {
+            writer << "uint32_t lane_idx = threadIdx.x & 0x1f; \n";
```
csullivan: nit: How about 31, or even easier to read `#define MAX_LANE_IDX 31` and 
```
            writer << "uint32_t lane_idx = threadIdx.x & MAX_LANE_IDX; \n";
```

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -498,6 +501,262 @@ void runtime::gpu::CudaKernelBuilder::get_softmax_op(codegen::CodeWriter& writer
...
+void runtime::gpu::CudaKernelBuilder::get_softmax_block_reduce_op(
+    codegen::CodeWriter& writer,
+    const std::string& name,
+    runtime::gpu::GPUKernelArgs& args,
+    const std::vector<std::string>& data_types,
+    size_t non_reduce_rank,
+    size_t reduce_rank,
+    size_t block_size_x)
+{
+    writer << runtime::gpu::nvrtc::helpers();
```
csullivan: nit: good opportunity for read-only loads since the same input addresses are loaded from several times
```suggestion
    writer << runtime::gpu::nvrtc::helpers();
    writer << runtime::gpu::nvrtc::define_non_coherent_load(data_types[1], "load");
```

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -498,6 +501,262 @@ void runtime::gpu::CudaKernelBuilder::get_softmax_op(codegen::CodeWriter& writer
...
+                                               "reduce_strides",
+                                               "reduce_strides_magic",
+                                               "reduce_strides_shift",
+                                               "reduce_strides_in_input",
+                                               "reduce_coordinate",
+                                               reduce_rank,
+                                               true,
+                                               "reduce_input_index");
+        writer << "input_idx = reduce_input_index + non_reduce_input_index;\n";
+        writer << "input_i = in[input_idx];\n";
```
csullivan:
```suggestion
        writer << "input_i = load(in, input_idx);\n";
```

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
```diff
@@ -326,9 +326,9 @@ void runtime::gpu::CudaKernelBuilder::get_softmax_op(codegen::CodeWriter& writer
             // loop through all reduction axis
```
csullivan: nit: refactor `:: get_softmax_op` to use the same lambda callbacks as `:: get_softmax_block_reduce_op`

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
fengleitian: good point.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
fengleitian: seems it's a little bit slower in this case.  `3,7xx us vs 3,5xx us`.

fengleitian: hmm, Ok, with more test, it's about the same on average. I guess it's because here data is been kicked out of cache anyway before the next read. So it does not help. But not hurt too. I might here is data size is smaller for each warp.

fengleitian: keep the load here.

csullivan: Makes sense. I had tested this previously in the nd case where reduce_stride !=1 (the `::get_softmax_op` impl. in the previous softmax PR) and it gave a ~20% perf. boost. That was using a smaller tensor from an rn50 json.

src/ngraph/runtime/gpu/gpu_cuda_kernel_builder.cpp
csullivan: Can we also add
#define MAX_LANE_IDX 31 here: https://github.com/NervanaSystems/ngraph/blob/f33317cc5538ddd7ff9e3174429846e80828b01c/src/ngraph/runtime/gpu/nvrtc/helpers.hpp#L58
And the replace the 0x1f above with it?

fengleitian:
@csullivan It's possible a reshape will worth it. Also possible won't worth it.  We could still add a fusion and collect more data to decide which way to go.

for example:
{61440, 2000} with axis{1}  37xx us;
{2000, 61440} with axis{0} 59xx us;
reshape 2700 us,  2700 + 37xx > 59xx us, which means a reshape does not worth it.

I'm sure there are other case reshape will worth it.

csullivan:
Right, it will require tuning depending on the graph.

src/ngraph/runtime/gpu/cuda_emitter.cpp
```diff
@@ -1651,79 +1657,158 @@ size_t runtime::gpu::CUDAEmitter::build_softmax(const std::vector<std::string>&
...
     }
+    else
+    {
+        uint32_t reduce_count = static_cast<uint32_t>(shape_size(reduce_shape));
+        uint32_t block_size_x = 1;
+        while ((block_size_x << 1) <= reduce_count)
+        {
+            block_size_x <<= 1;
+        }
+        block_size_x = fmin(512, block_size_x);
```
gcwenger: Infinite loop if reduce_count > 0x80000000.
This might be better done something like this:
```
uint32_t block_size_x = 1;
if (reduce_count < 512)
{
    while ((block_size_x << 1) <= reduce_count)
    {
        block_size_x <<= 1;
    }
}
else
{
    block_size_x = 512;
}
```

fengleitian: nice catch, fixed

PR #2103, opened 2018-11-22
Disable CF on CPU

Krovatkin:


penyung:
hi @Krovatkin 
I wonder why CF is disabled in CPU backend.
I enable it and it works like a charm.
Is there any concern here?

PR #2105, opened 2018-11-24
IntelGPU backend: Use clDNN matrix operations for nGraph::Dot

shssf:
- tests are aligned to current passrate
- used cldnn::gemm for nGraph::Dot in supported cases. Other cases are covered by custom kernel
- made uniwersal custom kernel (for Dot) instead of a set of custom kernels


Kase33:
Man I  am  so over my head.

PR #2106, opened 2018-11-24
copy tensor from one backend to another for hybrid and more

sasadep:
* copy tensor form one backend to another for hybrid
* add cpu and interpreter as fall back for GPU transformer


Kase33:
Wow

src/ngraph/runtime/tensor.cpp
```diff
@@ -70,3 +70,10 @@ void runtime::Tensor::set_stale(bool val)
 {
     m_stale = val;
 }
+
+void runtime::Tensor::copy_to(shared_ptr<runtime::Tensor> d, size_t offset, size_t n)
+{
+    void* copy = malloc(n);
+    read(copy, offset, n);
+    d->write(copy, offset, n);
+}
```
jbobba: free(copy)? The read will be an extra memcpy sometimes. We will have to optimize in the future.

sasadep: Absolutely ! 
I think there might be a few conditions we can optimize for..

src/ngraph/runtime/gpu/gpu_backend.cpp
```diff
@@ -20,11 +20,13 @@
 #include <cudnn.h>
 
 #include "ngraph/graph_util.hpp"
+#include "ngraph/runtime/cpu/cpu_backend.hpp"
```
rkimballn1: Don't modify the GPU backend. If you must create a wrapper which wraps the GPU/CPU/INTERPRETER in a new backend.
This change makes it a requirement that the CPU is built when the GPU is built which is very annoying for GPU developers because the CPU is very slow to build. Also, if you are going to do something like this you will need to update the cmake files to guarantee that everything that is required is built.

src/ngraph/runtime/tensor.cpp
```diff
@@ -70,3 +70,10 @@ void runtime::Tensor::set_stale(bool val)
 {
     m_stale = val;
 }
+
+void runtime::Tensor::copy_to(shared_ptr<runtime::Tensor> d, size_t offset, size_t n)
```
rkimballn1: The signature should be `copy_to(const runtime::Tensor&)` and no more. With this signature you can validate the call internally if you want and no need to pass offset and size as we know those from the tensors.
Don't pass shared_ptr around unless you really need to.

src/ngraph/runtime/tensor.cpp
```diff
@@ -70,3 +70,10 @@ void runtime::Tensor::set_stale(bool val)
...
+{
+    void* copy = malloc(n);
```
rkimballn1: You should use `unique_ptr<int64_t[]>` so that you get automatic memory freeing even if an exception occurrs somewhere and you get better memory alignment with int16_t which must be on 8 byte boundary

test/tensor.cpp
```diff
@@ -101,12 +101,67 @@ void test_read_write(const vector<T>& x)
...
+    auto s = a->get_size_in_bytes();
+    a->copy_to(b, 0, s);
+
+    vector<T> af_vector(x.size());
+    b->read(af_vector.data(), 0, af_vector.size() * sizeof(T));
+    ASSERT_EQ(af_vector, result);
+}
+
+template <typename T>
+void test_copy_to_other_backend(const vector<T>& x)
```
rkimballn1: Since there is nothing in copy_to related to backend then there is no need to test with two different backends. You only need to test that you can copy from one tensor to another.
I see no test for copy_to where the tensor sizes do not match. You can't just overwrite memory so you need to check for this condition in copy_to and you need a unit test to verify that you throw an exception when you try to copy to the wrong size tensor. You may want to compare size and shape.

test/tensor.cpp
```diff
@@ -101,12 +101,67 @@ void test_read_write(const vector<T>& x)
     ASSERT_EQ(result1, result2);
 }
 
+template <typename T>
+void test_copy_to_same_backend(const vector<T>& x)
```
rkimballn1: You should pass in shape as well so you can verify that test fails if shapes don't match.

WilliamTambellini:
Could you please fix the title: "copy tensor form ..." > "copy tensor from ..."

sasadep:
#2196 addressing the changes and more

PR #2113, opened 2018-11-26
[PRIVATE BRANCH] Jbobba/rnn fusion review

jbobba:


src/ngraph/runtime/cpu/pass/cpu_rnn_fusion.cpp
```diff
@@ -787,16 +720,16 @@ void ngraph::runtime::cpu::pass::MultiLayerRNNFusion::construct_multi_layer_rnn_
...
         {
             auto goe_nodes = op::get_output_elements(rnn_nodes[index]);
-            // if their is no GOE followed by the Lstm, their might be pattern match error
+            // if there is no GOE followed by the Lstm, their might be pattern match error
             // we will return safely
-            if (goe_nodes.size() == 0)
+            if (goe_nodes.size() != 2)
             {
-                return false;
+                throw ngraph_error("Expecting two outputs for each RNN node");
```
pruthviIntel: I think we need to remove the constraint here, for strictly checking it should equal to 2. A multilayer fused RNN need to have two GOE . for ex.

image

unlike LSTM cells , where each intermediate cell should have recurrent output state tensor . An timestep fused RNN cell will have `layer_rnn_ht` going in to next layer and only if their is any consumer of the last LSTM cell, it will have `layer_rnn_ht_ct`. I think it make sense to check if it has @least one `GOE`

PR #2119, opened 2018-11-26
Windows build

rkimballn1:
With this you can build nbench with the INTERPRETER backend. You can build unit test too but I did not run those tests.

Here are the commands to build:
```
cmake ../ngraph -DNGRAPH_CPU_ENABLE=FALSE  -G "Visual Studio 15 2017 Win64"
cmake --build . --target nbench --config Release
```

The `-G` option is specified so we can build 64 bit. 32 bit is the default :(


shssf:
@rkimballn1 Is it possible to create simple build scripts in the project root?
For example as it is in https://github.com/intel/clDNN (`create_unixmake_gcc.sh` for Linux build and `create_msvc_mscc.bat` for Windows)

src/ngraph/runtime/reference/topk.hpp
```diff
@@ -28,6 +28,17 @@ namespace ngraph
     {
         namespace reference
         {
+            template <typename T, typename U>
```
diyessi: Why are these changed?

test/util/test_tools.cpp
```diff
@@ -31,57 +31,90 @@ vector<float> read_float_vector(shared_ptr<runtime::Tensor> tv)
     if (element_type == element::boolean)
     {
         vector<char> vec = read_vector<char>(tv);
-        float_vec = vector<float>(vec.begin(), vec.end());
+        for (char value : vec)
```
diyessi: Why are these changes here?

jbobba:
Are we going to enable Windows CI builds to keep this from breaking?

rkimballn1:
Closing but not deleting branch. This PR is too large and must be split into multiple smaller PRs. The first of these is #2177

PR #2120, opened 2018-11-26
Implement ngraph::element::Type as a wrapper around an enum

rkimballn1:
I have been wanting a way to use switch() with element::Type for a long, long time. This allows me to do just that *and* is does not change any of the current element::Type API.

As an example I used the new enum Type in the Interpreter backend.


rkimballn1:
@diyessi and @aprocter I think your suggestions are great but really are a separate PR. This PR is a no-impact change to element::Type. By `no-impact` I mean that zero code requires change in order to use this class. Everything is exactly as it was before, only the internals of the Type class have changed. I only added a `new` API to get an enum of the Type. Zero impact.

PR #2122, opened 2018-11-27
Support arg_reduce for int32 for nvGPU

AyanmoI:


src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -158,13 +158,34 @@ cudnnDataType_t runtime::gpu::CUDNNEmitter::get_cudnn_datatype(const element::Ty
...
+    runtime::gpu::CUDNNEmitter::get_cudnn_workspace_size(const cudnnHandle_t& handle,
+                                                         const cudnnReduceTensorDescriptor_t& desc,
+                                                         const cudnnTensorDescriptor_t& input_desc,
+                                                         const cudnnTensorDescriptor_t& output_desc,
+                                                         size_t input_buffer_size)
+{
+    size_t workspace_size = 0;
+    CUDNN_SAFE_CALL(
+        cudnnGetReductionWorkspaceSize(handle, desc, input_desc, output_desc, &workspace_size));
+    if (workspace_size < input_buffer_size)
```
fengleitian: IS this still necessary?

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -158,13 +158,34 @@ cudnnDataType_t runtime::gpu::CUDNNEmitter::get_cudnn_datatype(const element::Ty
     return get_cudnn_datatype(dtype.c_type_string());
 }
 
+size_t
+    runtime::gpu::CUDNNEmitter::get_cudnn_workspace_size(const cudnnHandle_t& handle,
```
fengleitian: if this is only for reduce, should include the info in function name.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -180,7 +201,8 @@ size_t runtime::gpu::CUDNNEmitter::build_reduce_forward(const cudnnReduceTensorO
     }
 
     auto& desc = m_descriptors.build<cudnnReduceTensorDescriptor_t>();
-    cudnnDataType_t data_type = get_cudnn_datatype(input_type);
+    auto modified_input_type = (input_type == element::i32) ? element::f64 : input_type;
```
fengleitian: why convert i32 to f64?

src/ngraph/runtime/gpu/cudnn_emitter.cpp
AyanmoI: Not sure abt that... @csullivan Can you please point me to the tests that captures this case?

src/ngraph/runtime/gpu/cudnn_emitter.cpp
AyanmoI: Since f32 would not be able to capture the complete range of i32 (only upto 2^24), we need to use f64. Hence, when the input is i32, we convert it to f64, and create descriptors accordingly

src/ngraph/runtime/gpu/cudnn_emitter.cpp
csullivan: I would expect this isn't necessary any more. This is the test that revealed the bug `softmax_axis_3d_double`. That said, softmax is now instrumented with cuda, not cudnn, so this test doesn't cover the issue any longer.

AyanmoI: In that case I can remove this conditional. Also the wrapper function seems to redundant then.

src/ngraph/runtime/gpu/cudnn_emitter.cpp
```diff
@@ -304,8 +330,69 @@ size_t runtime::gpu::CUDNNEmitter::build_reduce_forward(const cudnnReduceTensorO
...
+                                                              input_f64_ptr,
+                                                              beta,
+                                                              output_desc,
+                                                              reduce_buffer));
+                            debug_sync();
+                        }});
+                }
+                else
+                {
+                    reduce.reset(new gpu::primitive{
```
csullivan: nit: maybe use callbacks to simplify the logic: https://github.com/NervanaSystems/ngraph/compare/amoitra/arg_reduce_int32...cs/amoitra/arg_reduce_int32

src/ngraph/runtime/gpu/cudnn_emitter.hpp
```diff
@@ -50,6 +50,7 @@ namespace ngraph
                 std::vector<int> get_vector_int_from_size_t(const std::vector<size_t>&);
             }
             class GPUPrimitiveEmitter;
+            class GPUAllocator;
```
csullivan: nit: Is this needed?

AyanmoI: Thnks :) Good catch! This was from a previous version of my code where I had wrapper function for getting the workspace.

PR #2126, opened 2018-11-27
generic layout methods for elementwise kernels.

jbobba:
Can wait till post-beta. I'm not sure if it improves perf on any of the models that we currently run but it is good to handle more cases.


src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -268,6 +266,59 @@ void runtime::cpu::pass::CPULayout::set_native_layouts(
...
+                    !mkldnn_utils::is_mkldnn_padded_layout(
+                        arg_mds[1], ngraph::get_default_order(node->get_input_shape(1)));
+    if (mkldnn_utils::use_mkldnn_kernel(node.get()) || md_check)
+    {
+        vector<memory::desc> i_mds;
+        vector<memory::desc> o_mds;
+        int select = 0;
+        if (std::getenv("NGRAPH_PASS_CPU_LAYOUT_ELTWISE") != nullptr)
+        {
+            const int user_select = std::atoi(std::getenv("NGRAPH_PASS_CPU_LAYOUT_ELTWISE"));
```
Krovatkin: should we make one of those static so we don't have to recompute it in every compilation?

src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -268,6 +266,59 @@ void runtime::cpu::pass::CPULayout::set_native_layouts(
...
+        o_mds.push_back(input_md);
+        set_output_layouts(node, o_mds);
+    }
+    else
+    {
+        set_native_layouts(external_function, node);
+    }
+}
+
+void set_layouts_binaryeltwise(ngraph::runtime::cpu::CPU_ExternalFunction* external_function,
```
pruthviIntel: why this is not made static?

PR #2144, opened 2018-11-29
Graph comparison - isolated per op testing

gcwenger:
Isolated per op testing when comparing graphs for better determination of source of accuracy divergence.


test/backend_graph_comparison.in.cpp
```diff
@@ -45,98 +45,178 @@ class serialized_graph_files : public ::testing::TestWithParam<string>
...
+                cout << "  inputs:" << endl;
+                for (auto& p : result_node->get_arguments())
                 {
-                    cout << " " << p->get_name() << endl;
-                    cout << "   nargs: " << p->get_arguments().size() << endl;
+                    cout << "    " << p->get_name() << " " << p->get_element_type() << endl;
                 }
             }
 
+            // Sanity check testing code
```
csullivan: Are these still desired? if so, change or remove the comment

test/backend_graph_comparison.in.cpp
```diff
@@ -185,7 +266,42 @@ NGRAPH_TEST_P(${BACKEND_NAME}, serialized_graph_files, compare_backends_with_gra
...
+                    isolated_node_to_original_node[isolated_param.get()] = arg;
+                }
+                else
+                {
+                    // It's not an output in the reference results - use the original arg
+                    isolated_op_args.push_back(arg);
+                }
+            }
+            auto isolated_op = n->copy_with_new_args(isolated_op_args);
+            isolated_results.push_back(isolated_op);
```
csullivan: Isolated op extraction looks good 👍

test/backend_graph_comparison.in.cpp
```diff
@@ -45,98 +45,178 @@ class serialized_graph_files : public ::testing::TestWithParam<string>
 public:
     void compare_results(NodeVector& result_nodes,
                          vector<shared_ptr<runtime::Tensor>> ref_results,
-                         vector<shared_ptr<runtime::Tensor>> bk_results)
+                         vector<shared_ptr<runtime::Tensor>> bk_results,
+                         vector<shared_ptr<runtime::Tensor>> bk_isolated_results)
```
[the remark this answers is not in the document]
gcwenger: - ref_results are the results from the backend designated as reference (e.g. INTERPRETER or CPU) with original native data flow (e.g. CPU op1->CPU op2-> CPU op3)
- bk_results are the results from the backend being tested (e.g. GPU) with original native data flow (e.g. GPU op1->GPU op2->GPU op3)
- bk_isolated_results are the results from the runtime being tested (e.g. GPU) with isolated ops where the input is fed in from the reference backend (e.g. GPU op1; CPU op1 results->GPU op2; CPU op2 results->GPU op3)

test/backend_graph_comparison.in.cpp
gcwenger: 👍

test/backend_graph_comparison.in.cpp
gcwenger: Comment tweaked for better clarity.

test/backend_graph_comparison.in.cpp
```diff
@@ -185,7 +266,42 @@ NGRAPH_TEST_P(${BACKEND_NAME}, serialized_graph_files, compare_backends_with_gra
...
+    // Collect set of parameters and results to run test backend with isolated
+    // inputs which will be copied from the reference backend outputs
+    ParameterVector isolated_parameters = func->get_parameters();
+    NodeVector isolated_results;
+    unordered_map<ngraph::Node*, std::shared_ptr<ngraph::Node>> isolated_node_to_original_node;
+    for (auto n : func->get_ordered_ops())
+    {
+        // Don't include op::Results otherwise Function c-tor will complain
+        if (!n->is_output() && !n->is_parameter() && !n->is_constant() &&
+            !(n->get_outputs().size() > 1))
```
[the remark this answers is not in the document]
csullivan: I could see removing `!(n->get_outputs().size() > 1)`, and then checking if it is a multi output op. If it is, add a GOE for each output and then add that to the result list. That way if a graph is only using a few of the outputs, we still check all outputs of an op.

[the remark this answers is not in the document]
csullivan: After posting my last comment about multi-output ops, I realized that I think the isolated op testing may not cover multi-output ops at all currently. I think with this conditional you will only be testing the GOEs (essentially piping a CPU op output through a GPU GOE). If so, we need to handle multi-output ops with an explicit conditional inside this scope before merge. Can you check? A graph that has op::BatchNormTraining in it would work.

PR #2154, opened 2018-12-03
Update TBB from 2019_U1 to 2019_U2.

silee2:


jbobba:
Curious, any reason for the update?

silee2:
Just staying up to date. Maintenance releases usually have bug fixes. I don't see anything for this version other than fixing an issue for older version of macos though. There is always a risk of breaking things with a new version but since we have released beta, now may be a good time to do it. Not a high priority PR.

silee2:
Do we have any guideline on updating external third party components? Like monthly/quarterly/semi-annual/annual update

silee2:
BTW, TBB 2019_U3 was just released today.

rkimballn1:
I would recommend as an update policy to never update unless there is something you need. If it is working why fix it?

jbobba:
Sometimes updates bring in fixes for bugs that we didn't know existed. But agree with @rkimballn1 that we dont update too often.

PR #2155, opened 2018-12-03
Fix compiler error GCC with 7.1

fboemer:
Previously: ```error: constructor delegates to itself
             : PartialShape({})```


fboemer:
Build failure seems unrelated:
`[BFA] Found failure cause(s):
[BFA] OutOfMemoryOrKill-9`

PR #2156, opened 2018-12-03
DEX Loop Kernel (updated)

Krovatkin:
passing tests

clean up

fix build breaks


test/cpu_fusion.cpp
```diff
@@ -1447,34 +1447,69 @@ TEST(cpu_fusion, backwards_maxpool_with_indices_n4_c1_hw4_2x2_max)
     ASSERT_TRUE(read_vector<float>(output) == expected);
 }
 
-#if 0
-TEST(cpu_fusion, loop_kernel_one_input_one_output)
+#if defined(NGRAPH_HALIDE)
```
diyessi: Why don't we keep the halide tests in a halide-specific file.

src/ngraph/runtime/cpu/builder/loop_kernel.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include "ngraph/op/subtract.hpp"
+
+#include "halide_generators.hpp"
+#include "ngraph/runtime/cpu/cpu_builder.hpp"
+#include "ngraph/runtime/cpu/op/loop_kernel.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+#define TI(x) type_index(typeid(x))
```
rkimballn1: How many places is this defined in our code? Current master the answer is 8. I think this makes 9. Hmm.

test/cpu_fusion.cpp
rkimballn1: If they are in a separate file then you can use cmake to add the test source file or not depending on halide/no halide

PR #2160, opened 2018-12-03
Allow CPU build without TBB.

silee2:
Standalone usage sometimes require small binary size. Providing an option to build CPU backend without TBB can help reducing size.


src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -195,7 +200,10 @@ if (NGRAPH_CPU_ENABLE)
     endif()
 
     add_dependencies(cpu_backend ext_mkldnn ext_eigen)
-    target_link_libraries(cpu_backend PUBLIC ngraph libmkldnn libeigen libjson libtbb)
+    target_link_libraries(cpu_backend PUBLIC ngraph libmkldnn libeigen libjson)
+    if (NGRAPH_TBB_ENABLE)
+        target_link_libraries(cpu_backend PUBLIC libtbb)
```
rkimballn1: I think this should be PRIVATE. Probably should make libmkldnn, libeigen and libjson PRIVATE too. codegen too.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -457,6 +461,7 @@ void runtime::cpu::CPU_ExternalFunction::compile()
     codegen::CodeWriter writer;
 
     writer << "// Generated by the nGraph CPU backend\n";
+#if defined(NGRAPH_TBB_ENABLE)
```
rkimballn1: Can we remove the runtime enable/disable for TBB? We probably don't need both.

src/ngraph/runtime/cpu/CMakeLists.txt
silee2: Had to make it PUBLIC again as it breaks compiling a unit test.

silee2:
Closing since TBB removal is not required yet.

PR #2145, opened 2018-11-30
Bug fix for invalid memory access to Constant (GPU backend)

xiezhq-hermann:
Fix the incorrect way to query the size of Constant tensor, which lead invalid memory access

PR #2161, opened 2018-12-03
fix tensor size bug that can cause invalid memory access

fengleitian:
apply change from this external PR:
https://github.com/NervanaSystems/ngraph/pull/2145


fengleitian:
close it for now.

PR #2163, opened 2018-12-04
Fix setup.py for CentOS

arogowie-intel:
Fixes error with `setuptools` to install numpy from `install_requires` - this tries to build `numpy` from sources which fails: https://github.com/numpy/numpy/issues/2434.
The _workaround_ is to add `numpy` dependency to `setup_requires`.

Moreover on CentOS libraries are placed at `ngraph_dist/lib64` not `ngraph_dist/lib` - so we check for this case.


python/setup.py
```diff
@@ -65,10 +65,16 @@ def find_pybind_headers_dir():
 
 
 NGRAPH_CPP_DIST_DIR = find_ngraph_dist_dir()
-PYBIND11_INCLUDE_DIR = find_pybind_headers_dir() + "/include"
-NGRAPH_CPP_INCLUDE_DIR = NGRAPH_CPP_DIST_DIR + "/include"
-NGRAPH_CPP_LIBRARY_DIR = NGRAPH_CPP_DIST_DIR + "/lib"
-
+PYBIND11_INCLUDE_DIR = find_pybind_headers_dir() + '/include'
+NGRAPH_CPP_INCLUDE_DIR = NGRAPH_CPP_DIST_DIR + '/include'
```
diyessi: Does this risk picking up a 32-bit lib ahead of a 64-bit?

python/setup.py
```diff
@@ -377,6 +383,7 @@ def build_extensions(self):
     packages=packages,
     cmdclass={'build_ext': BuildExt},
     data_files=data_files,
+    setup_requires=['numpy'],
```
silee2: Why does setup require numpy?

python/setup.py
arogowie-intel: Currently I've not seen that building nGraph produced two sets of libraries one for 32-bit and the other one for 64-bit. As far as I understand from `CMakeLists.txt`: https://github.com/NervanaSystems/ngraph/blob/9e885d05214baebe83b6a83d67f9903cf790ecb0/CMakeLists.txt#L147 The default location for libraries is `lib` directory under `CMAKE_INSTALL_PREFIX` https://github.com/NervanaSystems/ngraph/blob/9e885d05214baebe83b6a83d67f9903cf790ecb0/CMakeLists.txt#L156 unless: https://github.com/NervanaSystems/ngraph/blob/9e885d05214baebe83b6a83d67f9903cf790ecb0/CMakeLists.txt#L141-L142 which on CentOS sets the `CMAKE_INSTALL_LIBDIR` to `lib64`. So there is always only single set of libraries however it may be placed under `lib` or `lib64` directory which we check here.

PR #2169, opened 2018-12-04
Matcher skip

Krovatkin:
```
            // This env var allows one to specify node name patterns to abort pattern matching
            // at particular nodes. The upshot is that one can quickly zero in on an offending fusion by
            // disabling individual fusions or optimizations that use Matcher.
```


src/ngraph/pattern/matcher.cpp
```diff
@@ -177,6 +178,18 @@ namespace ngraph
             add_node(graph_node);
             size_t watermark = m_matched_list.size() - 1;
 
+            static const char* node_skip_cregex = std::getenv("NGRAPH_MATCHER_SKIP");
```
diyessi: @jbobba added a config table for setting/reading these sorts of things programmatically a few weeks ago, so we can set them in specific tests. Can you use/extend that here?

Krovatkin: @diyessi it might be a fairly substantial effort to make sure that a setting or settings from PassConfig are propagated to Matcher simply because Matcher isn't just used in one Pass but multiple passes, callbacks, handlers. It's fairly ubiquitous at this point. Also, the main use case of this is to *not* be invoked programmatically, but to give a **pass writer** a very quick way to disable individual fusions (more granular than disabling passes) without changing a single line of code in either ngraph or bridge.

PR #2171, opened 2018-12-04
Support dynamic scales for Qconv's and Dequantize

nbpatel:


jbobba:
There is a fair amount of code duplication. Can you refactor existing functions in mkldnn_emitter and reuse them?

PR #2175, opened 2018-12-05
fix failing bn test

pruthviIntel:


pruthviIntel:
yup, it was in the codegen..you missed it when you refactored batchnorm code. :)

PR #2180, opened 2018-12-05
Update slice kernels

jbobba:
Over 5X improvement in GNMT backprop. 

image

to 

image


test/cpu_fusion.cpp
```diff
@@ -3113,6 +3114,149 @@ TEST(cpu_fusion, fuse_leaky_relu)
...
 
+TEST(cpu_fusion, fuse_update_slice)
+{
+    auto make_function = [](bool fuse = true) {
+        auto input = std::make_shared<op::Parameter>(element::f32, Shape{4, 32, 16});
+        Shape lower_bounds{1, 0, 0};
+        Shape upper_bounds{2, 32, 16};
+        auto slice = std::make_shared<op::Slice>(
+            input, fuse ? lower_bounds : Shape{3, 0, 0}, fuse ? upper_bounds : Shape{4, 32, 16});
+        auto update = std::make_shared<op::Parameter>(element::f32, Shape{1, 32, 16});
```
gaurides: Trying to understand : shouldn't we add a GOE or some other node in between so that we don't fuse?
I don't understand how changing lower and upper bounds prevent fusion.

test/cpu_fusion.cpp
```diff
@@ -3113,6 +3114,149 @@ TEST(cpu_fusion, fuse_leaky_relu)
...
+    }
+    auto int_results = execute(int_f, args, "INTERPRETER");
+    auto cpu_results = execute(cpu_f, args, "CPU");
+    for (size_t i = 0; i < cpu_results.size(); i++)
+    {
+        EXPECT_TRUE(test::all_close(cpu_results.at(i), int_results.at(i)));
+    }
+}
+
+TEST(cpu_fusion, fuse_update_slice_strided)
```
gaurides: Do you see a need for fuse_update_slice_strided_inplace test?

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -427,6 +428,21 @@ namespace ngraph
                     replace_slice->set_op_annotations(op_annotations);
                 }
 
+                template <>
+                void CPUAssignment::ASSIGN_DECL(ngraph::op::UpdateSlice)
+                {
+                    auto replace_slice = static_cast<op::UpdateSlice*>(node);
```
gaurides: Can you please rename `replace_slice` to `update_slice`?

test/cpu_fusion.cpp
jbobba: fusion callback should check that the slice and replace_slice are indeed working on the same portion of the tensor. Otherwise, it shouldn't fuse.

test/cpu_fusion.cpp
jbobba: the inplace code path is similar for both kernels. so didn't add it in. but you are right that I should create one.

PR #2181, opened 2018-12-05
Ayzhuang/in place concat r1

ayzhuang:


aprocter:
I think I just got an auto-request because this PR merged in #2203, so removing the request.

jbobba:
Merged with #2210

PR #2184, opened 2018-12-05
Support for all_close_f w/ doubles

gcwenger:
Double support for all_close_f with parameterized testing of both float all_close_f and double all_close_f.


test/all_close_f.cpp
```diff
@@ -52,6 +71,230 @@ float bits_to_float(const string& s)
...
+
+            lower_bound = DoubleUnion(upper_bound_as_int | 0x8000000000000000).d;
+            past_lower_bound = DoubleUnion(past_upper_bound_as_int | 0x8000000000000000).d;
+        }
+    }
+
+    double expected;
+    int mantissa_bits;
+    int tolerance_bits;
+    double upper_bound;
```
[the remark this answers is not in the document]
gcwenger: Done.

test/util/all_close_f.cpp
```diff
@@ -127,6 +191,50 @@ uint32_t test::matching_mantissa_bits(uint32_t distance)
     return matching_matissa_bits;
 }
 
+uint64_t test::matching_mantissa_bits(uint64_t distance)
+{
+    uint64_t tolerance_needed = distance;
+
+    if (tolerance_needed < 0x8000000000000000)
```
fengleitian: This can be simplified in another PR as I discussed with @gcwenger .

test/util/all_close_f.cpp
```diff
@@ -203,6 +311,82 @@ bool test::all_close_f(const vector<float>& a,
...
+        {
+            max_distance = distances[i];
+            max_distance_index = i;
+        }
+        if (distances[i] < min_distance)
+        {
+            min_distance = distances[i];
+            min_distance_index = i;
+        }
+        bool is_close_f = distances[i] <= tolerance;
```
fengleitian: note here: Discussed with @gcwenger, is it possible to use ` <` instead of `<=`, currently when set tolerance bit to 0, it will still pass if there is 1 ulp diff.

test/util/all_close_f.cpp
gcwenger: +1

PR #2185, opened 2018-12-05
an env var to disable individual fusions

Krovatkin:


src/ngraph/pass/graph_rewrite.cpp
```diff
@@ -90,6 +92,48 @@ bool ngraph::pass::GraphRewrite::run_on_function(std::shared_ptr<ngraph::Functio
     return (NUM_TRIES - tries) > 1; //this means a graph was transformed
 }
 
+static const std::vector<std::regex> initialize_fusion_regexes()
+{
+    const char* cnsf = std::getenv("NGRAPH_SKIP_FUSION");
```
diyessi: It's more than just fusions. `NGRAPH_IGNORED_MATCHERS`

PR #2190, opened 2018-12-07
Disable large than cache test for INTERPRETER, enable for GPU

csullivan:


csullivan:
You, @gcwenger, and I discussed expanding tests to cover larger than cache tests. This is why I didn't do so. I'm disabling it for INTERPRETER because of the length of time it takes to run. If test-time is a concern, it likely is unfeasible to have systematic larger than cache tests for different hardware targets, since cache size vary significantly as does performance. For now, I can make this one GPU-only.

diyessi:
> You, @gcwenger, and I discussed expanding tests to cover larger than cache tests. This is why I didn't do so. I'm disabling it for INTERPRETER because of the length of time it takes to run. If test-time is a concern, it likely is unfeasible to have systematic larger than cache tests for different hardware targets, since cache size vary significantly as does performance. For now, I can make this one GPU-only.

Devices should have this sort of test, but I think the details are going to be device-specific. If you think this size is appropriate for other devices, we should have a comment in the manifest file saying the test is not appropriate for the interpreter, since we need to justify it being disabled.

csullivan:
Let's make it GPU only for now and address this class of tests explicitly in the future.

PR #2191, opened 2018-12-07
move GPU specific test to GPU only

rkimballn1:


src/ngraph/runtime/intelgpu/unit_test.manifest
```diff
@@ -148,4 +148,3 @@ max_3d_to_scalar_double
 argmin_trivial_in_i32
 argmax_4D_axis_3_i64_in_i32
 argmin_trivial_in_double
-maxpool_bprop_larger_than_cache
```
shssf: I would expect this test hangs until `echo 0 > /sys/module/i915/parameters/enable_hangcheck` implemented in CI testing.

PR #2195, opened 2018-12-07
change all_close tests to return gtest AssertionResult instead of bool

rkimballn1:
to allow for better error messages

This also silences all of the spurious output during test.

### DIFF
--- a/test/util/all_close.hpp
+++ b/test/util/all_close.hpp
@@ -20,6 +20,7 @@
 #include <memory>
 #include <vector>
 
+#include "gtest/gtest.h"
 #include "ngraph/type/element_type.hpp"
 #include "test_tools.hpp"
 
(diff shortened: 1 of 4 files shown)


test/util/all_close_f.cpp
```diff
@@ -361,7 +364,7 @@ bool test::all_close_f(const vector<double>& a, const vector<double>& b, int tol
     }
     if (!rc)
     {
-        NGRAPH_INFO << "diff count: " << diff_count << " out of " << a.size();
+        ar_fail << "diff count: " << diff_count << " out of " << a.size() << "\n";
```
gcwenger: Remove `if (!rc)` which is still using the old technique for limiting error reporting.

test/util/all_close.hpp
```diff
@@ -50,19 +52,18 @@ namespace ngraph
...
+                                << a[i] << " is not close to " << b[i] << " at index " << i << "\n";
                     }
                     count++;
                     rc = false;
                 }
             }
             if (!rc)
             {
-                NGRAPH_INFO << "diff count: " << count << " out of " << a.size();
+                ar_fail << "diff count: " << count << " out of " << a.size() << "\n";
```
gcwenger: Remove `if (!rc)` which is still using the old technique for limiting error reporting.

PR #2196, opened 2018-12-07
Framework for Hybrid GPU backend

rkimballn1:
Currently contains only the INTERPRETER backend. Most unit tests pass but GPUH unit tests are only run if you set NGRAPH_GPUH_ENABLE=TRUE in cmake.


src/ngraph/runtime/hybrid/hybrid_backend.cpp
```diff
@@ -80,8 +81,8 @@ runtime::Handle runtime::hybrid::HybridBackend::compile(shared_ptr<Function> fun
         instance.m_function = clone_function(*func);
 
         // Run placement pass
-        pass::Manager pass_manager;
-        pass_manager.register_pass<pass::AssignPlacement>(backend_list);
+        ngraph::pass::Manager pass_manager;
```
sasadep: is `ngraph::` needed even though we use `using namespace ngraph;`

src/ngraph/runtime/tensor.cpp
```diff
@@ -70,3 +73,21 @@ void runtime::Tensor::set_stale(bool val)
 {
     m_stale = val;
 }
+
+void runtime::Tensor::copy_from(const ngraph::runtime::Tensor& source)
```
sasadep: do we add tests for this later ?

PR #2197, opened 2018-12-07
Skip Broadcast in sigmoid fusion

gaurides:
Required to fuse Sigmoid in Wavenet.


src/ngraph/pass/core_fusion.cpp
```diff
@@ -92,11 +92,12 @@ void pass::CoreFusion::construct_sigmoid()
     auto exp_neg_input = std::make_shared<op::Exp>(neg_input);
 
     // broadcast input
-    auto constant = std::make_shared<pattern::op::Label>(element::f32, Shape{});
-    auto broadcast_constant = std::make_shared<op::Broadcast>(constant, Shape{3, 4}, AxisSet{0, 1});
+    auto constant = std::make_shared<pattern::op::Label>(element::f32, Shape{3, 4});
```
jbobba: in the callback, we need to check that the constant_label is indeed a constant and that it is equal to 1.

PR #2199, opened 2018-12-07
Bind cuda context to thread prior to compilation

csullivan:
I encountered a case in Tensorflow where the compilation thread failed with `CUDA_ERROR_INVALID_CONTEXT` as it was not the thread that initialized the backend and SetContextCurrent was not called until after compilation (in prepare_runtime_context).

### DIFF
--- a/src/ngraph/runtime/gpu/gpu_backend.cpp
+++ b/src/ngraph/runtime/gpu/gpu_backend.cpp
@@ -67,7 +67,7 @@ runtime::gpu::GPU_Backend::BackendContext::BackendContext()
     // Create context use driver API and make it current, the runtime call will pickup the context
     // http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html
     // #interoperability-between-runtime-and-driver-apis
-    m_cuda_manager->SetContextCurrent();
+    bind_cuda_to_thread();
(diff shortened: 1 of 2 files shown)


src/ngraph/runtime/gpu/gpu_backend.cpp
```diff
@@ -91,13 +91,18 @@ runtime::gpu::GPU_Backend::BackendContext::BackendContext()
...
-    //set context current each time in case thread changed
-    m_cuda_manager->SetContextCurrent();
+    // set context current each time in case thread changed
+    bind_cuda_to_thread();
     // add pointers to gpu primitives into the gpu runtime context
     m_runtime_context->gpu_primitives = m_primitive_emitter->get_primitives().data();
     m_runtime_context->gpu_memory_primitives = m_primitive_emitter->get_memory_primitives().data();
 }
 
+void runtime::gpu::GPU_Backend::BackendContext::bind_cuda_to_thread()
```
fengleitian: nit: maybe call `bind_cuda_context_to_thread`.

csullivan: Agreed that is more clear, I'll make the change.

PR #2200, opened 2018-12-10
Addressed some all_close_f loose ends

gcwenger:
* Tweaked new reporting method for all_close_f added by @rkimballn1 to also return message string on success (e.g. if being using with EXPECT_FALSE diagnostic info is reported; yes this is rare, but better form for completeness).
* Print very short inline info on success
* Fixed double printing for more precision as suggested by @fengleitian 
* Simplified all_close_f tolerance determination (bit manipulation code) as discussed with @fengleitian 
* Transitioned f64 graph comparison code to use double


test/util/all_close_f.cpp
```diff
@@ -299,20 +286,31 @@ ::testing::AssertionResult test::all_close_f(const vector<float>& a,
...
-            << min_distance_index << "])\n";
-    ar_fail << std::setprecision(std::numeric_limits<long double>::digits10 + 1)
-            << "loosest match:    " << matching_mantissa_bits(max_distance) << " mantissa bits ("
-            << a[max_distance_index] << " vs " << b[max_distance_index] << " at ["
-            << max_distance_index << "])\n";
-    ar_fail << "median match:     " << matching_mantissa_bits(median_distance)
-            << " mantissa bits\n";
-
-    return rc ? ::testing::AssertionSuccess() : ar_fail;
+    if (rc)
```
rkimballn1: Can we condition these with an env var or something? Unobtrusive is in the eye of the beholder.

gcwenger: Now quiet by default.

test/util/all_close_f.cpp
```diff
@@ -299,20 +286,33 @@ ::testing::AssertionResult test::all_close_f(const vector<float>& a,
...
+        << "])\n";
+    msg << std::setprecision(std::numeric_limits<long double>::digits10 + 1)
+        << "loosest match:    " << matching_mantissa_bits(max_distance) << " mantissa bits ("
+        << a[max_distance_index] << " vs " << b[max_distance_index] << " at [" << max_distance_index
+        << "])\n";
+    msg << "median match:     " << matching_mantissa_bits(median_distance) << " mantissa bits\n";
+
+    ::testing::AssertionResult res =
+        rc ? ::testing::AssertionSuccess() : ::testing::AssertionFailure();
+    res << msg.str();
```
rkimballn1: Does success print out these messages? If so I vote no, if not then why the change?

CMakeLists.txt
```diff
@@ -117,6 +118,7 @@ message(STATUS "NGRAPH_HYBRID_ENABLE:         ${NGRAPH_HYBRID_ENABLE}")
 message(STATUS "NGRAPH_NOP_ENABLE:            ${NGRAPH_NOP_ENABLE}")
 message(STATUS "NGRAPH_DISTRIBUTED_ENABLE:    ${NGRAPH_DISTRIBUTED_ENABLE}")
 message(STATUS "NGRAPH_DEBUG_ENABLE:          ${NGRAPH_DEBUG_ENABLE}")
+message(STATUS "NGRAPH_GTEST_INFO:            ${NGRAPH_GTEST_INFO}")
```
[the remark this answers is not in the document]
gcwenger: I likely misunderstood your suggestion about how to control the output ("Can we condition these with an env var or something?"). I'll follow up tomorrow on what you hand in mind.

test/util/all_close_f.cpp
gcwenger: Define "success". Success depends on what result is expected in a given test. For most tests this is ASSERT_TRUE and EXPECT_TRUE, which will be silent when rc is true and res gets ::testing::AssertionSuccess().

But ASSERT_FALSE and EXPECT_FALSE will print a message when rc is true and res gets ::testing::AssertionSuccess(). This actually can happen in the unit testing to verify the correct functioning of all_close_f which does EXPECT_TRUE for cases where all_close_f is expected to return true/AssertionSuccess and EXPECT_FALSE for cases where all_close_f is expected to return false/AssertionFailure.

If we're testing a condition where all_close_f is supposed to fail / return false/AssertionFailure (by using EXPECT_FALSE) and all_close_f instead returns true/AssertionSuccess (because for example a change has broken all_close_f) then the message will be printed.

CMakeLists.txt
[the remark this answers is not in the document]
gcwenger: Switched to environment variable.

PR #2202, opened 2018-12-10
[ONNX] Expose API enabling to get supported ONNX operators set.

arogowie-intel:


src/ngraph/frontend/onnx_import/onnx.cpp
```diff
@@ -99,6 +99,11 @@ namespace ngraph
             OperatorsBridge::register_operator(name, version, domain, std::move(fn));
         }
 
+        OperatorSet get_operator_set(std::int64_t version, const std::string& domain)
```
apwojcik: This is not what you need. Instead you supposed to provide a function e.g. `std::set<std::string> get_operators_list(version, domain)`.  You need the names of operators available (supported) for the given domain and version, and not the internals of ONNX importer exposed.

src/ngraph/frontend/onnx_import/onnx.hpp
```diff
@@ -40,6 +42,16 @@ namespace ngraph
...
                                Operator fn);
 
+        /// \brief      Return the set of names of supported operators.
+        ///
+        /// \param[in]  version  The requested version of ONNX operators set.
+        /// \param[in]  domain   The requested domain the operators where registered for.
+        ///
+        /// \return     The set containing names of supported operators.
+        ///
+        std::set<std::string> get_supported_operators(std::int64_t version,
```
postrational: We could add `get_supported_operators(std::int64_t version)` which assumes `ai.onnx` as the default domain.

apwojcik: This can be achieved by the default parameter `get_supported_operators(std::int64_t version, std::string domain = "")` and then the user may just call `get_supported_operators(version)`.
However, I see no use case for such a scenario. By ONNX specification an operator name combines `domain` name and `version`.

PR #2206, opened 2018-12-10
Skip --exclude-libs linker flag on macOS

aprocter:
The macOS linker does not seem to like the `--exclude-libs=ALL` flag used by codegen. This PR omits that flag on macOS. Not 100% sure it's the right way to handle it, but it seems to be what we are doing for `libngraph`.


silee2:
<--exclude-libs> is a GNU ld only option.

src/ngraph/codegen/CMakeLists.txt
```diff
@@ -84,9 +84,11 @@ if (NGRAPH_GPU_ENABLE OR (NGRAPH_CPU_ENABLE AND NOT NGRAPH_DEX_ONLY))
     add_dependencies(codegen header_resource)
     if (NGRAPH_CPU_ENABLE)
         add_dependencies(codegen libmkldnn libeigen)
-        # --exclude-libs=ALL prevents symbols from statically-linked libraries (LLVM, in this case)
-        # from being automatically exported
-        set_property(TARGET codegen APPEND PROPERTY LINK_FLAGS "-Wl,--exclude-libs=ALL")
+        if(NOT APPLE)
```
rkimballn1: This won't work on WIN32

silee2: `if(NOT LINUX)` May work.

aprocter:
#2217 incorporated this, so closing.

PR #2211, opened 2018-12-11
Move CPU ReshapeSinking to Core pass

jiminha:


jiminha:
NNPI also need this reshape sinking pass. Moving to Core per discussion with Nick.

PR #2212, opened 2018-12-11
Removing GOEE

Krovatkin:
This PR re-implements `GetOutputElementElimination` using `MemoryLayout` and `oi_pair` annotations infrastructure. It also disables GOEE on CPU thus completely switching to the new way of doing things.


src/ngraph/runtime/cpu/builder/get_output_element.cpp
```diff
@@ -0,0 +1,48 @@
...
+    {
+        namespace cpu
+        {
+            template <>
+            void Builder::BUILDER_DECL(ngraph::op::GetOutputElement)
+            {
+                auto& functors = external_function->get_functors();
+
+                auto functor = [&](CPURuntimeContext* ctx, CPUExecutionContext* ectx) {
+                    //nothing to do; GOEs' offsets should've been adjusted in MemoryLayout
```
jbobba: need a check and an assert here.

PR #2217, opened 2018-12-11
Any/all ops

aprocter:
Implementing phase 1 of the RFC to retire Reduce, ReduceWindow, SelectAndScatter.

Several things here:

1. `Any` and `All` ops. (`Any` is to `Or`, as `All` is to `And`, as `Sum` is to `Plus`.)
2. Unit tests and docs for the above
3. Implementation of `Any` and `All` for CPU backend.
3a. Added one of the new tests to CPU's unit test manifest, with explanation in a comment.
4. `AnyAllReplacement` pass that rewrites `Any` and `All` to `Reduce` as a compatibility stopgap.
5. Enabled `AnyAllReplacement` for `CPU` and `GPU`.
6. `AnyAllInsertion` pass that rewrites `Reduce`s to `Any`/`All` where possible (this is for @ransford2011).
7. Deprecation verbiage for the ops that we will be retiring.

Note that this PR branches from #2206.


src/ngraph/codegen/CMakeLists.txt
```diff
@@ -84,9 +84,11 @@ if (NGRAPH_GPU_ENABLE OR (NGRAPH_CPU_ENABLE AND NOT NGRAPH_DEX_ONLY))
     add_dependencies(codegen header_resource)
     if (NGRAPH_CPU_ENABLE)
         add_dependencies(codegen libmkldnn libeigen)
-        # --exclude-libs=ALL prevents symbols from statically-linked libraries (LLVM, in this case)
-        # from being automatically exported
-        set_property(TARGET codegen APPEND PROPERTY LINK_FLAGS "-Wl,--exclude-libs=ALL")
+        if(LINUX)
```
aprocter: This is some stuff I pulled in from #2206

src/ngraph/pass/any_all_insertion.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "ngraph/op/and.hpp"
+#include "ngraph/op/any.hpp"
+#include "ngraph/op/constant.hpp"
+#include "ngraph/op/or.hpp"
+#include "ngraph/op/reduce.hpp"
+#include "ngraph/util.hpp"
+
+using namespace ngraph;
+
+bool is_boolean_scalar_constant_with_val(std::shared_ptr<ngraph::Node> node, bool val)
```
rkimballn1: Functions should be static or in an anonymous namespace. Next function too.

PR #2221, opened 2018-12-12
Remove _backwards from conv backprop ops.

diyessi:
Move interpreter-specific code out of the convolution ops.

### DIFF
--- a/src/ngraph/op/convolution.cpp
+++ b/src/ngraph/op/convolution.cpp
@@ -270,13 +270,6 @@ void op::ConvolutionBackpropData::validate_and_infer_types()
     //
     // To _validate_, we simply need to check/infer the output shape of the forward convolution,
     // then check to make sure that the incoming delta has the same shape as the forward output.
-    //
-    // We will also compute and store the various parameters in the "backward" column above, since
-    // some backends need them. (TODO(amprocte): Is it just because of the way the reference works
-    // that this stuff is needed? If so, we can probably get rid of it and have conv_backprop
-    // reference kernels that do the calculations of the backward parameters internally, or supply
-    // utility functions to do it.)
-
     const PartialShape& filters_shape = get_input_partial_shape(0);
     element::Type filters_et = get_input_element_type(0);
     const PartialShape& delta_shape = get_input_partial_shape(1);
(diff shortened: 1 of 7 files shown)


src/ngraph/op/convolution.cpp
```diff
@@ -364,18 +323,32 @@ void op::ConvolutionBackpropData::generate_adjoints(autodiff::Adjoints& adjoints
...
-        window_dilation_strides.push_back(m_window_movement_strides_backward[i]);
-        padding_below.push_back(m_padding_below_backward[i]);
-        padding_above.push_back(m_padding_above_backward[i] -
-                                (m_padding_below_backward[i] +
-                                 (x_shape[i + 2] - 1) * m_data_dilation_strides_backward[i] +
-                                 m_padding_above_backward[i] -
-                                 (f_shape[i + 2] - 1) * m_window_dilation_strides_backward[i]) %
-                                    m_window_movement_strides_backward[i]);
-        data_dilation_strides.push_back(m_data_dilation_strides_backward[i]);
+        window_movement_strides.push_back(m_window_dilation_strides_forward[i]);
```
aprocter: `m_window_dilation_strides_forward[i]` here was originally
`m_window_dilation_strides_backward[i]` which was defined as
`m_window_dilation_strides_forward[i]`.

So this looks good.

src/ngraph/op/convolution.cpp
```diff
@@ -364,18 +323,32 @@ void op::ConvolutionBackpropData::generate_adjoints(autodiff::Adjoints& adjoints
...
+        window_dilation_strides.push_back(m_window_movement_strides_forward[i]);
```
aprocter: `m_window_movement_strides_forward[i]` here was originally
`m_window_movement_strides_backward[i]` which was defined as
`m_data_dilation_strides_forward[i]`.

So I think this should be CHANGED to `m_data_dilation_strides_forward[i]`.

src/ngraph/op/convolution.cpp
```diff
@@ -364,18 +323,32 @@ void op::ConvolutionBackpropData::generate_adjoints(autodiff::Adjoints& adjoints
...
+        ptrdiff_t padding_below_backward =
```
aprocter: Original definition of `m_padding_below_backward[i]` referenced no other `_backward`s, and this is the same definition, so I think we're okay.

src/ngraph/op/convolution.cpp
```diff
@@ -364,18 +323,32 @@ void op::ConvolutionBackpropData::generate_adjoints(autodiff::Adjoints& adjoints
...
+            (static_cast<ptrdiff_t>(filters_shape[i + 2]) - 1) * window_dilation_strides[i] -
+            m_padding_below_forward[i];
+        ptrdiff_t padding_above_backward =
```
aprocter: Original definition of `m_padding_above_backward[i]` referenced no other `_backward`s, and this is the same definition, so I think we're okay.

src/ngraph/op/convolution.cpp
```diff
@@ -364,18 +323,32 @@ void op::ConvolutionBackpropData::generate_adjoints(autodiff::Adjoints& adjoints
...
+              m_padding_above_forward[i] -
+              (static_cast<ptrdiff_t>(filters_shape[i + 2]) - 1) *
+                  m_window_dilation_strides_forward[i]) %
+             m_window_movement_strides_forward[i]) -
+            m_padding_above_forward[i];
+
+        padding_below.push_back(padding_below_backward);
+        padding_above.push_back(
+            padding_above_backward -
+            (padding_below_backward + (x_shape[i + 2] - 1) * m_window_movement_strides_forward[i] +
```
aprocter: `m_window_movement_strides_forward[i]` here was originally
`m_data_dilation_strides_backward[i]` which was defined as
`m_window_movement_strides_forward[i]`.

So I think this is good.

src/ngraph/op/convolution.cpp
```diff
@@ -364,18 +323,32 @@ void op::ConvolutionBackpropData::generate_adjoints(autodiff::Adjoints& adjoints
...
+             padding_above_backward - (f_shape[i + 2] - 1) * m_window_movement_strides_forward[i]) %
```
aprocter: `m_window_movement_strides_forward[i]` here was originally
`m_window_dilation_strides_backward[i]` which was defined as
`m_window_dilation_strides_forward[i]`.

So I think this should be CHANGED to `m_window_dilation_strides_forward[i]`.

src/ngraph/op/convolution.cpp
```diff
@@ -364,18 +323,32 @@ void op::ConvolutionBackpropData::generate_adjoints(autodiff::Adjoints& adjoints
...
+                m_data_dilation_strides_forward[i]);
```
aprocter: `m_data_dilation_strides_forward[i]` here was originally
`m_window_movement_strides_backward[i]` which was defined as
`m_data_dilation_strides_forward[i]`.

So I think this is good.

src/ngraph/op/convolution.cpp
```diff
@@ -364,18 +323,32 @@ void op::ConvolutionBackpropData::generate_adjoints(autodiff::Adjoints& adjoints
...
+        data_dilation_strides.push_back(m_window_movement_strides_forward[i]);
```
aprocter: `m_window_movement_strides_forward[i]` here was originally
`m_data_dilation_strides_backward[i]` which was defined as
`m_window_movement_strides_forward[i]`.

So I think this is good.

aprocter:
Just for the record: for the two changes I flagged, the Wiki seems to agree. (Took me a little time to untangle the notation.)

PR #2223, opened 2018-12-12
Retire FunctionCall, Reduce, ReduceWindow, SelectAndScatter

aprocter:
This removes the `FunctionCall`, `Reduce`, `ReduceWindow`, and `SelectAndScatter` ops, i.e., all the ops that contain embedded functions.

From a front-end point of view, the only potential I can see for functional regression from this PR is for backends (`GPU` and `PlaidML`) that do not yet support the `Any` and `All` ops. This should affect a very small number of topologies, and implementing the ops should be pretty quick and easy.

@earhart I have attempted to make changes to the PlaidML backend to remove its support for Reduce, but since PlaidML is not yet gating for CI and I don't have a way to build it (read: have been too lazy to set up a build environment), this may break the PlaidML build.

Note that this PR does not do things like remove the `get_functions()` function, nor attempt to simplify backend internals in light of the fact that we no longer can have embedded functions. We can do these things in subsequent PRs if we want to.


aprocter:
> I'm working on making it easier to set up a PlaidML build environment. :-)

Sounds good! Again, `too lazy` are the operative words here  ;)

src/ngraph/runtime/gpu/unit_test.manifest
```diff
@@ -67,7 +63,29 @@ product_vector_zero
...
+all_2x0_to_scalar
+all_2x3_eliminate_col_dim
+all_2x3_eliminate_row_dim
+all_2x2x3_eliminate_dim_0
+all_2x2x3_eliminate_dim_1
+all_2x2x3_eliminate_dim_2
+all_2x2x3_eliminate_dims_0_1
+all_2x2x3_eliminate_dims_0_2
+all_2x2x3_eliminate_dims_1_2
+all_2x2x3_eliminate_dims_0_1_2
```
Krovatkin: nitpick: missing a new line

aprocter: Fixed

PR #2225, opened 2018-12-12
enable visualize graph even if graphviz is not installed

rkimballn1:
because it just doesn't matter


src/ngraph/pass/visualize_tree.cpp
```diff
@@ -163,15 +162,12 @@ void pass::VisualizeTree::render() const
 
         stringstream ss;
 
-        ss << "dot -T" << get_file_ext() << " " << tmp_file << " -o " << m_name;
+        ss << "dot -T" << get_file_ext() << " " << dot_file << " -o " << m_name;
```
shssf: call `dot` might take very long time. It depends on size of the graph. Perhaps it is better to just write `*.dot` file to disk without conversion it to a picture or pdf.

rkimballn1: Personally I would really miss the feature where it runs dot for me. I use it all the time. Were it missing I would be sad. True, large graphs could be a problem but to date we only use this on smaller graphs for debugging.

PR #2226, opened 2018-12-12
Better all_close_f tolerance interpretability

gcwenger:


test/util/all_close_f.cpp
```diff
@@ -270,7 +258,7 @@ ::testing::AssertionResult test::all_close_f(const vector<float>& a,
             min_distance = distances[i];
             min_distance_index = i;
         }
-        bool is_close_f = distances[i] <= tolerance;
+        bool is_close_f = distances[i] < tolerance;
```
fengleitian: Should we merge this or archive this?

PR #2240, opened 2018-12-17
Hybrid GPU Backend

rkimballn1:
This GPUH backend is a hybrid of the GPU and INTERPRETER backends. This is the first functional hybrid backend with multiple devices.
I added a parent field to runtime::Tensor to track which backend created which tensor. This required adding this to all of the existing backends.
The GPUH is disabled by default, but can be enabled via NGRAPH_GPUH_ENABLE. It does pass unit tests but is not tested by CI until we enable it.


src/ngraph/runtime/intelgpu/intelgpu_tensor_view.hpp
```diff
@@ -38,7 +38,8 @@ class ngraph::runtime::intelgpu::IntelGPUTensorView : public ngraph::runtime::Te
     IntelGPUTensorView(const element::Type& element_type,
                        const Shape& shape,
                        const cldnn::engine& backend_engine,
-                       void* memory_pointer = nullptr);
+                       void* memory_pointer,
```
shssf: Is it necessary to eliminate default parameter?

rkimballn1: Strictly no, but I added a new parameter that I don't want to have a default. I could have added this new parameter between backend_engine and memory_pointer but that would have change the signature more than I wanted. I decided to remove the default parameter as this constructor is called in exactly two places, both within private parts of the intel gpu backend so the default parameter seems unnecessary. The backend parameter is required and therefore can't be optional and you can't have a required parameter after an optional one. The change seemed harmless but if you want I can change to something you suggest.

src/ngraph/runtime/hybrid/pass/fix_get_output_element.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include "ngraph/runtime/backend.hpp"
+
+using namespace ngraph;
+using namespace std;
+
+runtime::hybrid::pass::FixGetOutputElement::FixGetOutputElement()
+{
+}
+
+bool runtime::hybrid::pass::FixGetOutputElement::run_on_node(shared_ptr<Node> node)
```
Krovatkin: Could we use a little bit more specific name unless we expect this pass to do something else aside from setting placements on GOEs?

src/ngraph/runtime/hybrid/hybrid_util.cpp
```diff
@@ -240,15 +251,17 @@ pair<vector<shared_ptr<Function>>, unordered_map<shared_ptr<op::Parameter>, shar
         }
         auto sub_function = make_shared<Function>(res_vector, par_vector);
         sub_functions.push_back(sub_function);
+        // ngraph::pass::Manager pass_manager;
+        // pass_manager.register_pass<ngraph::pass::VisualizeTree>("subgraph_" + to_string(index++) +
+        //                                                         ".png");
+        // pass_manager.run_passes(sub_function);
```
Krovatkin: should we remove commented out code?

src/ngraph/runtime/hybrid/hybrid_backend.cpp
```diff
@@ -116,70 +97,106 @@ bool runtime::hybrid::HybridBackend::call(shared_ptr<Function> func,
...
         }
 
         // Call
-        backend->call_with_validate(sub_function, result_tvs, parameter_tvs);
+        backend->call(sub_function, results, parameters);
+
+        // Need to copy any results to the correct device
+        // backend_list[0] is the "default" backend and is the backend which created the
+        // input and output tensors.
+        backend = m_backend_list[0];
```
Krovatkin: `backend` isn't being used?

src/ngraph/runtime/hybrid/hybrid_backend.cpp
```diff
@@ -56,46 +41,42 @@ shared_ptr<runtime::Tensor>
...
         // Clone function
         FunctionInstance instance;
         instance.m_function = clone_function(*func);
 
         // Run placement pass
         ngraph::pass::Manager pass_manager;
-        pass_manager.register_pass<runtime::hybrid::pass::AssignPlacement>(backend_list);
+        pass_manager.register_pass<runtime::hybrid::pass::AssignPlacement>(m_backend_list);
+        pass_manager.register_pass<runtime::hybrid::pass::FixGetOutputElement>();
+        // pass_manager.register_pass<ngraph::pass::VisualizeTree>("graph.png");
```
Krovatkin: should we remove the commented-out code?

src/ngraph/runtime/hybrid/hybrid_backend.cpp
```diff
@@ -116,70 +97,106 @@ bool runtime::hybrid::HybridBackend::call(shared_ptr<Function> func,
...
+                if (it->second->get_parent() == backend.get())
+                {
+                    parameters.push_back(it->second);
+                }
+                else
+                {
+                    auto parameter = backend->create_tensor(parameter_node->get_element_type(),
+                                                            parameter_node->get_shape());
+                    parameter->copy_from(*(it->second));
+                    parameters.push_back(parameter);
```
Krovatkin: this logic is very similar to results' one? I was wondering if we could maybe refactor it into a helper?

src/ngraph/runtime/hybrid/hybrid_backend.cpp
diyessi: yes.

src/ngraph/runtime/hybrid/pass/fix_get_output_element.cpp
rkimballn1: That is the entire plan for this pass, just to fix placement.

src/ngraph/runtime/hybrid/hybrid_backend.cpp
rkimballn1: There is a lot of rework to go on here but that really belongs in a separate PR.

src/ngraph/runtime/hybrid/hybrid_util.cpp
rkimballn1: conditionally compiled

PR #2248, opened 2018-12-18
Fixed the benchmark file to reflect the updated compile function usage

ransford2011:
This crashes for backends if they have updated to the new nGraph function call semantics.


src/tools/nbench/benchmark.cpp
```diff
@@ -214,7 +214,7 @@ vector<runtime::PerformanceCounter> run_benchmark(shared_ptr<Function> f,
     {
         for (int i = 0; i < warmup_iterations; i++)
         {
-            backend->call(f, results, args);
+            backend->call(backend->compile(f), results, args);
```
rkimballn1: Move compile outside the for loop. You don't want to call compile for every iteration.

yimeisun123: Maybe create a variable to hold backend->compile(f) outside the loop, and pass this var to backend->call in warm up and in the run.

PR #2254, opened 2018-12-19
[NGCORE-270] UT for Softplus ONNX operator testing edge cases.

arogowie-intel:
The reference values are calculated as in ONNXRuntime UT.


test/onnx_import.cpp
```diff
@@ -1535,3 +1538,45 @@ TEST(onnx, model_matmul_vec_ten3d)
...
+    // FLT_MAX, -FLT_MAX, std::numeric_limits<float>::infinity()  max, -max, inf;
+    Inputs inputs{std::vector<float>{-1.0f,
+                                     0,
+                                     1.0f,
+                                     100.0f,
+                                     -100.0f,
+                                     1000.0f,
+                                     -1000.0f,
+                                     FLT_MIN,
+                                     FLT_MIN / 10,
```
diyessi: Let's use 16 here to avoid issues with low-precision number formats.

test/onnx_import.cpp
```diff
@@ -1536,6 +1539,74 @@ TEST(onnx, model_matmul_vec_ten3d)
...
 }
 
+TEST(onnx, model_softplus)
+{
+    auto function =
+        onnx_import::import_onnx_model(file_util::path_join(SERIALIZED_ZOO, "onnx/softplus.onnx"));
+
+    // -1.0f, 0, 1.0f, 10.f,                    normal input values for activation
+    // 100.0f, -100.0f, 1000.0f, -1000.0f,      input values that leads to exp() overflow
+    // FLT_MIN, FLT_MIN / 10, -FLT_MIN / 10,    min, denorm, -denorm
```
gcwenger: Comment not in sync with values you updated below. At request from @diyessi you updated to / 16. Need to update that here too.

PR #2255, opened 2018-12-19
As Cast (one pass)

Krovatkin:


src/ngraph/runtime/cpu/pass/cpu_collapse_dims.cpp
```diff
@@ -244,29 +245,29 @@ bool runtime::cpu::pass::CPUCollapseDims::run_on_function(std::shared_ptr<ngraph
     bool replaced = false;
     for (auto n : f->get_ordered_ops())
     {
-        if (std::dynamic_pointer_cast<op::Broadcast>(n))
+        if (as<OP_TYPEID::Broadcast>(n))
```
gcwenger: Given that the implementation of `as<OP_TYPEID::SomeOpType>(n))` uses `get_typeid(n->description())`, which returns an `OP_TYPEID` that could be used in a switch statement, why not simplify this?
Instead of the templatized `as` trick, why not provide a helper method `get_typeid` which takes an `Op` and produces the `OP_TYPEID`. Then switch off that?

src/ngraph/runtime/cpu/op/as_cast.hpp
```diff
@@ -0,0 +1,206 @@
...
+#define NGRAPH_OP(a, b) a,
+enum class OP_TYPEID
+{
+#include "ngraph/op/op_tbl.hpp"
+    BACKEND_OPS
+};
+#undef NGRAPH_OP
+
+template <OP_TYPEID tid>
+struct Hash
```
XWangIntel: `Hash` could be named better to reflect the fact that this class is used as template function, whose input is a number tid and output is a type. How about `MapTypeIDToOp`

src/ngraph/runtime/cpu/op/as_cast.hpp
```diff
@@ -0,0 +1,181 @@
...
+    auto it = typeid_map.find(s);
+    if (it == typeid_map.end())
+    {
+        throw unsupported_op("Unsupported op '" + s + "'");
+    }
+    return it->second;
+}
+
+template <OP_TYPEID tid>
+static std::shared_ptr<typename Hash<tid>::OutputType> as(std::shared_ptr<ngraph::Node> n)
```
rkimballn1: You need to document this file. Why was this file created? Why not just use dynamic_cast? The name `as` does not make sense to me but if you documented things perhaps it would.

diyessi: dynamic_cast doesn't work across a shared library boundary on Mac, so any transformer in a shared library can't cast from `Node`.

PR #2268, opened 2018-12-28
Initial backend for non-IA CPUs

rkimballn1:
This runs on Raspberry Pi. Performance is not bad.


src/ngraph/log.cpp
```diff
@@ -30,7 +30,7 @@ using namespace ngraph;
 
 void ngraph::default_logger_handler_func(const string& s)
 {
-    cout << s << endl;
+    cout << s + "\n";
```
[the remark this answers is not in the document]
rkimballn1: When doing print with multithreading if you use endl you get two prints, one for the string and one for the endl. With multithreaded you get the string printed separately from the endl. If we concat the string and the newline then it always prints correctly. Print strings interleave with endl and do not if the string is a single string.

CMakeLists.txt
```diff
@@ -108,13 +108,14 @@ option(NGRAPH_GPU_ENABLE "Control the building of the GPU backend" FALSE)
 option(NGRAPH_INTERPRETER_ENABLE "Control the building of the INTERPRETER backend" TRUE)
 option(NGRAPH_NOP_ENABLE "Control the building of the NOP backend" TRUE)
 option(NGRAPH_GPUH_ENABLE "Control the building of the Hybrid GPU backend" FALSE)
+option(NGRAPH_RPI_ENABLE "Enable build nGraph Raspberry Pi backend" FALSE)
```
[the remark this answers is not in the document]
rkimballn1: I like the idea of GENERIC but it is not enough. It is GENERIC CPU. Unfortunately we used CPU for IA CPU when we it should be IA_CPU or something. Then we could have ARM_CPU and GENERIC_CPU but no joy there. I will use GENERIC_CPU for now.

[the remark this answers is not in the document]
rkimballn1: I renamed it to GCPU in the runtime/generic_cpu directory

src/ngraph/log.cpp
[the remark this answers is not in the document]
diyessi: I thought I once long ago saw that `endl` did `\r\n` versus `\n` but the doc says it does `\n` followed by `flush()` so you're right.

CMakeLists.txt
```diff
@@ -105,13 +105,14 @@ option(NGRAPH_GPU_ENABLE "Control the building of the GPU backend" FALSE)
 option(NGRAPH_INTERPRETER_ENABLE "Control the building of the INTERPRETER backend" TRUE)
```
Krovatkin: please reset the perms back to 0644

src/ngraph/runtime/CMakeLists.txt
```diff
@@ -37,4 +37,8 @@ if (NGRAPH_GPUH_ENABLE)
     add_subdirectory(gpuh)
```
Krovatkin: perms

src/ngraph/runtime/generic_cpu/kernel/broadcast.hpp
```diff
@@ -0,0 +1,179 @@
...
+                    return std::make_tuple(start, finish);
+                }
+                template <typename T>
+                void broadcast_2d(const T* in,
+                                  T* out,
+                                  const Shape& in_shape,
+                                  const Shape& out_shape,
+                                  const AxisSet& broadcast_axes)
+                {
+                    size_t index[2];
```
Krovatkin: why not use Eigen for 2,3,4-D tensors?

src/ngraph/runtime/generic_cpu/kernel/reshape.hpp
```diff
@@ -0,0 +1,256 @@
...
+                                 const Shape& in_shape,
+                                 const AxisVector& in_axis_order,
+                                 const Shape& out_shape)
+                {
+                    *out = *in;
+                }
+
+                template <typename T>
+                void reshape_in1(const T* in,
+                                 T* out,
```
Krovatkin: why not Eigen?

src/ngraph/runtime/generic_cpu/node_wrapper.cpp
```diff
@@ -0,0 +1,44 @@
+//*****************************************************************************
+// Copyright 2017-2018 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
Krovatkin: don't we have something similar for Serializer? can we reuse/generalize that code?

src/ngraph/runtime/generic_cpu/kernel/broadcast.hpp
rkimballn1: I'm not very familiar with eigen and could not get it to work

PR #2277, opened 2019-01-02
add version.hpp to ngraph install files for external backends

rkimballn1:
version.hpp contains a string that looks like this:

`const char* NGRAPH_VERSION_NUMBER = "0.12.0-rc.0+b7f097e";`


src/ngraph/version.in.hpp
```diff
@@ -0,0 +1,17 @@
+//*****************************************************************************
+// Copyright 2017-2018 Intel Corporation
```
diyessi: 2019

PR #2278, opened 2019-01-02
Fix a throw in slice cpu_memory_optimization.cpp due to unsupported i64 in MKLDNN

Krovatkin:


src/ngraph/runtime/cpu/pass/cpu_memory_optimization.cpp
```diff
@@ -301,6 +301,15 @@ bool runtime::cpu::pass::CPUMemoryOptimization::run_on_function(std::shared_ptr<
                 continue;
             }
 
+            const auto& dtype = slice->get_input_element_type(0);
+            if (runtime::cpu::mkldnn_utils::get_mkldnn_data_type(dtype) ==
+                mkldnn::memory::data_type::data_undef)
+            {
+                NGRAPH_DEBUG << slice->get_input_element_type(0).c_type_string()
+                             << " isn't supported";
```
gaurides: can you please add ", no in place slice" to the debug message for clarity?

Krovatkin: will do

PR #2281, opened 2019-01-03
A backend property request

shssf:


src/ngraph/runtime/backend.hpp
```diff
@@ -124,6 +124,16 @@ class ngraph::runtime::Backend
     /// \returns true if the op is supported, false otherwise.
     virtual bool is_supported(const Node& node) const;
 
+    /// \brief A set of properties supported by a backend
+    enum class Property
+    {
+        memory_attach /// A tansor can attach the user memory
```
rkimballn1: You misspelled `tensor`

shssf: fixed

shssf:
@rkimballn1 Could you please look into new changes in CPU backend?
If it is ok, this PR ready to merge.

src/ngraph/runtime/backend.hpp
```diff
@@ -124,6 +124,16 @@ class ngraph::runtime::Backend
     /// \returns true if the op is supported, false otherwise.
     virtual bool is_supported(const Node& node) const;
 
+    /// \brief A set of properties supported by a backend
+    enum class Property
+    {
+        memory_attach /// A tensor can attach the user memory
```
jbobba: its not clear what this means. Could you perhaps explain in more detail?

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -145,3 +145,13 @@ vector<runtime::PerformanceCounter>
     }
     return rc;
 }
+
+bool runtime::cpu::CPU_Backend::is_supported_property(const Property prop) const
+{
+    if (prop == Property::memory_attach)
+    {
+        return true;
```
jbobba: why is this true? what about the CPU backend makes this property supported. Is there a way to test that a backend does indeed support the property.

src/ngraph/runtime/backend.hpp
shssf: @jbobba I explained these things in PPT I sent you some time ago. We discussed that.
or maybe you mean a comment in the code is not clear?

src/ngraph/runtime/cpu/cpu_backend.cpp
shssf: @jbobba As far as I see the CPU has memory attach ability as already used in nGraph-TF https://github.com/NervanaSystems/ngraph-tf/blob/master/src/ngraph_encapsulate_op.cc#L316

src/ngraph/runtime/backend.hpp
jbobba: i dont recall the PPT but would be nice to make the comment here more self-explanatory.

PR #2284, opened 2019-01-04
[DO NOT MERGE] MKLDNN upgrade for PP

Krovatkin:


cmake/external_mkldnn.cmake
```diff
@@ -43,15 +43,15 @@ endif()
...
-    set(MKL_SHA1_HASH 0d9cc8bfc2c1a1e3df5e0b07e2f363bbf934a7e9)
+    set(MKL_SHA1_HASH 155fe3ed6d30a772baedb224c41c0e9c784e4b46)
     set(MKL_LIBS libiomp5.so libmklml_intel.so)
 elseif (APPLE)
     set(MKLPACKAGE "mklml_mac_${MKLVERSION}.tgz")
-    set(MKL_SHA1_HASH 232787f41cb42f53f5b7e278d8240f6727896133)
+    set(MKL_SHA1_HASH d2652cdbefbebf1096153a4154cb30ff643ec8f5)
     set(MKL_LIBS libmklml.dylib libiomp5.dylib)
 elseif (WIN32)
     set(MKLPACKAGE "mklml_win_${MKLVERSION}.zip")
```
rkimballn1: Missing update to windows hash

silee2:
The first commit should be enough. relu api was removed in newer version of mkl-dnn but the alternative element_wise api was added a while back. So we don't need a newer version of mkl-dnn to use element_wise api for relu implementation.

silee2:
The first commit updates CPU backend to use element wise api and the other commits are just for updating mkl-dnn to a newer version.

silee2:
Created #2295 since we don't need to actually upgrade mkl-dnn to a newer version to use the updated relu implementation base on elementwise. I think we shouldn't have any issue merging #2295 since it is backward compatible.

diyessi:
Superseded.

PR #2285, opened 2019-01-04
Simplified all_close_f interface and tightened default criteria

gcwenger:
Removed specification of mantissa bits for all_close_f in favor of just specifying tolerance bits. Tightened up all_close_f default. Fixed LRN unit test which had insufficient result precision to pass tighter all_close_f tolerance.


test/all_close_f.cpp
```diff
@@ -368,7 +355,7 @@ INSTANTIATE_TEST_CASE_P(
 
 // Test the exact bounds near +0.f
 //
-// With mantissa_bits = 8, tolerance_bits = 2
+// With tolerance_bits = 18
```
rkimballn1: You should probably explain why 18 bits for bfloat is the same as 2 bits for float.

test/backend_test.in.cpp
```diff
@@ -1257,17 +1261,17 @@ NGRAPH_TEST(${BACKEND_NAME}, lrn)
...
-                           0.01869806f,
-                           0.06805293f,
-                           0.03287071f,
-                           0.00509002f,
-                           0.00356153f,
-                           0.00174719f,
-                           0.0012555f,
-                           0.00322708f,
-                           0.00235574f};
+                           0.3015113f,
```
rkimballn1: Where did these numbers come from? How can I generate new numbers?

test/backend_graph_comparison.in.cpp
```diff
@@ -104,13 +104,19 @@ class serialized_graph_files : public ::testing::TestWithParam<string>
                 msg << "Test backed op run w/ original graph dependencies:"
                     << "\n";
                 msg << get_results_str(ref_data_vector, bk_data_vector);
-                bool all_close_graph = test::all_close_f(ref_data_vector, bk_data_vector);
+                ::testing::AssertionResult all_close_graph = test::all_close_f(
+                    ref_data_vector, bk_data_vector, DEFAULT_FLOAT_TOLERANCE_BITS + 14);
```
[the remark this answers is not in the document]
gcwenger: Reworked to be a little more flexible & added some code comments in an attempt to clarify. Ultimately this is an area for future work.

test/backend_test.in.cpp
gcwenger: TF script I wrote, which first reproduced old numbers. Tweaked parameters to generate numbers that were less susceptible to float precision issues (e.g. used alpha of 3, because ngraph definition of lrn means that alpha / size - previously this resulted in using 1/3), and included 7 float digits (in contrast to previous 6).
Could add TF script in comments.

PR #2287, opened 2019-01-04
Fix signed conv op

nbpatel:


src/ngraph/builder/quantization.cpp
```diff
@@ -337,18 +338,19 @@ namespace ngraph
...
+                                                                            sum_input,
+                                                                            window_movement_strides,
+                                                                            window_dilation_strides,
+                                                                            padding_below,
+                                                                            padding_above,
+                                                                            data_dilation_strides,
+                                                                            requantization_scale,
+                                                                            sum_scale,
+                                                                            with_relu);
+            return make_shared<op::Convert>(qconv, element::u8);
```
Krovatkin: I would double check if `with_relu` is always set to true.

src/ngraph/op/experimental/quantized_conv_bias.cpp
```diff
@@ -192,9 +193,9 @@ op::QuantizedConvolutionBiasSignedAdd::QuantizedConvolutionBiasSignedAdd(
     // TODO: call ngraph util
     // util::validate_convbias_shapes(data_batch_shape, filters_shape, bias->get_shape());
 
-    auto output_et = with_relu ? element::u8 : element::i8;
+    assert(with_relu == true);
```
sayantan-nervana: Can you please throw an `ngraph_error` or `AssertionFailure` or some ngraph exception. In the bridge we catch ngraph-errors and print some error messages

PR #2290, opened 2019-01-07
Use plaidml cmake config

earhart:
This makes it a bit simpler to build the PlaidML backend, by making nGraph automatically build it if PlaidML's been installed.


CMakeLists.txt
```diff
@@ -139,7 +139,17 @@ if (NGRAPH_ONNX_IMPORT_ENABLE)
     option(NGRAPH_ONNXIFI_ENABLE "Enable ONNX Interface for Framework Integration" TRUE)
 endif()
 
-option(NGRAPH_PLAIDML_ENABLE "Enable the PlaidML backend" FALSE)
+option(NGRAPH_PLAIDML_ENABLE "Enable the PlaidML backend" ON)
+
+if (NGRAPH_PLAIDML_ENABLE)
+    find_package(PlaidML CONFIG)
+    if (NOT PLAIDML_FOUND)
+        message(STATUS "PlaidML not installed; disabling PlaidML backend")
```
rkimballn1: If the user specifies to build PlaidML then it should either build it or fail. In the past we had this sort of option for the GPU backend and for a couple of weeks our CI was not testing the GPU backend and was passing even though the GPU backend was requested for the build.

earhart: Something like this?

CMakeLists.txt
```diff
@@ -139,7 +139,17 @@ if (NGRAPH_ONNX_IMPORT_ENABLE)
     option(NGRAPH_ONNXIFI_ENABLE "Enable ONNX Interface for Framework Integration" TRUE)
 endif()
 
-option(NGRAPH_PLAIDML_ENABLE "Enable the PlaidML backend" FALSE)
+find_package(PlaidML CONFIG)
```
silee2: I don't see FindPlaidML.cmake in the repo. Is that file installed with PlaidML by a different installer?

earhart: Yes; there's a PlaidML Python wheel, and there may be other distribution mechanisms in the future.  On the PlaidML side, there's currently a hack SDK target that includes a working plaidml-config.cmake, and a PR-in-progress-once-I-fix-the-Windows-build that builds the plaidml-config.cmake into the Python wheel in a place where cmake expects to find it.

TL;DR: To build ngraph with plaidml, you do something like:
```
pip install plaidml
cmake ${NGRAPH_SRC_DIR}
make
```

Much easier than the current "Untar this, set this env var to this, set such-and-such cmake flag, &c" approach.  :-)

PR #2295, opened 2019-01-08
Replace relu implementation with element_wise

silee2:
Relu is removed from recent version of mkl-dnn.


silee2:
PaddlePaddle relies on externally built mkl-dnn and the whole #2284 is not needed to fix compilation issues. Can we merge this PR into master if it does not break anything? We would need to switch relu implementation when we upgrade mkl-dnn anyway.

silee2:
@mbrookhart Is this new implementation compatible with the mkl-dnn version that Amazon uses?

baojun-nervana:
@avijit-nervana @diyessi

mbrookhart:
@jbobba is out, remove him from the reviewers?

PR #2297, opened 2019-01-09
Addes backprop to BatchDot op, allows fusion in training.

louisfeng:
This PR implements BatchDot backprop to allow explicit BatchDot fusion in training. We observed good speed ups in GNMT training.


diyessi:
> @diyessi is `autodiff.in.cpp` for testing all backends? I think `BatchDot` is only available for CPU right now.

You could either do something that will be optimized into BatchDot on CPU, or do the same sort of test directly on BatchDot using the helper functions in autodiff.in.cpp. The numeric check helps catch errors like misplaced axes.

src/ngraph/runtime/cpu/pass/cpu_mat_fusion.cpp
```diff
@@ -578,16 +578,21 @@ bool runtime::cpu::pass::CPUBatchFusion::run_on_function(std::shared_ptr<Functio
         const Node& node = *n;
         if (TI(node) == TI(op::Concat))
         {
-            auto fused_node = fuse_batch_dot(n);
-            if (fused_node)
+            if (m_fusion_type & DIFFERENTIABLE_FUSIONS)
```
Krovatkin: seems a bit wasteful to run the whole fuse logic to realize later on it's the wrong fusion type. I'd reverse the checks

louisfeng:
@diyessi Added new test to `autodiff.in.cpp`.

louisfeng:
@csullivan I refactored `make_function()`

src/ngraph/runtime/cpu/pass/cpu_mat_fusion.hpp
```diff
@@ -29,12 +29,31 @@ namespace ngraph
...
                 public:
-                    bool run_on_function(std::shared_ptr<ngraph::Function> function) override;
+                    enum FusionType
+                    {
+                        //`DIFFERENTIABLE_FUSIONS` produce ops that support autodiff
+                        // i.e. implement `generate_adjoints`
+                        DIFFERENTIABLE_FUSIONS = 0x1,
+                        REGULAR_FUSIONS = 0x2,
+                        ALL = 0xFFFFFFFF
+                    };
```
jbobba: Can we avoid the duplication here and move the enum to a common place instead?

louisfeng:
@pruthviIntel Yes, should work the same way for RNN. I can enable it on the bridge side once it's merged in. Thanks.

PR #2299, opened 2019-01-10
[ONNX] Expose API to check whether ONNX Op is supported.

arogowie-intel:


src/ngraph/frontend/onnx_import/ops_bridge.hpp
```diff
@@ -62,6 +62,8 @@ namespace ngraph
         class OperatorsBridge
         {
         public:
+            static constexpr const int LATEST_SUPPORTED_OPSET_VERSION = 8;
```
tsocha: Please use `ONNX_OPSET_VERSION` instead.

PR #2301, opened 2019-01-10
Change interpreter to use dynamic memory allocation

rkimballn1:
Instead of using memory pool and pointers. Should make testing dynamic shapes easier.

### DIFF
--- a/src/ngraph/runtime/interpreter/int_backend.cpp
+++ b/src/ngraph/runtime/interpreter/int_backend.cpp
@@ -74,11 +74,7 @@ runtime::Handle runtime::interpreter::INTBackend::compile(shared_ptr<Function> f
         pass_manager.register_pass<pass::LikeReplacement>();
         pass_manager.register_pass<pass::AssignLayout<DenseTensorLayout>>();
         pass_manager.register_pass<pass::Liveness>();
-        pass_manager.register_pass<pass::MemoryLayout>(get_alignment());
         pass_manager.run_passes(function);
-
-        size_t memory_pool_size = function->get_temporary_pool_size();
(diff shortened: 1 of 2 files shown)


src/ngraph/runtime/interpreter/int_backend.cpp
```diff
@@ -159,42 +153,35 @@ bool runtime::interpreter::INTBackend::call(shared_ptr<Function> function,
...
         {
             descriptor::Tensor* tensor = op->get_output_tensor_ptr(i).get();
-            void* host_tensor = nullptr;
+            shared_ptr<HostTensor> host_tensor;
             auto it = tensor_map.find(tensor);
             if (it == tensor_map.end())
             {
-                auto offset = op->get_output_tensor(i).get_pool_offset();
-                host_tensor = instance.get_temporary_pointer(offset);
+                const Shape& shape = op->get_output_shape(i);
```
aprocter: For dynamic ops, the shape (and the element type, actually) can be dynamic at this point, and this is going to fail (assert) in that case. I think we'll either need to rig up a mechanism to delay creation of the `HostTensor` until runtime when the full shape is known, or make it possible for the `HostTensor` itself to have dynamic shape and delay allocation and/or allow resizing of its underlying buffer.

rkimballn1: Can we address this issue in a separate PR as it is beyond the scope of this PR. This PR is only to change the INTERPRETER from using a memory pool approach to using on-demand temporary tensor allocation.

aprocter: Sounds good!

PR #2303, opened 2019-01-11
update Dockerfile for new pyapi docs in 0.12

indie:


indie:
@aprocter got PythonAPI documentation to build with the `ngraph-core` pip package, something in the Dockerfile for CI needs updated, too, so the public-facing documentation will render properly.

indie:
Fixed in #2329 with a versioning hack instead.

PR #2304, opened 2019-01-11
sink pad and slice

Krovatkin:


test/reshape_sinking.cpp
```diff
@@ -163,3 +163,43 @@ TEST(reshape_sinking, nasnet_pooladd)
...
+TEST(reshape_sinking, slice_pad)
+{
+    Shape shape_a{1, 8, 8, 1};
+
+    AxisVector to_nhwc{0, 2, 3, 1};
+    AxisVector to_nchw{0, 3, 1, 2};
+
+    auto A = make_shared<op::Parameter>(element::f32, shape_a);
+    auto pad_value = op::Constant::create<float>(element::f32, Shape{}, std::vector<float>{0.0f});
+
```
diyessi: Better to not repeat values so you can always tell if the permutation worked correctly.

Krovatkin: addressed!

PR #2305, opened 2019-01-11
Move callback out of matcher

Krovatkin:
This PR moves the callback method out of the matcher for a better OO design (single responsibility). For a user of the GraphRewrite and Matcher classes, the change to use the new API is fairly simple. Previously, the callback function is passed to the constructor of the Matcher class. Now the callback should be passed to the `add_matcher()` method of the GraphRewrite class.

This PR does not change how the Matcher and GraphRewrite algorithms behaves.


test/pattern.cpp
```diff
@@ -196,9 +196,9 @@ class TestGraphRewrite : public ngraph::pass::GraphRewrite
         auto pattern = std::make_shared<pattern::op::Label>(iconst0);
 
         auto callback = [pattern](pattern::Matcher& m) {
-            NGRAPH_DEBUG << "In a callback for construct_add_zero against "
-                         << m.get_match_root()->get_name();
-            assert(m.get_match_root()->get_arguments().size() == 2);
+            std::cout << "In a callback for construct_add_zero against "
```
ayzhuang: why change to std::cout?

Krovatkin: a good catch. i'll switch it back!

aprocter:
Marking "Blocked" at the moment for UT failures

diyessi:
`ngraph::runtime::gpu::pass::LSTMFusion::construct_sigmoid` uses the callback.

louisfeng:
There are still improvements I want to make, but going to leave them for another PR.

louisfeng:
> Please add a description of what needs to be done to convert to the new arguments.

Description added to the PR.

src/ngraph/pass/graph_rewrite.cpp
```diff
@@ -64,35 +63,28 @@ bool pass::GraphRewrite::run_on_function(shared_ptr<Function> f)
...
+        // that need multiple passes. See comments above.
+        vector<MatchClosure> run_matchers{m_matchers};
         m_matchers.clear();
         for (auto node : f->get_ordered_ops())
         {
-            for (auto matcher : matchers)
+            for (auto& mc : run_matchers)
             {
-                if (is_dynamic_function &&
-                    (matcher->get_property(pass::PassProperty::REQUIRE_STATIC_SHAPE)))
```
aprocter: Do we want to keep the `REQUIRE_STATIC_SHAPE` logic? (Has it been moved elsewhere?)

src/ngraph/pass/graph_rewrite.cpp
```diff
@@ -139,41 +131,39 @@ bool pass::GraphRewrite::is_enabled(shared_ptr<pattern::Matcher> m)
...
-    bool is_dynamic_function = f->is_dynamic();
     do
     {
         for (auto node : f->get_ops())
         {
-            for (auto matcher : m_matchers)
+            for (auto& mc : m_matchers)
             {
-                if (is_dynamic_function &&
-                    (matcher->get_property(pass::PassProperty::REQUIRE_STATIC_SHAPE)))
```
aprocter: Do we want to keep the `REQUIRE_STATIC_SHAPE` logic? (Has it been moved elsewhere?)

src/ngraph/pass/graph_rewrite.cpp
louisfeng: I got rid of them because there's enough change in this PR that I need to make some adjustments to the GraphRewrite logic on handling dynamic shape. That's going to be in another PR.

louisfeng:
TEST NOW!

louisfeng:
TESTNOW

louisfeng:
TESTNOW

PR #2308, opened 2019-01-14
Skip denormal mode set iff clang

earhart:


earhart:
This broke the build on my Mac.

earhart:
That worked great -- thanks!

PR #2314, opened 2019-01-14
return `0` or `1` when converting to bool

apwojcik:
Only INTERPRETER and CPU backends are "working" right now. This change is required for ONNX, but it has a wider impact. MSFT ONNX runtime is expecting exactly this behavior when converting (casting) to `bool` from any integral or floating type.


src/ngraph/runtime/interpreter/int_backend.hpp
```diff
@@ -508,7 +508,7 @@ class ngraph::runtime::interpreter::INTBackend : public Backend
             {
             case element::Type_t::boolean:
                 reference::convert<T>(
-                    static_cast<const T*>(args[0]), static_cast<char*>(out[0]), element_count);
+                    static_cast<const T*>(args[0]), static_cast<bool*>(out[0]), element_count);
```
aprocter: I'm not sure this is the right way to do it, because the C++ standard does not require that `sizeof(bool) == 1`, but nGraph assumes that we are using one byte for each value in an `element::boolean` tensor. I think maybe we should have a special-case `reference::convert_to_bool` that does something like `static_cast<char>(static_cast<bool>(x))`.

(Similar comment applies to the conversion in the CPU backend, with the added awesomeness that I don't know how to effect the conversion we want through Eigen.)

test/backend_test.in.cpp
```diff
@@ -1453,12 +1453,12 @@ NGRAPH_TEST(${BACKEND_NAME}, convert_int32_bool)
 
     // Create some tensors for input/output
     auto a = backend->create_tensor(element::i32, shape);
-    copy_data(a, vector<int32_t>{1, 2, 3, 4});
+    copy_data(a, vector<int32_t>{0, 12, 23, 0});
```
aprocter: We might want to throw in some more extreme values here (e.g., `MAX_INT32 & (~0xFF)`).

src/ngraph/runtime/interpreter/int_backend.hpp
aprocter: Slight clarification: _technically_ nGraph only specifies element representations in host-side buffers---internally, a backend can store a boolean in 32 bits or 27 bits or whatever it wants---_but_ the CPU and interpreter backends have chosen to have internal storage follow the host-side conventions (at least at the per-element level).

PR #2316, opened 2019-01-14
fix parallel build of target `python_wheel`

apwojcik:

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -42,3 +42,7 @@ add_custom_command(
 )
 
 add_custom_target(python_wheel DEPENDS ngraph ${CMAKE_BINARY_DIR}/python/dist/)
+if (NGRAPH_CPU_ENABLE)
+    add_dependencies(python_wheel ext_mkldnn)
+endif()
+


python/CMakeLists.txt
```diff
@@ -42,3 +42,7 @@ add_custom_command(
 )
 
 add_custom_target(python_wheel DEPENDS ngraph ${CMAKE_BINARY_DIR}/python/dist/)
+if (NGRAPH_CPU_ENABLE)
+    add_dependencies(python_wheel ext_mkldnn)
```
silee2: Can you explain why you have to manually set dependency for only ext_mkldnn?

apwojcik: `python_wheel` target is different. It is a wrapper for `setup.py` of `pyngraph`, `make ngraph` and `package preparation` script. It seems it cannot handle dependencies correctly while building in parallel (a large number of tasks, i.e. just `make -j python_wheel`). The result is CPU backend cannot build because of the `pyngraph` installation routine installs MKLDNN to a different directory. When `ext_mkldnn` is building as the first target the `pyngraph` uses the part of CMake for preinstalled MKLDNN instead of part for an external project (fetching sources and building).
I tried to solve the problem different way - rearranging CMakeFiles.txt and changing the instructions for building, but I ended up in so many changes that it was not worthy (and it wasn't working perfectly). The simplest and working solution is to just add the dependency on `ext_mkldnn` to `python_wheel` target.

PR #2319, opened 2019-01-15
changes to get hybrid not dependent on any backends

rkimballn1:
and cpu configured as fallback cpu


aprocter:
Subbing @Krovatkin for @jbobba

PR #2321, opened 2019-01-15
Windows support.

jfdong:
Delete dependency of LLVM when building with MSVC. 
Use _popen on Windows instead of popen.

After this commit and importing dirent.h (Reason: It is a header file that is only on Linux system. Fix: https://github.com/tronkko/dirent is a project providing Linux compatible Dirent interface for Windows.), interpreter and unit test fully passed.

The build command with MSVC as follows:
 set path=%path%;C:\work\ngraph\source\ngraph\build\src\ngraph\Release
 cmake .. -G "Visual Studio 15 2017 Win64" -DGIT_EXECUTABLE="C:\Program Files\Git\cmd\git.exe" -DNGRAPH_CPU_ENABLE=OFF -DNGRAPH_DEX_ONLY=ON
 msbuild "ngraph.sln" /t:Rebuild /p:Configuration=Release /m


jfdong:
Current status after these fixes is CPU backend got compilation of all obj files passed, but failed when linking cpu_backend.dll.

src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -197,6 +197,11 @@ if (NGRAPH_CPU_ENABLE)
          message(WARNING "The build toolset doesn't support OpenMP. This will impact performance and lead to slowdowns.")
     endif()
 
+    if (MSVS)
```
rkimballn1: What is the difference between WIN32 and MSVS? Can we just use one or different meanings?

src/ngraph/runtime/cpu/builder/matmul_bias.cpp
```diff
@@ -262,8 +269,8 @@ namespace ngraph
                 size_t m = shape_a[1];
                 size_t k = shape_a[2];
                 size_t n = shape_b[2];
-                size_t lda = std::max(1UL, k);
-                size_t ldb = std::max(1UL, n);
+                size_t lda = max(1UL, k);
```
rkimballn1: Can you just use std::max<size_t> instead of the macros?

src/ngraph/runtime/cpu/CMakeLists.txt
jfdong: WIN32 works for both LLVM and VC. These compile options only works for VC.

silee2: MSVS is a variable that I created
See https://github.com/NervanaSystems/ngraph/blob/master/CMakeLists.txt#L67
It means Visual Studio is used. So it is different from WIN32
Windows build may use a different generator like Ninja if we add support in the future.

jfdong:
I pushed more fixes. now cpu_backend.dll can be successfully created, but lacking export function list, so it can't be used. @rkimballn1 please take a look.

cmake/external_tbb_prebuilt.cmake
```diff
@@ -37,11 +37,15 @@ ExternalProject_Add(
...
 
-set(TBB_LINK_LIBS
-    ${SOURCE_DIR}/lib/${CMAKE_SHARED_LIBRARY_PREFIX}clangTooling${CMAKE_SHARED_LIBRARY_SUFFIX}
-    ${SOURCE_DIR}/lib/${CMAKE_SHARED_LIBRARY_PREFIX}clangTooling${CMAKE_SHARED_LIBRARY_SUFFIX}
-    ${SOURCE_DIR}/lib/${CMAKE_SHARED_LIBRARY_PREFIX}clangTooling${CMAKE_SHARED_LIBRARY_SUFFIX}
-)
+if (WIN32)
+    set(TBB_LINK_LIBS ${SOURCE_DIR}/lib/intel64/vc14/tbb.lib)
+else()
+    set(TBB_LINK_LIBS
```
helloguo: nit: this `else` could be deleted because external_tbb_prebuilt.cmake is only invoked on Windows.

src/ngraph/runtime/cpu/cpu_builder.cpp
```diff
@@ -103,6 +103,12 @@
 #include "ngraph/type/element_type.hpp"
 #include "ngraph/util.hpp"
 
+#include "ngraph/runtime/cpu/mkldnn_invoke.hpp"
+#include "ngraph/runtime/cpu/mkldnn_utils.hpp"
+#include "ngraph/runtime/cpu/op/convert_layout.hpp"
```
helloguo: nit: `convert_layout.hpp` is included in line 100. so this one could be deleted.

jfdong: thanks for catching it. I will fix it.

cmake/external_tbb_prebuilt.cmake
jfdong: I would like to keep it there. Currently it is only used on Windows, but later it may be used on other system. It is always better to make codes portable.

jfdong:
@silee2 @diyessi @jbobba Could you please take a look.

src/ngraph/runtime/cpu/cpu_builder.cpp
```diff
@@ -360,6 +365,63 @@ namespace ngraph
...
+                auto& deps = mkldnn_emitter->get_primitive_deps(reorder_index);
+                auto functor = [&, reorder_index](CPURuntimeContext* ctx,
+                                                  CPUExecutionContext* ectx) {
+                    cpu::mkldnn_utils::set_memory_ptr(ctx, deps[0], arg_tensor);
+                    cpu::mkldnn_utils::set_memory_ptr(ctx, deps[1], out_tensor);
+                    cpu::mkldnn_utils::mkldnn_invoke_primitive(ctx, reorder_index);
+                };
+                functors.emplace_back(functor);
+            }
+
```
jbobba: This seems like a hack. Can we root-cause and do the right thing?

test/cpu_fusion.cpp
```diff
@@ -1348,10 +1348,13 @@ TEST(cpu_fusion, rnn_fusion_from_json_model)
     stringstream ss(json_string);
     shared_ptr<Function> func = ngraph::deserialize(ss);
     pass_manager.run_passes(func);
-    const size_t NUM_STEPS = 10;
+    // Replace variable NUM_STEP by its value to work around MSVC bug: 'NUM_STEPS' cannot be implicitly 
+    // captured because no default capture mode has been specified
+    //const size_t NUM_STEPS = 10;
```
jbobba: specify default capture mode (=) instead of commenting out NUM_STEPS.

CMakeLists.txt
```diff
@@ -315,10 +315,12 @@ include(cmake/external_eigen.cmake)
 include(cmake/external_mkldnn.cmake)
 include(cmake/external_cldnn.cmake)
 
-if (NGRAPH_USE_PREBUILT_LLVM OR DEFINED LLVM_TARBALL_URL)
-    include(cmake/external_llvm_prebuilt.cmake)
-else()
-    include(cmake/external_llvm.cmake)
+if (NOT WIN32)
```
diyessi: Rather than `(NOT WIN32)` and `if (WIN32) else ...` which assume everything not windows is the same, you should condition on Linux and Macos here; better would be to have a variable for using LLVM that was true for Linux and Macos. Same for the other similar changes.

src/ngraph/runtime/cpu/builder/matmul_bias.cpp
```diff
@@ -78,12 +78,12 @@ namespace ngraph
                             k,
                             1.0f,
                             static_cast<float*>(arg0_tensor),
-                            max(1UL, lda),
+                            max((size_t)1UL, lda),
```
diyessi: With the `size_t` there's no need for the `UL` so let's simplify to `1`.

test/tools.cpp
```diff
@@ -30,10 +30,19 @@ TEST(tools, nbench_functional)
     const string model_path = file_util::path_join(SERIALIZED_ZOO, model);
 
     stringstream ss;
-
+#ifdef _WIN32
```
diyessi: Let's set up a thin compatibility layer file with these functions that are slightly different; for example `port_open` does the `popen` sort of thing, etc. And this is another area to have a separate `ifdef` for linux/macos instead of an `else`.

src/ngraph/runtime/cpu/builder/matmul_bias.cpp
rkimballn1: Don't do old-style C casts. The right way to do this is `max<size_t>(1, lda)`

jfdong: Yes. fixed it.

src/ngraph/runtime/cpu/cpu_builder.cpp
jfdong: The reason of the redefined issue is because when cpu_builder.cpp using the following lines to register the build_dispatcher .  {TI(ngraph::runtime::cpu::op::ConvertLayout),
-                     &runtime::cpu::Builder::build<ngraph::runtime::cpu::op::ConvertLayout>}, MSVC create builder function for ConvertLayout using template function in cpu_build.hpp line 265. later it conflict with the function in builder/convert_layout.cpp. Is there other way to avoid this if not moving the explicit function definition before the use?

test/tools.cpp
jfdong: Fixed.

test/cpu_fusion.cpp
jfdong: Fixed.

src/ngraph/runtime/cpu/cpu_builder.cpp
jbobba: We can remove this line. This was probably accidentally left over from when we didn't have the registration mechanism.

jfdong: I pushed a better fix by creating a new template called REGISTER_CPU_OP_BUILDER and register the build function in convert_layout.cpp. The fix follows the design of keeping complex builder functions in separate file.  @jbobba Please take a look.

rkimballn1:
Supersceded by #2394

PR #2322, opened 2019-01-16
Add support for shape-specialization of functions

aprocter:
This adds a function to create a "shape-specialized" clone of a function.

For example, suppose that a function `f` has three parameters with partial shapes:

```
param0: ?
param1: {1,?,3}
param2: {?,?,4}
```

Shape specialization would allow us to create a clone of `f` where the shapes are (for example):

```
param0: {1,2}
param1: {1,5,3}
param2: {3,?,4}
```

But not (for example):

```
param1: {1,5,3,4}  // rank doesn't match {1,?,3}
param1: {2,?,3}    // the "2" doesn't match the "1"
param1: {?,?,3}    // the new shape is too relaxed: it doesn't require 1 for the first dim
```

Note that validation errors can potentially occur during cloning. For example:

```
n = Parameter{shape=?}
m = Parameter{shape=?}
x = n + m
f = Function(x,{n,m})
```

If we specialize `n` to the shape `{1,2,3}` and `m` to the shape `{4,5,6}`, cloning will fail because when we reconstruct the new `x` node, it will see that the shapes are inconsistent for elementwise add.

Specialization of element types is also possible: `element::dynamic` can be specialized to a concrete element type or left dynamic; but a concrete element type can only be specialized to itself (e.g., specialization does not allow you to change `element::i32` to `element::i64`).


src/ngraph/specialize_shapes.hpp
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+//*****************************************************************************
+
+#pragma once
+
+#include "ngraph/function.hpp"
+
+namespace ngraph
+{
+    /// \brief Creates a clone of a function, with the shapes of that function's parameters
```
rkimballn1: The description you have in the PR is excellent. So excellent I think it belongs in the header. It is more "tutorial" but I understood a lot from reading it.

PR #2327, opened 2019-01-16
[ONNX] UT for failing Slice op.

arogowie-intel:
Failing Slice op test (only on `INTERPRETER` backend) with err:
```
The start corner is out of bounds at axis 1
```
Slice attributes:
* axes: 1
* ends: 1000
* starts: 1000

ONNX Importer passes to nGraph `Slice` op parameters:
- lower_bounds: [0, 10, 0],
- upper_bounds: [20, 10, 5]


aprocter:
This is a fun one!

1. The `CoordinateIterator` junk requires that the start coordinate be strictly in bounds.
2. I think I intended shape validation for `Slice` to require the same thing, but it doesn't.
3. If we resolve the inconsistency of (1)-(2) by changing the shape validation rules to match the current `CoordinateIterator` assumptions, I think you would have to special-case the truncation of the slice bounds in the ONNX importer. (In this example, low bounds would have to be truncated to 9 rather than 10, but then so would high-bounds.)

So I think the thing to do is update `CoordinateIterator` to allow lower bounds to be less than _or equal_ to the bounds.

Could you open a JIRA for this?

arogowie-intel:
Great thanks @aprocter  for investigation! This indeed is a fun case :) ONNX standard states that `Slice` operator is _similar to `numpy`_. Therefore passing value greater than particular axis size is truncated to its size (in this example `1000` is truncated to 10). This results as you noticed in _zero-width slice taken out of bounds_ which in `numpy` terms has size of zero elements...  I'm not wondering on any rational usage for this :) 

I guess that in this special case we should just return _empty tensor_ with .. uh.. correct shape? Which hope your proposition would let to be able.

arogowie-intel:
I thought that It would be enough to change lines 136-137 in `coordinate_transform.cpp` file from:
```c++
if (source_start_corner[i] >= padded_upper_bounds[i] &&
            !(source_start_corner[i] == 0 && source_shape[i] == 0))
```
to: 
```c++
if (source_start_corner[i] > padded_upper_bounds[i])
```
However this change results in `Segmentation fault` in both `CPU` and `INTERPRETER`. So I'm giving-up at this point.

aprocter:
This seems to do the trick:

```
        if (source_start_corner[i] >= padded_upper_bounds[i] &&
            source_start_corner[i] != source_shape[i])
```

In other words, it's not allowed for the start corner to be past the upper bounds _unless_ this axis has zero "virtual" dimension.

I'll PR that separately.

arogowie-intel:
@aprocter  Unfortunatelly I still see segmentation faults both on cpu and interpreter: https://aipg-rancher.intel.com/jenkins/algo/blue/organizations/jenkins/ngraph-ci-premerge/detail/PR-2327/4/pipeline/74#step-467-log-358

test/onnx_import.in.cpp
```diff
@@ -38,6 +38,26 @@ static std::string s_manifest = "${MANIFEST}";
 using Inputs = std::vector<std::vector<float>>;
 using Outputs = std::vector<std::vector<float>>;
 
+namespace
```
postrational: Please copy this function into a separate PR, so we can merge it into `master`.

arogowie-intel: https://github.com/NervanaSystems/ngraph/pull/2423

diyessi:
Should be fixed by #2576

PR #2332, opened 2019-01-17
Use appropriate function to test NAN.

arogowie-intel:
Fix used function for testing whether provided value is NaN.


src/ngraph/util.cpp
```diff
@@ -440,7 +440,7 @@ void ngraph::check_fp_values_isnan(const char* name, const float* array, size_t
 {
     for (size_t i = 0; i < n; i++)
     {
-        if (std::isinf(array[i]))
+        if (std::isnan(array[i]))
```
rkimballn1: use `!std::isfinite(array[i])` to check for both nan and inf values

arogowie-intel: There is already a function checking only for `inf` here https://github.com/NervanaSystems/ngraph/blob/5eeb5dc89dde100c8afa143a678fd3a11a728f63/src/ngraph/util.cpp#L417
And what about the case when sb wants to check only for `NaN`?

The other thing is that both `check_fp_values_isinf` and `check_fp_values_nan` are not used anywhere in nGraph.

PR #2339, opened 2019-01-17
Divert dynamic offset to reference

nbpatel:


src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -728,8 +728,14 @@ namespace ngraph
                 void CPUAssignment::ASSIGN_DECL(ngraph::op::Dequantize)
                 {
                     auto dequantize = static_cast<op::Dequantize*>(node);
-                    auto offset_const_op =
-                        std::static_pointer_cast<ngraph::op::Constant>(dequantize->get_argument(2));
+                    auto offset_const_op = std::dynamic_pointer_cast<ngraph::op::Constant>(
+                        dequantize->get_argument(2));
```
jbobba: use is_constant instead of dynamic casting.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -761,8 +767,14 @@ namespace ngraph
                 {
                     auto quantize = static_cast<op::Quantize*>(node);
                     auto offset_const_op =
-                        std::static_pointer_cast<ngraph::op::Constant>(quantize->get_argument(2));
+                        std::dynamic_pointer_cast<ngraph::op::Constant>(quantize->get_argument(2));
```
jbobba: is_constant()

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
nbpatel: I use that to read the value of offset as well if its a constant. Not casting just for the check

jbobba: dynamic_casts dont work as expected on Mac. So check for is_constant and then do a static_cast.

nbpatel:
TESTNOW

PR #2346, opened 2019-01-18
converted test cases based on TF DO data

adstraw:


test/onnx_import.cpp
```diff
@@ -1580,7 +1580,7 @@ class lin_quant_model_param_test : public testing::TestWithParam<std::uint32_t>
 TEST_P(lin_quant_model_param_test, model_resnet50)
 {
     auto function = onnx_import::import_onnx_model(
-        file_util::path_join(SERIALIZED_ZOO, "onnx/final_int8_resnet50_static_input.onnx"));
+        file_util::path_join(SERIALIZED_ZOO, "onnx/final_int8_resnet50_static_shapes.onnx"));
```
arogowie-intel: I guess you also want to rename the model?

arogowie-intel: Why we use this old (first version) model? Isn't the "final_int8_resnet50_static_input.onnx" version newer?

adstraw: Nice catch.  You are correct.  I had gone back to test the old model to see what results it produced on the new (DO) data and forgot to change back.

PR #2347, opened 2019-01-19
Check if the pointer returned by malloc is nullptr.

ayzhuang:
Otherwise, it would cause segmentation fault at other place.

### DIFF
--- a/src/ngraph/runtime/aligned_buffer.cpp
+++ b/src/ngraph/runtime/aligned_buffer.cpp
@@ -17,7 +17,7 @@
 #include <cstdlib> // llvm 8.1 gets confused about `malloc` otherwise
 #include <memory>
 
-#include "ngraph/log.hpp"
+#include "ngraph/except.hpp"
 #include "ngraph/runtime/aligned_buffer.hpp"
 
 using namespace ngraph;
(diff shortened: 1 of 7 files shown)


src/ngraph/runtime/aligned_buffer.cpp
```diff
@@ -35,6 +36,10 @@ runtime::AlignedBuffer::AlignedBuffer(size_t byte_size, size_t alignment)
     {
         size_t allocation_size = m_byte_size + alignment;
         m_allocated_buffer = static_cast<char*>(malloc(allocation_size));
+        if (m_allocated_buffer == nullptr)
```
diyessi: Throw is very likely to fail if `malloc` has failed, but sometimes you get lucky.

ayzhuang: Is "NGRAPH_ERR + exit(1)" better than throw?

rkimballn1: I would throw an std::bad_alloc, or if you really want to go to the effort you can throw an ngraph_bad_alloc which does not yet exist.
The code will exit one way or the other and as @diyessi says you might get lucky. Throw is the correct behavior.

PR #2348, opened 2019-01-19
Support scalar inputs

nbpatel:
ngraph wants scalar as Shape{} and mkldnn wants it as Shape{1}. So changing the shape in the backend. Might not be the cleanest solution


src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -256,10 +261,16 @@ static void set_native_layouts(runtime::cpu::CPU_ExternalFunction* external_func
...
+        if (shape == Shape{})
+        {
+            shape = Shape{1};
+            strides = Strides{1};
+        }
+        if (mkldnn_utils::can_create_mkldnn_md(shape, strides, et))
         {
-            auto native_md =
-                mkldnn_utils::create_blocked_mkldnn_md(shape, layout->get_strides(), et);
+            auto native_md = mkldnn_utils::create_blocked_mkldnn_md(shape, strides, et);
```
jbobba: would be cleaner/less code duplication to modify create_blocked_mkldnn_md instead of creating Shape{1} here.

diyessi:
No longer needed.

PR #2350, opened 2019-01-21
[ONNX] Enable deselected supported opset domain when needed.

tsocha:


src/ngraph/frontend/onnx_import/core/graph.cpp
```diff
@@ -35,18 +35,22 @@ namespace ngraph
                 return result;
             }
 
+            inline std::string get_node_domain(const onnx::NodeProto& node_proto)
+            {
+                return (node_proto.domain().empty() ? "" : node_proto.domain());
+            }
+
             inline std::string to_string(const onnx::NodeProto& node_proto)
```
apwojcik: btw. please remove here, too

src/ngraph/frontend/onnx_import/core/graph.cpp
```diff
@@ -75,7 +79,11 @@ namespace ngraph
             {
                 if (!m_model->is_operator_available(node_proto))
                 {
-                    unknown_operator_types.emplace(detail::to_string(node_proto));
+                    m_model->enable_opset_domain(detail::get_node_domain(node_proto));
```
apwojcik: Can we separate this? The intention of the loop is to build the set of unsupported operators. This is a one-time operation, so having separate it is cleaner and more readable. Please don't forget to add an appropriate comment explaining what each of the loops does.

src/ngraph/frontend/onnx_import/ops_bridge.hpp
```diff
@@ -74,6 +74,11 @@ namespace ngraph
                 return instance()._get_operator_set(version, domain);
             }
 
+            static OperatorSet get_operator_set(const std::string& domain)
```
apwojcik: The version is missing.

src/ngraph/frontend/onnx_import/ops_bridge.hpp
```diff
@@ -107,6 +112,8 @@ namespace ngraph
                                     const std::string& domain,
                                     Operator fn);
             OperatorSet _get_operator_set(std::int64_t version, const std::string& domain);
+            OperatorSet _get_operator_set(const std::string& domain);
```
apwojcik: Redundant definition. Having those two methods introduces confusion and inconsistency about the API. Why single API call with default parameters is not enough?

src/ngraph/frontend/onnx_import/core/model.hpp
```diff
@@ -61,6 +61,8 @@ namespace ngraph
             /// \return `true` if the operator is available, otherwise it returns `false`.
             bool is_operator_available(const onnx::NodeProto& node_proto) const;
 
+            void enable_opset_domain(const std::string& domain);
```
arogowie-intel: Add documentation string.

src/ngraph/frontend/onnx_import/ops_bridge.cpp
```diff
@@ -154,6 +154,12 @@ namespace ngraph
             return result;
         }
 
+        OperatorSet OperatorsBridge::_get_operator_set(const std::string& domain)
+        {
+            std::int64_t version = 10;
```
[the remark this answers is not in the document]
arogowie-intel: Wouldn't be better to just set the `version` to `OperatorsBridge::LATEST_SUPPORTED_OPSET_VERSION` in the function call below?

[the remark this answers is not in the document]
arogowie-intel: Agree with Artur comments below. This function overload is redundant. Since this is pimpl just modify _private_ `_get*` function to accept `domain` as a first parameter and set `version` parameter default value equal to: `OperatorsBridge::LATEST_SUPPORTED_OPSET_VERSION`.

src/ngraph/frontend/onnx_import/core/graph.cpp
arogowie-intel: Moreover if you would do this out of this loop you could iterate only over the `unknown_operator_types` list.

src/ngraph/frontend/onnx_import/core/graph.cpp
```diff
@@ -25,26 +26,29 @@ namespace ngraph
...
-                return (node_proto.domain().empty() ? "" : node_proto.domain() + ".") +
-                       node_proto.op_type();
+                return (node_proto.domain().empty() ? "" : node_proto.domain());
             }
-        }
 
-        Graph::Graph(const onnx::GraphProto& graph_proto,
-                     const Model& model,
-                     const Weights& weights)
+            std::string to_string(const onnx::NodeProto& node_proto)
```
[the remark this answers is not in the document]
arogowie-intel: Actually your proposition seems to me a bit confusing - it's hard to understand what `reverse_dns` part relates to without reading the doc you linked. In fact the `reverse_dns` is only a form of operators domain. 

The operator is uniquely identified by the tuple (_domain_, _op\_type_, _since\_version_) https://github.com/onnx/onnx/blob/2e7099ee7c37b196c197c9a084a97698a41da232/onnx/onnx-operators.in.proto#L74. At this point we don't have _since\_version_ information, thus I would rename it to `get_op_uid` since we use the resulting value as a operator unique identificator.

PR #2353, opened 2019-01-22
add OpenMPI support  besides MLSL

sasadep:
a quick fix to have OpenMPI(for NNP + CPU) and  MLSL ( for Intel CPU  Only backend). A particular build can either have *openmpi* or *mlsl* and not both as we get runtime issues when having both *intelmpi* and *openmpi* at the same time.   
* the use of `NGRAPH_DISTRIBUTED_OMPI_ENABLE` enables openmpi 
   - this PR uses OpenMPI pre-existing library in the system. If not present on the system, the user is required to install `Open MPI 2.1.1`. 
   - support for building it from source will be added in consequent PR's.  
* the use of `NGRAPH_DISTRIBUTED_MLSL_ENABLE`  enables MLSL.


wenzhe-nrv:
@rsketine nnp needs openmpi for its multi-chip. so we need have both MLSL (for CPU only) and OpenMPI in the validation.

CMakeLists.txt
```diff
@@ -107,6 +107,7 @@ option(NGRAPH_NOP_ENABLE "Control the building of the NOP backend" TRUE)
 option(NGRAPH_GPUH_ENABLE "Control the building of the Hybrid GPU backend" FALSE)
 option(NGRAPH_GENERIC_CPU_ENABLE "Enable build nGraph for generic CPU backend" FALSE)
 option(NGRAPH_DISTRIBUTED_ENABLE "Add distributed mode to the CPU backend" FALSE)
+option(NGRAPH_DISTRIBUTED_MLSL_ENABLE "Add distributed mlsl mode to the for CPU only backend" FALSE)
```
silee2: This options is used like a sub option to NGRAPH_DISTRIBUTED_ENABLE. I think the option should only exist when NGRAPH_DISTRIBUTED_ENABLE is TRUE/ON.

sfvaroglu:
Two independent flags, `NGRAPH_DISTRIBUTED_MPI_ENABLE` and `NGRAPH_DISTRIBUTED_MLSL_ENABLE` will flow better and will be easier to maintain (both for validation and dev). Also, you are removing `NGRAPH_DISTRIBUTED` from several places. I'd recommend using a separate definition for each library.

src/ngraph/codegen/CMakeLists.txt
```diff
@@ -47,8 +47,14 @@ if ((NGRAPH_GPU_ENABLE OR NGRAPH_CPU_ENABLE) AND NOT NGRAPH_DEX_ONLY)
...
 
     if(NGRAPH_DISTRIBUTED_ENABLE)
-        get_target_property(MLSL_INCLUDE_DIR libmlsl INTERFACE_INCLUDE_DIRECTORIES)
-        list(APPEND HEADER_SEARCH_DEFINES MLSL_HEADER_PATH="${MLSL_INCLUDE_DIR}")
+        if (NGRAPH_DISTRIBUTED_MLSL_ENABLE)
+            get_target_property(MLSL_INCLUDE_DIR libmlsl INTERFACE_INCLUDE_DIRECTORIES)
+            list(APPEND HEADER_SEARCH_DEFINES MLSL_HEADER_PATH="${MLSL_INCLUDE_DIR}")
+        else()
+            find_package(MPI REQUIRED)
+            target_compile_definitions(codegen PRIVATE NGRAPH_DISTRIBUTED)
```
sfvaroglu: Why is `NGRAPH_DISTRIBUTED` specific to non-MLSL case? I think we should use separate flags to avoid these issues.

src/ngraph/runtime/cpu/builder/allreduce.cpp
```diff
@@ -13,9 +13,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //*****************************************************************************
-#ifdef NGRAPH_DISTRIBUTED
+#ifdef NGRAPH_DISTRIBUTED_ENABLE
```
jianyinglang: Maybe it is better to unify the format of `#ifdef NGRAPH_DISTRIBUTED_ENABLE` or `#if defined NGRAPH_DISTRIBUTED_ENABLE`?

rsketine:
@aslepko here one option is to change our premerge to use the following
CPU_distributed_clang_ubuntu1604.log   to ---> NGRAPH_DISTRIBUTED_MPI_ENABLE
CPU_distributed_gcc ----> to NGRAPH_DISTRIBUTED_MLSL_ENABLE

I am hoping this is ok with all. 

@wenzhe-nrv and @sasadep my assumption is that we do not need to add any lib in our docker file everything is taken care of in nGraph as we did for MLSL.

I see only one test case "distributed_cpu.allreduce" is this check good enough for us to figure out if this is running distributed or not.

sfvaroglu:
@rsketine @wenzhe-nrv @sasadep I may not be up-to-date on this, but do you still have OpenMPI in docker? MLSL downloads and installs any libraries it requires, but this PR uses `find_package` to find OpenMPI installation on the system.

As for "distributed_cpu.allreduce", is it run via `mpirun` on multiple instances? That would be the preferred test case.

rsketine:
@sfvaroglu we had that before but moved away from it and not using it. We would like this lib to be downloaded and setup as we do for mlsl. This is what we use in nGraph would like to get this 3rd party lib setup done internally with specific versions etc so that it is easier for us to maintain.

src/ngraph/codegen/CMakeLists.txt
sasadep: removed the use of `NGRAPH_DISTRIBUTED`

src/ngraph/distributed.cpp
```diff
@@ -14,37 +14,66 @@
...
+    MPI_Initialized(&flag);
+    if (!flag)
+    {
+        MPI_Init(NULL, NULL);
+    }
+#endif
 }
 
-ngraph::Distributed::~Distributed()
+void ngraph::Distributed::finalize()
```
wenzhe-nrv: why don't keep it as a destructor? for RAII

test/distributed.in.cpp
```diff
@@ -31,12 +30,13 @@ using namespace ngraph;
 
 TEST(distributed_${BACKEND_NAME}, allreduce)
 {
+    Distributed dist_instance;
```
jianyinglang: One comment for the unit test is: set the environment variable NGRAPH_CODEGEN=TRUE to check the code from cpu_emitter. The default only checks the code from builder (IIRC).

src/ngraph/distributed.cpp
sasadep: Addressed it for now, yet to see end to end integration

src/ngraph/CMakeLists.txt
```diff
@@ -179,9 +179,14 @@ add_library(ngraph SHARED ${SRC})
 
 if(NGRAPH_DISTRIBUTED_ENABLE)
     target_sources(ngraph PRIVATE distributed.cpp)
-    target_compile_definitions(ngraph PRIVATE NGRAPH_DISTRIBUTED)
-    target_include_directories(ngraph SYSTEM PRIVATE libmlsl)
-    target_link_libraries(ngraph PRIVATE libmlsl)
+    if(NGRAPH_DISTRIBUTED_MLSL_ENABLE)
+        target_include_directories(ngraph SYSTEM PRIVATE libmlsl)
+        target_link_libraries(ngraph PRIVATE libmlsl)
+    else()
```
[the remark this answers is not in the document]
sasadep: I like the idea of error at the end of toolchain! 
@wenzhe-nrv @sfvaroglu @jianyinglang any comments ?

test/main.cpp
```diff
@@ -22,15 +22,15 @@
...
 #endif
 
 int main(int argc, char** argv)
 {
-#ifdef NGRAPH_DISTRIBUTED
-    ngraph::Distributed dist;
-#endif
+    // #ifdef NGRAPH_DISTRIBUTED_ENABLE
+    //     ngraph::Distributed dist;
+    // #endif
```
indie: @wenzhe-nrv  Does this change mean that we can update [this documentation](http://ngraph.nervanasys.com/docs/latest/howto/distribute-train.html) to say "The default for making an nGraph build with the `DNGRAPH_DISTRIBUTED_ENABLE=TRUE` compilation flag will use two or more devices."  ?  

Or something similar.

src/ngraph/CMakeLists.txt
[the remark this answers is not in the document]
sfvaroglu: sounds good. create `*_OPENMPI_ENABLE` for the `elif` case to make it clear.

test/main.cpp
```diff
@@ -22,15 +22,15 @@
...
+#ifdef NGRAPH_DISTRIBUTED_ENABLE
 #include "ngraph/distributed.hpp"
 #endif
 
 int main(int argc, char** argv)
 {
-#ifdef NGRAPH_DISTRIBUTED
-    ngraph::Distributed dist;
-#endif
+    // #ifdef NGRAPH_DISTRIBUTED_ENABLE
```
sfvaroglu: how do you test `distributed` w/o this?

src/ngraph/CMakeLists.txt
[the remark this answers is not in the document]
sasadep: @aslepko likes having separate flag for each library

test/main.cpp
sasadep: was testing CI job

src/ngraph/CMakeLists.txt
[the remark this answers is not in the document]
sasadep: Addressed this comment of using separate flags

src/ngraph/codegen/CMakeLists.txt
```diff
@@ -47,9 +47,13 @@ if ((NGRAPH_GPU_ENABLE OR NGRAPH_CPU_ENABLE) AND NOT NGRAPH_DEX_ONLY)
     list(APPEND HEADER_SEARCH_DEFINES NGRAPH_HEADERS_PATH="${NGRAPH_INCLUDE_PATH}")
 
     if(NGRAPH_DISTRIBUTED_ENABLE)
-        get_target_property(MLSL_INCLUDE_DIR libmlsl INTERFACE_INCLUDE_DIRECTORIES)
-        list(APPEND HEADER_SEARCH_DEFINES MLSL_HEADER_PATH="${MLSL_INCLUDE_DIR}")
-        add_definitions(-DNGRAPH_DISTRIBUTED)
+        if (NGRAPH_DISTRIBUTED_MLSL_ENABLE)
+            get_target_property(MLSL_INCLUDE_DIR libmlsl INTERFACE_INCLUDE_DIRECTORIES)
+            list(APPEND HEADER_SEARCH_DEFINES MLSL_HEADER_PATH="${MLSL_INCLUDE_DIR}")
+        else()
```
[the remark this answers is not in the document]
sasadep: made changes to use standalone flags and throw an error when a library flag is not mentioned

CMakeLists.txt
```diff
@@ -123,6 +123,20 @@ if (NGRAPH_GPUH_ENABLE)
...
 endif()
 
+if (NGRAPH_DISTRIBUTED_ENABLE)
+    option(NGRAPH_DISTRIBUTED_MLSL_ENABLE "Add distributed mlsl mode to the for CPU only backend" FALSE)
+    option(NGRAPH_DISTRIBUTED_OMPI_ENABLE "Add distributed open-MPI mode for all backend" FALSE)
+    if ( (NOT NGRAPH_DISTRIBUTED_MLSL_ENABLE) AND (NOT NGRAPH_DISTRIBUTED_OMPI_ENABLE) )
+    message(FATAL_ERROR "Distributed Library not mentioned use NGRAPH_DISTRIBUTED_MLSL_ENABLE for mlsl or NGRAPH_DISTRIBUTED_OMPI_ENABLE for open MPI")
+    endif()
+endif()
+if (NGRAPH_DISTRIBUTED_MLSL_ENABLE)
```
sfvaroglu: Throw an error if both MLSL and OMPI flags are set.

CMakeLists.txt
```diff
@@ -265,6 +281,16 @@ if (NGRAPH_PLAIDML_ENABLE)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNGRAPH_PlaidML_ENABLE")
 endif()
 
+if (NGRAPH_DISTRIBUTED_ENABLE)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNGRAPH_DISTRIBUTED_ENABLE")
+    if (NGRAPH_DISTRIBUTED_MLSL_ENABLE)
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNGRAPH_DISTRIBUTED_MLSL_ENABLE")
+    endif()
+    if (NGRAPH_DISTRIBUTED_OMPI_ENABLE)
```
sfvaroglu: This should be elseif.

CMakeLists.txt
```diff
@@ -341,11 +367,11 @@ if (WIN32 OR APPLE)
 else()
     include(cmake/external_tbb.cmake)
 endif()
-if (NGRAPH_DISTRIBUTED_ENABLE)
+
+if (NGRAPH_DISTRIBUTED_ENABLE AND NGRAPH_DISTRIBUTED_MLSL_ENABLE)
```
sfvaroglu: Do you need to check NGRAPH_DISTRIBUTED_ENABLE? Looks like you set it to true if NGRAPH_DISTRIBUTED_MLSL_ENABLE is set.

src/ngraph/codegen/compiler.cpp
```diff
@@ -472,8 +472,14 @@ void codegen::CompilerCore::configure_search_path()
     add_header_search_path(CUDNN_HEADER_PATHS);
 #endif
 
-#ifdef NGRAPH_DISTRIBUTED
+#ifdef NGRAPH_DISTRIBUTED_ENABLE
```
sfvaroglu: Not needed? The same comment applies to the rest of the files.

sasadep: if that's done then you will throw an error when not in distributed mode

src/ngraph/runtime/cpu/builder/allreduce.cpp
sasadep: staying with `#ifdef`

test/main.cpp
sasadep: @indie yes, the documentation  would have to change 
now the user enables the flags based on the library of choice 
i.e. 
* the use of `NGRAPH_DISTRIBUTED_OMPI_ENABLE` enables OpenMPI
* the use of `NGRAPH_DISTRIBUTED_MLSL_ENABLE`  enables MLSL.

Note: this choice applies to `CPU` and `Interpreter` backend only
 For all other backends we support only `OpenMPI` at present. 
We can't use both the libraries simultaneously at present. 
We recommend the use of `Intel  MLSL` if the user is using CPU only backend.

test/backend_binary_elementwise.in.cpp
```diff
@@ -242,8 +242,11 @@ NGRAPH_TEST(${BACKEND_NAME}, divide_by_zero_float32)
                              std::numeric_limits<float>::infinity()}),
               read_vector<float>(result));
 }
-
+#ifdef NGRAPH_DISTRIBUTED_OMPI_ENABLE
```
sasadep: https://jira.devtools.intel.com/browse/NGCPU-343 to be resolved

jianyinglang:
LGTM except minor comments: you might want to change `#ifdef NGRAPH_DISTRIBUTED` to `#ifdef NGRAPH_DISTRIBUTED_ENABLE` in file `src/ngraph/runtime/generic_cpu/gcpu_backend.hpp` since `NGRAPH_DISTRIBUTED` is no longer defined. Also the file `doc/examples/mnist_mlp/CMakeLists.txt` needs to be updated with the new flags.

src/ngraph/distributed.cpp
```diff
@@ -14,37 +14,90 @@
...
+    if (this_init_comm == true)
+    {
+        finalize();
+    }
+}
+
+void ngraph::Distributed::finalize()
+{
+#ifdef NGRAPH_DISTRIBUTED_MLSL_ENABLE
     if (MLSL::Environment::GetEnv().IsInitialized())
```
AlekseyMarchuk: We could rely on `this_init_comm` flag and skip library specific check, e.g.
```
void ngraph::Distributed::finalize()
{
    if (this_init_comm)
    {
#ifdef NGRAPH_DISTRIBUTED_MLSL_ENABLE
        MLSL::Environment::GetEnv().Finalize();
#elif NGRAPH_DISTRIBUTED_OMPI_ENABLE
        MPI_Finalize();
#endif
    }
}
```

src/ngraph/runtime/generic_cpu/gcpu_backend.hpp
```diff
@@ -139,7 +139,7 @@
 #include "ngraph/runtime/tensor.hpp"
 #include "ngraph/state/rng_state.hpp"
 
-#ifdef NGRAPH_DISTRIBUTED
+#ifdef NGRAPH_DISTRIBUTED_ENABLE
```
sasadep: @jianyinglang good catch !

src/tools/nbench/nbench.cpp
```diff
@@ -298,8 +298,14 @@ OPTIONS
...
 
-#ifdef NGRAPH_DISTRIBUTED
-    ngraph::Distributed dist;
+#if defined NGRAPH_DISTRIBUTED_ENABLE
+    auto dist = new ngraph::Distributed;
+    bool delete_instance = true;
+    if (dist->get_size() == 1)
+    {
+        delete dist;
+        delete_instance = false;
```
AlekseyMarchuk: I would suggest a small improvement - either use smart pointer e.g. unique_ptr or set dist to nullptr if distribution size is equal to 1, e.g.
```
if (dist->get_size() == 1)
{
    delete dist;
    dist = nullptr;
}
```
Later we could delete `dist` in condition `if(dist) delete dist;`

sasadep: done !

src/ngraph/distributed.cpp
sasadep: I would like to leave this as is, for additional reliability :)

src/ngraph/runtime/reference/allreduce.hpp
```diff
@@ -16,10 +16,12 @@
 
 #pragma once
 
-#ifdef NGRAPH_DISTRIBUTED
-
+#ifdef NGRAPH_DISTRIBUTED_ENABLE
+#ifdef NGRAPH_DISTRIBUTED_MLSL_ENABLE
```
avijit-nervana: Instead of using MPI/MLSL both in this file - create two files - one for MPI and one for MLSL. Then include them based on the specific transport.

jianyinglang:
LGTM

src/ngraph/CMakeLists.txt
```diff
@@ -178,9 +179,16 @@ add_library(ngraph SHARED ${SRC})
 
 if(NGRAPH_DISTRIBUTED_ENABLE)
     target_sources(ngraph PRIVATE distributed.cpp)
-    target_compile_definitions(ngraph PRIVATE NGRAPH_DISTRIBUTED)
-    target_include_directories(ngraph SYSTEM PRIVATE libmlsl)
-    target_link_libraries(ngraph PRIVATE libmlsl)
+    if(NGRAPH_DISTRIBUTED_MLSL_ENABLE)
+        target_include_directories(ngraph SYSTEM PRIVATE libmlsl)
```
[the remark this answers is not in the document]
sasadep: We build [MLSL](https://github.com/NervanaSystems/ngraph/blob/sandeep/distributed-mlsl-openmpi/cmake/external_mlsl.cmake) from source which is the only option at present, that mentioned [here](https://github.com/NervanaSystems/ngraph/blob/sandeep/distributed-mlsl-openmpi/CMakeLists.txt#L368) 

OpenMPI we will be adding support for building from source option in the consequent PR's

src/ngraph/runtime/reference/allreduce.hpp
sasadep: I'll leave it as is for now (as it unifies the design on this PR), address this part of refactoring for this file and others in the consequent PR.

sasadep:
@avijit-nervana 
@indie's PR #2417 address you comment on prerequisites If any.  
We'll be adding building from source option support for OpenMPI in the consequent PR's.

src/ngraph/distributed.cpp
```diff
@@ -14,37 +14,90 @@
...
+    int size;
+    MPI_Comm_size(MPI_COMM_WORLD, &size);
+    return size;
+#else
+    throw ngraph_error("Distributed Library not supported/mentioned");
+#endif
 }
 
-size_t ngraph::Distributed::get_rank() const
+int ngraph::Distributed::get_rank() const
```
avijit-nervana: Is rank ever going to be negative? If not - why not keeping size_t?

src/ngraph/distributed.cpp
```diff
@@ -14,37 +14,90 @@
...
 {
-    return MLSL::Environment::GetEnv().GetProcessIdx();
+#ifdef NGRAPH_DISTRIBUTED_MLSL_ENABLE
+    return static_cast<int>(MLSL::Environment::GetEnv().GetProcessIdx());
```
avijit-nervana: Is the meaning of MLSL same as MPI rank?

src/ngraph/distributed.cpp
```diff
@@ -14,37 +14,90 @@
 // limitations under the License.
 //*****************************************************************************
 
-#ifdef NGRAPH_DISTRIBUTED
+#ifdef NGRAPH_DISTRIBUTED_ENABLE
```
avijit-nervana: Consider splitting this into multiple files - one for each type of transport (i.e., MLSL vs. MPI)

sasadep:
Thanks @avijit-nervana
addressing these comments as a part of [ticket](https://jira.devtools.intel.com/browse/NGCORE-318)

PR #2355, opened 2019-01-23
Update PR to nGraph master.

arogowie-intel:
Moreover add more meaningfull names to models and enable running on multiple backends.

Except last 4 commits everything else is already on master branch.
Now run with:
```
./test/unit-test --gtest_filter=*onnx*CPU*model_resnet50*
```


arogowie-intel:
Limited the list of reviewers since most of the code is from merging to master.

PR #2357, opened 2019-01-23
IntelGPU backend: Min and Max operations fix

shssf:


src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -58,6 +58,56 @@ string runtime::intelgpu::get_opencl_type_name(const element::Type& ngraph_type)
     }
 }
 
+string runtime::intelgpu::get_opencl_type_min_max_value(const element::Type& ngraph_type,
+                                                        bool is_min)
+{
+    if (ngraph_type == ngraph::element::f32)
```
rkimballn1: element::Type has a new method get_type_enum() which returns an enum::Type_t which can be used in a switch statement. You can see an example use here https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/interpreter/int_backend.cpp#L252

PR #2359, opened 2019-01-24
[ONNX] TopK operator

tsocha:


src/ngraph/frontend/onnx_import/ops_bridge.cpp
```diff
@@ -94,6 +94,8 @@
 #include "op/tan.hpp"
 #include "op/tanh.hpp"
 #include "op/thresholded_relu.hpp"
+#include "op/topk.cpp"
```
arogowie-intel: You probably would not want this here. Please remove ;)

src/ngraph/frontend/onnx_import/op/topk.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace ngraph
+{
+    namespace onnx_import
+    {
+        namespace op
+        {
+            namespace set_1
+            {
+                NodeVector topk(const Node& node);
```
arogowie-intel: Some simple documentation would be helpfull.

src/ngraph/frontend/onnx_import/op/topk.cpp
```diff
@@ -0,0 +1,64 @@
...
+        {
+            namespace set_1
+            {
+                NodeVector topk(const Node& node)
+                {
+                    auto data = node.get_ng_inputs().at(0);
+                    std::int64_t axis{node.get_attribute_value<std::int64_t>("axis", -1)};
+                    std::int64_t k{node.get_attribute_value<std::int64_t>("k")};
+
+                    if (axis < 0)
```
arogowie-intel: You may use: `convert_negative_axis` https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/frontend/onnx_import/utils/common.hpp#L149 here.

src/ngraph/frontend/onnx_import/op/topk.cpp
```diff
@@ -0,0 +1,64 @@
...
+                    {
+                        axis += data->get_shape().size();
+                    }
+
```
arogowie-intel: Still you need here `ASSERT_VALID_ARGUMENT` to check whether `axis` is not negative.

src/ngraph/frontend/onnx_import/op/topk.cpp
```diff
@@ -0,0 +1,64 @@
...
+                    std::shared_ptr<ngraph::Node> top_k =
+                        std::make_shared<ngraph::op::TopK>(data, axis, element::i64, k);
+
+                    std::shared_ptr<ngraph::Node> indices =
+                        std::make_shared<ngraph::op::GetOutputElement>(top_k, 0);
```
arogowie-intel: I'm not sure whether this is correct? Please add some UT similar to this: https://github.com/onnx/onnx/blob/master/docs/Operators.md#examples-104

src/ngraph/frontend/onnx_import/op/topk.cpp
```diff
@@ -0,0 +1,64 @@
...
+                    std::shared_ptr<ngraph::Node> values =
+                        std::make_shared<ngraph::op::GetOutputElement>(top_k, 1);
```
arogowie-intel: Does nGraph's `TopK` returns both values and indices?

tsocha: Yes, but documentation in .cpp file don't say it and Python API does.

src/ngraph/frontend/onnx_import/op/topk.cpp
tsocha: I don't need assert because axis is checked in if statement and default value for axis is -1.

src/ngraph/frontend/onnx_import/op/topk.cpp
arogowie-intel: Still we do not have one in our nGraph repository. If onnx backend tests output correctness then you may always check whether we behave as we expect for incorrect values.

src/ngraph/frontend/onnx_import/op/topk.cpp
tsocha: I agree that I should check if axis > "number of axis", but this function take the last one.
I will add an assert to check it.

tsocha:
> Please add UT.

This code is tested with tests from `ngraph-onnx` there is no need to duplicate tests.

src/ngraph/frontend/onnx_import/op/topk.cpp
tsocha: These tests are integrated with CI process another ut is not needed in my opinion.

arogowie-intel:
> 
> This code is tested with tests from `ngraph-onnx` there is no need to duplicate tests.

But I can find only single simple test in onnx backend tests where there is used default `axis` attribute. And you may at least make a test where you provide all attributes, not to mention edge values for them.

src/ngraph/frontend/onnx_import/op/topk.cpp
arogowie-intel: What if someone pass value `-1000000` ? will this still be correct?

src/ngraph/frontend/onnx_import/op/topk.cpp
tsocha: Truncate provided axis means the same as take the last one because axis is used to select axis.

src/ngraph/frontend/onnx_import/op/topk.cpp
tsocha: I fixed this case in last commit(check range after adding value).

PR #2360, opened 2019-01-24
Add error check for user setting CMAKE_BUILD_TYPE

rkimballn1:


CMakeLists.txt
```diff
@@ -27,6 +27,12 @@ if("${CMAKE_GENERATOR}" STREQUAL "Ninja")
     message(FATAL_ERROR "Ninja generator not supported.")
 endif()
 
+set(RELEASE_TYPES Debug Release RelWithDebInfo MinSizeRel)
+list(FIND RELEASE_TYPES ${CMAKE_BUILD_TYPE} INDEX_FOUND)
+if (${INDEX_FOUND} EQUAL -1)
```
silee2: CMAKE_BUILD_TYPE could be empty which is the default value. Build will fail.

rkimballn1: Thanks for catching that. I added a check to only perform this if CMAKE_BUILD_TYPE is set.

PR #2361, opened 2019-01-24
Convert PlaidML Tile op to generic ngraph passthrough op

earhart:


earhart:
I've updated this to provide a general passthrough operation, instead of a Tile-specific operation -- there was a request on the PlaidML side for supporting arbitrary languages.  I've also added serialization support, and changed the class to derive from `Op`.

earhart:
N.B. This should go in after #2445; that PR isn't actually part of this one.

PR #2363, opened 2019-01-24
Refactor to create MKLDNN primitives on the first iteration:

ayzhuang:
add, avg_pool, batch_norm, bounded_relu, concat, convert_layout,
  leaky_relu, lrn, max_pool, quantized_avg_pool, quantized_max_pool,
  relu, sigmoid, slice, softmax.


src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -601,6 +968,7 @@ namespace ngraph
                     {
                         return 4;
                     }
+                    return -1;
```
jbobba: shouldn't this be an assert?

src/ngraph/runtime/cpu/builder/avg_pool.cpp
```diff
@@ -130,23 +124,23 @@ namespace ngraph
...
-                        apb->get_padding_above());
-
+                    auto avg_pool_fwd_desc =
+                        mkldnn_emitter->get_avg_pooling_forward_desc<ngraph::op::AvgPoolBackprop>(
+                            node, true);
+                    auto avg_pool_desc =
+                        mkldnn_emitter->get_avg_pooling_backward_desc<ngraph::op::AvgPoolBackprop>(
+                            node);
+
+                    size_t avg_pool_index = mkldnn_emitter->primitive_init(3);
```
dcaballe: How is the 'count' value determined? If there is an op->constant correspondence, maybe we could use an enum to make the code more auto-descriptive and group all the values together.

src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -823,6 +1278,166 @@ namespace ngraph
...
+                    {
+                        window_dilation_strides_adjusted.push_back(s - 1);
+                    }
+
+                    auto weights_desc = mkldnn_utils::get_input_mkldnn_md(node, 0);
+                    // MKLDNN relies on named formats for kernel selection
+                    if (weights_desc.data.format == mkldnn_nchw)
+                        weights_desc.data.format = mkldnn_oihw;
+                    if (weights_desc.data.format == mkldnn_ncdhw)
+                        weights_desc.data.format = mkldnn_oidhw;
```
dcaballe: should we use curly braces here, according to coding standard or single line then/else/for are ok without them?

src/ngraph/runtime/cpu/mkldnn_emitter.hpp
dcaballe: do we have something like llvm_unreachable ([link](https://llvm.org/docs/ProgrammersManual.html#programmatic-errors)) to document regions of code that should never be executed, like this one?

src/ngraph/runtime/cpu/builder/avg_pool.cpp
pruthviIntel: @ayzhuang i guess can we add utility function for getting the count of the number of primitives_desc? . we know how many primitive desc we need in the `get_<mkldnn>_desc()` . can we maintain another private data structure  `m_primitive_desc_count` . which uniquely index to primitive and its `count of number of desc` ?.

ayzhuang: Yes, we could add a utility function to return the number of primitives needed for each op based on op type. I don't think we need another private data structure since this number is only needed once.

src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -14,6 +14,10 @@
 // limitations under the License.
 //*****************************************************************************
 
+// For direct execution, we reserve space for primitives then create those primitives the first
+// time functor is called. This could be extended to create primitives when shapes are changed.
+// Different ops need different numbers of primitives.
+
```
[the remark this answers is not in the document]
ayzhuang: codegen will be modified in another PR, maybe we should wait till then to add comments.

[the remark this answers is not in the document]
dcaballe: Sure, no problem

PR #2364, opened 2019-01-24
Pad changes to support different modes

gaurides:


gaurides:
accidentally created against master.

PR #2368, opened 2019-01-25
Fuse Q->QCB->DQ->RELU to QCBR op

nbpatel:

### DIFF
--- a/src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
+++ b/src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
@@ -30,9 +30,12 @@
 #include "ngraph/op/concat.hpp"
 #include "ngraph/op/constant.hpp"
 #include "ngraph/op/convolution.hpp"
+#include "ngraph/op/dequantize.hpp"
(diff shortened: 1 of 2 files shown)


src/ngraph/op/experimental/quantized_conv.hpp
```diff
@@ -33,7 +33,9 @@ namespace ngraph
                                  const CoordinateDiff& padding_below,
                                  const CoordinateDiff& padding_above,
                                  const Strides& data_dilation_strides,
-                                 const std::shared_ptr<Node> scale);
+                                 const std::shared_ptr<Node> scale,
+                                 const std::shared_ptr<Node> input_scale = nullptr,
```
aprocter: Does this really work when `input_scale == nullptr`? In the ctor implementation over in `quantized_conv.cpp` you are going directly to calling `check_single_output_args` on this, and that function does not contain any `nullptr` checks.

aprocter:
Just realized this is not going into `master`.

postrational:
@adstraw Should we merge this into `arogowiec/lin_quant` for tests?

PR #2369, opened 2019-01-25
[WIP] Merge core quant fuse patterns into onnx_import

nivas-x86:
WIP for testing quant patterns


postrational:
Why is this PR made against `tsocha/lin_quant`?
Shouldn't we be working on `arogowiec/lin_quant`?

@tsocha Can we delete `tsocha/lin_quant`?

PR #2378, opened 2019-01-27
New Backend API attempt #3

rkimballn1:
This is a vastly scaled-down PR of #2273, which encompassed 92 files with +5,629 and -4,803 changes.
This PR on the other hand currently modifies only 31 files with +649 and -440 changes thanks mostly to the suggestion  of @shssf to do all I can to leave lines where they are. Most of the changes in the old PR were from moving lines, not changing.

Motivations:

Currently the Backend uses a `shared_ptr<Function>` as the handle passed to the `call` method to execute a compiled object. This PR changes the `compile` function to return a `shared_ptr<Executable>` which is an object with methods to `validate`, `call`, `get_performance_data`, etc. Adding an Executable object allows for things like saving/loading an executable to/from storage. It also creates a clear line of when a Function is compiled, making the internals of the Backend and Executable easier to implement.

This is the major change for the new Backend API. More PRs will follow but those can be much smaller and localized to single backends.


src/ngraph/runtime/intelgpu/intelgpu_backend.hpp
```diff
@@ -68,12 +86,19 @@ class ngraph::runtime::intelgpu::IntelGPUBackend : public runtime::Backend
...
         double m_compilation_time = 0.0;
         double m_consumed_memory = 0.0;
-    };
-
-    std::map<std::shared_ptr<Function>, FunctionInstance> ocl_networks;
-    std::shared_ptr<cldnn::engine> ocl_engine;
+        std::shared_ptr<Function> m_function;
+    } m_function_instance;
 
+    bool m_profile_enable = false;
```
shssf: Variables are duplicated in `Backend` and `Executable`. But this is not a problem for this PR. I will take care about this later.
Please proceed with main idea of the PR.

rkimballn1: Thank you. Yes, it is not an ideal solution but I believe you can see the path to get to a better solution. We can do that as individual, backend focused PRs.

src/ngraph/runtime/backend.cpp
```diff
@@ -39,78 +39,127 @@ vector<string> runtime::Backend::get_registered_devices()
...
-    // then override this method and enhance.
     return false;
 }
+
+void runtime::Backend::remove_compiled_function(std::shared_ptr<Executable> exec)
+{
+}
+
+bool runtime::Backend::call_with_validate(
+    std::shared_ptr<Executable> exec,
```
dcaballe: Should this be a reference or a raw pointer or are we extending the ownership of `exec`?

src/ngraph/runtime/backend.hpp
```diff
@@ -133,8 +99,64 @@ class ngraph::runtime::Backend
...
+    /// They will be removed in a follow-on PR
+    bool call_with_validate(std::shared_ptr<Executable> handle,
+                            const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
+                            const std::vector<std::shared_ptr<runtime::Tensor>>& inputs);
+    bool call_with_validate(const std::unique_ptr<Executable>& handle,
+                            const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
+                            const std::vector<std::shared_ptr<runtime::Tensor>>& inputs);
+};
+
+class ngraph::runtime::Executable
```
dcaballe: It would be great if you could add part of the PR description to this class as doxygen doc

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -114,35 +122,27 @@ bool runtime::cpu::CPU_Backend::call(shared_ptr<Function> func,
     return rc;
 }
 
-void runtime::cpu::CPU_Backend::remove_compiled_function(shared_ptr<Function> func)
-{
-    m_function_map.erase(func);
-}
-
-void runtime::cpu::CPU_Backend::enable_performance_data(shared_ptr<Function> func, bool enable)
+void runtime::cpu::CPU_Backend::remove_compiled_function(shared_ptr<Executable> exec)
```
dcaballe: Same here. It looks like we want a reference since we are not extending the ownership.

src/ngraph/runtime/gpu/gpu_backend.cpp
```diff
@@ -175,33 +192,18 @@ bool runtime::gpu::GPU_Backend::call(shared_ptr<Function> func,
...
-void runtime::gpu::GPU_Backend::enable_performance_data(shared_ptr<Function> func, bool enable)
-{
-    FunctionInstance& instance = m_function_map[func];
-    if (instance.m_compiled_function != nullptr)
-    {
-        throw runtime_error("Performance data collection must be enabled prior to compiling.");
-    }
-    instance.m_performance_counters_enabled = enable;
-}
+// void runtime::gpu::GPU_Backend::remove_compiled_function(shared_ptr<Function> func)
```
dcaballe: forgot to remove?

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -1879,41 +1911,32 @@ bool runtime::intelgpu::IntelGPUBackend::call(shared_ptr<Function> func,
...
-        remove_compiled_function(func);
+        m_compilation_time = 0.0;
+        m_consumed_memory = 0.0;
     }
 
     return true;
 }
 
-void runtime::intelgpu::IntelGPUBackend::remove_compiled_function(shared_ptr<Function> func)
+void runtime::intelgpu::IntelGPUBackend::remove_compiled_function(shared_ptr<Executable> exec)
```
dcaballe: Same. reference?

PR #2386, opened 2019-01-31
[ONNX] Add null node.

tsocha:
Optional inputs in ONNX standard are represented by empty string.
We need a placeholder to keep information which inputs are not provided.


src/ngraph/frontend/onnx_import/core/node.cpp
```diff
@@ -122,7 +123,14 @@ namespace ngraph
             NodeVector result;
             for (const auto& name : m_node_proto->input())
             {
-                result.push_back(m_graph->get_ng_node_from_cache(name));
+                if (name != "")
```
arogowie-intel: If this is `string` consider using `!name.empty()`

tsocha:
> 
> 
> I think you also need a function like `virtual bool Node::is_null_node() const` to be able to check whether we are dealing with this class.

@diyessi what do you think about it?

diyessi:
I like that better than checking for the class type.

src/ngraph/frontend/onnx_import/core/null_node.hpp
```diff
@@ -25,6 +25,7 @@ namespace ngraph
 {
     namespace onnx_import
     {
+        /// \brief Placeholder for expected but missing node in graph.
```
arogowie-intel: This is kind of misleading. We won't put anything in place of this node, are we? If it is expected then decidedly it is an error if it is missing.

postrational: Please use a docstring such as this:

> Represents a missing optional input or output of an ONNX node
> 
> Some ONNX operators have inputs or outputs that are marked as optional, which means that a referring node MAY forgo providing values for such inputs or computing these outputs. An empty string is used in place of a name of such input or output.
> 
> See:
> https://github.com/onnx/onnx/blob/master/docs/IR.md#optional-inputs-and-outputs
>

src/ngraph/frontend/onnx_import/core/null_node.hpp
```diff
@@ -0,0 +1,48 @@
...
+        /// An empty string is used in place of a name of such input or output.
+        ///
+        /// More:
+        /// https://github.com/onnx/onnx/blob/master/docs/IR.md#optional-inputs-and-outputs
+        class NullNode : public ngraph::Node
+        {
+        public:
+            NullNode();
+
+            virtual bool is_null() const { return true; }
```
arogowie-intel: Please add `override` keyword, see [here why](https://stackoverflow.com/questions/13880205/is-the-override-keyword-just-a-check-for-a-overridden-virtual-method). Moreover the `final` keyword also is suitable. If you let anything derive from `NullNode` then you probably don't want it to change it's behavior of `is_null`, since it no longer will be a `NullNode`.

tsocha:
> @diyessi 
> Please add a comment to `null_node.hpp` explaining what this is for.

Comment has been added.

src/ngraph/frontend/onnx_import/core/null_node.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "ngraph/node_vector.hpp"
+
+namespace ngraph
+{
+    namespace onnx_import
+    {
+        NullNode::NullNode()
+            : ngraph::Node("NullNode", {}, 0)
+        {
+            m_name = "NullNode";
```
rkimballn1: Overriding the Node's m_name in this way is a bad idea. Why is this done? A comment why would be nice if this is really necessary.

tsocha: Good point I will remove it.

PR #2388, opened 2019-01-31
[WIP] [ONNX Runtime] Quantize operators

postrational:


arogowie-intel:
If we're merging to master, then maybe we should get rid of all those binary data files?

postrational:
> @postrational If we're merging to master, then maybe we should get rid of all those binary data files?

We are only merging parts as separate PRs, this WIP is just a reference.

src/ngraph/op/experimental/quantized_conv_bias.hpp
```diff
@@ -40,7 +40,8 @@ namespace ngraph
                                      const CoordinateDiff& padding_above,
                                      const Strides& data_dilation_strides,
                                      const std::shared_ptr<Node> scale,
-                                     const bool with_relu = false);
+                                     const bool with_relu = false,
+                                     const std::shared_ptr<Node> output_scale = nullptr);
```
aprocter: Won't this simply crash if called with `output_scale==nullptr`? (The constructor calls `check_single_output_args` on a vector including `output_scale`, which should deref the null pointer.) If so, we should remove the default value.

src/ngraph/op/experimental/quantized_conv.hpp
```diff
@@ -33,7 +33,10 @@ namespace ngraph
                                  const CoordinateDiff& padding_below,
                                  const CoordinateDiff& padding_above,
                                  const Strides& data_dilation_strides,
-                                 const std::shared_ptr<Node> scale);
+                                 const std::shared_ptr<Node> scale,
+                                 const std::shared_ptr<Node> input_scale = nullptr,
```
aprocter: Similar issue to my other comment: I don't think we should have a default value of `nullptr` here, because this is just going to segfault once `check_single_output_args` is called.

adstraw:
all relevant code is now in master via other PRs

PR #2389, opened 2019-01-31
upgraded MKLDNN version to V0.17.2

pruthviIntel:


cmake/external_mkldnn.cmake
```diff
@@ -49,7 +49,7 @@ endif()
 
 # This section sets up MKL as an external project to be used later by MKLDNN
 
-set(MKLURLROOT "https://github.com/intel/mkl-dnn/releases/download/v0.17/")
+set(MKLURLROOT "https://github.com/intel/mkl-dnn/releases/download/v0.17.2/")
 set(MKLVERSION "2019.0.1.20180928")
```
silee2: You could use 2019.0.1.20181227
Any reason for using an older version?

silee2:
Just noticed that 0.17.3 is sticking to `2019.0.1.20180928`

PR #2396, opened 2019-02-05
handle channelwise scale mask

ashokei:


src/ngraph/runtime/cpu/builder/quantization.cpp
```diff
@@ -73,7 +73,8 @@ namespace ngraph
                                 vector<float> dyn_scales;
                                 dyn_scales.assign(static_cast<float*>(arg1_tensor),
                                                   static_cast<float*>(arg1_tensor) + scales_size);
-                                attr.set_output_scales(0, dyn_scales);
+                                const int mask = scales_size == 1 ? 0 : 1;
```
jbobba: If we are dequantizing an {N, C, H, W} data tensor, channel is second dimension. So per-channel would require a mask of 2^1 = 2.  Do we need to add the per-channel scaling dimension as an attribute to the op?

jbobba: At the very least, we need to add a comment/note on why this needs to be '1' and '2' in the conv op.

ashokei: currently , mkldnn applies mask according to "logical" dims, and conv.ops its always nchw, so 2 makes sense.

Above mask 1 is for `quantize` op, which we expect to use with no-channels (mask 0) or with-channels (primarily for weights/bias, which use oihw format), since channel is across dim 0 (2^0=1) we use that here. 

I recall nishant mentioned he oihw expected for weights.

https://intel.github.io/mkl-dnn/ex_int8_simplenet.html (search for mask)

PR #2398, opened 2019-02-06
Fix cmake TBB lib path on Windows.

jfdong:


rkimballn1:
This is the original fork for what is now #2415

PR #2406, opened 2019-02-08
 [ONNX] Support for initializers without corresponding inputs

postrational:
Add support for this change in ONNX:
> ONNX dropped the constraint that the initializers must be a subset of graph inputs. An initializer is allowed to have the same name as a graph-input, but no longer required to be a graph-input.

https://github.com/onnx/onnx/pull/1718


src/ngraph/frontend/onnx_import/core/tensor.hpp
```diff
@@ -387,6 +388,43 @@ namespace ngraph
...
+                case onnx::TensorProto_DataType::TensorProto_DataType_UINT32:
+                    return make_ng_constant<uint32_t>(element::u32);
+                case onnx::TensorProto_DataType::TensorProto_DataType_UINT64:
+                    return make_ng_constant<uint64_t>(element::u64);
+                default: throw error::tensor::unsupported_data_type{m_tensor_proto->data_type()};
+                }
+            }
+
+            template <typename T>
+            std::shared_ptr<ngraph::op::Constant> make_ng_constant(const element::Type& type) const
```
arogowie-intel: This function should be at least `protected`.

PR #2408, opened 2019-02-08
[ONNX] Enhance LSTM support.

arogowie-intel:
Added support for features:
- `clip` attribute
- `activations` attribute - currently supported only required `sigmoid`, `tanh` and `relu` functions.
- `input_forget` - coupling input and forget gates.
- `direction` - supported `forward`, `reverse` and `bidirectional` modes.


src/ngraph/frontend/onnx_import/op/lstm.cpp
```diff
@@ -55,23 +64,63 @@ namespace ngraph
...
+                }
+
+                std::string& to_lower_case(std::string&& s)
+                {
+                    std::transform(std::begin(s), std::end(s), std::begin(s), ::tolower);
+                    return s;
+                }
 
-                std::shared_ptr<ngraph::Node> sigmoid(const std::shared_ptr<ngraph::Node>& arg)
+                std::string& to_lower_case(std::string& s)
```
rkimballn1: There is already a to_lower function in src/ngraph/util.hpp. Use that one.

src/ngraph/frontend/onnx_import/op/lstm.cpp
```diff
@@ -55,23 +64,63 @@ namespace ngraph
...
                 {
-                    return std::make_shared<ngraph::op::Sigmoid>(arg);
+                    std::transform(std::begin(s), std::end(s), std::begin(s), ::tolower);
+                    return s;
                 }
 
-                std::shared_ptr<ngraph::Node> tanh(const std::shared_ptr<ngraph::Node>& arg)
+                std::vector<std::string>& to_lower_case(std::vector<std::string>&& vs)
```
rkimballn1: If you are modifying the source vector then why return a vector? You should add a comment about the fact that this modifies the source vector.

arogowie-intel:
This PR needs #2386 to work with missing optional inputs for LSTM.

src/ngraph/frontend/onnx_import/op/lstm.cpp
arogowie-intel: My intent was:
* to modify input vector (which will be a returned value from other function) in-place
* return a reference to modified vector, since it will be used as initializer for struct member in constructor initializer list. See here:
https://github.com/NervanaSystems/ngraph/blob/985c7d6f4713263bde98293bc618c59891c56292/src/ngraph/frontend/onnx_import/op/lstm.cpp#L259-L261
* Avoid copying of modified vector.

arogowie-intel:
Need PR #2434 to pass ONNX backend tests.

PR #2409, opened 2019-02-08
Fix memory leak

gaurides:
Running BERT training model on CPU, showed memory leak. 
There was memory leak in cblas::cblas_sgemm. 
Following this link to free the buffers and fix the memory leak https://software.intel.com/en-us/mkl-linux-developer-guide-avoiding-memory-leaks-in-intel-mkl.


src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -36,6 +36,7 @@ MKLDNNEmitter::~MKLDNNEmitter()
 {
     for (auto p : m_mkldnn_primitives)
         delete p;
+    mkldnn_utils::mkl_serv_free_buffers();
```
pruthviIntel: @gaurides Can you please add a description , explaining  the role of `serv_free_buffers` ?

PR #2411, opened 2019-02-09
Clean up cmake project options

postrational:
Move `find_package(PlaidML CONFIG)` inside `if (NGRAPH_PLAIDML_ENABLE)` to avoid warning when PlaidML is not installed.


rkimballn1:
I'm not sure we want the behavior to be automatic building. We don't do this for other backends like the nvidia GPU. I dragged @diyessi in here to get his opinion.

earhart:
FWIW -- I don't have a strong feeling about this, but it was a deliberate decision (mimicking the on-by-default aspect of the CPU backend), not an accident; if the goal is just to prevent the warning, we can do that without changing the current behavior.

diyessi:
I would rather make PlaidML required than pull it in if it happens to be there. If it's required, we know we can use it.

earhart:
👍

PR #2412, opened 2019-02-09
[ONNX] Sign operator support

tomdol:


test/onnx_import.in.cpp
```diff
@@ -1889,3 +1889,21 @@ TEST(onnx_${BACKEND_NAME}, model_cosh)
 
     EXPECT_TRUE(test::all_close_f(expected_outputs.front(), outputs.front()));
 }
+
+TEST(onnx_${BACKEND_NAME}, model_sign)
+{
+    auto function =
+        onnx_import::import_onnx_model(file_util::path_join(SERIALIZED_ZOO, "onnx/sign.onnx"));
+
```
diyessi: since sign(-1) is -1, sign(1) is 1, you should use more challenging values, like -2.5 and 4.25

tomdol: That is a very good suggestion, thanks. I've updated the code.

tomdol: Guys, I'm not sure what the protocol is - should I resolve this comment after I've made the changes or does the requester normally do it?

PR #2416, opened 2019-02-11
up the required cmake version to 3.5 from 3.1

rkimballn1:
Since cmake 3.1 is not correct I updated the required cmake version to 3.5 which is what is supplied on ubuntu 16. I also modified the external cmake projects to not check for versions < 3.2 since we are now at 3.5


rkimballn1:
Please include any other reviewers you folks want.

ransford2011:
We need to ensure that CentOS 7.6 supports this version of Cmake before merging 😃

baojun-nervana:
It seems PaddlePaddle cmake minimum required is 3.0. Let me check if there is any issue on Paddle side. They use a range of old and new systems.

silee2:
"Since cmake 3.1 is not correct" Do you mean we are already using cmake features that require 3.2+ ? Or cmake 3.1 has some sort of bug?

silee2:
Looks current nGraph needs at least 3.3+ to avoid
`add_dependencies Cannot add target-level dependencies to INTERFACE library`

baojun-nervana:
I checkout the PR and was able to build with paddle develop branch on skx (Ubuntu16.04). Let me checkout with Baidu on Tuesday evening meeting on this.

rkimballn1:
@silee2 I was referring to the error you posted here when I was saying that 3.2 did not work. The user was using 3.2 and got the error so we need 3.3+. I did not want to test which version worked so I chose 3.5 which is the default for ubuntu 16. On centos you will likely need to build cmake from source just as you are now. Centos uses 2.x cmake.

baojun-nervana:
I submitted a PR to test through paddle CI https://github.com/PaddlePaddle/Paddle/pull/15709/files
After it passes, I will confirm with Baidu folks on this.

silee2:
@rkimballn1 I've played around with different versions for cmakes and noticed the following. 3.3 is sufficient for -DNGRAPH_DEX_ONLY=TRUE. 3.4.3 is needed if LLVM needs to be built from source. 3.5 is needed for -DNGRAPH_ONNX_IMPORT_ENABLE.  With #2420, external project are now configured only if the are enabled by options. So how do you think about the following middle ground. Keep minumum required at 3.3 for top CMakeLists.txt, and add a higher minimum required in external projects if the external project needs a higher version. That would make the minimum requried checking dynamic based on the actual nGraph build options used. PaddlePaddle officially has 3.0 as minimum but their key engineer is saying it is more like 3.3 minimum that they are using. Since the paddle bridge is only using CPU DEX, we won't disrupt their CI.

baojun-nervana:
Thanks @silee2 My PR to paddle found one system uses 3.4.3. So 3.5 will work for them.

baojun-nervana:
3.5 will NOT work for them.

PR #2419, opened 2019-02-11
Replace halide patch with cmake options.

silee2:


jbobba:
Have you tried this out? Ngraph cmake options didn't work for me when I was looking into this issue last.

PR #2420, opened 2019-02-11
Silee2/subdirectories

silee2:
Include external projects only if needed.


CMakeLists.txt
```diff
@@ -141,6 +141,10 @@ if (NGRAPH_ONNX_IMPORT_ENABLE)
     option(NGRAPH_ONNXIFI_ENABLE "Enable ONNX Interface for Framework Integration" TRUE)
 endif()
 
+if (NGRAPH_UNIT_TEST_ENABLE AND NOT NGRAPH_TOOLS_ENABLE)
```
rkimballn1: Can we just disable the tool unit test if tools are disabled and tests are enabled? I don't think there should be a cmake error related to config of tools or test.

PR #2238, opened 2018-12-16
Reuse memory for CPU backend.

ayzhuang:
1.Reuse memory for intermediates for CPU backend.
2.Use NGRAPH_REUSE_MEMORY=1 to enable memory reuse.
3.Test result for mxnet_densenet121_inference_batch1_float32.json:
   without reuse:
   cpu_memory_assignment: max allocated in total is 173740032.
   with reuse:
   cpu_memory_assignment: max allocated in total is 74346496.
4.With modified liveness analysis
  Test result for mxnet_densenet121_inference_batch1_float32.json:
   without reuse:
   cpu_memory_assignment: max allocated in total is 132820992.
   with reuse:
   cpu_memory_assignment: max allocated in total is 71942144.
5.With buffer sets and new liveness analysis
 1)mxnet_densenet121_inference_batch1_float32.json: 132812800 vs 69328896.
2)mxnet_dcgan-g_inference_batch1_float32.json: 16044032 vs 15335424.
 3)mxnet_squeezenet1.1_inference_batch1_float32.json: 13582336 vs 6815744.

PR #2421, opened 2019-02-12
 [Standalone] Introduce CPURuntimeContextCG for standalone codegen generation.

dcaballe:
This patch introduces CPURuntimeContextCG. This class is aimed at
removing the dependency between nGraph and the generated code in
codegen mode. It will be used to hold the runtime context in
codegen mode and it will be emitted in the generated code. For now,
CPURuntimeContextCG only contains TBB's graph and global context.
Follow-up patches will migrate more members in CPURuntimeContext to
CPURuntimeContextCG for codegen mode.

Testing results:
  - Before: NGRAPH_CODEGEN=1 test/unit-test
    [==========] 2503 tests from 54 test cases ran. (290406 ms total)
    [  PASSED  ] 2490 tests.

  - After: NGRAPH_CODEGEN=1 test/unit-test
    [==========] 2503 tests from 54 test cases ran. (412616 ms total)
    [  PASSED  ] 2490 tests.


dcaballe:
> It is my understanding that the codegen backend is not tested during CI. What is the plan to address this?

Jayaram mentioned that we will use `PassConfig` changes [here](https://github.com/NervanaSystems/ngraph/pull/2238/files#diff-c9d0a05c1687eee71d3b5543dfa90c36R80) to introduce a flag to enable codegen testing. Then the idea is to progressively enable backend_* tests so that they also run in codegen mode. I can take care of that once the `PassConfig` code is in place.

src/ngraph/runtime/cpu/cpu_call_frame.cpp
```diff
@@ -30,16 +30,30 @@ using namespace std;
...
     : m_external_function(external_function)
+    , m_compiled_init_ctx_func(compiled_init_ctx_func)
+    , m_compiled_destroy_ctx_func(compiled_destroy_ctx_func)
     , m_compiled_function(compiled_function)
 {
     setup_runtime_context();
+    if (!m_external_function->is_direct_execution())
+    {
+        // Invoke codegen runtime context initialization function.
+        assert(m_compiled_init_ctx_func && "compiled_init_ctx_func cannot be null.");
```
ayzhuang: There is NGRAPH_ASSERT defined in ngraph/assertion.hpp.

dcaballe:
TESTNOW

PR #2425, opened 2019-02-12
Add doc for new backend api

indie:
Doc changes needed from PR #2378 including --

* inheritance diagram
* new classes and methods
* how to work with backend -> executable
* plus anything else anybody else thinks would be useful


indie:
P.S.  Yes,  this contains a version change on the root doc to v0.15 so don't merge until that is approved and ready.

PR #2434, opened 2019-02-13
[ONNX] Handle trimmed optional outputs.

arogowie-intel:


src/ngraph/frontend/onnx_import/core/graph.cpp
```diff
@@ -132,8 +133,14 @@ namespace ngraph
...
                 m_nodes.emplace_back(node_proto, *this);
                 const Node& node{m_nodes.back()};
+
                 NodeVector ng_nodes{node.get_ng_nodes()};
-                for (int i = 0; i < ng_nodes.size(); i++)
+                // Iterate over the minimum from requested number of outputs for given node and
+                // the number of outputs we return.
+                // Some of them may be optional and trimmed. See:
+                // https://github.com/onnx/onnx/blob/master/docs/IR.md#optional-inputs-and-outputs
+                std::size_t output_count = std::min(ng_nodes.size(), node.get_outputs_size());
```
postrational: Let's try to avoid this hack. Perhaps we should try to return `NullNode` in all ops which currently don't return the expected number of outputs.

arogowie-intel: Done, looks, like `Dropout` was the only case.

src/ngraph/frontend/onnx_import/op/dropout.hpp
```diff
@@ -0,0 +1,45 @@
...
+    namespace onnx_import
+    {
+        namespace op
+        {
+            namespace set_1
+            {
+                inline NodeVector dropout(const Node& node)
+                {
+                    // First value is actual output of Dropout,
+                    // the second one is just a plumb for optional trailing output.
```
postrational: `plumb` -> `placeholder`

PR #2436, opened 2019-02-13
[ONNX] Inverse hyperbolic trigonometric functions.

arogowie-intel:


src/ngraph/frontend/onnx_import/CMakeLists.txt
```diff
@@ -39,14 +39,20 @@ add_library(onnx_import STATIC
         core/value_info.hpp
         exceptions.hpp
         op/acos.hpp
+        op/acosh.cpp
+        op/acosh.hpp
```
silee2: Any reason for including header files in the source list?

silee2: By the way, this is not a change request. Just curious.

arogowie-intel: @silee2 This is only a developer convenience thing. Some of our team is using IDEs like Visual Studio or CLion with CMake based projects. Adding headers to sources list enables them to easily navigate and search through project files. Otherwise these IDEs can't find those files.

PR #2441, opened 2019-02-13
Fixed the unit test build

avijit-nervana:
Currently if `NGRAPH_TOOLS_ENABLE` is set to `NO` but `NGRAPH_UNIT_TEST_ENABLE` is set to `YES` then that results in broken build. 
This fix should be applied to v0.14 release also.


silee2:
@avijit-nervana FYI #2420 Has the same fix bit not merge yet

rkimballn1:
@silee2 should we close this one in favor of @2420?

PR #2442, opened 2019-02-13
Fixed the build which currently breaks if tools are not enabled but u…

silee2:
…nit test is


silee2:
merge #2441 to r0.14

silee2:
#2441 not needed for r0.14

PR #2443, opened 2019-02-13
adding quantized convolution builders for ONNX

adstraw:


src/ngraph/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ set (SRC
     autodiff/adjoints.cpp
     builder/autobroadcast.cpp
     builder/numpy_transpose.cpp
+	builder/quantization/quantized_linear_convolution.cpp
```
jbobba: extra space?

jbobba: Also not sure if we need a new directory here.

adstraw: my plan is to move all quant builders into a new directory.  I thought I would start with this PR.  but, a wholesale move later could work.  I will fix the extra space.

PR #2446, opened 2019-02-14
Inplace convert for bitcasting ops

jbobba:
Snuck in a few changes for fused quantized convolutions since MKLDNN is the only available option for those ops.


src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -670,58 +671,40 @@ namespace ngraph
                 template <>
                 void CPUAssignment::ASSIGN_DECL(ngraph::op::QuantizedConvolutionRelu)
                 {
-                    if (node->get_input_element_type(0) == element::u8 &&
-                        node->get_input_element_type(1) == element::i8)
-                    {
-                        runtime::cpu::mkldnn_utils::assign_mkldnn_kernel(node);
-                    }
+                    runtime::cpu::mkldnn_utils::assign_mkldnn_kernel(node);
```
gaurides: Why did you remove the type check? Is it because we know QuantizedConvolutionRelu will be called only for u8 or i8 and the check is happening elsewhere maybe in ctor?
for ngraph::op::ConvolutionRelu(), `can_use_mkldnn_conv()` is called which checks if the type is `f32`.

jbobba: Right. we need to create these fused kernels only if they are supported by mkldnn

PR #2456, opened 2019-02-15
Unit tests for relevant resnet50 integer ops

AyanmoI:
nnp-transformer manifest changes in PR:
https://github.com/NervanaSystems/nnp-transformer/pull/366


test/CMakeLists.txt
```diff
@@ -133,6 +133,7 @@ set(MULTI_TEST_SRC
     backend_arg_reduce.in.cpp
     backend_test.in.cpp
     backend_unary_elementwise.in.cpp
+    backend_nnp_datatype_test.in.cpp
```
rkimballn1: bfloat16 is not an nnp datatype. nnp does use bfloat16 but we should treat it as just another datatype. Upcoming CPUs support bfloat16. The file name should be something different or these tests should be included in other files.

test/backend_binary_elementwise.in.cpp
```diff
@@ -148,6 +148,28 @@ NGRAPH_TEST(${BACKEND_NAME}, divide)
...
+
+    auto A = make_shared<op::Parameter>(element::i32, shape);
+    auto B = make_shared<op::Parameter>(element::i32, shape);
+    auto f = make_shared<Function>(make_shared<op::Divide>(A, B), ParameterVector{A, B});
+
+    auto backend = runtime::Backend::create("${BACKEND_NAME}");
+
+    // Create some tensors for input/output
+    auto a = backend->create_tensor(element::i32, shape);
+    copy_data(a, vector<int32_t>{2, 4, 8, 16});
```
rkimballn1: These number are easily represented by all datatypes. To exercise int32 you should use numbers that explore a larger space of numbers. Numbers like 0x40000004 are valid int32 numbers but neither bfloat16 nor float can represent it accurately. To test that int32 divide actually works correctly you must test numbers that can't be represented by smaller datatypes.

test/backend_nnp_datatype_test.in.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include "util/test_tools.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+static string s_manifest = "${MANIFEST}";
+
+NGRAPH_TEST(${BACKEND_NAME}, scalar_constant_int64_bfloat_fail)
+{
+    auto r = op::Constant::create(element::i64, Shape{}, {501});
```
rkimballn1: Just as with int32 you should be using numbers that exercise int64. 501 can't fit in a bfloat but it can fit in a float or an int32. The numbers you test should only pass if they are implemented as 64 bit integers internally.
The same argument holds for the other tests.

test/backend_nnp_datatype_test.in.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include "util/random.hpp"
+#include "util/test_control.hpp"
+#include "util/test_tools.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+static string s_manifest = "${MANIFEST}";
+
+NGRAPH_TEST(${BACKEND_NAME}, scalar_constant_int64_bfloat_fail)
```
rkimballn1: The test name should just be scalar_constant_int64

test/CMakeLists.txt
diyessi: We also have some float16 hardware, so we need to be careful with naming.

test/backend_test.in.cpp
```diff
@@ -1791,6 +1791,21 @@ NGRAPH_TEST(${BACKEND_NAME}, scalar_constant_int64)
     EXPECT_EQ(vector<int64_t>{2112}, read_vector<int64_t>(result));
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, scalar_constant_int64_1)
```
rkimballn1: This test should be named `scalar_constant_int64` and should replace the existing test.

test/backend_test.in.cpp
```diff
@@ -1791,6 +1791,21 @@ NGRAPH_TEST(${BACKEND_NAME}, scalar_constant_int64)
...
+{
+    auto r = op::Constant::create(element::i64, Shape{}, {0x80000002});
```
rkimballn1: Use int64 input values for the test. This is an int32 value.

test/backend_test.in.cpp
```diff
@@ -1807,17 +1822,27 @@ NGRAPH_TEST(${BACKEND_NAME}, tensor_constant_float32)
     EXPECT_EQ((vector<float>{4.75f, 4.5f, -5.25f, 0.0f}), read_vector<float>(result));
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, tensor_constant_int64_1)
```
rkimballn1: Same comment. Rename test to replace existing. Use int64 input values.

test/backend_comparison.in.cpp
```diff
@@ -101,6 +101,27 @@ NGRAPH_TEST(${BACKEND_NAME}, greater)
...
+    Shape shape{2, 2, 2};
+    auto A = make_shared<op::Parameter>(element::i64, shape);
+    auto B = make_shared<op::Parameter>(element::i64, shape);
+    auto f = make_shared<Function>(make_shared<op::Greater>(A, B), ParameterVector{A, B});
+
+    auto backend = runtime::Backend::create("${BACKEND_NAME}");
+
+    // Create some tensors for input/output
+    auto a = backend->create_tensor(element::i64, shape);
+    copy_data(a, vector<int64_t>{0x80000000, 0x80000002, -8, 17, -5, 5, 2, 1});
```
rkimballn1: use int64 values

test/backend_comparison.in.cpp
```diff
@@ -164,6 +185,27 @@ NGRAPH_TEST(${BACKEND_NAME}, lesseq)
     EXPECT_EQ((vector<char>{1, 0, 1, 0, 1, 1, 0, 1}), read_vector<char>(result));
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, lesseq_int32)
```
csullivan: This `lesseq` test only tests less than, doesn't cover the equality case.

AyanmoI: Good point..

test/backend_sum.in.cpp
```diff
@@ -403,6 +403,32 @@ NGRAPH_TEST(${BACKEND_NAME}, sum_3d_to_scalar)
...
+        vector<int32_t>{
+            0x40000001, 10, 19, 4,  13, 22, 7,  16, 25, 2,  11, 20, 5, 14,
+            23,         8,  17, 26, 3,  12, 21, 6,  15, 24, 9,  18, 27}); // 1073742202 == 0x4000017A
+    auto result = backend->create_tensor(element::i32, shape_rt);
+
+    auto handle = backend->compile(f);
+    handle->call_with_validate({result}, {a});
+    EXPECT_EQ((vector<int32_t>{0x40000001 + 10 + 19 + 4 + 13 + 22 + 7 + 16 + 25 + 2 + 11 + 20 + 5 +
+                               14 + 23 + 8 + 17 + 26 + 3 + 12 + 21 + 6 + 15 + 24 + 9 + 18 +
+                               27}), // NNP output 1073741824 == 0x40000000
```
csullivan: Is this comment about NNP left in nGraph intentionally? I would guess we should remove it.

test/backend_comparison.in.cpp
AyanmoI: However, the equality condition will alway pass for something like `0x40000005` since they will fall back to `0x40000000`. But it makes sense to add this anyway.

test/backend_sum.in.cpp
AyanmoI: Yes ...this was intentional but I can remove it.

AyanmoI:
@ransford2011 Used hex so that it is gives an understanding/sense of which bits are turned on. It helps visualize the number while testing for the dynamic range. What I mean by that is `0x40000000` is `01000000000000000000000000000000` and  `0x40000001` is `01000000000000000000000000000001`, so on and so forth.

PR #2465, opened 2019-02-15
IntelGPU backend: Quantize operations

shssf:


src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -1519,6 +1519,191 @@ void runtime::intelgpu::do_reshape_operation(cldnn::topology& topology,
...
+            break;
+        }
+        case ngraph::op::Quantize::RoundMode::ROUND_NEAREST_TOWARD_ZERO:
+        {
+            writer
+                << real_type_str << " abs_qvalue = fabs(qvalue);\n"
+                << real_type_str << " abs_qvalue_toward_zero = ceil(abs_qvalue - 0.5);\n"
+                << "qvalue = (qvalue < 0.0) ? -abs_qvalue_toward_zero : abs_qvalue_toward_zero;\n";
+        }
+        break;
```
rkimballn1: You should be consistent where you put the break; statement. In the previous case you put it inside the brackets while for this case you put it outside the brackets. For consistency with the rest of the code please put break; inside brackets. This applies to all cases here.

shssf: @rkimballn1 Good catch! Usually I put the `break` after brackets but it depends on the project code style.  This particular statement placement was a typo.

PR #2472, opened 2019-02-16
Function call working

rkimballn1:
This is a departure from the old hybrid transformer. Now instead of splitting the graph in order to change from one backend to a different backend we wrap all of the other backend ops in a FunctionCall op and insert that into the original graph.

Advantages:
1) No splitting of the graph. No need to pull all temporaries off of one device only to put them back after the "splitting" op
2) Calls to the secondary backend are all via pre-compiled objects so no run-time compiles.

So this (The yellow ops run on a different backend than the green ops):
graph

is converted to this (FunctionCall_16 contains all of the yellow ops):
f

where FuncitonCall_16 contains this:
sub_function


src/ngraph/runtime/interpreter/int_executable.hpp
```diff
@@ -740,6 +740,29 @@ class ngraph::runtime::interpreter::INTExecutable : public Executable
                 args[0]->get_data_ptr<const T>(), out[0]->get_data_ptr<T>(), element_count);
             break;
         }
+        case OP_TYPEID::FunctionCall:
```
shssf: Sorry, I do not understand a purpose of that operation.
Is it an attempt to emulate "graph level inlining"?
Do we have any performance statistics about diffrent approaches (generate two large graphs with similar part and the same with "FunctionCall")?
In particular, INTELGPU backend has no ability for native support this. Any FunctionCall will be "unrolled", or executed as independent graph with all required overhead (load to runtime stack and memory copy of the input parameters).

rkimballn1: FunctionCall is a collection of ops which are not supported by the current running backend. An example is integer ops with the NNP transformer. Integer ops must fall back to run on the CPU in that case. The FunctionCall is specifically because backends like the INTELGPU have no native support for some things, all op that are not supported natively are wrapped up in the FunctionCall and then called on an appropriate backend which does support those ops. The intent is to execute this "as independent graph with all required overhead (load to runtime stack and memory copy of the input parameters)". This is the hybrid transformer.
Currently two other approaches have been tried. One is "graph splitting" which ends up copying too much data back and forth between backends, and the "one op at a time" approach of the NNP transformer. Both of those have performance issues that are addressed with this PR.

jbobba:
I haven't gone through the hybrid backend code fully but i was hoping to find a request for default (row-major) layout on the result nodes (https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/result.hpp#L41). Otherwise backends can return the tensor in a layout that the other backends might not understand.

sasadep:
@rkimballn1 correct me if I'm wrong. 
@jbobba currently there is an extra copy when moving a tensor from one backend to another, by using the function [copy_from](https://github.com/NervanaSystems/ngraph/blob/bob/h4/src/ngraph/runtime/tensor.cpp#L77) 
which is an optimization after functionality

sasadep:
@rkimballn1 do you think it would be nice if `FunctionCall_16` on second graph in the PR description to be outlined in green but filled with yellow.
So it represents it represents `FunctionCall_16` is on another device ?

PR #2475, opened 2019-02-18
[ONNX] Import functions input validation

tomdol:
Two unit tests checking the pesimistic path when using `import_onnx_model`


src/ngraph/frontend/onnx_import/onnx.cpp
```diff
@@ -35,15 +35,15 @@ namespace ngraph
...
+                        : ngraph_error{"Failure opening file: " + path}
                     {
                     }
                 };
 
                 struct stream_parse : ngraph_error
                 {
                     explicit stream_parse(std::istream&)
-                        : ngraph_error{"failure parsing data from the stream"}
+                        : ngraph_error{"Failure parsing data from the stream"}
```
postrational: Let's try to make this message a bit easier to understand. For example: change `the stream` to `input stream`

PR #2483, opened 2019-02-20
Update mkl-dnn build script.

silee2:


silee2:
Explicitly linking cpu backend against MKL since cpu backend depends on MKL and MKL-DNN may be built without MKLl

silee2:
Added some logic for Windows. But have not tested yet.

PR #2485, opened 2019-02-21
IntelGPU backend: Convert operation fix to be friendly with CPU float…

shssf:
…ing point conversion

### DIFF
--- a/src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
+++ b/src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
@@ -59,6 +59,7 @@ string runtime::intelgpu::get_opencl_type_min_max_value(const element::Type& ngr
     case element::Type_t::u32: return is_min ? "0" : "UINT_MAX";
     case element::Type_t::i16: return is_min ? "SHRT_MIN" : "SHRT_MAX";
     case element::Type_t::u16: return is_min ? "0" : "USHRT_MAX";
+    case element::Type_t::boolean:
     case element::Type_t::i8: return is_min ? "CHAR_MIN" : "CHAR_MAX";
     case element::Type_t::u8: return is_min ? "0" : "UCHAR_MAX";
     }
@@ -1229,8 +1230,37 @@ void runtime::intelgpu::do_convert_operation(cldnn::topology& topology,
     {
         gws = generate_loops(writer, output_shape, true);
 
-        writer << "output" << access_dims(output_shape) << " = convert_" << output_type_name
-               << "(input0" << access_dims(output_shape) << ");\n";
+        if ((input_type.get_type_enum() == element::Type_t::f64) ||
+            (input_type.get_type_enum() == element::Type_t::f32))
+        {
+            // this is the workaround for OpenCL to be same as with CPU floating point operations
+            writer << input_type_name << " input_var = input0" << access_dims(output_shape) << ";\n"
+                   << output_type_name << " output_var = 0;\n";
+
+            writer << "if (input_var > " << get_opencl_type_min_max_value(output_type, false)
+                   << ")\n";
+            writer.block_begin();
+            {
+                writer << "output_var = " << get_opencl_type_min_max_value(output_type, true)
+                       << ";\n";
+            }
+            writer.block_end();
+
+            writer << "else\n";
+
+            writer.block_begin();
+            {
+                writer << "output_var = convert_" << output_type_name << "(input_var);\n";
+            }
+            writer.block_end();
+
+            writer << "output" << access_dims(output_shape) << " = output_var;\n";
+        }
+        else
+        {
+            writer << "output" << access_dims(output_shape) << " = convert_" << output_type_name
+                   << "(input0" << access_dims(output_shape) << ");\n";
+        }
 
         generate_loops(writer, output_shape, false);
     }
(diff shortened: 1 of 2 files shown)


test/backend_test.in.cpp
```diff
@@ -1382,6 +1382,66 @@ NGRAPH_TEST(${BACKEND_NAME}, convert_int32_float32)
...
+                             std::numeric_limits<double>::lowest(),
+                             INFINITY,
+                             -INFINITY,
+                             4.0});
+    auto result = backend->create_tensor(element::i32, shape);
+
+    auto handle = backend->compile(f);
+    backend->call_with_validate(handle, {result}, {a});
+    EXPECT_EQ((vector<int>{0,
+                           std::numeric_limits<int>::lowest(),
```
rkimballn1: If I c++ static_cast a max double into an int I get the answer 2147483647 which is std::numeric_limits<int>::max(). 
So I think the results should be:
```
0
std::numeric_limits<int>::max()
std::numeric_limits<int>::lowest()
std::numeric_limits<int>::max()
std::numeric_limits<int>::lowest()
4
```
Same comment for the float test below.
Although I really don't know what the correct answer is for the INFINITY

shssf: Which version of the compiler you used?
All these things are not defined well by the compiler standard. 
I just want to provide a backend behavior as it CPU backend does (this is expected by the user).
Perhaps it is impossible due to diffrent compilers provide diffrent results.

rkimballn1: This was with gcc 7.3 which is the default on ubuntu 18
I can try again with clang

rkimballn1: I only tested because converting a large positive float/double to an in should be a positive result, not a negative result.

shssf: floating point arithmetic are in the grey area of diffrent compilers.

shssf: "I only tested because converting a large positive float/double to an int should be a positive result, not a negative result"
This is not quite right. The maximum in double datatype can't represented in int datatype. There is no representations in this case. The CPU instructions used for this return "error status" but the exception is masked by the CPU. So, If we ask to convert "double::maximum" to int - it will be seen as "int::minimum".

rkimballn1: This is what the C99 spec has to say on the issue:

`When a finite value of real floating type is converted to an integer type other than _Bool, the fractional part is discarded (i.e., the value is truncated toward zero). If the value of the integral part cannot be represented by the integer type, the behavior is undefined.`

https://stackoverflow.com/questions/526070/handling-overflow-when-casting-doubles-to-integers-in-c

shssf: @rkimballn1 "If the value of the integral part cannot be represented by the integer type, the behavior is undefined."
This is exactly part that I told.
But the question is - what we should do with it? I think the user would expect the operation behavior exactly as it done by CPU backend. OpenCL (the background compiler under INTELGPU) has it's own understanding of these things.
This is why I tried to make this PR.

diyessi: I think the user shouldn't be expecting any consistent behavior when they drift outside of what the language definition; they should fix their model to stay within the defined behavior.
However, our users are actually python users, so python behavior, not C/C++ behavior, is what we should be emulating. What does Python do here?

shssf: Ok, I changed this PR to be INTELGPU specific.
We need these changes to pass, at least, TF test system using the backend.

shssf:
@rkimballn1 @diyessi Could you please approve this PR? I make it the backend specific. Th is PR make several test passed in TF unit tests with the backend.

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -59,6 +59,7 @@ string runtime::intelgpu::get_opencl_type_min_max_value(const element::Type& ngr
     case element::Type_t::u32: return is_min ? "0" : "UINT_MAX";
     case element::Type_t::i16: return is_min ? "SHRT_MIN" : "SHRT_MAX";
     case element::Type_t::u16: return is_min ? "0" : "USHRT_MAX";
+    case element::Type_t::boolean:
     case element::Type_t::i8: return is_min ? "CHAR_MIN" : "CHAR_MAX";
```
diyessi: CHAR_MIN and CHAR_MAX don't seem like the right values for boolean. Even if they are, I don't think it should fall through to i8 and a comment about what is going on is needed.

shssf: @diyessi According to this https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/type/element_type.cpp#L66
the nGraph::boolean is the most close to nGraph::i8. Same size, same sign.

shssf:
Gentlemen, this PR is 20 days old. Do you see really blocked things to merge this?

src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
diyessi: That's because std::vector<T> is something completely different thing when T is bool than when it is any other type. ngraph bool as two values, 0 and 1.

shssf:
@diyessi All comments are addressed

PR #2490, opened 2019-02-22
rename and document the node name methods

rkimballn1:


postrational:
I would say that the different types of names are confusing. 
We now have `name`, `friendly_name`, `unique_name`...

@diyessi Could we just call the unique value `id` and the user-designated label `name`?

dcaballe:
Please, let me add my (unsolicited) opinion to this interesting discussion 🙂. 
IMO, I think we should distinguish between three concepts: opcode, type and name. We should pay special attention to the memory footprint of these three, since Node is the core of our representation and we should be able to scale to thousands/million of them. For that reason, a single byte can make a difference:

1) Opcode: the ID that identifies the operation (Add, Mul, Conv, etc.). We shouldn't use `std::string` for it since it's not very efficient in space and we don't want to replicate the same `std::string` in every node with the same opcode. A simple `enum`/`unsigned char` (preferably the last one) should be enough (we could use op_tbl for that, right?). We could have `get_opcode` or `get_operation_id` returning the numerical id and `get_opcode_name` or `get_operation_name` returning the `string`/`char*`. The latter should be implemented with a map or switch mapping the numerical id to the string. No string should be allocated per Node, to save memory.

2) Type: in other compilation frameworks, this corresponds to the data type of the operation (float, double, int, vector, tensor, ...). Maybe it would be a good idea to leave that term for that purpose, for future use and avoid confusion in the interaction with other frameworks.

3) Name: ID for each particular node instance. We could have something like `get_node_name`. It should have a default name but the clients should be able to "contribute" to that name. I say "contribute" because I'm not sure how the client could guarantee that the name provided is not already in use. IMO, we should have a naming mechanism that turns client's names into unique names by appending some numerical counter, if needed. Regarding the implementation, we should use a `char *` instead of `std::string` because of the larger memory footprint of an empty `std::string`.

IIRC, what LLVM is doing in this regard is storing only the indispensable name information (i.e., client provided) and "producing" the unique name of each node every time the node is printed. They use a counter to assign a default unique id to each node and also make client provided names into unique. This means that a node name may change between passes but it guarantees that counter values are reused when a node is removed/optimized away, preventing the counter to unnecessarily reach a high (unreadable) value and eventually overflow. I guess this wouldn't fulfill our needs.

I hope this is useful in some way.
Thanks!

PR #2496, opened 2019-02-25
[ONNX] numpy broadcasting refactoring

tsocha:


src/ngraph/frontend/onnx_import/utils/broadcasting.cpp
```diff
@@ -66,10 +40,25 @@ static std::vector<ngraph::Shape> get_numpy_broadcast_shape(ngraph::Shape left_s
 static std::pair<ngraph::Shape, std::vector<ngraph::Shape>>
     get_numpy_broadcast_shapes(const ngraph::NodeVector& inputs)
 {
-    auto shape_left_fold = [](const ngraph::Shape& accumulator,
+    auto shape_left_fold = [](ngraph::Shape& accumulator,
```
tomdol: When operating on a non-const reference to the accumulator, you're left padding the internal object held as an accumulator by std::accumulate. I think it would be safer, especially without unit tests, to pass it to this labda by value.

src/ngraph/frontend/onnx_import/utils/broadcasting.cpp
```diff
@@ -141,21 +130,6 @@ namespace ngraph
...
-            const auto& numpy_shapes = get_numpy_broadcast_shape(left_shape, right_shape);
-            auto output_shape = numpy_shapes.at(0);
-            auto left_full_shape = numpy_shapes.at(1);
-            auto right_full_shape = numpy_shapes.at(2);
-
-            return {broadcast_node_numpy_style(left, output_shape, left_full_shape),
-                    broadcast_node_numpy_style(right, output_shape, right_full_shape)};
-        }
-
         NodeVector numpy_style_broadcast(NodeVector inputs)
```
tomdol: That's a leftover from my previous task, but this function should probably take a const reference instead of a copy.

src/ngraph/frontend/onnx_import/utils/broadcasting.cpp
```diff
@@ -185,14 +159,20 @@ namespace ngraph
         {
             const auto& left_shape = left->get_shape();
             const auto& right_shape = right->get_shape();
-            // Broadcast only _stack of matrices_ axes.
-            const auto& numpy_shapes = get_numpy_broadcast_shape(
-                Shape{std::begin(left_shape), std::next(std::end(left_shape), -2)},
+            // Make temporary Parameters to pass shapes into broadcast function.
```
postrational: Can we refactor `get_numpy_broadcast_shapes` to split it not 2 functions: one which takes `NodeVector` another which takes `vector<Shape>`? That way you could pass `Shape`s here instead of making temporary `Parameter` nodes.

PR #2507, opened 2019-02-27
Normalize the printing of settings of bools to ON/OFF

rkimballn1:
Now we get this. All bools are normalized to TRUE or FALSE so no more mixture of ON/OFF
```
-- NGRAPH_UNIT_TEST_ENABLE:        ON
-- NGRAPH_TOOLS_ENABLE:            ON
-- NGRAPH_CPU_ENABLE:              OFF
-- NGRAPH_INTELGPU_ENABLE:         OFF
-- NGRAPH_GPU_ENABLE:              OFF
-- NGRAPH_INTERPRETER_ENABLE:      ON
-- NGRAPH_NOP_ENABLE:              ON
-- NGRAPH_GPUH_ENABLE:             OFF
-- NGRAPH_GENERIC_CPU_ENABLE:      OFF
-- NGRAPH_DISTRIBUTED_MLSL_ENABLE: OFF
-- NGRAPH_DISTRIBUTED_OMPI_ENABLE: OFF
-- NGRAPH_DEBUG_ENABLE:            OFF
-- NGRAPH_ONNX_IMPORT_ENABLE:      OFF
-- NGRAPH_DEX_ONLY:                OFF
-- NGRAPH_CODE_COVERAGE_ENABLE:    OFF
-- NGRAPH_LIB_VERSIONING_ENABLE:   OFF
-- NGRAPH_PYTHON_BUILD_ENABLE:     OFF
-- NGRAPH_USE_PREBUILT_LLVM:       ON
-- NGRAPH_PLAIDML_ENABLE:          OFF
```


rkimballn1:
Do we want TRUE/FALSE or ON/OFF? It is trivial to change.

silee2:
I think ON/OFF is the norm. Not sure why we ended up with a lot of TRUE/FALSE.

diyessi:
Would it make sense to be consistent on `ENABLE`, i.e. `NGRAPH_DEX_ENABLE` and `NGRAPH_CODEGEN_ENABLE` and `NGRAPH_PREBUILT_LLVM_ENABLE` ?
Upon looking at the list, isn't `NGRAPH_ENABLE_X` more common than `NGRAPH_X_ENABLE`?

rkimballn1:
I have never liked the DEX_ONLY flag. It does not make sense. It would be much more clear if the CPU backend were really two proper and separate CPU backends, the interpreted version and the compiled version. Two backends with two names.
I also don't like the distributed flags NGRAPH_DISTRIBUTED_MLSL_ENABLE and NGRAPH_DISTRIBUTED_OMPI_ENABLE and would much prefer NGRAPH_DISBRIBUTED_ENABLE to be one of OFF/OMPI/MLSL as a single option. Two options allows for the state where both MLSL and OMPI are requested which is an error.
All of this is at least one more separate PR.

diyessi:
`NGRAPH_DISTRIBUTED_ENABLE` as on/off
`NGRAPH_DISTRIBUTED_KIND` as MLSL, OMPI, etc. with an error if one is not supplied when enabled.

rkimballn1:
And a new backend, something like CODEGENCPU or CCPU with CPU being the DEX version of the CPU backend.

PR #2508, opened 2019-02-27
Add info about lib versions in an easy to find place

diyessi:


cmake/external_protobuf.cmake
```diff
@@ -21,14 +21,15 @@ include(ExternalProject)
 # Download and install Google Protobuf ...
 #------------------------------------------------------------------------------
 
-set(PROTOBUF_GIT_REPO_URL https://github.com/google/protobuf.git)
-set(PROTOBUF_GIT_BRANCH origin/3.5.x)
+# This version of PROTOBUF is required by Microsoft ONNX Runtime.
+set(NGRAPH_PROTOBUF_GIT_REPO_URL "https://github.com/google/protobuf.git")
```
silee2: Repo name as changed to "https://github.com/protocolbuffers/protobuf" Ri-direction happens correctly but would be better to use the updated official URL here.

cmake/external_protobuf.cmake
```diff
@@ -21,14 +21,15 @@ include(ExternalProject)
...
+set(NGRAPH_PROTOBUF_GIT_BRANCH "origin/3.5.x")
```
silee2: Can we use a tag instead of a branch. "v3.5.2" is the latest tag on the branch "origin/3.5.x"

cmake/external_mkldnn.cmake
```diff
@@ -16,6 +16,11 @@
 
 include(ExternalProject)
 
+# Includes blas 3.8.0 in mkldnn 
+set(NGRAPH_MKLDNN_VERSION "v0.18-rc")
+set(NGRAPH_MKLDNN_SUB_VERSION "2019.0.3.20190125")
+set(NGRAPH_MKLDNN_GIT_TAG "08bd90c")
```
silee2: Can we use a symbolic tag instead? "v0.18-rc" is the same as  "08bd90c"

cmake/external_tbb_prebuilt.cmake
```diff
@@ -16,12 +16,15 @@
 
 include(ExternalProject)
 
-set(ARCHIVE_FILE_BASE tbb2019_20181203oss)
+
+set(NGRAPH_TBB_VERSION "2019_U3")
```
rkimballn1: Since this is used in two places (this file and external_tbb.cmake) can we move it up one level so it is only set in one place? The two places will get out of sync and lead to unexpected behavior

cmake/external_mkldnn.cmake
rkimballn1: We should be able to use git to get the hash of v0.18-rc so they are always in sync.

cmake/external_tbb_prebuilt.cmake
diyessi: Can we do this in a separate PR that maybe switches to just using the pre-built tbb, i.e. eliminating the second place?

PR #2511, opened 2019-02-27
mkl-dnn use CMAKE_INSTALL_LIBDIR to handle lib vs lib64 directory nam…

silee2:
…e difference.


silee2:
CentOS test are failing since it tries to copy libmkldnn.so from lib instead of lib64

PR #2513, opened 2019-02-28
Support CPU DEX on Windows

silee2:
Along with downloading LLVM from tar ball and some work towards making codegen work on Windows.


cmake/external_gtest.cmake
```diff
@@ -76,8 +76,16 @@ target_include_directories(libgtest SYSTEM INTERFACE
     ${SOURCE_DIR}/googletest/include
     ${SOURCE_DIR}/googlemock/include)
 
-target_link_libraries(libgtest INTERFACE
-    debug ${GTEST_OUTPUT_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}gtestd${CMAKE_STATIC_LIBRARY_SUFFIX}
-    debug ${GMOCK_OUTPUT_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}gmockd${CMAKE_STATIC_LIBRARY_SUFFIX}
-    optimized ${GTEST_OUTPUT_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}gtest${CMAKE_STATIC_LIBRARY_SUFFIX}
-    optimized ${GMOCK_OUTPUT_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}gmock${CMAKE_STATIC_LIBRARY_SUFFIX})
+if(NOT WIN32)
```
diyessi: Be explicit about platforms -- check for macos or linux here, windows where you do the else. That way no surprises if something else comes along.

cmake/external_llvm.cmake
```diff
@@ -25,8 +25,8 @@ endif()
 ExternalProject_Add(
     ext_clang
     PREFIX clang
-    GIT_REPOSITORY https://github.com/llvm-mirror/clang.git
-    GIT_TAG 26cac19a0d622afc91cd52a002921074bccc6a27
+    URL http://releases.llvm.org/5.0.2/cfe-5.0.2.src.tar.xz
```
diyessi: Please move these versions up to `NGRAPH_` variables at the top of the file like in the mkldnn and protobuf cmakes. This makes it easier for validation to go through SDL.

src/resource/util.cpp
```diff
@@ -165,13 +170,44 @@ void iterate_files_worker(const string& path,
...
+            }
+            else
+            {
+                string file_name = path_join(path, data.cFileName);
+                func(file_name, false);
+            }
+        } while (FindNextFile(hFind, &data));
+        FindClose(hFind);
+    }
+#else
```
diyessi: Use a separate if for the non-windows platforms.

silee2: I don't get it. The "#else" part is handling non-windows platforms.
I don't know a way to express "if (APPLE or LINUX)" like we do in cmake for a c/c++ preprocessor.

silee2:
Opened #2531 with same change to fix Windows build issues on r01.5

silee2:
Merged by #2567

PR #2514, opened 2019-02-28
Fix bug introduced by #2238 on master

silee2:


silee2:
#2515 is the same patch going to r0.15

PR #2515, opened 2019-02-28
Fix bug introduced by #2238 on r0.15

silee2:


silee2:
#2514 is the same patch going to master

PR #2516, opened 2019-02-28
Extend compile API to take in PassConfig object

jbobba:
`ngraph::pass::PassConfig pass_config;
pass_config.set_compilation_mode(ngraph::pass::CompilationMode::CODEGEN);`
`auto handle = backend->compile(f, pass_config);`

We need to find a better way to create a small testsuite for codegen. Hopefully in a future PR.


src/ngraph/pass/pass_config.hpp
```diff
@@ -23,21 +23,28 @@ namespace ngraph
...
+    PassConfig(CompilationMode mode = CompilationMode::DEX);
     const std::map<std::string, bool>& get_enables() { return m_pass_enables; }
     void set_pass_enable(std::string name, bool enable);
     bool get_pass_enable(std::string name);
     const std::map<std::string, bool>& get_pass_attributes() { return m_pass_attributes; }
     void set_pass_attribute(std::string name, bool enable);
     bool get_pass_attribute(std::string name);
-
+    CompilationMode get_compilation_mode() const { return m_compilation_mode; }
+    void set_compilation_mode(CompilationMode mode) { m_compilation_mode = mode; }
```
dcaballe: I wonder if we should remove the setter so that we only allow setting the compilation mode through the constructor. This would prevent clients to change it randomly. We might have initialization based on the compilation mode in the future.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1703,6 +1703,11 @@ shared_ptr<ngraph::runtime::cpu::CPU_CallFrame>
     runtime::cpu::CPU_ExternalFunction::make_call_frame(ngraph::pass::PassConfig& pass_config)
 {
 #if !defined(NGRAPH_DEX_ONLY)
+    // Override DEX if pass_config requests CODEGEN
+    if (pass_config.get_compilation_mode() == ngraph::pass::CompilationMode::CODEGEN)
+    {
+        m_direct_execution = false;
```
dcaballe: To prevent unexpected behavior, shouldn't we better assert here? Like "CodeGen mode not supported"?

src/ngraph/pass/pass_config.hpp
jbobba: i'm ok either way. i can remove it for now and we can add later if needed.

src/ngraph/runtime/cpu/cpu_external_function.cpp
jbobba: In my mind, pass_config is requests from the user that we will do our best to adhere to. May be a warning instead of assert?

dcaballe: Sure! It should be fine as long as the user can see "hey, you said CG but I'm running DEX instead".

python/pyngraph/runtime/backend.cpp
```diff
@@ -17,12 +17,20 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
+#include "ngraph/pass/pass_config.hpp"
```
postrational: We should remove this `include` or add another `compile` wrapper function using the config.

PR #2523, opened 2019-02-28
Avijit/add diag 4 distributed debugging

avijit-nervana:
* Added a new debug logging macro: `NGRAPH_DIST_DEBUG(...)` which prints high precision timestamp and MPI rank before every message to be able to correlate logs from various computers for debugging the AllReduce bug
* Added environment variable `NGRAPH_DIST_DISABLE_LOGGING` to turn this off when not needed
* Added necessary debug messages in the AllReduce op kernel for CPU backend. These messages are very useful debugging the out of order execution issues
* All of these are under the `NGRAPH_DISTRIBUTED_ENABLE` - so doesn't affect non-distributed use cases.

The debug macro can be extended to the rest of the nGraph code to enable `on-demand` debug logging (as opposed to the current model of compile time selection)

### DIFF
--- a/src/ngraph/runtime/cpu/builder/allreduce.cpp
+++ b/src/ngraph/runtime/cpu/builder/allreduce.cpp
@@ -21,70 +21,9 @@
 #include <mpi.h>
 #endif
 
+#include "ngraph/log.hpp"
 #include "ngraph/op/allreduce.hpp"
 #include "ngraph/runtime/cpu/cpu_builder.hpp"
-
-#include <cstdarg>
-#include <iomanip>
-#include <locale>
-#include <sys/time.h>
-#include <unistd.h> /*for CLOCK_REALTIME? */
-
-#include "ngraph/distributed.hpp"
-
-std::string get_timestamp()
-{
-    using namespace std::chrono;
-
-    // get current time
-    auto now = system_clock::now();
-
-    // get number of nanoseconds for the current second
-    // (remainder after division into seconds)
-    auto ns = duration_cast<nanoseconds>(now.time_since_epoch()) % 1000000;
-
-    // convert to std::time_t in order to convert to std::tm (broken time)
-    auto timer = system_clock::to_time_t(now);
-
-    // convert to broken time
-    std::tm bt = *std::localtime(&timer);
-
-    std::ostringstream timestamp;
-    timestamp << std::put_time(&bt, "%H:%M:%S"); // HH:MM:SS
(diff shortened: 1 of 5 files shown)


src/ngraph/runtime/cpu/builder/allreduce.cpp
```diff
@@ -24,6 +24,68 @@
...
+    std::vsnprintf(buf.data(), buf.size(), fmt, args2);
+    va_end(args2);
+
+    ngraph::Distributed dist;
+    std::printf("%s [RANK: %d]: %s\n", get_timestamp().c_str(), dist.get_rank(), buf.data());
+}
+#define NGRAPH_DIST_DEBUG(fmt, ...)                                                                \
+    if (!DISABLE_LOGGING)                                                                          \
+    LogPrintf(fmt, ##__VA_ARGS__)
+
```
jbobba: Can we move this code to ngraph/log.hpp?

avijit-nervana: Moved to the ngraph/log.hpp.

avijit-nervana:
Moved the implementation to the ngraph/log.hpp/.cpp as this mechanism may be used by other code modules. Please note the following:
1. This implementation is not available on Windows as the timestamp APIs are different on Windows(used to be at least)
2. This implementation should be extended with the following:
    - Add support for a log level
    - Add support for log areas
Also - current implementation uses printf style but another possibility is to use C++ stringstream. Should be discussed and addressed in a future PR.

avijit-nervana:
@jianyinglang For now we print the logs by default (by deign). Once all the necessary features are implemented (i.e., log levels and areas perhaphs) - we can change the default.

silee2:
```
/home/dockuser/ngraph/src/ngraph/log.cpp:104:58: error: format string is not a string literal [-Werror,-Wformat-nonliteral]

    std::vector<char> buf(1 + std::vsnprintf(nullptr, 0, fmt, args1));

                                                         ^~~

/home/dockuser/ngraph/src/ngraph/log.cpp:106:44: error: format string is not a string literal [-Werror,-Wformat-nonliteral]

    std::vsnprintf(buf.data(), buf.size(), fmt, args2);

                                           ^~~

2 errors generated.

src/ngraph/CMakeFiles/ngraph.dir/build.make:518: recipe for target 'src/ngraph/CMakeFiles/ngraph.dir/log.cpp.o' failed
```

PR #2525, opened 2019-03-01
Silee2/external project rpath

silee2:
Set all shared library rpath on linux to "$ORIGIN". All shared libraries are currently placed in a single directory ${NGRAPH_BUILD_DIR}.


silee2:
Blocking until Linux CI has patchelf installed.

CMakeLists.txt
```diff
@@ -58,12 +58,8 @@ set(NGRAPH_INCLUDE_PATH
...
 
 if (APPLE)
-    # Suppress an OS X-specific warning.
-    if (POLICY CMP0042)
-        cmake_policy(SET CMP0042 NEW)
-    else()
-        set(MACOS_RPATH ON)
-    endif()
+    # Enable MACOS_RPATH by default.
+    cmake_policy(SET CMP0042 NEW)
```
silee2: CMake required version is at least 3.1 so no need to check if POLICY CMP0042 is available.

diyessi:
If we need patchelf, we'll need to update the doc and some other things.

silee2:
@diyessi patchelf is for patching prebuilt libraries. mklml does not have an rpath set. It is assuming libiomp5 is a system library. The whole OpenMP situation is not pretty. libngraph is forced to link against libiomp5 and "-Wl,--no-as-needed" is used to prevent gcc from linking against libgomp. So far, libngraph is loaded before mklml so mklml missing rpath was not an issue as libiomp5 would be already loaded in memory but the whole setup looks fragile.

silee2:
@diyessi We could ask mkl team to set $ORIGIN as rpath for the libmklml bundled with mkl-dnn. Then we don't need patchelf.

diyessi:
It would be good to ask mkl to fix their linking, but, in the mean time, if we have a new build dependency we need to update the build instructions to get the patchelf script (if I remember, it's a simple sed script since it's just replacing a constant's value in the binary).

silee2:
looks like clear linux does not have patchelf. Will have to remove patchelf usage from this PR and add a workaround in a future PR.

PR #2526, opened 2019-03-01
Fallback to default pooling kernels when mkldnn doesn't support input…

jbobba:
… format

### DIFF
--- a/src/ngraph/runtime/cpu/pass/cpu_layout.cpp
+++ b/src/ngraph/runtime/cpu/pass/cpu_layout.cpp
@@ -995,10 +995,12 @@ namespace ngraph
                             auto default_format = arg0_shape.size() == 4
                                                       ? mkldnn::memory::format::nchw
                                                       : mkldnn::memory::format::ncdhw;
-                            auto default_desc = mkldnn_utils::create_default_mkldnn_md(
+                            auto default_desc_i = mkldnn_utils::create_default_mkldnn_md(
                                 node.get(), 0, false, default_format);
-                            i_mds.push_back(default_desc);
-                            o_mds.push_back(default_desc);
+                            auto default_desc_o = mkldnn_utils::create_default_mkldnn_md(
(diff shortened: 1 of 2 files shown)


src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -990,8 +990,21 @@ namespace ngraph
...
-                                           to_string(input_desc.data.format) + e.message);
+                        if (arg0_shape.size() == 4 || arg0_shape.size() == 5)
+                        {
+                            auto default_format = arg0_shape.size() == 4
+                                                      ? mkldnn::memory::format::nchw
+                                                      : mkldnn::memory::format::ncdhw;
+                            auto default_desc = mkldnn_utils::create_default_mkldnn_md(
+                                node.get(), 0, false, default_format);
+                            i_mds.push_back(default_desc);
+                            o_mds.push_back(default_desc);
```
ayzhuang: Should default_desc be mkldnn_utils::create_default_mkldnn_md(node.get(), 0, true, default_format) for o_mds? Or you just let the next node to take care of it?

pruthviIntel: I guess here it make sense, since mkldnn keeps the result in the same layout as that of input.

ayzhuang: The output shape is not always the same as the input shape for pooling.

nmostafa: I would move that code out of the catch block as a special-handling in the try block.

jbobba: didn't quite follow your comment. We only need this if MKLDNN doesn't support the incoming layout and the only way to know that for now is to try and create a primitive descriptor.

PR #2531, opened 2019-03-01
Update build script to fix Windows build issues and add some work tow…

silee2:
…ards making codegen work on Windows.

PR #2532, opened 2019-03-01
Fixes on Windows

jfdong:
The changes include two parts: 
1. Enable using pre-installed MKLDNN on Window.
2. Use "git apply" instead of "patch" to have less dependency.  what's more, "patch" doesn't work very well on Windows.


silee2:
The PR will be merged in to #2531 and then #2531 will be merged into r0.15

silee2:
@rkimballn1 Removed you from reviewer since you will be reviewing #2531 which combine this PR

PR #2535, opened 2019-03-01
Cyphers/distdebug

diyessi:


postrational:
I think Travis managed to find something on CentOS:
```
[ 33%] Building CXX object src/ngraph/CMakeFiles/ngraph.dir/log.cpp.o
/root/ngraph/src/ngraph/log.cpp: In function 'std::string ngraph::get_timestamp()':
/root/ngraph/src/ngraph/log.cpp:91:18: error: 'put_time' is not a member of 'std'
     timestamp << std::put_time(&bt, "%H:%M:%S"); // HH:MM:SS
                  ^
```

diyessi:
`std::put_time` is in C++11 and comes from <iomanip> which is included. What version of centos is travis using?

postrational:
Travis is using the latest version of CentOS 7, which should mean 7.6

The entire environment is defined in this Dockerfile:
https://github.com/NervanaSystems/ngraph/blob/master/.ci/travis/centos/Dockerfile#L1

avijit-nervana:
@diyessi It turns out that gcc 4.8.5 (default compiler on CentOS 7) doesn't support the std::put_time - hence the error. I replaced that call and now the build should pass on CentOS. I am also going to fix this on r0.15 as well (as this would chock the TensorFlow bridge as well which is built with gcc 4.8).
@postrational Thank for bringing this up.

PR #2537, opened 2019-03-01
New macros for faster node validation checks

aprocter:
The current `NODE_VALIDATION_ASSERT` machinery is unfortunately kind of a drag on performance, because the use of streams forces us to build error messages even when we are not actually going to throw them. This adds a new macro called `NODE_VALIDATION_CHECK`, which pushes the explanation strings into the arguments of the macro, allowing them to be evaluated lazily.

This results in a big speedup of node validation for ops which have a lot of checks (like `Convolution`), and even a measurable improvement on ops that don't (like `Add`):

```
Time to construct 1,000,000 nodes
=================================

+------------+-----------------+----------------+
|            |  With old macros| With new macros|
|Op          |       Time (sec)|      Time (sec)|
+------------+-----------------+----------------+
|Add         |             5.2 |            3.4 |
|Convolution |            36.4 |            4.8 |
+------------+-----------------+----------------+
```

The underlying mechanism could eventually be used to replace `NGRAPH_ASSERT` as well.


aprocter:
BTW: we can do more ops in a follow-on PR. Wanted to do just the minimum to demonstrate that this was beneficial.

aprocter:
TESTNOW

silee2:
Strange failure
```
OutOfMemoryOrKill-9

detect hudson status 137 which indicates an out-of-memory condition has been encountered
Indication 1 
```

silee2:
Rebuilding python-ci

PR #2538, opened 2019-03-01
Use cpu kernel for constant folding.

ayzhuang:


src/ngraph/pass/constant_folding.cpp
```diff
@@ -193,16 +234,33 @@ void ngraph::pass::ConstantFolding::construct_constant_reshape()
...
 
-    runtime::reference::broadcast<T>(constant->get_vector<T>().data(),
-                                     out_vec.data(),
-                                     constant->get_shape(),
-                                     out_shape,
-                                     broadcast->get_broadcast_axes());
+    auto handler = m_cfmap.find(type_index(typeid(ngraph::op::Broadcast)));
+    if (handler != m_cfmap.end())
+    {
+        vector<T> in_vec = constant->get_vector<T>();
```
rkimballn1: Constant stores the data in an array of T internally. Use `constant->get_data_ptr<T>()` to get a pointer to the raw data. This will save a copy.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -260,52 +318,103 @@ void ngraph::pass::ConstantFolding::construct_constant_broadcast()
 template <class T>
 shared_ptr<op::Constant> make_constant_binary(shared_ptr<op::Constant> a,
                                               shared_ptr<op::Constant> b,
-                                              shared_ptr<Node> binary)
+                                              shared_ptr<Node> binary,
+                                              ngraph::BuildCFMap& m_cfmap)
 {
     auto out_shape = binary->get_shape();
     vector<T> out_vec(shape_size(out_shape));
+    vector<T> in_vec1 = a->get_vector<T>();
```
rkimballn1: get_data_ptr here too

src/ngraph/pass/constant_folding.cpp
```diff
@@ -390,25 +499,56 @@ bool is_supported_unary_op(std::shared_ptr<Node> n)
 
 template <class T>
 shared_ptr<op::Constant> make_constant_unary(shared_ptr<op::Constant> constant,
-                                             shared_ptr<Node> unary)
+                                             shared_ptr<Node> unary,
+                                             ngraph::BuildCFMap& m_cfmap)
 {
     auto out_shape = unary->get_shape();
     vector<T> out_vec(shape_size(out_shape));
+    vector<T> in_vec = constant->get_vector<T>();
```
rkimballn1: and here

src/ngraph/pass/constant_folding.hpp
```diff
@@ -40,9 +41,18 @@ class ngraph::pass::ConstantFolding : public ngraph::pass::GraphRewrite
         QUANTIZE
     };
 
-    ConstantFolding()
+    ConstantFolding(ngraph::Backend backend = ngraph::Backend::ANY)
         : GraphRewrite()
     {
+        switch (backend)
```
rkimballn1: This belongs in runtime/cpu/pass, not in ngraph pass

src/ngraph/util.hpp
```diff
@@ -214,12 +217,33 @@ namespace ngraph
...
     * the forward prop are added to the output of fprop and the input of the bprop
-    * to avoid repeat calcualtions.
+    * to avoid repeat calculations.
     * The last argument is the adjoints coming into the bprop function, the output
     * bprop function will have these nodes as the first N input parameters
     **/
     FpropCache cache_fprop(std::shared_ptr<Function> fprop, std::shared_ptr<Function> bprop);
 
+    // constant folding functions
+    enum class Backend
```
rkimballn1: If backends have specific constant folding algorithms then they belong in the backend, not in core. This enum should be removed. It is not complete.

ayzhuang: how about passing a bool variable use_cpu_kernel whose default value is false?

src/ngraph/pass/constant_folding.cpp
```diff
@@ -260,57 +332,64 @@ void ngraph::pass::ConstantFolding::construct_constant_broadcast()
 template <class T>
 shared_ptr<op::Constant> make_constant_binary(shared_ptr<op::Constant> a,
                                               shared_ptr<op::Constant> b,
-                                              shared_ptr<Node> binary)
+                                              shared_ptr<Node> binary,
+                                              ngraph::BuildNodeExecutorMap& m_cfmap)
```
jbobba: cfmap. m_ is only used for member variables.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -260,57 +332,64 @@ void ngraph::pass::ConstantFolding::construct_constant_broadcast()
...
+    outputs.push_back(out_vec.data());
+
+    auto& node = *binary;
+    if (!m_cfmap.empty())
     {
-        runtime::reference::maximum<T>(a->get_vector<T>().data(),
-                                       b->get_vector<T>().data(),
-                                       out_vec.data(),
-                                       shape_size(out_shape));
+        auto handler = m_cfmap.find(type_index(typeid(node)));
```
jbobba: why is the lookup done here instead of in the callback?

src/ngraph/pass/constant_folding.cpp
```diff
@@ -390,29 +469,47 @@ bool is_supported_unary_op(std::shared_ptr<Node> n)
...
+    inputs.push_back(const_cast<void*>(constant->get_data_ptr()));
+    vector<void*> outputs;
+    outputs.push_back(out_vec.data());
+
+    auto& node = *unary;
+    if (!m_cfmap.empty())
     {
-        runtime::reference::relu<T>(
-            constant->get_vector<T>().data(), out_vec.data(), shape_size(out_shape));
+        auto handler = m_cfmap.find(type_index(typeid(node)));
```
jbobba: can we lookup in callback and pass func* to this method?

src/ngraph/util.hpp
```diff
@@ -214,12 +217,19 @@ namespace ngraph
...
     * and turns them into clone functions where the intermediate values of
     * the forward prop are added to the output of fprop and the input of the bprop
-    * to avoid repeat calcualtions.
+    * to avoid repeat calculations.
     * The last argument is the adjoints coming into the bprop function, the output
     * bprop function will have these nodes as the first N input parameters
     **/
     FpropCache cache_fprop(std::shared_ptr<Function> fprop, std::shared_ptr<Function> bprop);
 
+    // node executor
```
jbobba: A slightly longer comment would be useful. "NodeExecutors are used in compiler optimization passes like ConstantFolding to execute a node using the supplied input and output memory locations. A BuildNodeExecutor returns a backend-specific NodeExecutor for a given Node type"

src/ngraph/runtime/cpu/builder/pad.cpp
```diff
@@ -100,6 +100,69 @@ namespace ngraph
...
+                {
+                    std::function<decltype(runtime::cpu::kernel::pad<float, 1>)> kernel;
+
+                    SELECT_KERNEL_BY_RANK(kernel,
+                                          pad->get_input_element_type(0),
+                                          arg_shape.size(),
+                                          runtime::cpu::kernel::pad);
+
+                    auto functor = [kernel, arg_shape, out_shape, padding_below, padding_above](
+                        const std::vector<void*> inputs, std::vector<void*> outputs) {
```
dcaballe: I see we are passing vector by value instead of by reference here and in many other places of this code. Is this intentional? Should we use references instead?

test/cpu_test.cpp
```diff
@@ -919,3 +921,203 @@ TEST(cpu_test, rotated_pooling)
...
+
+    ASSERT_EQ(count_ops_of_type<op::Reshape>(f), 0);
+    ASSERT_EQ(count_ops_of_type<op::Constant>(f), 1);
+
+    auto new_const =
+        std::dynamic_pointer_cast<op::Constant>(f->get_results().at(0)->get_argument(0));
+    ASSERT_TRUE(new_const);
+    auto values_out = new_const->get_vector<float>();
+
+    ASSERT_EQ(values_in, values_out);
```
diyessi: Use `all_close_f` in this test and the ones below.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -157,29 +198,38 @@ void pass::ConstantFolding::construct_constant_reshape()
         auto constant_match = static_pointer_cast<op::Constant>(pattern_map[constant_label]);
         auto reshape_match = static_pointer_cast<op::Reshape>(m.get_match_root());
 
+        NodeExecutorTy func = nullptr;
+        if (!m_cfmap.empty())
+        {
+            auto handler = m_cfmap.find(type_index(typeid(ngraph::op::Reshape)));
+            NGRAPH_ASSERT(handler != m_cfmap.end())
```
rkimballn1: I believe there are more modern versions of NGRAPH_ASSERT, @aprocter?

aprocter: Sorry, I missed this tag. `NGRAPH_ASSERT` has not been modernized yet. This is the right thing to use.

test/cpu_test.cpp
gcwenger: @ayzhuang made this change.

PR #2546, opened 2019-03-03
Switch everything from NODE_VALIDATION_ASSERT to NODE_VALIDATION_CHECK; remove NODE_VALIDATION_ASSERT

aprocter:
This switches every use of `NODE_VALIDATION_ASSERT` over to the new, faster, `NODE_VALIDATION_CHECK`, and removes `NODE_VALIDATION_ASSERT`. (See #2537 for more details on the new macro.)

The vast majority of changes here are completely mechanical. The one exception is that `NODE_VALIDATION_ASSERT` itself has actually been removed; that makes this an API-breaking change.

Any client code (mainly backend-custom ops) doing something like this:

```
NODE_VALIDATION_ASSERT(node, cond) << "foo" << "bar";
```

will have to switch to:

```
NODE_VALIDATION_CHECK(node, cond, "foo", "bar");
```


src/ngraph/node.hpp
```diff
@@ -321,12 +308,5 @@ namespace ngraph
...
-#define NODE_VALIDATION_ASSERT(node, cond)                                                         \
-    NGRAPH_ASSERT_STREAM_WITH_LOC(                                                                 \
-        ::ngraph::NodeValidationError, cond, ::ngraph::node_validation_assertion_string(node))
-#define NODE_VALIDATION_FAIL(node)                                                                 \
-    NGRAPH_FAIL_STREAM_WITH_LOC(::ngraph::NodeValidationError,                                     \
-                                ::ngraph::node_validation_assertion_string(node))
-
 #define NODE_VALIDATION_CHECK(node, cond, ...)                                                     \
-    CHECK(::NodeValidationFailure, (node), (cond), __VA_ARGS__)
+    CHECK(::ngraph::NodeValidationFailure, (node), (cond), __VA_ARGS__)
```
aprocter: (No idea how it was working before without this, but whatever.)

PR #2552, opened 2019-03-04
Return 0 or 1 when converting to bool

arogowie-intel:
This PR is a update for #2314 It addresses Review comments.


src/ngraph/runtime/reference/convert.hpp
```diff
@@ -25,13 +25,22 @@ namespace ngraph
         namespace reference
         {
             template <typename TI, typename TO>
-            void convert(const TI* arg, TO* out, size_t count)
+            void convert(const TI* arg, TO* out, std::size_t count)
```
rkimballn1: We don't use std::size_t in ngraph. Please use just size_t.

src/ngraph/runtime/reference/convert.hpp
```diff
@@ -25,13 +25,22 @@ namespace ngraph
...
             {
-                for (size_t i = 0; i < count; i++)
+                for (std::size_t i = 0; i < count; ++i)
```
rkimballn1: Same here, just size_t

src/ngraph/runtime/reference/convert.hpp
```diff
@@ -25,13 +25,22 @@ namespace ngraph
...
                 {
                     out[i] = static_cast<TO>(arg[i]);
                 }
             }
+
+            template <typename T>
+            void convert_to_bool(const T* arg, char* out, std::size_t count)
```
rkimballn1: size_t

src/ngraph/runtime/reference/convert.hpp
```diff
@@ -25,13 +25,22 @@ namespace ngraph
...
+            {
+                for (std::size_t i = 0; i < count; ++i)
```
rkimballn1: size_t

PR #2553, opened 2019-03-04
[ONNX] Expose onnx_import functions through Python API

postrational:


python/setup.py
```diff
@@ -397,7 +394,6 @@ def build_extensions(self):
     cmdclass={'build_ext': BuildExt},
     data_files=data_files,
     setup_requires=['numpy'],
-    install_requires=requirements,
```
postrational: We need to put this back, perhaps in another form.

diyessi:
Obsolete

PR #2575, opened 2019-03-07
Silee2/fix prebuilt mkldnn

silee2:
pre-installed mkldnn support was broken. This PR fixes it. Same as #2574 for r0.15


aslepko:
TESTNOW

PR #2576, opened 2019-03-08
Don't segfault trying to be informative about empty tensors

diyessi:


test/util/all_close_f.cpp
```diff
@@ -238,6 +238,11 @@ ::testing::AssertionResult
     {
         return ::testing::AssertionFailure() << "a.size() != b.size() for all_close_f comparison.";
     }
+    if (a.size() == 0)
+    {
+        msg << "No elements to compare";
```
gcwenger: Nothing being done w/ msg in this path.
Either use it like this:.
`return ::testing::AssertionSuccess() << msg.str();`
or skip it like this:
`return ::testing::AssertionSuccess() << "No elements to compare";`

test/util/all_close_f.cpp
```diff
@@ -338,6 +343,12 @@ ::testing::AssertionResult
...
```
gcwenger: Nothing being done w/ msg in this path.
Either use it like this:.
return ::testing::AssertionSuccess() << msg.str();
or skip it like this:
return ::testing::AssertionSuccess() << "No elements to compare";

PR #2582, opened 2019-03-08
Fix debugger tests crash for CODEGEN.

ayzhuang:


src/ngraph/runtime/cpu/cpu_debugger.cpp
```diff
@@ -129,8 +129,17 @@ bool runtime::cpu::CPU_Debugger::delete_breakpoint(std::shared_ptr<Node> op)
 
 void* runtime::cpu::CPU_Debugger::inspect(std::shared_ptr<Node> op, size_t output_index)
 {
-    return m_callframe.m_external_function->get_tensor_data(op->get_name() + "_" +
-                                                            to_string(output_index));
+    if (std::getenv("NGRAPH_CODEGEN") == nullptr ||
```
jbobba: use m_external_function->is_direct_execution()

test/cpu_debugger.cpp
```diff
@@ -38,8 +38,14 @@
 using namespace ngraph;
 using namespace std;
 
+// These tests are for DEX mode only.
 TEST(debugger, add_breakpoint)
 {
+    if (std::getenv("NGRAPH_CODEGEN") != nullptr &&
```
jbobba: Can we extract this out into a static method? Gtest probably has a better way to do this.

test/cpu_debugger.cpp
```diff
@@ -38,8 +38,14 @@
...
+        std::string(std::getenv("NGRAPH_CODEGEN")) != "0")
+    {
+        return;
```
dcaballe: My only concern is that now we are returning PASSED when the test wasn't actually run. Is there a way to just ignore the test based on NGRAPH_CODEGEN? I did a quick search and found GTEST_SKIP. Maybe it's worth having a look.

ayzhuang: Thank you. GTEST_SKIP was introduced last year. But our gtest is older version and does not have it. Need to update gtest.

ayzhuang: The latest gtest release dates back to Aug 31, 2018, which doe not have GTEST_SKIP. Will create another PR when there is a new release.

dcaballe: What could we do in the meantime? Should we raise an assert saying that the test is not supported in codegen mode better than return pass? I feel that if we return pass when then test didn't actually run we will lose track of which tests actually pass in codegen mode.

src/ngraph/runtime/cpu/cpu_debugger.cpp
```diff
@@ -129,8 +129,16 @@ bool runtime::cpu::CPU_Debugger::delete_breakpoint(std::shared_ptr<Node> op)
 
 void* runtime::cpu::CPU_Debugger::inspect(std::shared_ptr<Node> op, size_t output_index)
 {
```
diyessi: Can we make this a virtual on callframe?

test/cpu_debugger.cpp
```diff
@@ -68,6 +79,24 @@ TEST(debugger, add_breakpoint)
...
+        {
+            EXPECT_EQ(e.what(), std::string("_Map_base::at"));
+            return;
+        }
+        catch (...)
+        {
+            FAIL() << "Expected std::exception";
+        }
+    }
+
```
nmostafa: can we outline this into a function ? Too much code duplication.

test/cpu_debugger.cpp
```diff
@@ -68,6 +79,24 @@ TEST(debugger, add_breakpoint)
...
+        catch (const std::exception& e)
+        {
+            EXPECT_EQ(e.what(), std::string("_Map_base::at"));
+            return;
+        }
+        catch (...)
+        {
+            FAIL() << "Expected std::exception";
+        }
+    }
```
dcaballe: What was the conclusion about unsupported tests? Make them pass?
It we cannot do any better for now, I would suggest that we just return.  Code in line 87/91 seems very fragile to me. If the way of "crashing" changes, we would have to update the test to the new way of crashing :). What do you think?

dcaballe: If agreed, I would move this to the beginning of the test.

ayzhuang: Yes, you can move it. Please put a TODO there to switch to skip once there is a new release of gtest.

dcaballe: Sorry, my bad. In my previous message, I wanted to suggest that you could move the code as part of this PR. I didn't mean I wanted to intrude on your changes and do the coding myself, sorry :). Since you are modifying this code, could you please take care of that?

PR #2584, opened 2019-03-08
nvgpu backend check if the ptr pass to create_tensor is device pointer

fengleitian:


src/ngraph/runtime/gpu/gpu_backend.cpp
```diff
@@ -114,6 +115,10 @@ shared_ptr<runtime::Tensor>
 shared_ptr<runtime::Tensor> runtime::gpu::GPU_Backend::create_tensor(
     const element::Type& element_type, const Shape& shape, void* memory_pointer)
 {
+    if (memory_pointer != nullptr && !is_device_pointer(memory_pointer))
+    {
+        throw ngraph_error("The pointer passed to create_tensor is not a device pointer.");
```
rkimballn1: We can create a device tensor here and copy the data over, but that might be a separate PR.

src/ngraph/runtime/gpu/gpu_tensor.cpp
```diff
@@ -39,11 +39,15 @@ runtime::gpu::GPUTensor::GPUTensor(const ngraph::element::Type& element_type,
         std::make_shared<ngraph::descriptor::layout::DenseTensorLayout>(*m_descriptor));
 
     m_buffer_size = shape_size(shape) * element_type.size();
-    if (memory_pointer != nullptr)
+    if (memory_pointer != nullptr && is_device_pointer(memory_pointer))
```
rkimballn1: Have one if with memory_pointer != nullptr then inside check for is_device_pointer. It will look a little cleaner.

src/ngraph/runtime/gpu/gpu_util.cpp
```diff
@@ -62,6 +62,17 @@ void runtime::gpu::free_gpu_buffer(void* buffer)
     }
 }
 
+bool runtime::gpu::is_device_pointer(const void *ptr)
+{
+  cudaPointerAttributes attributes;
```
rkimballn1: Indenting looks wrong.

src/ngraph/runtime/gpu/gpu_util.cpp
```diff
@@ -62,6 +62,17 @@ void runtime::gpu::free_gpu_buffer(void* buffer)
...
+  CUDA_RT_SAFE_CALL_NO_THROW(cudaPointerGetAttributes(&attributes, ptr));
+  if(attributes.devicePointer != NULL)
```
rkimballn1: nullptr, not NULL

fengleitian: sharp eye. missed "style-apply"

fengleitian: fixed.

src/ngraph/runtime/gpu/gpu_util.cpp
fengleitian: sharp eye. missed "style-apply"

src/ngraph/runtime/gpu/gpu_tensor.cpp
fengleitian: agree. fixed.

src/ngraph/runtime/gpu/gpu_backend.cpp
fengleitian: ya, I was also think in that way. one concern was user might assume the pointer will be used by backend. which is not true. it's likely user will miss the data sync.

fengleitian: We can handle that in another PR.

AyanmoI:
LGTM

PR #2585, opened 2019-03-09
CPU Codegen running unit tests

rkimballn1:
I wanted to try out an idea to get the codegen cpu running unit tests for CI. This is what I came up with. Not complete in that it does not handle the case where you use PassConfig on the codegen backend but that should be easy to add. I wanted to get your opinions on this.

This makes a new backend called `CPU_CODEGEN`. This new backend is created by a simple copy of the `libcpu_backend.so` file to `libcpu_codegen_backend.so`. Once you have that the rest is straightforward.

The codegen backend does not pass CI and segfaults at one point so I set DEX_ONLY to ON.


jbobba:
Instead of copying backends, can we leverage backend attributes? https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/backend_manager.cpp#L74
We will either have to change the separator to '_' or make changes to the unit test cmake file to understand "CPU:CODEGEN"

Some secondary concerns
- We need to make unit_test.manifest work for this as well.
- I dont think we can run all the tests under codegen in CI. It would slow things down too much.
If we want to run under CI, we need to select/enable a subset (perhaps through the manifest)

PR #2587, opened 2019-03-09
Cyphers/convback WIP

diyessi:


diyessi:
#2593 is a cleaned up replacement.

PR #2591, opened 2019-03-11
Adds basic support for provenance

tachyon77:
Adds methods in Node class to add, merge provenance tags
Adds logic to merge tags in ngraph::replace_node()


src/ngraph/node.cpp
```diff
@@ -165,6 +165,29 @@ void Node::set_placement_index(size_t placement)
...
+{
+    m_provenance_tags.insert(tag);
+}
+
+size_t Node::remove_provenance_tag(const std::string& tag)
+{
+    return m_provenance_tags.erase(tag);
+}
+
+void Node::merge_provenance_tags_from(std::shared_ptr<const Node> source)
```
aprocter: We usually pass shared pointers by constant reference, i.e. `const std::shared_ptr<const Node>&`.

src/ngraph/node.hpp
```diff
@@ -245,6 +245,13 @@ namespace ngraph
         /// Set device placement
         void set_placement_index(size_t placement);
 
+        std::unordered_set<std::string> get_provenance_tags() const;
```
aprocter: Let's return a const reference here.

src/ngraph/node.hpp
```diff
@@ -245,6 +245,13 @@ namespace ngraph
...
+        void add_provenance_tag(const std::string& tag);
+        size_t remove_provenance_tag(const std::string& tag);
```
aprocter: I don't think there's any need to return the `size_t` returned by `erase` --- let's make it `void`.

test/op.cpp
```diff
@@ -41,3 +43,35 @@ TEST(op, is_parameter)
...
+    node->add_provenance_tag(tag2);
+
+    node->remove_provenance_tag(tag1);
+
+    auto tags = node->get_provenance_tags();
+    ASSERT_TRUE(tags.find(tag1) == tags.end());
+    ASSERT_TRUE(tags.find(tag2) != tags.end());
+}
+
+// TODO: Need to mock Node, Op etc to be able to unit test functions like replace_node().
```
aprocter: (We've already chatted a bit about this on Slack, just summarizing my understanding on the PR thread.) Let's work without mocks for now, until we have a chance to look into applying them more broadly.

src/ngraph/graph_util.cpp
```diff
@@ -165,6 +165,7 @@ void ngraph::replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> re
             input->replace_output(replacement->get_outputs().at(i));
         }
     }
+    replacement->merge_provenance_tags_from(target);
```
aprocter: Yeah, this thing here is exactly the piece that will get more complex as we go. :)

test/op.cpp
tachyon77: In my updated push, I left my commented code for future reminder if that's ok.

aprocter:
BTW, doesn't matter this time, but we usually put our username (or some semblance thereof) in the branch names. For example, instead of `foo` I would call a branch `aprocter/foo`.

src/ngraph/node.hpp
```diff
@@ -268,6 +275,7 @@ namespace ngraph
         std::string m_friendly_name;
         const std::string m_unique_name;
         static std::atomic<size_t> m_next_instance_id;
+        std::unordered_set<std::string> m_provenance_tags;
```
adstraw: would it make sense to make this a set of *const* std::string?  would we ever want to change a provenance tag once it is added?

tachyon77: It did sound more robust to me too, however my finding is, it is not a valid C++. I did a local testing in my VS (unordered_set<const string> s;) and it says "error C2338: The C++ Standard forbids containers of const elements because allocator<const T> is ill-formed." I was also not able to get it compiled on the linux server in our lab.

PR #2592, opened 2019-03-11
QuantizedDot and QuantizedDotBias ops CPU impl

ashokei:


diyessi:
#2593 is a cleaned up replacement.

diyessi:
Clicked on wrong line.

nbpatel:
where is the scale getting computed? Also, don't see quantization.cpp

ashokei:
@nbpatel similar to QuantizedConv , the scale is computed in the builders. 
Added builders, scale computation, and unittests.

src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -42,8 +42,7 @@ set(SRC
     builder/concat.cpp
     builder/convert.cpp
     builder/convert_layout.cpp
-    builder/quantized_conv.cpp
-    builder/quantized_concat.cpp
+    builder/quantized_dot.cpp
```
silee2: Remove this line. It also shows up at line 66.

PR #2593, opened 2019-03-11
Remove _backwards from conv backprop ops.

diyessi:
This was a PR several months ago to do some cleanups to convolution ops and reference implementations, which ran into trouble on igpu. Everything seems to work now. There are additional name cleanups to be done in the op, but that should be a separate PR.

### DIFF
--- a/src/ngraph/op/convolution.hpp
+++ b/src/ngraph/op/convolution.hpp
@@ -320,6 +320,6 @@ namespace ngraph
                                                  size_t output_channel_axis_filters,
                                                  size_t batch_axis_result,
                                                  size_t output_channel_axis_result);
-        } // namespace util
-    }     // namespace op
-} // namespace ngraph
+        }
+    }
+}
--- a/src/ngraph/op/convolution.hpp
+++ b/src/ngraph/op/convolution.hpp
@@ -220,31 +220,9 @@ namespace ngraph
                 return m_data_dilation_strides_forward;
             }
 
-            /// \return The window movement strides for the backward prop.
-            const Strides& get_window_movement_strides_backward() const
-            {
-                return m_window_movement_strides_backward;
-            }
-            /// \return The window dilation strides for the backward prop.
-            const Strides& get_window_dilation_strides_backward() const
-            {
-                return m_window_dilation_strides_backward;
-            }
-            /// \return The padding-below sizes (possibly negative) for the backward prop.
-            const CoordinateDiff& get_padding_below_backward() const
-            {
-                return m_padding_below_backward;
-            }
-            /// \return The padding-above sizes (possibly negative) for the backward prop.
-            const CoordinateDiff& get_padding_above_backward() const
-            {
-                return m_padding_above_backward;
-            }
-            /// \return The input data dilation strides for the backward prop.
-            const Strides& get_data_dilation_strides_backward() const
-            {
-                return m_data_dilation_strides_backward;
-            }
+            // Compute the pad_above values to be used if in a convolution
+            CoordinateDiff compute_backward_delta_out_pad_above() const;
+            CoordinateDiff compute_backward_delta_out_pad_below() const;
 
         protected:
             Shape m_data_batch_shape;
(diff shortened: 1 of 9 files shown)


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -1728,11 +1728,11 @@ shared_ptr<runtime::Executable>
 
             const shared_ptr<op::ConvolutionBackpropData> conv_op =
                 static_pointer_cast<op::ConvolutionBackpropData>(op);
-            const Strides& win_stride = conv_op->get_window_movement_strides_backward();
-            const CoordinateDiff& pad_below = conv_op->get_padding_below_backward();
-            const CoordinateDiff& pad_above = conv_op->get_padding_above_backward();
-            const Strides& win_dilation = conv_op->get_window_dilation_strides_backward();
-            const Strides& data_dilation = conv_op->get_data_dilation_strides_backward();
+            const Strides& win_stride = conv_op->get_data_dilation_strides_forward();
+            CoordinateDiff pad_below = conv_op->compute_backward_delta_out_pad_below();
```
shssf: I don't understand why you removed "&" (and `const`). Is it necessary to do `memvove` here?

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -1629,11 +1629,11 @@ shared_ptr<runtime::Executable>
             const shared_ptr<op::ConvolutionBackpropFilters> conv_op =
                 static_pointer_cast<op::ConvolutionBackpropFilters>(op);
 
-            const Strides& win_stride = conv_op->get_window_movement_strides_backward();
-            const CoordinateDiff& pad_below = conv_op->get_padding_below_backward();
-            const CoordinateDiff& pad_above = conv_op->get_padding_above_backward();
-            const Strides& win_dilation = conv_op->get_window_dilation_strides_backward();
-            const Strides& data_dilation = conv_op->get_data_dilation_strides_backward();
+            const Strides& win_stride = conv_op->get_window_dilation_strides_forward();
```
shssf: Logically, if you operate with "backprop" you expect something like "_backwards". If all these "backward" and "forward" has no meaning, please make the name shorter, without suffixes.

src/ngraph/op/convolution.hpp
```diff
@@ -377,6 +320,6 @@ namespace ngraph
                                                  size_t output_channel_axis_filters,
                                                  size_t batch_axis_result,
                                                  size_t output_channel_axis_result);
-        }
-    }
-}
+        } // namespace util
+    }     // namespace op
```
rkimballn1: The indenting of the namespace comment is inconsistent on all files which have it. I would just remove these comments. We don't use them elsewhere in ngraph core

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
rkimballn1: If I had to guess I would say it is because the function that uses it below did not mark this input as const and a non-const reference is bad style unless you really want that.

diyessi: No, it's because it's returning a computed value; you have the option of computing that value once and stashing it yourself, but using a version stashed in the op isn't a good thing since in the future the op might not be there when you execute.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
diyessi: That's the second cleanup I alliuded too; "window movement" versus "filter" etc.

src/ngraph/op/convolution.hpp
diyessi: clang-format seems to be inserting those comments. I remove them and reformat and they come back.

diyessi: Or it's VSC adding them; I am removing them but they are likely to return in the future.

PR #2594, opened 2019-03-11
(Dynamic) Transpose op

aprocter:
This adds an experimental `Transpose` op, which takes the axis permutation as an input of type `int64`.

Currently this is just the construction API---no `INTERPRETER` implementation, and no execution tests as yet. The reason for this is that we'll need to lay some groundwork in the form of some shape-specialization passes before we can implement those. I think that is best done in a separate PR, but it presents a bit of a chicken-and-egg problem.


src/ngraph/op/experimental/transpose.cpp
```diff
@@ -0,0 +1,59 @@
...
+op::Transpose::Transpose(const shared_ptr<Node>& arg, const shared_ptr<Node>& input_order)
+    : Op("Transpose", check_single_output_args({arg, input_order}))
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::Transpose::validate_and_infer_types()
+{
+    NODE_VALIDATION_CHECK(this,
+                          get_output_element_type(0).compatible(element::i64),
```
louisfeng: Should this be `get_input_element_type(1)`?

aprocter: Yes! Good catch. Please request changes.

PR #2601, opened 2019-03-12
Converted some unit tests to all_close_f

gcwenger:


test/convolution_test.in.cpp
```diff
@@ -86,7 +87,11 @@ NGRAPH_TEST (${BACKEND_NAME}, convolution_2d_1item)
 
     auto handle = backend->compile(function);
     handle->call_with_validate({result}, {a, b});
-    EXPECT_TRUE(test::all_close<float>(vector<float>{expected_result}, read_vector<float>(result), 1.0e-4f, 1.0e-6f));
+    // for float this will be 18 bits matching
```
rkimballn1: Can we make this into a util function or something? It is a lot of copied code

gcwenger: Reworked.

PR #2606, opened 2019-03-13
[ONNX] LSTM mixed length sequences.

arogowie-intel:


test/onnx_import.in.cpp
```diff
@@ -2227,6 +2227,101 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, import_non_existing_file)
...
+    handle->call_with_validate(result_tensors, arg_tensors);
+
+    Outputs outputs;
+    for (auto rt : result_tensors)
+    {
+        outputs.push_back(read_vector<float>(rt));
+    }
+
+    EXPECT_TRUE(outputs.size() == expected_output.size());
+    for (std::size_t i{0}; i < expected_output.size(); ++i)
```
tomdol: Just a very low importance comment: `std::size_t -> size_t` for consistency

arogowie-intel: Actually I'm adding this prefix few lines above. Within this file we're using `std::` prefixes.

PR #2607, opened 2019-03-13
add unit tests for the two versions of Backend create_tensor

rkimballn1:
There are currently no unit tests to check that backends implement the three parameter version of create_tensor. This PR adds a test for each of the two create_tensor methods to verify that backends handle them correctly.


silee2:
[  FAILED  ] 1 test, listed below:
[  FAILED  ] GPU.create_tensor_2

 1 FAILED TEST
  YOU HAVE 115 DISABLED TEST

PR #2610, opened 2019-03-13
Added sqrt to constant folding

fboemer:
Before, the `CoreFusion` pass followed by `ConstantFolding` pass doesn't simplify `BatchNormInference` all the way, due to the `sqrt` op.

Before: 
after

After:
after


src/ngraph/pass/constant_folding.cpp
```diff
@@ -410,6 +412,11 @@ shared_ptr<op::Constant> make_constant_unary(shared_ptr<op::Constant> constant,
         runtime::reference::relu<T>(
             constant->get_vector<T>().data(), out_vec.data(), shape_size(out_shape));
     }
+    else if (std::dynamic_pointer_cast<op::Sqrt>(unary))
+    {
+        runtime::reference::sqrt<T>(
```
diyessi: Check for negative values and throw an error if there are any, and add a test for it.

fboemer: Done

PR #2612, opened 2019-03-14
Add NodeInput and NodeOutput classes

aprocter:
This PR adds two classes:

- `NodeInput` is a handle for a particular input to a node. Internally it is just a `<Node *,size_t>` pair, where the `size_t` component represents the index of the input being referred to.
- `NodeOutput` is a handle for a particular output from a node. Internally it is just a `<shared_ptr<Node>,size_t>` pair, where the `size_t` component represents the index of the output being referred to.

They also contain accessor functions for things like shapes, element types, connected sources/targets, etc.

The idea is that these classes can be used in future PRs to:

1. replace direct access to `descriptor::Input` and `descriptor::Output`;
2. eliminate the need for `GetOutputElement` by changing node constructor signatures to take `NodeOutput`s as arguments (an implicit conversion from `shared_ptr<Node>` to `NodeOutput`, with a default output index of 0, makes it almost completely backwards compatible with existing graph-constructing code).

This is the first in a series of about 10 or 12 planned PRs (with number 2 and number 3 already in the works) aimed at simplifying the `Node` class and eliminating `GetOutputElement`.


src/ngraph/node_input.hpp
```diff
@@ -0,0 +1,95 @@
...
+        NodeInput(Node* node, size_t index)
+            : m_node(node)
+            , m_index(index)
+        {
+        }
+
+        /// \brief Constructs a NodeInput.
+        /// \param node Shared pointer to the node for the input handle.
+        /// \param index The index of the input.
+        NodeInput(std::shared_ptr<Node> node, size_t index)
```
rkimballn1: How about we just get rid of this signature? It is easy enough to call the other signature. I'm on a campaign to stomp out shared_ptr everywhere it is not required.

src/ngraph/node_output.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include "ngraph/node.hpp"
+
+namespace ngraph
+{
+    /// \brief A handle for one of a node's outputs.
+    class NodeOutput
+    {
+    public:
+        /// \brief Constructs a NodeOutput.
+        /// \param node Pointer to the node for the output handle.
```
rkimballn1: shared_ptr<Node>, not pointer. Ugh, shared_ptr. So the output it the tie that binds. That which holds the graph together. The asymmetry between the NodeInput constructor signatures and the output of get_node on the two classes is ugly. Symmetry is better. Any way to achieve that? 
What horrible things would happen if instead of spewing shared_ptr everywhere we used raw pointer and did things the old-fashioned way, with new and delete?

src/ngraph/node_output.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+#pragma once
+
+#include "ngraph/descriptor/input.hpp"
+#include "ngraph/node.hpp"
+
+namespace ngraph
+{
+    /// \brief A handle for one of a node's outputs.
+    class NodeOutput
```
rkimballn1: I would love to see both NodeInput and NodeOutput derived some the same common abstract base class. There is so much in common or that should be in common that it really would make things nice I think.

src/ngraph/node_input.hpp
aprocter: I'd be fine with that!

src/ngraph/node_output.hpp
aprocter: > shared_ptr, not pointer.

True, will change :)

In general I don't much like our use of `shared_ptr` either. I have some ideas on how to remedy that, but I think it can be addressed more or less independently of this work. One thing we _could_ do, I suppose, is store a raw pointer internally and leverage `shared_from_this`. That feels a bit like trading a venial sin for a mortal one, but then again it'd get us the symmetry we want and force users of `NodeOutput` to at least choose explicitly which flavor of pointer they want.

src/ngraph/node_output.hpp
aprocter: This is an interesting idea (maybe something like `NodePort` with `is_input` and `is_output` virtuals? though the virtual means every one of these objects now has its own vtable)---let me think about it and get back to you.

src/ngraph/node_input.hpp
aprocter: Done

src/ngraph/node_output.hpp
diyessi: I am a bit against it unless the abstract class is some internal class; the implementations are similar, but the API/semantics are separate.

jbobba:
Are these meant to be temporary handles that are used locally or things that would sit around in the Node?

aprocter:
@jbobba For the moment, temporary handles. The next step after this is to make `descriptor::Input` and `descriptor::Output` private, meaning they could eventually be eliminated and possibly replaced by internal storage of `NodeOutput` vectors as the owning reference.

PR #2614, opened 2019-03-14
[ONNX] Fix default axis for legacy broadcasting

tsocha:


tsocha:
> 
> 
> In order not to repeat this code, we should fold calculating the `axis` into the `legacy_style_broadcast_for_binary_operation` function.

This code is used for default `value` for `axis` and because of that I made a new function instead of calculating axis in `legacy_style_broadcast_for_binary_operation`

src/ngraph/frontend/onnx_import/utils/broadcasting.hpp
```diff
@@ -125,5 +125,10 @@ namespace ngraph
                 new_shape,
                 calculate_broadcast_axes(new_shape, node->get_shape(), start_match_axis));
         }
+
+        inline std::size_t get_default_axis(std::vector<std::shared_ptr<ngraph::Node>> inputs)
```
arogowie-intel: What if the size of `inputs` is larger than 2?

src/ngraph/frontend/onnx_import/op/add.hpp
```diff
@@ -31,7 +31,8 @@ namespace ngraph
             {
                 inline NodeVector add(const Node& node)
                 {
-                    auto axis = node.get_attribute_value<int64_t>("axis", 0);
+                    auto axis = node.get_attribute_value<int64_t>(
```
arogowie-intel: Please use `std::` prefix.

src/ngraph/frontend/onnx_import/op/div.hpp
```diff
@@ -31,7 +31,8 @@ namespace ngraph
             {
                 inline NodeVector div(const Node& node)
                 {
-                    auto axis = node.get_attribute_value<int64_t>("axis", 0);
+                    auto axis = node.get_attribute_value<int64_t>(
```
arogowie-intel: Please use `std::` prefix.

src/ngraph/frontend/onnx_import/op/mul.hpp
```diff
@@ -32,7 +32,8 @@ namespace ngraph
             {
                 inline NodeVector mul(const Node& node)
                 {
-                    auto axis = node.get_attribute_value<int64_t>("axis", 0);
+                    auto axis = node.get_attribute_value<int64_t>(
```
arogowie-intel: Please use `std::` prefix.

src/ngraph/frontend/onnx_import/op/sub.hpp
```diff
@@ -31,7 +31,8 @@ namespace ngraph
             {
                 inline NodeVector sub(const Node& node)
                 {
-                    auto axis = node.get_attribute_value<int64_t>("axis", 0);
+                    auto axis = node.get_attribute_value<int64_t>(
```
arogowie-intel: Please use `std::` prefix.

src/ngraph/frontend/onnx_import/utils/broadcasting.hpp
```diff
@@ -125,5 +125,10 @@ namespace ngraph
...
+        {
+            return inputs.at(0)->get_shape().size() - inputs.at(1)->get_shape().size();
```
arogowie-intel: If you use only two elements from the vector please enforce it appropriately.

src/ngraph/frontend/onnx_import/utils/broadcasting.hpp
tsocha: The legacy style broadcasting force broadcast from right to left operand and there is always only 2 elements.

PR #2625, opened 2019-03-15
Do not allow in place slice if the arg is in CONSTANT buffer set.

ayzhuang:


test/cpu_test.cpp
```diff
@@ -871,6 +871,29 @@ TEST(cpu_test, memory_reuse_in_place_slice_after_in_place_concat)
...
+    auto neg = make_shared<op::Negative>(slice);
+    auto f = make_shared<Function>(neg, ParameterVector{});
+
+    auto backend = runtime::Backend::create("CPU");
+
+    auto result = backend->create_tensor(element::f32, shape);
+
+    auto handle = backend->compile(f);
+    handle->call_with_validate({result}, {});
+    EXPECT_EQ((vector<float>{-5., -6., -7., -8.}), read_vector<float>(result));
```
diyessi: Use all_close_f with strictest tolerance for the comparison.

PR #2632, opened 2019-03-18
Make convolution test generator work with py3

diyessi:
Also correct formatting.


aprocter:
@diyessi GitHub seems to think that the diff touches some unrelated files. Could you merge from `master` to see if that unconfuses it?

diyessi:
It doesn't. I'll start over.

PR #2634, opened 2019-03-18
Concat Elimination and Self Concat Fusion pass

AyanmoI:
Pass in action shown in resnet50 fprop graph:
rsnet_50

Examples of the transform are discussed below using the test cases.
The tests added deal with the following cases:
1) Single branch

single_branch

2) Multiple branches

multiple_branches_1

3) Multiple branches but one of the branches has a single self concat which is not worth replacing

multiple_branches_2

4) There is a pattern of self concats, but one of them concats along an axis where the shape val is not 1. Hence, the fusion is skipped

non_fusable_self_concat

5) There is a self concat pattern which is fusable but there is a fan out. Hence, the pattern is not fused (fused accordingly)

self_concat_with_fan_out

 - Opportunity for further optimizations:
1) Reshape elimination where a group of contiguous reshapes, reshape the initial input back to the same shape (core pass).
2) Broadcast fusion (elimination). In case 3, a fusable broadcast pattern can be seen (core pass)

Haven't seen a broadcast fusion opportunity in a real graph yet, but there is a pattern where reshapes can be eliminated in resnet50.


AyanmoI:
Still pushing commits but more or less ready for review

src/ngraph/pass/concat_fusion.cpp
```diff
@@ -0,0 +1,304 @@
...
+#include "ngraph/pattern/op/label.hpp"
+#include "ngraph/pattern/op/skip.hpp"
+#include "ngraph/util.hpp"
+
+using namespace ngraph;
+
+bool check_self_concat_op(const std::shared_ptr<Node>& op)
+{
+    auto input_args = op->get_arguments();
+    std::set<std::shared_ptr<Node>> input_args_set(input_args.begin(), input_args.end());
```
XWangIntel: The comparison of `std::shared_ptr<Node>` is kind of tricky. Use raw pointer Node* or node name is more readable.

src/ngraph/pass/concat_fusion.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+    auto callback = [op_label](pattern::Matcher& m) {
+        NGRAPH_DEBUG
+            << "concat_elimination: In callback for construct_concat_elimination against node = "
+            << m.get_match_root()->get_name();
+        auto pattern_map = m.get_pattern_map();
+        auto op = pattern_map[op_label];
+
+        auto root = std::dynamic_pointer_cast<op::Concat>(m.get_match_root());
+        if (root && (root->get_input_shape(0) == root->get_output_shape(0)))
```
XWangIntel: The input shape and output shape are the same for `Concat`?

src/ngraph/pass/concat_fusion.cpp
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    if (!concat_source_found)
+    {
+        this->m_concat_pattern_vectors.push_back(
+            make_pair(NodeVector{concat_op}, std::vector<size_t>{concat_axis}));
+    }
+}
+
+void ngraph::pass::SelfConcatFusion::remove_single_concat_op_pattern()
```
XWangIntel: Do we really need to remove single concat? What will happen if concat fusion work on a single concat?

src/ngraph/pass/concat_fusion.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+private:
+    void update_concat_pattern_vectors(const std::shared_ptr<Node>&, size_t);
+    void remove_single_concat_op_pattern();
+    void construct_concat_patterns(const std::shared_ptr<pattern::Matcher>&,
+                                   const std::shared_ptr<pattern::op::Label>&,
+                                   const std::shared_ptr<Node>&);
+    bool replace_patterns(const std::pair<NodeVector, std::vector<size_t>>&);
+
+    std::vector<std::pair<NodeVector, std::vector<size_t>>> m_concat_pattern_vectors;
```
XWangIntel: Instead of keep all the self concat in a NodeVector, can we just keep the last `Concat` in the vector, or unordered_set, which can simplify the code a lot. The parent concat nodes can be easily deduced from the last one.

src/ngraph/pass/concat_fusion.cpp
AyanmoI: Yes..for elimination of a concat in this case, the shapes will be same...as in the concat does nothing. Actually this check is redundant but I put it in to drive home the point. Technically, any concat that has 1 input must have the same input and output shape.

src/ngraph/pass/concat_fusion.cpp
AyanmoI: If I do not remove the single self concat, it will get replaced by a reshape and a broadcast. As you had mentioned once, that a single self concat is handled as a special case using arAssign, IMO replacing it with a reshape and a broadcast will be redundant and might even be slightly inefficient.

src/ngraph/pass/concat_fusion.hpp
AyanmoI: I do need the first and the last concat of the pattern plus all their concat axes. I can create a something that will just store the first and last at each iteration where the match happens, but I thought it might be easier to just have the nodevector. Based on this design, I definitely need the argument to the first concat and the last concat to replace the pattern. I thought it is simpler to keep track of the concat nodes as a set. Please let me know what you think.

src/ngraph/pass/concat_fusion.cpp
AyanmoI: Let me know if you think replacing a single self concat with reshape+broadcast would be equally/more efficient. I am not sure of that.

src/ngraph/pass/concat_fusion.cpp
AyanmoI: The idea here is that a self concat will have multiple instances of the same node in `input_args`. Creating a set out of that vector is a quick way of determining whether it is a self concat since the set will have just one instance of the node. Can you please elaborate which comparison you are referring to? Sorry I am a little confused.

src/ngraph/pass/concat_fusion.cpp
```diff
@@ -0,0 +1,304 @@
...
+        NGRAPH_DEBUG
+            << "concat_elimination: In callback for construct_concat_elimination against node = "
+            << m.get_match_root()->get_name();
+        auto pattern_map = m.get_pattern_map();
+        auto op = pattern_map[op_label];
+
+        auto root = std::dynamic_pointer_cast<op::Concat>(m.get_match_root());
+        if (root && (root->get_input_shape(0) == root->get_output_shape(0)))
+        {
+            NGRAPH_DEBUG << " elmimited " << m.get_match_root() << "\n";
```
ransford2011: Nitpick: `eliminated`

AyanmoI:
That sounds good...I will add comments.

src/ngraph/pass/concat_fusion.cpp
XWangIntel: I mean `std::set<Node*>` will be more straight forward, as we know what to compare. For `std::set<std::shared_ptr<Node>>` I need to check the manual to make sure that `shared_ptr` is compared by the raw pointer it wrapped.
But I understand the implementation will be less elegant. No a big issue though.

src/ngraph/pass/concat_fusion.cpp
XWangIntel: reshape and broadcast is also implemented as `arAssign`. But since it is a ngraph pass I'm OK with what you have right now. Can simplify it later if necessary.

src/ngraph/pass/concat_fusion.hpp
XWangIntel: Just feel this data structure is too deeply nested, and therefore less straightforward to understand what you like to do.
Since it is necessary to keep the whole concat node vector, do we still need the `second` in the pair? Cannot it be derived from the node by `get_concatenation_axis()`?

src/ngraph/pass/concat_fusion.cpp
```diff
@@ -0,0 +1,304 @@
...
+#include "ngraph/op/parameter.hpp"
+#include "ngraph/op/reshape.hpp"
+#include "ngraph/pattern/matcher.hpp"
+#include "ngraph/pattern/op/label.hpp"
+#include "ngraph/pattern/op/skip.hpp"
+#include "ngraph/util.hpp"
+
+using namespace ngraph;
+
+bool check_self_concat_op(const std::shared_ptr<Node>& op)
```
rkimballn1: Either make these functions static, add them as member methods of the class, or put them in an anonymous namespace. As they are they are public symbols.
This goes for all public functions here.

src/ngraph/pass/concat_fusion.hpp
AyanmoI: Yes..I can be derived from the `NodeVector` using another loop. This is what I was going to do. ButI figured why not do it in one go and bind it to the corresponding `NodeVector`. Do you want me change the implementation or is ok to keep it this way?

src/ngraph/pass/concat_fusion.cpp
```diff
@@ -0,0 +1,308 @@
...
+        auto input_shape = concat_op->get_input_shape(0);
+        size_t concat_axis =
+            std::static_pointer_cast<op::Concat>(concat_op)->get_concatenation_axis();
+
+        return (input_shape[concat_axis] == 1);
+    }
+
+    bool check_concat_has_no_fan_out(const std::shared_ptr<Node>& op)
+    {
+        auto users = op->get_users();
```
pruthviIntel: would be good to use `get_users(true)` , so it check only on the reachable nodes?

pruthviIntel:
minor comment, apart from that LGTM. glad if our discussions helped :)

src/ngraph/pass/concat_fusion.hpp
AyanmoI: Made the changes...Please take a look

AyanmoI:
> minor comment, apart from that LGTM. glad if our discussions helped :)

Thanks Pruthvi...our discussion really helped.

PR #2637, opened 2019-03-19
[ONNX] Fix convpool padding by max(x, 0) on pad calculation

tsocha:


src/ngraph/frontend/onnx_import/utils/convpool.cpp
```diff
@@ -82,8 +82,9 @@ namespace ngraph
                     Shape pad_shape;
                     for (std::size_t idx = 0; idx < input.size(); ++idx)
                     {
-                        pad_shape.emplace_back((output.at(idx) - 1) * strides.at(idx) +
-                                               kernel.at(idx) - input.at(idx));
+                        pad_shape.emplace_back(std::max((output.at(idx) - 1) * strides.at(idx) +
+                                                            kernel.at(idx) - input.at(idx),
+                                                        0ULL));
```
tomdol: How about `size_t{0}` instead of `0ULL`?

pruthviIntel:
```node_name: Convolution_1108
input_shape: 1, 256, 6, 40, 40
filter_shape: 512, 256, 1, 1, 1
result_shape: 1, 512, 6, 20, 20
mkldnn_filter_strides: 1, 2, 2
mkldnn_dialted_strides: 0, 0, 0
mkldnn_padding_below: 0, -1, -1
mkldnn_padding_above: 0, 0, 0

Exception caught on '/nfs/site/home/pthoreho/keras_lotus_resnet3D/keras_lotus.json'
std::exception
```

still throws the exception and passes -ve values for padding in ngraph cone

pruthviIntel:
```
{
        "control_deps": [],
        "data_dilation_strides": [
          1,
          1
        ],
        "inputs": [
          "Reshape_2517",
          "Constant_452"
        ],
        "name": "Convolution_2518",
        "op": "Convolution",
        "outputs": [
          "Convolution_2518_0"
        ],
        "padding_above": [
          -9223372036854775808,
          -9223372036854775808
        ],
        "padding_below": [
          9223372036854775807,
          9223372036854775807
        ],
        "window_dilation_strides": [
          1,
          1
        ],
        "window_movement_strides": [
          2,
          2
        ]
      },
```

padding values are overflowed in serialized graph dumped by the onnx-ngraph serializer

src/ngraph/frontend/onnx_import/utils/convpool.cpp
```diff
@@ -82,8 +82,10 @@ namespace ngraph
                     Shape pad_shape;
                     for (std::size_t idx = 0; idx < input.size(); ++idx)
                     {
-                        pad_shape.emplace_back((output.at(idx) - 1) * strides.at(idx) +
-                                               kernel.at(idx) - input.at(idx));
+                        pad_shape.emplace_back(
+                            std::max(static_cast<long>((output.at(idx) - 1) * strides.at(idx) +
+                                                       kernel.at(idx) - input.at(idx)),
```
pruthviIntel: @tsocha can you please give me some more info what happens if their is no casting, is it something to do with integer promotion?

tsocha: yes, because without casting we can get a negative value when type is size_t.
Then std::max compare overflowed size_t to 0 and result is overflowed > 0.

PR #2638, opened 2019-03-19
[ONNX] Change default broadcast dim value, gemm improvement

tsocha:


tomdol:
This PR is a result of investigation why the onnxruntime `GemmOpTest.GemmEmptyTensor` test ended with a segfault https://github.com/Microsoft/onnxruntime/blob/master/onnxruntime/test/providers/cpu/math/gemm_test.cc#L198

The reason was that an exception was thrown in onnx imported broadcasting code because the shapes specified in the test (`{0,3}` for example) were not accepted by nG code.

@diyessi any thoughts? Should we accept such shapes?

src/ngraph/frontend/onnx_import/utils/broadcasting.cpp
```diff
@@ -45,9 +45,9 @@ ngraph::Shape calculate_broadcast_shape(ngraph::Shape left_shape, ngraph::Shape
     auto max_rank = std::max(left_rank, right_rank);
 
     // left-pad the left_shape with ones
```
arogowie-intel: Update comment.

src/ngraph/frontend/onnx_import/utils/broadcasting.cpp
```diff
@@ -45,9 +45,9 @@ ngraph::Shape calculate_broadcast_shape(ngraph::Shape left_shape, ngraph::Shape
...
-    left_shape.insert(std::begin(left_shape), max_rank - left_rank, 1);
+    left_shape.insert(std::begin(left_shape), max_rank - left_rank, 0);
     // left-pad the right_shape with ones
```
arogowie-intel: Update comment.

PR #2640, opened 2019-03-20
Fallback to ref convolution for negative padding

pruthviIntel:


src/ngraph/runtime/cpu/mkldnn_utils.hpp
```diff
@@ -97,6 +97,23 @@ namespace ngraph
                         if (s != 1)
                             return false;
                     }
+                    // MKLDNN doesnt support negative padding
+                    for (auto s : convolution->get_padding_above())
+                    {
+                        if (s == -1)
```
diyessi: `s < 0` here and beyond.

PR #2642, opened 2019-03-21
INTELGPU: Fix names duplication of generated supporting primitives fo…

dmyershov:
…r Convolution* operations


dmyershov:
Of course, I've checked that resnet20 is functional. It works.
But model cannot be trained because total_loss is equal nan after several iterations. The issue is in clDNN primitives and will be solved on clDNN side.

PR #2644, opened 2019-03-22
Add naive int64-indexing to EmbeddingLookup

cconvey:


src/ngraph/runtime/cpu/builder/embedding_lookup.cpp
```diff
@@ -77,6 +78,19 @@ namespace ngraph
                                 in_shape);
                         };
                     }
+                    else if (index_element_type == element::i64)
+                    {
+                        functor = [&, in_shape, element_count](CPURuntimeContext* ctx,
+                                                               CPUExecutionContext* ectx) {
+
+                            ngraph::runtime::reference::embedding<float, long>(
+                                static_cast<long*>(arg0_tensor),
```
rkimballn1: Use types like `int64_t` instead of `long` since long is only guaranteed to be at least 32 bits

src/ngraph/runtime/cpu/builder/embedding_lookup.cpp
```diff
@@ -111,6 +125,19 @@ namespace ngraph
                                 in_shape);
                         };
                     }
+                    else if (index_element_type == element::i64)
+                    {
+                        functor = [&, in_shape, element_count](CPURuntimeContext* ctx,
+                                                               CPUExecutionContext* ectx) {
+
+                            ngraph::runtime::reference::embedding<int, long>(
+                                static_cast<long*>(arg0_tensor),
```
rkimballn1: same here, int64_t

src/ngraph/runtime/intelgpu/unit_test.manifest
```diff
@@ -8,6 +8,7 @@ backwards_slice
 batch_norm_bprop_n4c3h2w2
 embedding_lookup_10x1_arbitrary
 embedding_lookup_10x1_arbitrary_index_type_int
+embedding_lookup_10x1_arbitrary_index_type_long
```
rkimballn1: name should be *_int64

test/backend_embedding_lookup.in.cpp
```diff
@@ -106,3 +106,27 @@ NGRAPH_TEST(${BACKEND_NAME}, embedding_lookup_10x1_arbitrary_index_type_int)
...
+    auto A = make_shared<op::Parameter>(element::i64, shape);
+    auto B = make_shared<op::Parameter>(element::f32, rshape);
+    auto embed = make_shared<op::EmbeddingLookup>(A, B);
+    auto f0 = make_shared<Function>(NodeVector{embed}, ParameterVector{A, B});
+
+    auto backend = runtime::Backend::create("${BACKEND_NAME}");
+
+    // Create some tensors for input/output
+    auto a = backend->create_tensor(element::i64, shape);
+    copy_data(a, vector<long>{9, 2, 1, 0, 3, 5, 4, 6, 8, 7});
```
rkimballn1: again, don't use long

src/ngraph/runtime/cpu/builder/embedding_lookup.cpp
cconvey: @rkimballn1 : Good catch.  Looks like my assumption was [pretty safe for 64-bit OSX and Linux, but not for Win64](http://nickdesaulniers.github.io/blog/2016/05/30/data-models-and-word-size/).

Issue #2646, opened 2019-03-22
CPU Executable not releasing memory when destroyed


hildebrandmw:
Hi all,

I've noticed an issue with memory resources not being freed when calling `remove_compiled_function` on the CPU Backend. I do not have a great MWE for this, but the issue comes from `runtime::cpu::CPU_CallFrame::cleanup_runtime_context()` never being called.

I've managed to fix it for my purposes by putting `cleanup_runtime_context()` back in the destructor for `CPU_CallFrame` and will open a pull request shortly with that fix.

Mark

jbobba:
Thanks @hildebrandmw It was an accidental deletion.
Your PR has been merged.

PR #2647, opened 2019-03-22
Cleanup runtime context in CPU_CallFrame Destructor

hildebrandmw:
I noticed that `cleanup_runtime_context();` was removed from the `CPU_CallFrame` destructor in #2421. However, this causes runtime context to never get cleaned up leading to a memory leak (#2646), so I put it back.

If it's easier for you all to do this internally, or if you would prefer an alternative fix, please feel free to close this pull request.

Mark


diyessi:
Replaced by #2651

PR #2648, opened 2019-03-22
Do not set num_threads_per_pool to 1 in cpu executor.

ayzhuang:
Ignore cacheability in memory assignment when not reusing memory.


src/ngraph/runtime/cpu/cpu_executor.cpp
```diff
@@ -68,13 +68,11 @@ namespace ngraph
                     for (int i = 0; i < num_thread_pools; i++)
                     {
                         int num_threads_per_pool;
-#if defined(EIGEN_OPENMP)
+
                         // Eigen threadpool will still be used for reductions
                         // and other tensor operations that dont use a parallelFor
-                        num_threads_per_pool = 1;
-#else
```
nmostafa: Any idea why it was set to 1 ?

jbobba: It was based on an incorrect understanding of EIGEN_OPENMP on my part. When this was set I assumed all Eigen execution is going to go through openmp and hence reduced the number of threads in the eigen thread pool.

src/ngraph/runtime/cpu/pass/cpu_memory_assignment.cpp
```diff
@@ -712,21 +712,16 @@ bool runtime::cpu::pass::CPUMemoryAssignment::run_on_function(shared_ptr<ngraph:
...
-                                {
-                                    NGRAPH_DEBUG << "cpu_memory_assignment: cacheable input, no "
-                                                    "destructive oi";
-                                    continue;
-                                }
                                 // when reusing memory, ops with different cacheabilities are using different memory manager
                                 // and should not share the same buffer.
-                                else if (!m_disable_memory_sharing &&
-                                         op_annotations->is_cacheable())
+                                if (!m_disable_memory_sharing &&
```
gaurides: just a trivial comment/question for later: should we rename the variable as `m_memory_shared_enabled` instead of using double negatives most of the times? `!` and `disable` ?

diyessi: `m_enable_memory_sharing` would be better, but it is a good idea.

PR #2650, opened 2019-03-22
Add erf op, interpreter kernel and unit tests

sfvaroglu:


test/backend_test.in.cpp
```diff
@@ -2687,6 +2687,12 @@ NGRAPH_TEST(${BACKEND_NAME}, zero_sized_ceiling)
     make_unary_empty_test<op::Ceiling>("${BACKEND_NAME}");
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, zero_sized_erf)
+{
+    // TODO: Fix when backends are ready
+    make_unary_empty_test<op::Erf>("INTERPRETER");
```
aprocter: Instead of hard-coding to `INTERPRETER` we should put `${BACKEND_NAME}` and disable the test in the other backends' manifests.

test/backend_unary_elementwise.in.cpp
```diff
@@ -211,6 +211,27 @@ NGRAPH_TEST(${BACKEND_NAME}, cosh)
...
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, erf)
+{
+    Shape shape{8};
+    auto A = make_shared<op::Parameter>(element::f32, shape);
+    auto f = make_shared<Function>(make_shared<op::Erf>(A), ParameterVector{A});
+
+    // TODO: Fix when backends are ready
+    auto backend = runtime::Backend::create("INTERPRETER");
```
aprocter: Instead of hard-coding to `INTERPRETER` we should put `${BACKEND_NAME}` and disable the test in the other backends' manifests.

PR #2655, opened 2019-03-23
Setting the ABI when building gtest

rkimballn1:
This worked...for any OS that was not Linux or Macos. Fixed it for Linux. Someone can tell me if it works for Macos.


test/CMakeLists.txt
```diff
@@ -191,13 +191,23 @@ endforeach()
 add_executable(unit-test ${SRC})
 
 target_include_directories(unit-test PRIVATE ".")
-
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DCURDIR=\\\"${CMAKE_CURRENT_SOURCE_DIR}\\\"")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DJSON_INCLUDES=\\\"${JSON_INCLUDE_DIR}\\\"")
+target_compile_options(unit-test
+    PRIVATE
+        -std=c++11
```
silee2: What about MSVC? This flag should only be set for non MSVC target.
I've noticed we set CMAKE_CXX_STANDARD in the top level CMakeLists.txt for WIN32.
https://github.com/NervanaSystems/ngraph/blob/master/CMakeLists.txt#L269
And I think this flag already gets inherited from top-level CMakeList (https://github.com/NervanaSystems/ngraph/blob/master/CMakeLists.txt#L272)

rkimballn1: I did not add this flag, it was already there. I did not remove any MSVC flags since they were not there. No changes with flags. All I did was reformat some ugly cmake code into less ugly cmake code. I can remove the beautification if you want since it is not part of the PR.

rkimballn1:
@avijit-nervana do you still get an error? CI is passing and it builds with gcc 4.8

PR #2656, opened 2019-03-25
Adds extra case for conv affine folding

jbobba:
tensorflow inception resnetv2 has a slightly different broadcasting pattern for scale.


test/cpu_fusion.cpp
```diff
@@ -2051,6 +2051,48 @@ TEST(cpu_fusion, convbias_affine_folding)
...
+    for (shared_ptr<op::Parameter> param : cpu_f->get_parameters())
+    {
+        vector<float> tensor_val(shape_size(param->get_shape()));
+        rng.initialize(tensor_val);
+        args.push_back(tensor_val);
+    }
+
+    auto int_results = execute(int_f, args, "INTERPRETER");
+    auto cpu_results = execute(cpu_f, args, "CPU");
+    EXPECT_TRUE(test::all_close(cpu_results.at(0), int_results.at(0)));
```
gaurides: Can you please add a check to make sure the fusion/folding happened? 
expect 1 op::ConvolutionBias and 0 Multiply?

PR #2659, opened 2019-03-26
Sarkars/bypass ngcore

sayantan-nervana:
Request to add an API in ngraph `backend` to return a handle to the backend shared library.


src/ngraph/runtime/backend.hpp
```diff
@@ -48,6 +49,10 @@ class ngraph::runtime::Backend
     ///   does not exist.
     static std::unique_ptr<Backend> create(const std::string& type);
 
+    /// \brief Get handle to backend shared library
+    /// \returns backend handle
+    static DL_HANDLE get_handle(const std::string& type);
```
avijit-nervana: If the backend is not found then how is the error propagated? This function should behave exactly the same way as the create_backend(). 

Better yet - consider using the ngraph::Backend* as the key to the registry - not the name. That way - the caller needs to make the "create_backend()" first to get the backend created - and then get its handle. Then it's failsafe.

src/ngraph/runtime/backend_manager.cpp
```diff
@@ -66,6 +66,22 @@ vector<string> runtime::BackendManager::get_registered_backends()
...
+}
+
+void runtime::BackendManager::register_handle(const string& name, DL_HANDLE h)
+{
+    get_handle_registry()[name] = h;
+}
+
+DL_HANDLE runtime::BackendManager::get_handle(const std::string& type)
+{
+    return get_handle_registry()[type];
```
avijit-nervana: As commented above - this implementation needs to add proper error handling.

src/ngraph/runtime/backend_manager.hpp
```diff
@@ -57,10 +57,18 @@ class ngraph::runtime::BackendManager
     /// \returns A vector of all registered devices.
     static std::vector<std::string> get_registered_backends();
 
+    static DL_HANDLE get_handle(const std::string& type);
```
avijit-nervana: Add comments.

src/ngraph/runtime/backend.cpp
```diff
@@ -29,6 +29,11 @@ runtime::Backend::~Backend()
 {
 }
 
+DL_HANDLE runtime::Backend::get_handle(const std::string& type)
```
diyessi: I'd prefer returning a defined interface that backends can implement rather than leaving the keys to the backdoor of the dll the under the mat like a handle does.

src/ngraph/runtime/backend.cpp
```diff
@@ -29,6 +29,12 @@ runtime::Backend::~Backend()
 {
 }
```
diyessi: You should call it `get_backend_op` since it is returning a node that is an op, as opposed to a node that is some other thing like a pattern.

avijit-nervana: Yes - that makes sense. However, this bubbles up an interesting question - the return type should be ngraph::Op and `not` ngraph::Node - is there such a thing?

aprocter: @avijit-nervana That does exist, yes.

avijit-nervana: @aprocter That's what I thought - so in this case the return type should be changed to ngraph::Op to make it fail safe.

sayantan-nervana: `Op` is a child class of `Node`. At the bridge we mostly use `Op`s
Yes, we can change the return type to `Op`

src/ngraph/runtime/backend.hpp
```diff
@@ -109,4 +109,9 @@ class ngraph::runtime::Backend
     virtual bool is_supported_property(const Property prop) const;
 
     virtual void remove_compiled_function(std::shared_ptr<Executable> exec);
+
+    // \brief Return a requested op from the backend if possible
+    // \param op_name is the name of the backend specific op
+    // \returns a shared pointer to the op if found, else nullptr
+    virtual std::shared_ptr<ngraph::op::Op> get_backend_op(const std::string& op_name, ...);
```
rkimballn1: I don't like the `...` here. Why not pass a second string which is the config string, or encode the config in the op_name and call it something else. If not then you really really need to put a description in the comment block on what this is. How do I use it? From the current comment I have no ideal now to implement `get_backend_op`.

avijit-nervana: @rkimballn1 Adding more comments sounds like a great idea.

src/ngraph/runtime/backend.cpp
```diff
@@ -29,9 +29,9 @@ runtime::Backend::~Backend()
...
 
-std::shared_ptr<ngraph::Node> runtime::Backend::get_backend_node(const std::string& node_name, ...)
+std::shared_ptr<ngraph::op::Op> runtime::Backend::get_backend_op(const std::string& op_name, ...)
```
diyessi: Safer to return `std::shared_ptr<Node>` or you're likely to encounter a C++ types issue. Just call the function `get_backend_op` but leave it returning a Node.

avijit-nervana: @diyessi since ngraph::Op is a derived class of ngraph::Node - what `you're likely to encounter a C++ types issue` are you referring to?

src/ngraph/runtime/backend.hpp
sayantan-nervana: Added more comments

rkimballn1: Since this is so backend specific, why not put it only in the backend in question. We don't need a generic Backend definition for a specific backend function. If you were to cast the backend you get back from Backend::create("CPU") into a CPUBackend class then you could directly access any custom op constructors without the need for variadic args

PR #2660, opened 2019-03-26
gauss error function (erf) implementation through eigen for DEX and CODEGEN

pruthviIntel:


test/cpu_test.cpp
```diff
@@ -1009,3 +1010,32 @@ TEST(cpu_test, conv_negative_padding)
     };
     compare_backends(make_f(), make_f(), "CPU", "INTERPRETER");
 }
+
+TEST(cpu_test, guass_error_function_erf)
```
rkimballn1: Can we make this a general erf unit test for all backends? I know that most don't yet implement it so they would need the test disabled via their manifest file.

pruthviIntel: @rkimballn1 we have this https://github.com/NervanaSystems/ngraph/blob/master/test/backend_unary_elementwise.in.cpp#L214, unit test already for `Erf` which test for all the backend.

PR #2662, opened 2019-03-26
Use official Eigen repo.

silee2:
Instead of using https://github.com/NervanaSystems/eigen/tree/patched (Which is the same as official ded1e7b4960f0074fa147a8ed1c9926174958092 with some extra commits ) move to https://github.com/eigenteam/eigen-git-mirror/tree/ded1e7b4960f0074fa147a8ed1c9926174958092 with manual patch with CMake.
End result is identical Eigen code used by nGraph.


jbobba:
Looks good but lets get some performance numbers to make sure upstream eigen doesn't have changes that slow things down.

silee2:
@jbobba There is not code change. You get the exact same Eigen code. It is just structured differently. Old: Eigen Fork + some commit. New: Official Eigen SHA/commit + patch.

silee2:
@jbobba the patch is created by doing a diff between https://github.com/eigenteam/eigen-git-mirror/tree/ded1e7b4960f0074fa147a8ed1c9926174958092 and our fork.

jbobba:
sounds good. didn't realize you were using the exact same commit upstream

PR #2663, opened 2019-03-26
header resource line by line

silee2:
Handle one header line at time instead of entire file to work around the 64K string literal size limit in MSVC.


silee2:
Problem is not really solved. Adding WIP tag.

PR #2664, opened 2019-03-26
[Standalone] Set MKLDNN kernel flag in ConvertLayout op.

dcaballe:
ConvertLayout is an MKLDNN specific op and the MKLDNN kernel flag was not
set. This led to assume that this operation was not supported by
MKLDNN.


jbobba:
Can you add this in cpu_assignment.cpp instead of the op constructor.

dcaballe:
Ok. I thought this op only made sense in the context of mkldnn. I'll move it there for consistency.

dcaballe:
Sorry, rebase problem

PR #2665, opened 2019-03-26
General-purpose recurrent reshape elimination pass

AyanmoI:
Motivated by a pattern in resnet 50, I added this pass to fuse recurrent reshapes for the following pattern:
`{reshape(any_axis_order)->reshape(default_axis_order)->reshape(default_axis_order)->......} ----------->{ Single reshape}`
The following describes the pattern and the fusion:
recurrent_reshapes

With concat fusion (https://github.com/NervanaSystems/ngraph/pull/2634) and the current PR, part of the resnet50 graph is optimized as follows:

rsnet_50

As a reminder, after concat fusion, the graph looked as follows (now fused to the final state as shown above). The circled part is fused and eliminated to produce the final state:
after_concat_fusion

There are various cases that have been handled such as:
1) Multiple Branches in the graph
multiple_branches
2) Complete reshape elimination, i.e, recurrent reshape fusion followed by reshape elimination(existing pass)
reshape_elimination

Screen Shot 2019-04-03 at 2 53 34 PM

3) Recurrent reshapes with fan out
Screen Shot 2019-04-03 at 2 55 18 PM

Screen Shot 2019-04-03 at 2 56 14 PM


src/ngraph/pass/reshape_elimination.cpp
```diff
@@ -192,3 +191,127 @@ void pass::ReshapeElimination::construct_dot_transpose_pattern()
     auto m = make_shared<pattern::Matcher>(preshape, callback);
     this->add_matcher(m);
 }
+
+void pass::RecurrentReshapeElimination::construct_recurrent_reshape()
+{
+    Shape shape_op{3};
+    Shape shape_r{1, 3};
+
+    auto no_fan_out = [](std::shared_ptr<Node> n) {
```
XWangIntel: Looks like this one can combine with the on in concat elimination as a helper function.

AyanmoI: That might be tricky since this is a lambda function. Refactoring this small piece of code might not be effective.

diyessi: You could make a helper where you passed the arguments to be captured by the lambda and returned the lambda.

src/ngraph/pass/reshape_elimination.cpp
```diff
@@ -192,3 +191,127 @@ void pass::ReshapeElimination::construct_dot_transpose_pattern()
...
+        else
+        {
+            NGRAPH_DEBUG << "recurrent_reshape_elimination: " << n->get_name() << " has fan out\n";
+            return false;
+        }
+    };
+
+    auto op = make_shared<pattern::op::Label>(element::f32, shape_op);
+    auto reshape = make_shared<op::Reshape>(op, AxisVector{0}, shape_r);
+    auto rehape_label = make_shared<pattern::op::Label>(reshape, no_fan_out, NodeVector{reshape});
```
aprocter: `rehape_label` -> `reshape_label`

src/ngraph/pass/reshape_elimination.cpp
```diff
@@ -192,3 +191,127 @@ void pass::ReshapeElimination::construct_dot_transpose_pattern()
...
+
+    auto callback = [op, rehape_label](pattern::RecurrentMatcher& m) {
+        NGRAPH_DEBUG << "In callback for construct_recurrent_reshape against node = "
+                     << rehape_label->get_argument(0)->get_name();
+        auto reshape_node_vector = m.get_bound_nodes_for_pattern(rehape_label);
+
+        // The bounded node vector is in reverse order. It is convinient to have the
```
aprocter: `convinient` -> `convenient`

src/ngraph/pass/reshape_elimination.cpp
```diff
@@ -192,3 +191,127 @@ void pass::ReshapeElimination::construct_dot_transpose_pattern()
...
+        // bounded node vector in the correct order
+        std::reverse(std::begin(reshape_node_vector), std::end(reshape_node_vector));
+
+        auto first_bounded_reshape_op = reshape_node_vector.front();
+        auto driver_op = first_bounded_reshape_op->get_argument(0);
+        auto last_bounded_reshape_op = reshape_node_vector.back();
+
+        // Need to check if the next to last bounded op is a reshape since the last reshape is allowed
```
aprocter: `next to last` -> `user of the last` (right?)

src/ngraph/pass/reshape_elimination.cpp
```diff
@@ -192,3 +191,127 @@ void pass::ReshapeElimination::construct_dot_transpose_pattern()
...
+    auto callback = [op, rehape_label](pattern::RecurrentMatcher& m) {
+        NGRAPH_DEBUG << "In callback for construct_recurrent_reshape against node = "
+                     << rehape_label->get_argument(0)->get_name();
+        auto reshape_node_vector = m.get_bound_nodes_for_pattern(rehape_label);
+
+        // The bounded node vector is in reverse order. It is convinient to have the
+        // bounded node vector in the correct order
+        std::reverse(std::begin(reshape_node_vector), std::end(reshape_node_vector));
+
+        auto first_bounded_reshape_op = reshape_node_vector.front();
```
aprocter: `bounded` -> `bound`

src/ngraph/pass/reshape_elimination.cpp
```diff
@@ -192,3 +191,127 @@ void pass::ReshapeElimination::construct_dot_transpose_pattern()
...
+            last_bounded_reshape_op = reshape_node_vector.back();
+        }
+
+        // Return if the recurrent matcher matches only one reshape
+        if (reshape_node_vector.size() == 1)
+        {
+            return false;
+        }
+
+        // The complete reshape node vector may not contain contigous reshapes that can be
```
aprocter: `contigous` -> `contiguous`

src/ngraph/pass/reshape_elimination.cpp
```diff
@@ -192,3 +191,127 @@ void pass::ReshapeElimination::construct_dot_transpose_pattern()
...
+            }
+            else
+            {
+                NGRAPH_DEBUG << r->get_name() << "does not have default axis order. "
+                             << "It might be part of a different subpattern";
+                sub_patterns.push_back(NodeVector{r});
+            }
+        }
+
+        // Remove the subpatterns with just one reshape in them
```
aprocter: Can we just use `continue` in the loop that follows, rather than modify the vector here?

test/reshape_elimination.cpp
```diff
@@ -106,3 +110,310 @@ TEST(reshape_elimination, dot_transpose_to_dot_w_transpose_args)
     ASSERT_EQ(gdot->get_argument(1)->get_argument(0), W);
     ASSERT_EQ(gdot->get_shape(), (Shape{1, 2}));
 }
+
+TEST(reshape_elimination, recurrent_reshapes)
```
aprocter: While there's nothing wrong with tests like this, I think it's best as a general rule to test that the _structure_ of the graph is actually modified as expected. As it stands, this test will pass even if the pass doesn't do anything. (Unless I am missing something.)

test/reshape_elimination.cpp
```diff
@@ -106,3 +110,310 @@ TEST(reshape_elimination, dot_transpose_to_dot_w_transpose_args)
...
+    };
+
+    auto baseline_f = generate_func();
+    auto optimized_f = generate_func();
+    auto baseline_input_shape = baseline_f->get_parameters().at(0)->get_shape();
+
+    pass::Manager pass_manager;
+    pass_manager.register_pass<pass::VisualizeTree>("before_recurrent_reshapes.pdf");
+    pass_manager.register_pass<pass::RecurrentReshapeElimination>();
+    pass_manager.register_pass<pass::ReshapeElimination>();
```
aprocter: Why is `ReshapeElimination` (as opposed to `RecurrentReshapeElimination` being called here? (Same question applies in several other places throughout the tests.)

test/reshape_elimination.cpp
```diff
@@ -106,3 +110,310 @@ TEST(reshape_elimination, dot_transpose_to_dot_w_transpose_args)
...
+    };
+
+    auto baseline_f = generate_func();
+    auto optimized_f = generate_func();
+    auto baseline_input_shape = baseline_f->get_parameters().at(0)->get_shape();
+
+    pass::Manager pass_manager;
+    pass_manager.register_pass<pass::VisualizeTree>("before_recurrent_reshapes_fan_out.pdf");
+    pass_manager.register_pass<pass::RecurrentReshapeElimination>();
+    //pass_manager.register_pass<pass::ReshapeElimination>();
```
aprocter: Why is this commented out?

test/reshape_elimination.cpp
```diff
@@ -106,3 +110,310 @@ TEST(reshape_elimination, dot_transpose_to_dot_w_transpose_args)
...
+    };
+
+    auto baseline_f = generate_func();
+    auto optimized_f = generate_func();
+    auto baseline_input_shape = baseline_f->get_parameters().at(0)->get_shape();
+
+    pass::Manager pass_manager;
+    pass_manager.register_pass<pass::VisualizeTree>("before_nonrecurrent_reshapes.pdf");
+    pass_manager.register_pass<pass::RecurrentReshapeElimination>();
+    //pass_manager.register_pass<pass::ReshapeElimination>();
```
aprocter: Why is this commented out?

test/reshape_elimination.cpp
AyanmoI: That is true! My intention is to add a check to count the final number of Reshapes.

test/reshape_elimination.cpp
AyanmoI: Called `ReshapeElimination` in addition to `RecurrentReshapeElimination` to show how it might work where the fusion creates a `Reshape` that can be eliminated completely by the existing pass. Actually there is just one test which shows that. It is kind of redundant in the other ones. i'll remove them from the other tests

test/reshape_elimination.cpp
AyanmoI: As I mentioned, this test will not leverage the `ReshapeElimination` pass. I'll just remove it avoid confusion and keep it only in the test `recurrent_reshapes_elimination`

test/reshape_elimination.cpp
AyanmoI: Same comment as above.

test/reshape_elimination.cpp
aprocter: Works for me. You might add a comment or something here.

src/ngraph/pass/reshape_elimination.cpp
AyanmoI: Sounds good :). Will do that.

src/ngraph/pass/reshape_elimination.cpp
AyanmoI: True..I just overcomplicated this . :\

src/ngraph/pass/reshape_elimination.cpp
AyanmoI: Yes

src/ngraph/pass/pass_util.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include "ngraph/pass/pass_util.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+std::function<bool(std::shared_ptr<Node>)> ngraph::pass::get_no_fan_out_function()
+{
+    auto ret_fun = [](std::shared_ptr<Node> n) {
+        auto users = n->get_users(true);
+        std::set<std::shared_ptr<Node>> user_set(users.begin(), users.end());
```
ransford2011: We should really decide whether we want to use the `std` namespace in the source files or not.  Typically that hasn't been the case, but its not a deal breaker 😉

test/reshape_elimination.cpp
```diff
@@ -106,3 +110,325 @@ TEST(reshape_elimination, dot_transpose_to_dot_w_transpose_args)
...
+        auto f = make_shared<Function>(r_6, ParameterVector{A});
+        return f;
+    };
+
+    auto baseline_f = generate_func();
+    auto optimized_f = generate_func();
+    auto baseline_input_shape = baseline_f->get_parameters().at(0)->get_shape();
+
+    pass::Manager pass_manager;
+    pass_manager.register_pass<pass::VisualizeTree>("before_recurrent_reshapes.pdf");
```
jbobba: I would prefer if we not create PDFs by default.

AyanmoI: That makes sense. I'll comment this out incase someone wants to use this later. Or do u think I shud remove these lines completely?

PR #2667, opened 2019-03-27
[Standalone] Decouple MKLDNN primitive build from code generation

dcaballe:
This patch introduces a new pass, MKLDNNPrimitiveBuildPass, which
iterates over all the ops assigned to MKLDNN and builds their
corresponding primitives. Primitive indexes are stored in MKLDNNEmitter
and can easily be retrieved with the get_primitive_index(node)
interface. This decouples the creation of primitives from codegen and
fixes the problem of MKLDNN primitives being created twice
(CommonFunctionCollection pass and codegen).

Current assertions only allow the creation of a single primitive per
node but it should be simple to remove this when needed. Using a pass
might not be the best approach here but I found it convenient for the
current needs and it should be straightforward to convert into a utility,
if needed.

These changes caused a conflict with recently introduced
'build_quantized_inner_product*'. These new build methods will be ported
in a follow up patch to new build approach.

Local testing with and without this patch is the same:

[==========] 2536 tests from 56 test cases ran. (565721 ms total)
[  PASSED  ] 2522 tests.
[  FAILED  ] 14 tests, listed below:
[  FAILED  ] CPU.sum_stable_acc
[  FAILED  ] CPU.one_hot_scalar_oob_in_3
[  FAILED  ] CPU.one_hot_vector_1_barely_oob
[  FAILED  ] builder.dynamic_scaled_QC
[  FAILED  ] builder.dynamic_scaled_QC_with_relu
[  FAILED  ] builder.dynamic_scaled_QC_with_bias
[  FAILED  ] builder.dynamic_scaled_QC_with_bias_add_and_relu
[  FAILED  ] builder.dynamic_scaled_QC_with_bias_signed_add_and_relu
[  FAILED  ] builder.dynamic_scaled_Q
[  FAILED  ] builder.dynamic_scaled_DQ
[  FAILED  ] builder.scaled_quantize_concat_unsigned
[  FAILED  ] builder.scaled_quantize_concat_signed
[  FAILED  ] builder.scaled_quantize_concat_unsigned_varying
[  FAILED  ] cpu_quant_fusion.qconvbsa

14 FAILED TESTS
  YOU HAVE 21 DISABLED TESTS

### DIFF
--- a/src/ngraph/runtime/cpu/mkldnn_emitter.cpp
+++ b/src/ngraph/runtime/cpu/mkldnn_emitter.cpp
@@ -19,18 +19,47 @@
 
 #include "mkldnn_emitter.hpp"
 
+#include "ngraph/op/add.hpp"
+#include "ngraph/op/avg_pool.hpp"
+#include "ngraph/op/batch_norm.hpp"
+#include "ngraph/op/concat.hpp"
 #include "ngraph/op/constant.hpp"
+#include "ngraph/op/constant.hpp"
+#include "ngraph/op/convert.hpp"
+#include "ngraph/op/convolution.hpp"
+#include "ngraph/op/dequantize.hpp"
 #include "ngraph/op/dequantize.hpp"
 #include "ngraph/op/experimental/quantized_avg_pool.hpp"
+#include "ngraph/op/experimental/quantized_avg_pool.hpp"
+#include "ngraph/op/experimental/quantized_concat.hpp"
+#include "ngraph/op/experimental/quantized_conv.hpp"
+#include "ngraph/op/experimental/quantized_conv_bias.hpp"
+#include "ngraph/op/experimental/quantized_conv_relu.hpp"
 #include "ngraph/op/experimental/quantized_max_pool.hpp"
+#include "ngraph/op/experimental/quantized_max_pool.hpp"
+#include "ngraph/op/get_output_element.hpp"
+#include "ngraph/op/lrn.hpp"
+#include "ngraph/op/max_pool.hpp"
+#include "ngraph/op/quantize.hpp"
+#include "ngraph/op/relu.hpp"
+#include "ngraph/op/replace_slice.hpp"
+#include "ngraph/op/slice.hpp"
+#include "ngraph/op/softmax.hpp"
 #include "ngraph/runtime/cpu/cpu_executor.hpp"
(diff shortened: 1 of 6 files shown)


src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -2099,3 +2272,831 @@ size_t MKLDNNEmitter::build_quantized_inner_product_forward(
     m_primitive_deps[ip_index] = {input_data_index, weights_index, result_index};
     return ip_index;
 }
+
+// Translate inputs of \p node to TensorViewWrappers.
+static void translate_node_inputs(const Node* node, std::vector<TensorViewWrapper>& args)
```
dcaballe: With these `translate_*`, I'm trying to keep the original code as it was but I'm not sure if creating TensorViewWrapper is actually needed at this point. Could someone provide feedback on this?

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -2099,3 +2272,831 @@ size_t MKLDNNEmitter::build_quantized_inner_product_forward(
...
+    NGRAPH_ASSERT(args.empty()) << "Expected output vector";
+
+    for (const descriptor::Input& input : node->get_inputs())
+    {
+        const descriptor::Output& output = input.get_output();
+        std::shared_ptr<descriptor::Tensor> tv = output.get_tensor_ptr();
+        // TODO: We should use m_variable_name_map[tv->get_name()] but this map hasn't been
+        // populated when MKLDNN primitives are built. Neither it was when intrinsics where built
+        // for CommonFunctionCollection.
+        args.emplace_back(TensorViewWrapper(tv, /*m_variable_name_map[*/ tv->get_name() /*]*/));
```
dcaballe: As the TODO explains, `m_variable_name_map` is empty when the mkldnn primitives are created. It was also empty when mkldnn primitives were created the first time by `CommonFunctionCollection` so I'm not sure how critical this is. Populating this map before creating mkldnn primitives would require moving more code to the entry point of `cpu_external_function`. Not sure how feasible that is.

dcaballe:
I also messed up my local branch with the git rebase for the previous PR. I'm closing this this PR and opening it again with the updated version of the code.

PR #2669, opened 2019-03-27
[ONNX] Add verification of split attribute in split op

tsocha:
This check is required by onnxruntime for split verification.


src/ngraph/frontend/onnx_import/op/split.cpp
```diff
@@ -111,6 +116,15 @@ namespace ngraph
                         length_parts.assign(count_outputs, length_axis_to_split / count_outputs);
                     }
 
+                    std::size_t total_parts_length = 0;
+                    for (auto length : length_parts)
+                    {
+                        ASSERT_VALID_ARGUMENT(node, length > 0)
+                            << "Invalid value in 'split' attribute";
```
arogowie-intel: Please be specific about the cause of the error. ie Tell that the chunk length should be positive.

PR #2670, opened 2019-03-27
Add one hot new unit test

rkimballn1:
The new test has a large number of categories similar in number to imagenet


test/backend_one_hot.in.cpp
```diff
@@ -321,3 +321,36 @@ NGRAPH_TEST(${BACKEND_NAME}, one_hot_vector_1_fp_nonint)
         FAIL() << "Expected a std::out_of_range exception";
     }
 }
+
```
diyessi: Should really test for int64_t, int32_t, int_16_t, and int8_t.

PR #2674, opened 2019-03-28
Avijit/event traces

avijit-nervana:
Added Event generation feature. To create events, instantiate and object of ngraph::Event class for marking the start of an event (e.g., an expensive function call). Then call Event::stop() to mark the end of it. Finally call Event::write_event() to write the event traces to an output file called: `ngraph_event_trace.json`. 

The event writing is disabled by default. To enable set the following environment variable: `NGRAPH_ENABLE_EVENT_TRACING=1` 

The JSON file can be viewed with Chrome browser using the URL `chrome://tracing` and then loading the file.


src/ngraph/event_tracing.hpp
```diff
@@ -0,0 +1,106 @@
+/*******************************************************************************
```
rkimballn1: License header.

test/event_tracing.cpp
```diff
@@ -0,0 +1,63 @@
+/*******************************************************************************
```
rkimballn1: License header

src/ngraph/event_tracing.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include "event_tracing.hpp"
+#include "nlohmann/json.hpp"
+
+using namespace std;
+
+mutex ngraph::Event::s_file_mutex;
+ofstream ngraph::Event::s_event_log;
+
+namespace ngraph
```
rkimballn1: Remove this namespace since you already have appended ngraph:: to the methods. Normally we add a using namespace ngraph; up around line 23 and don't add ngraph::

This change saves one level of indenting.

src/ngraph/event_tracing.cpp
```diff
@@ -0,0 +1,74 @@
...
+{
+    void ngraph::Event::write_trace(const ngraph::Event& event)
+    {
+        lock_guard<mutex> lock(s_file_mutex);
+        if (!is_tracing_enabled())
+            return;
+
+        static bool initialized = false;
```
rkimballn1: static var so s_initialized

src/ngraph/event_tracing.cpp
```diff
@@ -0,0 +1,74 @@
...
+mutex ngraph::Event::s_file_mutex;
+ofstream ngraph::Event::s_event_log;
+
+namespace ngraph
+{
+    void ngraph::Event::write_trace(const ngraph::Event& event)
+    {
+        lock_guard<mutex> lock(s_file_mutex);
+        if (!is_tracing_enabled())
+            return;
```
rkimballn1: curly braces around all close blocks. You could change this to `if (is_tracing_enabled()) and put the code in the code block rather than early return.

src/ngraph/event_tracing.cpp
```diff
@@ -0,0 +1,74 @@
...
+            // Open the file
+            s_event_log.open("ngraph_event_trace.json", ios_base::trunc);
+            s_event_log << "[\n";
+            s_event_log << event.to_json() << "\n";
+            initialized = true;
+            return;
+        }
+
+        s_event_log << ",\n";
+        s_event_log << event.to_json() << "\n" << flush;
```
rkimballn1: Remove the `flush`

src/ngraph/event_tracing.hpp
```diff
@@ -0,0 +1,106 @@
...
+            , m_args(args)
+        {
+            m_stop = m_start;
+            m_pid = getpid();
+        }
+
+        void Stop()
+        {
+            if (m_stopped)
+                return;
```
rkimballn1: curly braces around return

src/ngraph/event_tracing.hpp
```diff
@@ -0,0 +1,106 @@
...
+            m_stopped = true;
+            m_stop = std::chrono::high_resolution_clock::now();
+        }
+
+        static void write_trace(const Event& event);
+        static bool is_tracing_enabled()
```
rkimballn1: add a set_tracing_enabled(bool) method

test/event_tracing.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "gtest/gtest.h"
+#include "ngraph/event_tracing.hpp"
+#include "ngraph/file_util.hpp"
+
+using namespace std;
+
+TEST(event_tracing, event_file)
+{
+    // Set the environment variable to ensure logging
+    putenv((char*)"NGRAPH_ENABLE_TRACING=1");
```
rkimballn1: Use the set_tracing_enable() function to enable tracing.

rkimballn1:
My dumb idea for the day. Put this code in a separate repo. Then we can add this external repo to all sorts of projects rather than copying the code around.

src/ngraph/event_tracing.cpp
avijit-nervana: The flush is there by design. If the software using these Events crashe (which often does) then there is a trace of the events preserved in the file. So we need the flush here.

src/ngraph/event_tracing.cpp
```diff
@@ -0,0 +1,74 @@
+/*******************************************************************************
```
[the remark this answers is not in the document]
avijit-nervana: Done

src/ngraph/event_tracing.cpp
rkimballn1: What is the performance impact of the flush? That is my concern

src/ngraph/event_tracing.hpp
avijit-nervana: I have added `enable_tracing()` and `disable_tracing()`

src/ngraph/event_tracing.cpp
avijit-nervana: Based on the measurements - it's in the micro seconds range. Considering where these Events are going to be used - the cost is insignificant. If we do see any performance issues then to would be easy to convert these to a circular queue/atomic instructions to minimize any overhead. But in a future PR if needed.

src/ngraph/event_tracing.hpp
```diff
@@ -0,0 +1,116 @@
...
+                return;
+            }
+            m_stopped = true;
+            m_stop = std::chrono::high_resolution_clock::now();
+        }
+
+        static void write_trace(const Event& event);
+        static bool is_tracing_enabled()
+        {
+            if (std::getenv("NGRAPH_ENABLE_TRACING") != nullptr)
```
aprocter: Not sure how expensive this `getenv` is, but we might want to cache this. No need to address in this PR.

test/event_tracing.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "gtest/gtest.h"
+#include "ngraph/event_tracing.hpp"
+#include "ngraph/file_util.hpp"
+
+using namespace std;
+
+TEST(event_tracing, event_file)
+{
+    // Set the environment variable to ensure logging
+    ngraph::Event::enable_event_tracing();
```
aprocter: I think we should disable at the end of the test (or, even better, restore to the original state).

src/ngraph/event_tracing.cpp
```diff
@@ -0,0 +1,76 @@
...
+    if (!is_tracing_enabled())
+    {
+        return;
+    }
+
+    static bool so_initialized = false;
+    if (!so_initialized)
+    {
+        // Open the file
+        s_event_log.open("ngraph_event_trace.json", ios_base::trunc);
```
aprocter: Would be nice to make this filename customizable somehow, perhaps through an environment variable. No need to address in this PR.

src/ngraph/event_tracing.hpp
```diff
@@ -0,0 +1,116 @@
...
+        explicit Event(const std::string& name,
+                       const std::string& category,
+                       const std::string& args)
+            : m_start(std::chrono::high_resolution_clock::now())
+            , m_name(name)
+            , m_category(category)
+            , m_args(args)
+        {
+            m_stop = m_start;
+            m_pid = getpid();
```
aprocter: Minor, but do we actually need to store this? Can always getpid() when we write out. No need to address in this PR.

src/ngraph/event_tracing.hpp
```diff
@@ -0,0 +1,116 @@
...
+        static void disable_event_tracing() { s_tracing_enabled = false; }
+        std::string to_json() const;
+
+        Event(const Event&) = delete;
+        Event& operator=(Event const&) = delete;
+
+    private:
+        std::chrono::time_point<std::chrono::high_resolution_clock> m_start;
+        std::chrono::time_point<std::chrono::high_resolution_clock> m_stop;
+        bool m_stopped{false};
```
aprocter: I think it's probably better to move these initializers to the constructor (that's what we tend to do in `ngraph` anyway).

src/ngraph/event_tracing.hpp
```diff
@@ -0,0 +1,116 @@
...
+        std::string m_name;
+        std::string m_category;
+        std::string m_args;
+        int m_pid{0};
```
aprocter: Can we move the initializer to the constructor?

src/ngraph/event_tracing.hpp
avijit-nervana: Changed.

test/event_tracing.cpp
avijit-nervana: Done

src/ngraph/event_tracing.cpp
avijit-nervana: We will address this in a future PR.

src/ngraph/event_tracing.hpp
avijit-nervana: Yes - as this is a system call - better to store this.

src/ngraph/event_tracing.hpp
avijit-nervana: Done

src/ngraph/event_tracing.hpp
avijit-nervana: Done

src/ngraph/event_tracing.hpp
aprocter: Not sure I agree---we only write the event out once, right? Doesn't matter, though.

avijit-nervana: That's actually true. Will remove it in the next PR

PR #2677, opened 2019-03-28
save compiled function instead of parameters and results, add get_compiled_function()

fengleitian:
what this PR did?
1. using `clone_function()` for input graph for all backend, so input graph will not be modified.
2. add `get_compiled_function()` to `ngraph::runtime::Executable`.
3. add `nodemap` to `ngraph::runtime::Executable` so the compiled node in compiled function can be found with original node.
4. fix unit tests after change.
5. fixed a issue in `cpu_memory_assignment.cpp`, there do have cases the tensor can not be found in `m_tensor_to_bufferID`. my thought is some ops is been eliminated but not been cleaned completely. so when propagate through `op_output_input`, it get the tensor not in map. but could be some other bugs in passes.


src/ngraph/runtime/executable.hpp
```diff
@@ -61,6 +61,10 @@ class ngraph::runtime::Executable
     void validate(const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
                   const std::vector<std::shared_ptr<runtime::Tensor>>& inputs);
 
+    /// \brief Query the compiled function
+    /// \returns an nshared_ptr<Function> of compiled function
```
rkimballn1: shared_ptr not nshared_ptr

fengleitian: fixed.

fengleitian:
@jbobba I agree with @mbrookhart , this should not break fprop-cache.

fengleitian:
@sayantan-nervana Yes.

fengleitian:
close this PR as the other PR solve the original problem.

fengleitian:
https://github.com/NervanaSystems/ngraph/pull/2790

PR #2678, opened 2019-03-28
[Warnings] Remove unused variable hso

dcaballe:
And its annoying warning. Would it be better to remove all the commented code instead?


src/ngraph/codegen/compiler.cpp
```diff
@@ -571,7 +571,7 @@ void codegen::CompilerCore::configure_search_path()
 void codegen::CompilerCore::load_headers_from_resource()
 {
     const std::string builtin_root = "";
-    HeaderSearchOptions& hso = m_compiler->getInvocation().getHeaderSearchOpts();
+    //HeaderSearchOptions& hso = m_compiler->getInvocation().getHeaderSearchOpts();
```
rkimballn1: yes, remove the commented out code.

PR #2686, opened 2019-03-29
[Standalone] Set MKLDNN kernel flag in ConvertLayout op.

dcaballe:
ConvertLayout is an MKLDNN specific op and the MKLDNN kernel flag
was not set. This led to assume that this operation was not
supported by MKLDNN.

### DIFF
--- a/src/ngraph/graph_util.cpp
+++ b/src/ngraph/graph_util.cpp
@@ -243,6 +243,15 @@ std::list<std::shared_ptr<ngraph::Node>>
             }
             auto cloned_node = node->copy_with_new_args(cloned_args);
 
+            // Copy op annotations.
+            if (auto cloned_op = dynamic_pointer_cast<ngraph::op::Op>(cloned_node))
+            {
+                if (auto op_annotations = cloned_op->get_op_annotations())
+                {
+                    cloned_op->set_op_annotations(op_annotations->clone());
+                }
+            }
+
             //copy control dependencies
             for (auto cdep : node->get_control_dependencies())
             {
--- a/src/ngraph/op/op.hpp
+++ b/src/ngraph/op/op.hpp
@@ -40,10 +40,9 @@ namespace ngraph
 
             virtual bool is_op() const override { return true; }
         protected:
-            Op(const std::string& node_type, const NodeVector& arguments);
-
-        private:
             std::shared_ptr<ngraph::op::util::OpAnnotations> m_op_annotations;
+
+            Op(const std::string& node_type, const NodeVector& arguments);
         };
     }
 }
--- a/src/ngraph/op/op.hpp
+++ b/src/ngraph/op/op.hpp
@@ -40,9 +40,10 @@ namespace ngraph
 
             virtual bool is_op() const override { return true; }
         protected:
-            std::shared_ptr<ngraph::op::util::OpAnnotations> m_op_annotations;
-
             Op(const std::string& node_type, const NodeVector& arguments);
+
+        private:
+            std::shared_ptr<ngraph::op::util::OpAnnotations> m_op_annotations;
         };
     }
 }
--- a/src/ngraph/op/util/op_annotations.hpp
+++ b/src/ngraph/op/util/op_annotations.hpp
@@ -35,6 +35,7 @@ namespace ngraph
             class OpAnnotations
             {
             public:
+                OpAnnotations() = default;
                 virtual ~OpAnnotations() = default;
 
                 void add_in_place_oi_pair(const struct oi_pair& oi)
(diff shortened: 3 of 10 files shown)


dcaballe:
I found a bug in existing code (`ConvertLayout::copy_with_args`) that made op annotations not be propagated to the copy. I'm adding the fix as part of this PR since the PR wouldn't be fully functional otherwise.

src/ngraph/runtime/cpu/op/convert_layout.cpp
```diff
@@ -33,7 +36,14 @@ shared_ptr<Node>
     {
         throw ngraph_error("Incorrect number of new arguments");
     }
-    return make_shared<ConvertLayout>(new_args.at(0), output_layout);
+
+    auto copy = make_shared<ConvertLayout>(new_args.at(0), output_layout);
+
+    if (m_op_annotations)
+      // Copy annotations
+      copy->set_op_annotations(m_op_annotations->clone());
```
aprocter: I don't like the idea of copying the op annotations here. The purpose of `copy_with_new_args` is to produce a new node that is semantically equivalent to the old node, up to reassigning of the inputs. If there is some semantically relevant information being stored in the op annotations, it should be moved out of the annotations.

Alternatively, if we want to change the contract so that it's part of `copy_with_new_args`'s job to preserve annotations, we should refactor more broadly so that all nodes behave this way. Inconsistent treatment of the annotations is risky, IMHO.

src/ngraph/op/op.hpp
```diff
@@ -40,10 +40,9 @@ namespace ngraph
 
             virtual bool is_op() const override { return true; }
         protected:
-            Op(const std::string& node_type, const NodeVector& arguments);
-
-        private:
             std::shared_ptr<ngraph::op::util::OpAnnotations> m_op_annotations;
```
aprocter: I don't think there is any need to move this to `protected`. We have `get_op_annotations()`, and can `clone` from that if we need to.

src/ngraph/op/util/op_annotations.hpp
```diff
@@ -56,6 +57,17 @@ namespace ngraph
 
                 bool is_cacheable() const { return m_cacheable; }
                 void set_cacheable(bool val) { m_cacheable = val; }
+                /// \returns a deep copy of this OpAnnotations object.
+                virtual std::shared_ptr<OpAnnotations> clone()
+                {
+                    auto* clone = new OpAnnotations(*this);
```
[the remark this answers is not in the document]
dcaballe: > Why not just return std::make_shared<OpAnnotations>(*this) ?
Because the copy constructor was intentionally made protected so that we only use `clone` to create copies, taking into account the specific object of the class hierarchy. For that reason, we cannot do `std::make_shared<OpAnnotations>(*this)`.

> This also looks like a shallow copy to me.
At this point, shallow copy and deep copy are equivalent in this class but that might change in the future if we add more members to it.

Does it make sense?

src/ngraph/runtime/cpu/op/convert_layout.cpp
dcaballe: Thanks Adam. I don't have a strong opinion here but I think things can get buggy if we only do a selective copy of the node information. Propagating annotations outside of this call might also be error prone. I would lean towards your second suggestion but I'll wait for more feedback.

src/ngraph/op/op.hpp
dcaballe: Ok, I can do that instead but I'm not sure I understand the rationale. Why should we hide `m_op_annotations` in the subclass and use a public interface if exposing it it's not compromising the correctness of the implementation?

aprocter: My first instinct is always to keep data members private, except in extremely limited circumstances, on the theory that it provides better separation of implementation and interface.

src/ngraph/graph_util.cpp
```diff
@@ -243,6 +243,15 @@ std::list<std::shared_ptr<ngraph::Node>>
...
 
+            // Copy op annotations.
+            if (auto cloned_op = dynamic_pointer_cast<ngraph::op::Op>(cloned_node))
+            {
+                if (auto op_annotations = cloned_op->get_op_annotations())
+                {
+                    cloned_op->set_op_annotations(op_annotations->clone());
+                }
+            }
+
```
dcaballe: Moved the copy of op annotations to this `clone_nodes` utility. I guess it makes more sense since we are also copying control dependencies (below). If it doesn't, I could set the annotation information in `CovertLayout` constructor as I had initially but it would be a bit inconsistent with the rest of ops. No strong opinion from my side.

dcaballe:
Unfortunately, the new copy approach seems to be problematic for other ops that do not expect the op annotations to be copied. New tests are failing now. I had a look and fix is not trivial. This seems to be an issue beyond the scope of this small change so I'm going to cancel this PR and proceed with the initial changes in the `ConvertLayout` constructor (jbobba was ok with this).

PR #2687, opened 2019-03-29
Add support for QDotInteger

nbpatel:


test/builder_quantization.cpp
```diff
@@ -1182,6 +1184,31 @@ TEST(builder, scaled_quantize_concat_unsigned_varying)
               read_vector<uint8_t>(result));
 }
 
+TEST(builder, scaled_QDotInteger)
+{
+    Shape shape_a{3, 3}; // input shape
```
rkimballn1: If you named your variable `input_shape` then the comment would not be needed and the code would be that much more readable. You could also rename a_data to input_data, etc..

diyessi: It would be `input/data_shape`, `filter_shape`, `output/result_shape` and `input/data`, `filter`, `output/result`, but not `input_data` which is redundant.

test/builder_quantization.cpp
```diff
@@ -1182,6 +1184,31 @@ TEST(builder, scaled_quantize_concat_unsigned_varying)
...
+    vector<uint8_t> a_data = {1, 2, 2, 3, 4, 5, 1, 0, 1};
+    Shape shape_b{3, 3}; // filter shape
```
jbobba: Change to {2, 4} or something where the dim sizes are not equal. This will help with whether QD is [m,k] * [n,k] or [m,k] * [k,n]

src/ngraph/builder/quantization/quantized_linear_dot.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+    namespace builder
+    {
+        namespace quantization
+        {
+            shared_ptr<Node> QuantizedDotInteger(shared_ptr<Node> input, shared_ptr<Node> filter)
+            {
+                auto output_scale = make_constant(element::f32, Shape{}, 1);
+                auto reshape_filter =
+                    make_shared<op::Reshape>(filter, AxisVector{1, 0}, filter->get_shape());
```
jbobba: It is weird for a builder to change the order of axis. Cant we stick with the convention used by QuantizedDot? and let bridges deal with reordering if needed. 

Also add a comment in the definition of QuantizedDot that it is [m,k] * [n,k]

src/ngraph/op/experimental/quantized_dot.cpp
```diff
@@ -45,6 +47,8 @@ op::QuantizedDot::QuantizedDot(const shared_ptr<Node>& data,
                           " weights shape ",
                           weights_shape);
 
-    auto output_et = requantize ? (with_relu ? element::u8 : element::i8) : element::f32;
+    auto output_et = is_output_i32
```
adstraw: This nested trinary is a little dense for me.  Prefer that you unpack it into a few lines.

PR #2688, opened 2019-03-29
Adding support for fused ops that are decomposable to core ngraph ops

jbobba:
Adds a FusedOp Op variant and a concrete implementation of PRelu using the decomposition in Onnx frontend + shape inference on the decomposed graph.  Also copies the broadcasting utility functions from there.

- [x] Add FusedOpDecomposition pass for existing backends to support FusedOps
- [x] Add unit tests to test FusedOp execution
- [x] Figure out serializer support 

In Future PRs:
- Autodiff for fused ops
- Multi-output FusedOpDecomposition
- Add backend-configurability to FusedOpDecomposition
- Move util/broadcasting into builder/autobroadcast


src/ngraph/op/util/broadcasting.cpp
```diff
@@ -0,0 +1,296 @@
+//*****************************************************************************
```
aprocter: This seems to duplicate what's already in `src/ngraph/builder/autobroadcast.[ch]pp`. If there's any functionality you need that's not covered by those builders, could you incorporate that functionality into them and remove this?

src/ngraph/op/fused/prelu.cpp
```diff
@@ -0,0 +1,81 @@
...
+    else if (data_shape != slope_shape)
+    {
+        slope = numpy_style_broadcast({slope, data})[0];
+    }
+
+    // x <  0 => f(x) = x * slope
+    // x >= 0 => f(x) = x
+
+    std::shared_ptr<ngraph::Node> zero_node = std::make_shared<ngraph::op::Constant>(
+        data->get_element_type(), ngraph::Shape{}, std::vector<double>{0});
```
nmostafa: Why not create const node with desired shape directly ?
```suggestion
        data->get_element_type(), data->get_shape(), std::vector<double>{0});
```

src/ngraph/op/util/fused_op.cpp
```diff
@@ -0,0 +1,49 @@
...
+using namespace ngraph;
+
+op::util::FusedOp::FusedOp(const std::string& node_type, const NodeVector& args)
+    : Op(node_type, args)
+{
+}
+
+void op::util::FusedOp::validate_and_infer_types()
+{
+    auto subgraph_outputs = decompose_op();
```
nmostafa: Are we decomposing here only to do type inference ? If so, then why not instead infer for the fused node instead (we know its semantics). If not, and we are inferring types on-the-fly while decomposing, then why do we annotate the original node with the inferred types instead of removing it ?

jbobba: We would like to leave the choice to the backends to decide whether they want to deal with fused ops or not. As for type inference directly instead of decomposing, op implementers can choose to do so by overriding this implementation.

src/ngraph/op/util/broadcasting.cpp
aprocter: I actually didn't realize ONNX had its own implementation either... two's company, three's a crowd, I guess ;)

src/ngraph/graph_util.hpp
```diff
@@ -53,6 +53,11 @@ namespace ngraph
                         std::function<void(std::shared_ptr<Node>)> f,
                         bool include_control_deps);
 
+    void traverse_nodes(const NodeVector& subgraph_results,
```
louisfeng: Would be nice to have a short comment on what this version of `traverse_nodes()` do that's different from others.

src/ngraph/pass/fusedop_decomposition.cpp
```diff
@@ -0,0 +1,52 @@
+//*****************************************************************************
```
aprocter: Should probably name this file `fused_op_decomposition.cpp`

src/ngraph/graph_util.cpp
```diff
@@ -110,6 +110,50 @@ void ngraph::traverse_nodes(const NodeVector& io_nodes,
     }
 }
 
+void ngraph::traverse_nodes(const NodeVector& subgraph_results,
+                            const NodeVector& subgraph_params,
```
adstraw: This version of `traverse_nodes` is mostly copy/paste from the existing (above) version.  Why not just add `subgraph_params` or "stop nodes" as I call them as a default argument to the existing version of `traverse_nodes`.  It does not have to be a default parameter --- this just simplifies the API, no need to change other call sites.

src/ngraph/graph_util.hpp
jbobba: Based on @adstraw feedback, i'm extending the current method instead of adding a new one. Will expand on the comment in the cpp file.

src/ngraph/op/fused/prelu.cpp
```diff
@@ -0,0 +1,81 @@
...
+op::PRelu::PRelu(const shared_ptr<Node>& data, const shared_ptr<Node>& slope)
+    : FusedOp("PRelu", {data, slope})
+{
+    constructor_validate_and_infer_types();
+}
+
+NodeVector op::PRelu::decompose_op() const
+{
+    auto data = get_argument(0);
+    auto data_shape = data->get_shape();
```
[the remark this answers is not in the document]
jbobba: you raise a good point. This is going to be tricky though. We will need to construct a graph for internal broadcast shape computation, which will take some time and a separate PR. For now, we could either (a) require FusedOps to use only fully defined shapes or (b) bail out on the decomposition if it runs into partial shapes, but require the op definition to provide its own  validate_and_infer_types() method. Thoughts?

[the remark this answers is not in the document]
aprocter: As of right now I don’t think we have all the ops in place that we would need to do the in-graph shape computation. I think (a) is probably the best solution for now. Or maybe we could just have validate_and_infer_types bail out with a totally unknown shape (PartialShape::dynamic()) in this case.

src/ngraph/graph_util.hpp
```diff
@@ -45,13 +45,15 @@ namespace ngraph
     void traverse_nodes(const std::shared_ptr<const Function> p,
                         std::function<void(std::shared_ptr<Node>)> f,
                         bool include_control_deps = false);
+
     void traverse_nodes(const Function* p,
                         std::function<void(std::shared_ptr<Node>)> f,
                         bool include_control_deps);
 
-    void traverse_nodes(const NodeVector& io_nodes,
+    void traverse_nodes(const NodeVector& subgraph_results,
```
rkimballn1: This needs doxygen comments.

src/ngraph/op/fused_op_tbl.hpp
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+// This collection contains one entry for each fused op.
+//
+
+#ifndef NGRAPH_OP
```
rkimballn1: This is not needed. If NGRAPH_OP is not defined then you will get a compile error.

PR #2689, opened 2019-03-29
Change OneHot to accept only integral types

aprocter:
OneHot currently accepts arbitrary input types for the indices. This changes it to only accept non-real types.

EDIT: er, well, types where `is_real()` is false, i.e., integers. You know what I mean. :)

EDIT EDIT: now we have an `is_integral()` which is, y'know, nice


aprocter:
@postrational I noticed that ONNX actually _does_ allow float types on the input here, but by definition these types are cast to int64 anyway, so I'm thinking that can be covered in the importer. Let me know what you think.

test/type_prop.cpp
```diff
@@ -4343,12 +4343,32 @@ TEST(type_prop, one_hot_deduce_matrix_2)
...
+
 TEST(type_prop, one_hot_deduce_floating_point)
 {
     auto param = make_shared<op::Parameter>(element::f32, Shape{12, 24});
-    auto oh = make_shared<op::OneHot>(param, Shape{12, 24, 8}, 2);
-    ASSERT_EQ(oh->get_element_type(), element::f32);
-    ASSERT_EQ(oh->get_shape(), (Shape{12, 24, 8}));
+    try
+    {
+        auto oh = make_shared<op::OneHot>(param, Shape{12, 24, 8}, 3);
```
rkimballn1: Use EXPECT_THROW or whatever that call is. Your check is too strict in that it expects an exact string match. It should throw an `invalid_argument` exceptions since one of the arguments is invalid. Don't check for string match.

aprocter: Can we revisit this some other time? This pattern is the same for all 378 billion type prop tests, and I do think we need a way to be sure that a reasonably informative error string has been produced (which `EXPECT_THROW` doesn't seem to help with), since users are relying on these error messages.

doc/sphinx/source/ops/one_hot.rst
```diff
@@ -15,11 +15,11 @@ Description
...
 ------
 
-+-----------------+-------------------------+---------------------------------------------------------+
-| Name            | Element Type            | Shape                                                   |
-+=================+=========================+=========================================================+
-| ``arg``         | Any                     | :math:`d_1,\dots,d_{m-1},d_{m+1},\dots,d_n)~(n \geq 0)` |
-+-----------------+-------------------------+---------------------------------------------------------+
++-----------------+-----------------------------+---------------------------------------------------------+
+| Name            | Element Type                | Shape                                                   |
++=================+=============================+=========================================================+
```
diyessi: Non-floating is too general; non-negative integral, or integral if we must.
So adding an `is_integral` would be the right thing.

src/ngraph/op/one_hot.cpp
```diff
@@ -34,6 +34,10 @@ void op::OneHot::validate_and_infer_types()
     PartialShape arg_shape = get_input_partial_shape(0);
     Rank arg_rank = arg_shape.rank();
 
+    NODE_VALIDATION_CHECK(this,
+                          arg_et.is_dynamic() || !arg_et.is_real(),
```
diyessi: `is_integral`

aprocter: What do you mean by `is_integral` here? It's not something that is defined yet in `element::Type`.

test/type_prop.cpp
rkimballn1: It is convenient that English is the only language used in the world, otherwise this could be difficult.

aprocter: These are type prop tests, not backend tests.

test/type_prop.cpp
```diff
@@ -9731,9 +9751,9 @@ TEST(type_prop, topk_invalid_index_type)
     }
     catch (const NodeValidationFailure& error)
     {
-        EXPECT_HAS_SUBSTRING(
```
aprocter: This and several other unrelated `TopK` tests got tripped up by the change to `element::Type`. It looks like these tests are actually screwed up anyway; I'll fix them up in a separate PR.

test/type_prop.cpp
```diff
@@ -4343,12 +4343,32 @@ TEST(type_prop, one_hot_deduce_matrix_2)
     ASSERT_EQ(oh->get_shape(), (Shape{12, 24, 2}));
 }
 
+TEST(type_prop, one_hot_deduce_et_dynamic)
```
indie: > `one_hot_deduce_et_dynamic` 

Curious what the `_et` is referring to here: "element type" or maybe "element test" to determine dynamism?

Since one of the possible implementations for dynamic shapes is "track whether each node depends on dynamic shape information. If none of the results in a function depend on dynamic shapes, then the function does not." seems like this could be ambiguous enough to break graphs that *could* be handled dynamically.  I don't know what this means for this test, though.

aprocter: It stands for element type. I don’t understand the second paragraph of your comment. (Sorry, it’s early and no caffeine yet ;)) What is it that is ambiguous, and what do you mean by breaking graphs?

indie: Some of the drafts on the Intelpedia page for nGraph "Designs" mentioned the various ways that dynamic shapes and types might be implemented.  We also have some notes on how TensorFlow wants things to be done, but there might be other ways.  That's all

indie: Also, we haven't yet added any docs about dynamic graphs or shapes so I probably just don't yet understand these ops well enough.

aprocter: In any case I think this test should be okay :)

PR #2691, opened 2019-03-30
[Dynamic Shape] Moving BatchDot to Core Op

louisfeng:
I still need to add some more unit tests. But most of the implementation are there. Does not do backprop if the inputs have dynamic shapes (per discussion with @aprocter).


src/ngraph/op/experimental/batch_dot.cpp
```diff
@@ -0,0 +1,139 @@
...
+    : Op("BatchDot", check_single_output_args({arg0, arg1}))
+    , m_transpose_arg0(transpose_arg0)
+    , m_transpose_arg1(transpose_arg1)
+{
+    constructor_validate_and_infer_types();
+}
+
+shared_ptr<Node> op::BatchDot::copy_with_new_args(const NodeVector& new_args) const
+{
+    if (new_args.size() != 2)
```
aprocter: This `if` block can be replaced with `check_new_args_count(this, new_args);`

src/ngraph/op/experimental/batch_dot.cpp
```diff
@@ -0,0 +1,139 @@
...
+                          ".");
+
+    size_t dot_dim_arg0 = (m_transpose_arg0) ? 1 : 2;
+    size_t dot_dim_arg1 = (m_transpose_arg1) ? 2 : 1;
+
+    // We expect output shape always have rank 3
+    PartialShape output_shape(PartialShape::dynamic(3));
+
+    // Construct output shape with more information if avalible.
+    if (arg0_shape.rank().is_static() && arg1_shape.rank().is_static() &&
```
aprocter: You can rewrite this condition as:

`if (arg0_shape.rank().same_scheme(3) && arg1_shape.rank().same_scheme(3))`.

(The nomenclature here is odd, perhaps we should change it. Perhaps something like `necessarily_eq`?)

src/ngraph/op/experimental/batch_dot.cpp
```diff
@@ -0,0 +1,139 @@
...
+                              arg0_shape[dot_dim_arg0].compatible(arg1_shape[dot_dim_arg1]),
+                              "Product dimensions are not equal while creating BatchDot.");
+        output_shape =
+            PartialShape{arg0_shape[0], arg0_shape[3 - dot_dim_arg0], arg1_shape[3 - dot_dim_arg1]};
+    }
+
+    set_output_type(0, get_input_element_type(0), output_shape);
+}
+
+void op::BatchDot::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
```
aprocter: Can we add some tests to `test/autodiff.in.cpp`?

src/ngraph/op/experimental/batch_dot.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "ngraph/dimension.hpp"
+#include "ngraph/log.hpp"
+#include "ngraph/op/experimental/dyn_reshape.hpp"
+#include "ngraph/op/reshape.hpp"
+#include "ngraph/util.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+op::BatchDot::BatchDot(const shared_ptr<Node>& arg0,
```
aprocter: You know I'm not sure it's worth bikeshedding, but I wonder if `BatchMatMul` is a better name :)

src/ngraph/op/experimental/batch_dot.cpp
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    return make_shared<BatchDot>(
+        new_args.at(0), new_args.at(1), m_transpose_arg0, m_transpose_arg1);
+}
+
+void op::BatchDot::validate_and_infer_types()
+{
+    // Check input types
+    auto arg0_et = get_input_element_type(0);
```
aprocter: Might be better to use `auto&` here.

test/cpu_fusion.cpp
```diff
@@ -3069,6 +3069,7 @@ TEST(cpu_fusion, fuse_batch_dot_forward)
     auto cpu_f = make_function_from_file(file_name);
     auto int_f = make_function_from_file(file_name);
     pass_manager.run_passes(cpu_f);
+    pass_manager.run_passes(int_f);
```
jbobba: Why do we need to run fusion on interpreter side?

src/ngraph/op/experimental/batch_dot.hpp
```diff
@@ -22,27 +22,35 @@ namespace ngraph
...
-            BatchDot(std::shared_ptr<Node> a,
-                     std::shared_ptr<Node> b,
-                     bool transpose_a,
-                     bool transpose_b);
+            /// \brief Constructs a batch dot product operation.
+            ///
+            /// \param arg0 The node producing the first argument.
+            /// \param arg1 The node producing the second argument.
+            /// \param transpose_0 Apply transpose to arg0.
+            /// \param transpose_1 Apply transpose to arg1.
```
jbobba: Can we add a more mathematical definition? Or at the very least, the axis order convention.

test/cpu_fusion.cpp
louisfeng: I still need to write more unit tests to over different use cases. But in this case, the input graph hare multiple `Dot` operators that need to be fused into `BatchDot`.

src/ngraph/op/experimental/batch_dot.hpp
louisfeng: Sure, I'll add these. Thanks.

src/ngraph/op/experimental/batch_dot.cpp
louisfeng: I see a good point here since we always deal with matrices here. I'll rename it.

src/ngraph/op/experimental/batch_dot.cpp
louisfeng: Fixed. It looks to the logic in `same_scheme()` is essentially what I would call `equal`.

src/ngraph/op/experimental/batch_dot.cpp
louisfeng: fixed.

src/ngraph/op/experimental/batch_dot.cpp
louisfeng: @jbobba If I rename this to `BatchMatMul`, I think the backend should be renamed too? What do you think? Keep it as `BatchDot` or rename?

src/ngraph/op/experimental/batch_dot.hpp
louisfeng: Fixed.

src/ngraph/op/experimental/batch_dot.cpp
diyessi: `BatchMatMul` seems to be the name frameworks use, so we should consistently go with that.

src/ngraph/op/experimental/batch_dot.hpp
```diff
@@ -0,0 +1,62 @@
...
+        /// `(batch_size, n, m)`, and `BatchDot(a, b)[i] = Dot(a[i], b[i])`.
+        class BatchDot : public Op
+        {
+        public:
+            /// \brief Constructs a batch dot product operation.
+            ///
+            /// \param arg0 The node producing the first argument.
+            /// \param arg1 The node producing the second argument.
+            /// \param transpose_0 Apply transpose to arg0.
+            /// \param transpose_1 Apply transpose to arg1.
```
diyessi: We have ops for doing these transposes, so let's use them instead of complicating this op. Backends can fuse them if they want to.

louisfeng: Sounds good, I'll leave make the core op simple without transpose and leave the CPU version of batch dot and fusion unchanged in this PR. Additional fusion can be done later, but don't want to block PDPD to start using batchmatmul.

src/ngraph/op/experimental/batch_dot.cpp
aprocter: What I was thinking is that it's important to distinguish between the case where s1 and s2 are necessarily the same shape at runtime (meaning, really, that both are static and equal), and the case where s1 and s2 represent quantities that statically aren't yet known to be different but could be at runtime (e.g. a [?,2] over here and [?,2] over there could respectively wind up being [3,2] and [4,2] in a given iteration).

louisfeng:
Rename and changed interface based on feedback from @diyessi, it also impacted new changes in CPU transformer for @jbobba to review.

src/ngraph/op/experimental/batch_dot.hpp
louisfeng: Fixed.

src/ngraph/op/experimental/batch_dot.cpp
louisfeng: Fixed.

test/backend_test.in.cpp
```diff
@@ -7371,3 +7376,74 @@ NGRAPH_TEST(${BACKEND_NAME}, quantize_dynamic_offset)
     EXPECT_EQ((vector<output_c_type>{1, 1, 2, 3, 3, 3, 4, 5, 5, 5, 6, 7}),
               read_vector<output_c_type>(y));
 }
+
+#if defined(BACKEND_TEST_CPU) || defined(BACKEND_TEST_INTERPRETER)
```
jbobba: If you move batch_mat_mul to fused ops and provide a decomposition, then you wouldn't need to make this exemption.

PR #2695, opened 2019-04-01
Add windows support for event trace

rkimballn1:
Also enabled multi-threaded windows build so it should go faster.


rkimballn1:
@jfdong I added windows.h

PR #2699, opened 2019-04-02
[Warnings] Silence more warnings

dcaballe:
Unused variables and missing switch case.

### DIFF
--- a/src/ngraph/runtime/host_tensor.hpp
+++ b/src/ngraph/runtime/host_tensor.hpp
@@ -26,8 +26,6 @@ namespace ngraph
 {
     namespace runtime
     {
-        static __attribute__((unused)) size_t alignment = 64;
-
         class HostTensor;
     }
 }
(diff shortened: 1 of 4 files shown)


src/ngraph/runtime/host_tensor.hpp
```diff
@@ -26,8 +26,6 @@ namespace ngraph
 {
     namespace runtime
     {
-        static size_t alignment = 64;
-
         class HostTensor;
     }
 }
```
dcaballe: Moved the static global var to the .cpp file since only host_tensor.cpp is using it. We can turn it into a member if this alignment is a property of HostTensor but I think the current change should be fine.

I also found interesting that we are doing a fw declaration of HostTensor (line 29) to immediately declare it (line 33). Is this a common practice for some reason? Just to avoid the namespace indentation?

diyessi: Probably something got removed that used to be in between the two.

PR #2705, opened 2019-04-03
[ONNX] Unit test models for QLinearConv

postrational:


src/ngraph/runtime/cpu/builder/quantized_conv.cpp
```diff
@@ -79,7 +87,46 @@ namespace ngraph
...
-                    throw ngraph_error("unsupported parameters for QuantizedConvolution via DEX");
+                    // Needs generalizing
+                    std::function<decltype(
+                        runtime::cpu::kernel::quant_convolution<uint8_t, uint8_t, uint8_t>)>
+                        kernel;
+                    kernel = runtime::cpu::kernel::quant_convolution<uint8_t, uint8_t, uint8_t>;
+
+                    auto scale_const_op =
+                        std::dynamic_pointer_cast<ngraph::op::Constant>(node->get_arguments()[2]);
+                    auto scale_val = scale_const_op->get_vector<float>();
```
jbobba: This might fail if the scale argument is not a constant.

src/ngraph/runtime/reference/quantized_convolution.hpp
```diff
@@ -0,0 +1,240 @@
...
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace reference
+        {
+            // in: NC_I...
+            // filter: C_OC_I...
+            // out: NC_O...
+            template <class INPUT, class FILTER, class OUTPUT>
```
jbobba: Scott can decide but looks like this can easily fit into the convolution.hpp by modifying the general_convolution there.

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
```diff
@@ -79,7 +87,46 @@ namespace ngraph
...
+                               arg1_shape,
+                               result_shape,
+                               window_movement_strides,
+                               window_dilation_strides,
+                               padding_below,
+                               padding_above,
+                               data_dilation_strides,
+                               scale_val[0]);
+                    };
+                    functors.emplace_back(functor);
```
jbobba: Can you add a similar fallback to cpu_emitter.cpp as well?

src/ngraph/runtime/reference/quantized_convolution.hpp
nbpatel: For u8u8u8, we can probably use the existing conv ref kernel with some minimal changes, but in case of different input/filter/output type, we might need some additional changes to the definition of original ref kernel. but yes certainly doable, but this just seemed straightforward given we had a deadline to chase. will wait on @diyessi to make the call

src/ngraph/op/experimental/quantized_conv.cpp
```diff
@@ -40,15 +40,17 @@ op::QuantizedConvolution::QuantizedConvolution(const shared_ptr<Node>& data_batc
     , m_data_dilation_strides(data_dilation_strides)
     , m_requantize(requantize)
 {
-    constructor_validate_and_infer_types();
```
aprocter: If possible, please put this back in. Otherwise, please add a comment on why we are skipping it.

adstraw:
@aprocter and @jbobba the comments are outdated.  I had merged a change @nbpatel that is now reverted.  if you want to comment on that change see #2713

src/ngraph/builder/quantization/quantized_linear_convolution.cpp
```diff
@@ -56,6 +61,48 @@ namespace ngraph
                                                              requantization_scale);
             }
 
+            shared_ptr<Node> QuantizedLinearConvolution(shared_ptr<Node> input,
```
jbobba: QuantizedLinearConvolutionFallback? or Add comments on why we are falling back to fp32.

test/onnx_import.in.cpp
```diff
@@ -2375,3 +2375,147 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_quant_conv_linear)
...
     EXPECT_TRUE(test::all_close(expected_output.front(), outputs.front()));
 }
+
+NGRAPH_TEST(onnx_${BACKEND_NAME}, model_quant_conv_linear_2d)
+{
+    auto function = onnx_import::import_onnx_model(
+        file_util::path_join(SERIALIZED_ZOO, "onnx/qlinear_conv_2d.prototxt"));
+
+    auto x =
+        read_binary_file<uint8_t>(file_util::path_join(SERIALIZED_ZOO, "onnx/qlinearconv2d/x.bin"));
```
arogowie-intel: Would be nice if you could move all these binary files to `test/files/onnx/` directory. Then just use `TEST_FILES` symbol instread of `SERIALIZED_ZOO`:
```suggestion
        read_binary_file<uint8_t>(file_util::path_join(TEST_FILES, "onnx/qlinearconv2d/x.bin"));
```

adstraw:
@diyessi let's go ahead and merge this since all feedback is nice-to-have.  this way we can proceed on with the release.  I will follow up with a PR to address the feedback.

PR #2710, opened 2019-04-04
TopK: return the lower index for equivalent values

tomdol:
This PR is a result of analysis of a test in [onnxruntime](https://github.com/Microsoft/onnxruntime/blob/master/onnxruntime/test/providers/cpu/math/topk_op_test.cc#L42-L49)

When ran using nGraph backend, the computed indices were `{3, 2}` while the test expects `{3, 1}`.

According to the [ONNX TopK specs](https://github.com/onnx/onnx/blob/master/docs/Operators.md#TopK): 
> Given two equivalent values, this operator uses the indices along the axis as a tiebreaker. That is, the element with the lower index will appear first.


src/ngraph/runtime/reference/topk.hpp
```diff
@@ -33,7 +33,18 @@ namespace ngraph
             template <typename T, typename U>
             static bool compare_max(const std::tuple<T, U>& a, const std::tuple<T, U>& b)
             {
-                return a > b;
+                if (std::get<0>(a) > std::get<0>(b))
```
diyessi: Tuple comparison is supposed to be lexical so I'm not sure why you should need all this, or even `compare_max`.

tomdol: As discussed on slack - will go back the solution with direct comparison of elements at index 0 and disable the clang warning using pragma.

tomdol:
Michał and Adam removed from the reviewers - they are aware of this PR but are flying back to PL and that would block the merge.

aprocter:
@avijit-nervana @mbrookhart @ransford2011 Please give a quick up-or-down on the change to semantics, no need for full review.

PR #2713, opened 2019-04-04
[WIP] Ref conv kernel for Quantized conv 

nbpatel:
Not mature yet. Needs generalizing and some edge case handling.

### DIFF
--- a/src/ngraph/op/experimental/quantized_conv.cpp
+++ b/src/ngraph/op/experimental/quantized_conv.cpp
@@ -42,6 +42,8 @@ op::QuantizedConvolution::QuantizedConvolution(const shared_ptr<Node>& data_batc
 {
     //TODO(nbpatel): Add checks.
 
+    constructor_validate_and_infer_types();
+
     auto& data_batch_shape = data_batch->get_shape();
     auto& filters_shape = filters->get_shape();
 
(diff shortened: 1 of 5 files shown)


src/ngraph/op/experimental/quantized_conv.cpp
```diff
@@ -40,15 +40,17 @@ op::QuantizedConvolution::QuantizedConvolution(const shared_ptr<Node>& data_batc
     , m_data_dilation_strides(data_dilation_strides)
     , m_requantize(requantize)
 {
-    constructor_validate_and_infer_types();
-
     //TODO(nbpatel): Add checks.
```
diyessi: This should all be done in `validate_and_infer_types`, which is when element type and shape propagation will happen in the near future. Likewise, you need to be calling `constructor_validate_and_infer_types` which will become a no-op and then be removed in the near future, but, currently, does the type/shape propagation.

src/ngraph/runtime/reference/quantized_convolution.hpp
```diff
@@ -0,0 +1,240 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
diyessi: I think I might be able to get the other convolution template to do this.

adstraw:
I created some more unit tests for QLinearConv in PR #2705 .  Let's plan to merge the changes here through that PR.

nbpatel:
#2840

PR #2716, opened 2019-04-05
DeconvBias

gaurides:
Fuses ConvolutioBackpropData + BatchNormInference -> DeconvolutionBias
set env variable `NGRAPH_DECONV_FUSE=1` to enable the fusion.
and the unit test cases are disabled for now.
Currently, the performance of deconvolution is not better than ConvolutionBackpropData. Mostly likely due to the kernels. Will keep monitoring and enable when the kernel performance is improved.


src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -1484,6 +1568,52 @@ namespace ngraph
...
+                    // MKLDNN relies on named formats for kernel selection
+                    if (weights_desc.data.format == mkldnn_nchw)
+                    {
+                        weights_desc.data.format = mkldnn_oihw;
+                    }
+                    if (weights_desc.data.format == mkldnn_ncdhw)
+                    {
+                        weights_desc.data.format = mkldnn_oidhw;
+                    }
+                    weights_desc.data.format = mkldnn_any;
```
jbobba: needs a comment here.

src/ngraph/runtime/cpu/op/deconv.cpp
```diff
@@ -0,0 +1,169 @@
+//*****************************************************************************
+// Copyright 2017-2018 Intel Corporation
```
jbobba: change to 2017-2019

src/ngraph/runtime/cpu/op/deconv.hpp
```diff
@@ -0,0 +1,129 @@
+//*****************************************************************************
+// Copyright 2017-2018 Intel Corporation
```
jbobba: 2017-2019

test/cpu_fusion.cpp
```diff
@@ -1108,6 +1109,88 @@ TEST(cpu_fusion, conv_add)
...
+                  ? std::make_shared<op::BatchNormInference>(goe_bn, gamma, beta, mean, var, eps)
+                  : std::make_shared<op::BatchNormInference>(conv, gamma, beta, mean, var, eps);
+
+    return make_shared<Function>(NodeVector{bn},
+                                 ParameterVector{filters, out_delta, gamma, beta, mean, var});
+}
+
+// Note: enable deconv tests, when the deconv perf is better than convbackpropdata
+TEST(cpu_fusion, DISABLED_fuse_deconv)
+{
```
jbobba: Instead of disabling, you could explicitly set the environment flag to ensure that the code path is tested https://github.com/NervanaSystems/ngraph/blob/master/test/cpu_test.cpp#L132-L137

PR #2717, opened 2019-04-06
allow multiple thread safe calls to compiled function in DEX mode

ayzhuang:
1. Use NGRAPH_CONCURRENCY=m to create m copies of cpu runtime contex. Default is 1.
2. Use multiple threads to call the same compiled function. Please see the unit test: cpu_test.thread_safe_calls_convolution_2d_2items.


avijit-nervana:
Added @sayantan-nervana to review the details as this may impact the TensorFlow bridge.

PR #2718, opened 2019-04-06
Event Trace enhancement

rkimballn1:
- First concern was performance. Added benchmark.event_tracing unit test to measure. Starting time was 14,901ms. After optimization benchmark time is now 2,625ms or a 5.7x performance increase.
- Performance when disabled is about 160x faster than before.
- Moved event tracing to the ngraph::event namespace
- Renamed Event class to event::Duration
- Added event::Object which tracks object lifetimes but I can't figure out how to display anything in chrome://tracing. The json files look good though.
- Fix event unit test to generate a valid chrome tracing json file and completed the unit test by checking the results.
- Event file is 1/2 the size


src/ngraph/event_tracing.cpp
```diff
@@ -19,72 +19,191 @@
...
+#include "ngraph/log.hpp"
 #include "nlohmann/json.hpp"
 
 using namespace std;
+using namespace ngraph;
 
 static bool read_tracing_env_var()
 {
-    return (std::getenv("NGRAPH_ENABLE_TRACING") != nullptr);
+    static const bool is_enabled = (getenv("NGRAPH_ENABLE_TRACING") != nullptr);
```
diyessi: Longer term it would make more sense to have a way for a bridge to turn this on and off.

mbrookhart: 👍

rkimballn1: There are already calls to do this
```
ngraph::event::Manager::enable_event_tracing()
ngraph::event::Manager::disable_event_tracing()
```

PR #2724, opened 2019-04-08
Deprecate direct access to descriptor::Input and descriptor::Output

aprocter:
This PR is part of a longer march towards simplifying the `Node` class and strengthening its encapsulation, and getting rid of `GetOutputElement`.

This PR adds:

* Accessor classes that allow us to replace direct access to the `descriptor::Input` and `descriptor::Output` classes with access via short-lived "handle" objects. These new classes are called `Input` and `Output`, without the `descriptor::` in front.
  - Actually introduced in #2612, but overhauled here: renamed them, templatized them to allow the underlying `Node` object to be `const`, and ditched internal storage of `shared_ptr`.
* "Optional deprecation" (see below) of several old APIs in the Node class. (In particular, the only way to get at a node's Input/Output descriptors is through now-deprecated functions.)
  - Deprecated functions are:
    * `get_inputs()`
    * `get_outputs()`
    * `get_output_inputs()`
    * `get_output_tensor()`
    * `get_output_tensor_ptr()`
  - APIs that *will* be deprecated, but are not yet:
    * `get_element_type()`
    * `get_shape()`
    * `get_output_element_type()`
    * `get_output_shape()`
    * `get_output_partial_shape()`
    * `get_input_element_type()`
    * `get_input_shape()`
    * `get_input_partial_shape()`
  - "Optional deprecation" means that there is a compile flag, defaulting to OFF, to turn on actual deprecation of these functions. I made this optional because (1) CPU backend hasn't been updated yet, and (2) didn't want to disrupt things for the moment for backends that don't live in the `ngraph` repo. (But on the other hand, being able to turn on deprecation is nice if you are trying to search and destroy instances of the deprecated APIs).
  - To turn on the optional deprecation, run cmake with `-DNGRAPH_DEPRECATE_OLD_NODE_APIS=ON`. This defaults to `OFF`.
* Updates to eliminate all uses of deprecated APIs, **outside of the CPU backend**.
  - I'm treating the CPU backend as out of bounds for this PR, since the PR is already rather large and the optional deprecation means we can attack it separately.
* Documented replacements for all deprecated functions.

I would recommend focusing initial attention on `node.hpp`, and working outward from there.


src/ngraph/node.hpp
```diff
@@ -37,30 +37,30 @@
 #include "ngraph/node_vector.hpp"
 #include "ngraph/placement.hpp"
 
+// There is work in progress on deprecating some old APIs here, and for development purposes it is
+// nice to have a compile-time option to turn on actual deprecation.
+//
+// Enable the deprecation with -DNGRAPH_DEPRECATE_OLD_NODE_APIS=ON at cmake time.
+#ifdef NGRAPH_DEPRECATE_OLD_NODE_APIS
```
rkimballn1: This could be made more general by calling it NGRAPH_DEPRECATED and put it in deprecated.hpp. Then it could be used in other places where it really should be used.

src/ngraph/node.hpp
```diff
@@ -324,6 +376,309 @@ namespace ngraph
...
+        const PartialShape& get_partial_shape() const
+        {
+            return m_node->get_input_partial_shape(m_index);
+        }
+        /// \return A handle to the output that is connected to this input.
+        Output<Node> get_source_output() const;
+        /// \return A reference to the tensor descriptor for this input.
+        descriptor::Tensor& get_tensor() const
+        {
+            return m_node->m_inputs.at(m_index).get_output().get_tensor();
```
rkimballn1: Implement these by calling proper Node methods to get an output. Might not even need to be friends of Node, because who really needs friends?
This holds for next couple of methods, too.
If you don't want to use methods than at least select between .at(x) and [x] in these methods.

src/ngraph/node.hpp
```diff
@@ -324,6 +376,309 @@ namespace ngraph
...
+        /// \return A `shared_ptr` to the node referred to by this output handle.
+        ///
+        /// TODO: Make a plan to deprecate this.
+        std::shared_ptr<NodeType> get_node_shared_ptr() const { return m_node->shared_from_this(); }
+        /// \return The index of the output referred to by this output handle.
+        size_t get_index() const { return m_index; }
+        /// \return A reference to the tensor descriptor for this output.
+        descriptor::Tensor& get_tensor() const
+        {
+            return m_node->m_outputs.at(m_index).get_tensor();
```
rkimballn1: Use node methods to get access to output. Same for all other methods in this class which directly access Node's innards

src/ngraph/op/replace_slice.cpp
```diff
@@ -176,11 +176,10 @@ void op::ReplaceSlice::generate_adjoints(autodiff::Adjoints& adjoints, const Nod
 {
     auto delta = deltas.at(0);
 
-    auto x = get_inputs().at(0).get_output().get_node();
-    auto& y_input = get_inputs().at(1);
-    auto y = y_input.get_output().get_node();
-    auto& y_element_type = y_input.get_element_type();
-    auto y_shape = y_input.get_shape();
+    auto x = get_argument(0);
```
rkimballn1: OMG I can actually read this. What sorcery is this? :)

src/ngraph/graph_util.cpp
```diff
@@ -168,14 +168,11 @@ void ngraph::replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> re
     // For each of target's output O with replacement output O_rep:
     //     For each O's connected downstream input I:
     //         Change I's connected upstream output to O_rep
-    for (size_t i = 0; i < target->get_outputs().size(); i++)
+    for (size_t i = 0; i < target->get_output_size(); i++)
```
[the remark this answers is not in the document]
aprocter: I agree, but could we do it in a later PR? It's going to touch a lot of files and drag in a lot of `CODEOWNERS`.

src/ngraph/graph_util.cpp
```diff
@@ -357,10 +354,19 @@ pair<shared_ptr<op::Result>, shared_ptr<op::Parameter>>
...
+    auto& dst_input = dst_inputs[0];
+
+    std::vector<Output<Node>> src_outputs = get_outputs_to(*src_node, *dst_node);
+    NGRAPH_ASSERT(src_outputs.size() == 1) << "insert_result_parameter_split encountered more than "
+                                              "one output between the source and destination nodes";
+    auto& src_output = src_outputs[0];
+
+    src_output.remove_target_input(dst_input); // Remove [0]
+    dst_input.replace_source_output(
+        par_node->output(0)); // Remove [0] (again), add [8], remove [1], add [9]
```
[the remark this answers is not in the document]
aprocter: Will fix

src/ngraph/graph_util.cpp
```diff
@@ -530,18 +545,18 @@ size_t ngraph::get_user_count(Node* node)
 
 bool ngraph::possibly_overwritten(Node* node)
 {
-    for (const descriptor::Output& output : node->get_outputs())
+    for (auto& output : node->outputs())
```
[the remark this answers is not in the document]
aprocter: I think _technically_ no according to our coding standards (which I don't agree with on this point anyway), because `get_` is supposed to be used for trivial getters, which this is not (at least not yet). I kinda feel like it reads better as `outputs()`, but I wouldn't have major aesthetic objections to changing it.

For now the problem is that `get_outputs()`, which returns `deque<descriptor::Output*>`, already exists (albeit deprecated) and can't be overloaded.

src/ngraph/graph_util.cpp
```diff
@@ -596,3 +611,44 @@ void ngraph::plot_graph(
     pass_manager.register_pass<ngraph::pass::VisualizeTree>(filename, attributes);
     pass_manager.run_passes(f);
 }
+
+std::vector<Input<Node>> ngraph::get_inputs_from(Node& src, Node& dst)
```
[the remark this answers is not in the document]
aprocter: Honestly I've been doing it that way pretty consistently all along. :)

src/ngraph/node.hpp
```diff
@@ -324,6 +376,309 @@ namespace ngraph
...
+        /// \param new_source_output A handle for the output that will replace this input's source.
+        void replace_source_output(const Output<Node>& new_source_output) const;
+
+        bool operator==(const Input& other) const
+        {
+            return m_node == other.m_node && m_index == other.m_index;
+        }
+        bool operator!=(const Input& other) const
+        {
+            return m_node != other.m_node || m_index != other.m_index;
```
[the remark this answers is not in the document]
aprocter: Will do

src/ngraph/node.hpp
aprocter: I like it.

src/ngraph/node.hpp
aprocter: Doing it this way makes me sad, but:
- We still need the `descriptor::*` classes to be around for now, because existing deprecated APIs return pointers to them.
- We need a single source of truth for input/output info, and for the moment I think that has to be the (legacy) descriptor classes.
- The accessors are deprecated, so if we use them here we would be throwing deprecation warnings in core code (which I was trying hard to get rid of in this PR).
- Once we actually retire the `descriptor::*` accessors, we can come up with a cleaner interface here.

src/ngraph/op/replace_slice.cpp
aprocter: In fairness, it could also have been written more readably with the old APIs :)

src/ngraph/node.hpp
aprocter: (Same comment)

src/ngraph/node.hpp
aprocter: Have created an `NGRAPH_DEPRECATED` macro in `deprecated.hpp`, enabled at cmake time with `-DNGRAPH_DEPRECATED_ENABLE=ON`.

src/ngraph/node.hpp
aprocter: Think I've addressed the `at` vs `[]` thing, BTW.

src/ngraph/graph_util.cpp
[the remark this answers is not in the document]
rkimballn1: I agree with you that I disagree with the coding standard. The issue I have is that if I change the `implementation` of a method I may then need to change the `method name`. Exposing the implementation indirectly this way is not needed and not followed in our coding. The fact that this rule is often blatantly ignored means it is a bad rule and should be changed.

PR #2727, opened 2019-04-09
Replace NGRAPH_ASSERT with faster NGRAPH_CHECK; remove all casserts

aprocter:
Extension of #2537 to supersede `NGRAPH_ASSERT`. This requires minor changes to callers.

Before:

```
NGRAPH_ASSERT(cond) << some_error_message_info << some_more_error_message_info;
```

After:

```
NGRAPH_CHECK(cond, some_error_message_info, some_more_error_message_info);
```

What we get in return is lazy evaluation of the error string, which means that `NGRAPH_CHECK` is suitable for frequently-evaluated checks, unlike `NGRAPH_ASSERT`.

I've left the old `NGRAPH_ASSERT` infrastructure in place for now because (1) ONNX is using it for some custom macros; and (2) `NGRAPH_ASSERT` itself is being used in some backends.

As long as I was here, I decided to get rid of all remaining uses of `assert`, with one exception inside some generated code (I added a `TODO` to take another look at that).

**Heads up**: This requires the use of a GNU extension to allow zero-argument invocations of variadic macros. I believe this extension is available in all of our supported compilers, but we will have to verify this. If it turns out there are problems with compiler compatibility, I will have to go back to the drawing board and come back with something slightly less elegant.

Backend folks: the old `NGRAPH_ASSERT` is still around so this should not break anything for you. We may deprecate more noisily in a forthcoming release, and remove it sometime after that.


src/ngraph/check.hpp
```diff
@@ -66,14 +66,17 @@ namespace ngraph
...
 // shadowing here.
-#define NGRAPH_CHECK(exc_class, ctx, check, ...)                                                   \
+#define NGRAPH_CHECK_HELPER(exc_class, ctx, check, ...)                                            \
     do                                                                                             \
     {                                                                                              \
         if (!(check))                                                                              \
         {                                                                                          \
             ::std::stringstream ss___;                                                             \
-            ::ngraph::write_all_to_stream(ss___, __VA_ARGS__);                                     \
+            ::ngraph::write_all_to_stream(ss___, ##__VA_ARGS__);                                   \
```
aprocter: This `##` here is the "token-pasting hack" I was referring to in the PR comment. It basically means "backspace the comma if the number of varargs is zero."

src/ngraph/assertion.hpp
```diff
@@ -22,6 +22,16 @@
 
 #include "ngraph/except.hpp"
 
+// ****************************
```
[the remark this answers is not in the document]
aprocter: It’s being used in the ONNX importer (which I suppose I could update in this PR) as well as in some separate repos (which is harder to synchronize the update with, hence my suggestion to maintain as deprecated for now)

src/ngraph/check.hpp
diyessi: @aslepko can you run Windows CI on this PR?

test/util/ndarray.hpp
```diff
@@ -23,7 +23,6 @@
 #pragma once
```
aprocter: @diyessi You might want to check this file to make sure what I've done makes sense.

aslepko:
TESTNOW

src/ngraph/check.hpp
```diff
@@ -57,23 +57,101 @@ namespace ngraph
...
             throw exc_class(                                                                       \
                 (::ngraph::CheckLocInfo{__FILE__, __LINE__, #check}), (ctx), ss___.str());         \
         }                                                                                          \
     } while (0)
+
+/// \brief Macro to check whether a boolean condition holds.
+/// \param cond Condition to check
+/// \param ... Additional error message info to be added to the error message via the `<<`
```
arogowie-intel: Probably you should change `<<` to _varargs_

src/ngraph/check.hpp
```diff
@@ -57,23 +57,101 @@ namespace ngraph
             std::stringstream ss;
             ss << "Check '" << check_loc_info.check_string << "' failed at " << check_loc_info.file
                << ":" << check_loc_info.line << ":" << std::endl;
```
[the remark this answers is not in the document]
aprocter: Done, thanks!

PR #2733, opened 2019-04-10
Moves some fused convolution ops to core FusedOps

jbobba:
- Adds support for decomposing and replacing multi-output FusedOps
- Adds query callbacks to FusedOpDecomposition to check if a FusedOp is
  supported by a backend
- Adds core fusion patterns for FusedOps
-


src/ngraph/pass/core_fusion.hpp
```diff
@@ -29,17 +29,25 @@ namespace ngraph
 class ngraph::pass::CoreFusion : public ngraph::pass::GraphRewrite
 {
 public:
-    CoreFusion()
+    CoreFusion(ngraph::pass::FusionType fusions = ngraph::pass::REGULAR_FUSIONS)
```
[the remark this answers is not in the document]
jbobba: I should probably add a comment. REGULAR_FUSIONS dont create any non-core ops. FOP_FUSIONS create FusedOps that backends might not support. The recommendation when specifying FOP_FUSIONS is to follow it up with a FusedOpDecomposition pass which has the ability to break down ops that are not supported by the backend. 

We have a way to disable individual fusions buried in here: https://github.com/NervanaSystems/ngraph/blob/13fc556ead765e9a04613a45e7bbfbbd6a2ddf9f/src/ngraph/pass/graph_rewrite.cpp#L100 I can look into making it more accessible to backends in a separate PR.

src/ngraph/op/fused/conv_fused.cpp
```diff
@@ -16,64 +16,56 @@
```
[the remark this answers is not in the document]
jbobba: i was going to bring more ops like conv_relu and conv_add here.

src/ngraph/op/fused/conv_fused.cpp
```diff
@@ -92,38 +84,92 @@ op::ConvolutionBias::ConvolutionBias(const shared_ptr<Node>& data_batch,
...
+                      Strides())
+{
+}
+
+/// Overrides the default shape inference utility provided by FusedOp
+/// based on FusedOp decomposition.
+///
+/// This implementation handles partial shapes and adjust conv attributes
+/// to support simplified ConvolutionBias op construction
+void op::ConvolutionBias::validate_and_infer_types()
```
aprocter: Is this override going to allow/disallow things that are not allowed/disallowed by the default implementation?

src/ngraph/op/fused/conv_fused.cpp
```diff
@@ -220,7 +297,6 @@ op::ConvolutionBiasBackpropFiltersBias::ConvolutionBiasBackpropFiltersBias(
             "match");
```
aprocter: The check for element type mismatch should be done in `validate_and_infer_types()`, not in the constructor.

jbobba: Will remove it since its redundant. convbprop_filters will check for this.

src/ngraph/op/fused/conv_fused.cpp
jbobba: Does a few more things compared to default. Primarily, setting strides, padding if they are not already specified.

PR #2736, opened 2019-04-10
IntelGPU backend: TopK operation

aalbersk:


src/ngraph/runtime/intelgpu/intelgpu_op_custom_kernels.cpp
```diff
@@ -2168,6 +2168,149 @@ void runtime::intelgpu::do_dequantize_operation(cldnn::topology& topology,
...
+    writer.block_end();
+
+    const cldnn::custom_gpu_primitive op_topk(output_name,
+                                              {input_name},
+                                              {writer.get_code()},
+                                              entry_point_name,
+                                              get_kernel_args(1, 1),
+                                              "",
+                                              layout,
+                                              gws);
```
shssf: as far as @dmyershov mentioned, `gws` is not initialized here. Perhaps it is need to be initialized by some values before (at least {1})

shssf:
we also need to implement `TopK` operation properties in `visualize_tree.cpp` for graph visualization

PR #2739, opened 2019-04-10
Serializer supports new pad api

fboemer:
Previously, a Pad op without the `padding_interior` field would result in a parsing error.

### DIFF
--- a/src/ngraph/serializer.cpp
+++ b/src/ngraph/serializer.cpp
@@ -981,10 +981,10 @@ static shared_ptr<ngraph::Function>
                 // This is a legacy field whose functionality is no longer supported. The new
                 // behavior is equivalent to interior padding of 0, so we will accept it under
                 // those conditions.
-                auto it = node_js.find("padding_interior");
-                if (it != node_js.end())
+                auto padding_interior_maybe = node_js.find("padding_interior");
+                if (padding_interior_maybe != node_js.end())
                 {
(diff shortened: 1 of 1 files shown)


src/ngraph/serializer.cpp
```diff
@@ -981,11 +981,15 @@ static shared_ptr<ngraph::Function>
                 // This is a legacy field whose functionality is no longer supported. The new
                 // behavior is equivalent to interior padding of 0, so we will accept it under
                 // those conditions.
-                auto padding_interior = node_js.at("padding_interior").get<vector<size_t>>();
-                NGRAPH_ASSERT(std::all_of(padding_interior.begin(),
-                                          padding_interior.end(),
-                                          [](size_t s) { return s == 0; }))
-                    << "Legacy padding_interior field must be zero everywhere.";
+                auto padding_interior_maybe = node_js["padding_interior"];
+                if (!padding_interior_maybe.empty())
```
aprocter: 🤦‍♂️ my bad, thanks for the fix :)

rkimballn1: Do something like this instead. This pattern is used in other places.

```
            auto it = node_js.find("padding_interior");
            if (it != node_js.end())
            {
                auto padding_interior = it->get<vector<size_t>>();
                <more stuff here>
            }
```

PR #2740, opened 2019-04-10
remove ngraph::aligned_alloc and replaced with AlignedBuffer

rkimballn1:
AlignedBuffer is RAII
AlignedBuffer is easier to use. aligned_alloc has the requirement that the input size must be an integral multiple of the alignment while AlignedBuffer has no such requirement.


src/ngraph/op/constant.hpp
```diff
@@ -44,8 +44,8 @@ namespace ngraph
                 : Node("Constant", {})
                 , m_element_type(type)
                 , m_shape(shape)
-                , m_data(ngraph::aligned_alloc(m_element_type.size(),
-                                               shape_size(m_shape) * m_element_type.size()))
+                , m_data(
+                      new runtime::AlignedBuffer(shape_size(m_shape) * m_element_type.size(), 64))
```
jbobba: Minor nit. Macro to define 64?

rkimballn1: Made a static constant to hold this magic number

PR #2743, opened 2019-04-11
Adding auto pad to convolution and pooling

jbobba:
- [x] Move auto pad computations into separate utility methods 
- [x] Add auto pad to pooling ops
- [x] More unit tests
- [x] Serializer support


src/ngraph/op/util/attr_types.hpp
```diff
@@ -0,0 +1,53 @@
...
+        };
+
+        /// \brief Padding Type used for `Convolution` and `Pooling`
+        ///
+        /// Follows ONNX padding type definitions
+        /// EXPLICIT   - Pad dimensions are explicity specified
+        /// SAME_UPPER - Pad dimensions computed to match input shape
+        ///              Ceil(num_dims/2) at the beginning and
+        ///              Floor(num_dims/2) at the end
+        /// SAME_UPPER - Pad dimensions computed to match input shape
```
mbrookhart: Should this be SAME_LOWER?

src/ngraph/validation_util.cpp
```diff
@@ -589,3 +589,27 @@ std::tuple<element::Type, PartialShape, PartialShape>
...
+        int64_t filter_size = (static_cast<int64_t>(filter_shape[i]) - 1) * filter_dilations[i] + 1;
+        int64_t filter_stride = static_cast<int64_t>(filter_strides[i]);
+        auto output_size = (image_size + filter_stride - 1) / filter_stride;
+
+        auto padding_needed =
+            std::max(int64_t(0), (output_size - 1) * filter_stride + filter_size - image_size);
+        auto padding_lhs = padding_needed / 2;
+        auto padding_rhs = padding_needed - padding_lhs;
+        padding_below.push_back(pad_type == op::PadType::SAME_UPPER ? padding_lhs : padding_rhs);
+        padding_above.push_back(pad_type == op::PadType::SAME_UPPER ? padding_rhs : padding_lhs);
```
mbrookhart: I'm not sure this logic will always return padding that produces the output shape you want.

mbrookhart: Should one of them be SAME_LOWER?

jbobba: padding_lhs is the smaller number. SAME_UPPER will assign the smaller number to padding_below and larger number to padding_above. SAME_LOWER reverses this.

jbobba:
> LGTM.
> 
> current behaviour bridge side is that we are currently using `PadType& pad_type = PadType::EXPLICIT`. Since that is the default no changes required in the bridge.
> 
> @jbobba: Currently I see this is implemented for Conv, MaxPool and AvgPool, in the future will this be supported for other paddable ops as well, such as AvgPoolBackprop etc?

Yes, we will need to do that.

src/ngraph/validation_util.cpp
mbrookhart: So you're assuming whenever this function is called, the pad_type is SAME_LOWER or SAME_UPPER? Can you add a check for that/

PR #2747, opened 2019-04-11
Adds provenance tag propagation

tachyon77:
Adds ability to enable/disable provenance using environment variable.


src/ngraph/graph_util.cpp
```diff
@@ -136,7 +136,9 @@ void ngraph::traverse_functions(std::shared_ptr<ngraph::Function> p,
     }
 }
 
-void ngraph::replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> replacement)
+void ngraph::replace_node(std::shared_ptr<Node> target,
+                          std::shared_ptr<Node> replacement,
+                          std::shared_ptr<ProvenanceConfig> provenance_config)
```
rkimballn1: This function does not need a shared_ptr<ProvenanceConfig> since it only used for the lifetime of this function. Just pass in a `const ProvenanceConfig*`

src/ngraph/graph_util.hpp
```diff
@@ -72,7 +73,10 @@ namespace ngraph
     void traverse_functions(std::shared_ptr<Function> p,
                             std::function<void(std::shared_ptr<Function>)> f);
 
-    void replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> replacement);
+    void replace_node(
+        std::shared_ptr<Node> target,
+        std::shared_ptr<Node> replacement,
+        std::shared_ptr<ProvenanceConfig> provenance_config = std::make_shared<ProvenanceConfig>());
```
rkimballn1: Initialize provenance_config to nullptr. No need to create an object.

src/ngraph/provenance_config.cpp
```diff
@@ -0,0 +1,25 @@
...
+//*****************************************************************************
+
+#include "ngraph/provenance_config.hpp"
+#include <cstdlib>
+
+using namespace ngraph;
+
+bool ProvenanceConfig::is_enabled() const
+{
+    return std::getenv("NGRAPH_PROVENANCE_ENABLE");
```
rkimballn1: make a static variable inside this function and store
```
bool s_provenanace_enabled = std::getenv("NGRAPH_PROVENANCE_ENABLE") != nullptr;
```
then return s_provenance_enabled

src/ngraph/provenance_config.cpp
```diff
@@ -0,0 +1,25 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include "ngraph/provenance_config.hpp"
+#include <cstdlib>
```
rkimballn1: <> included go before ""

src/ngraph/provenance_config.hpp
```diff
@@ -0,0 +1,30 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#pragma once
+
+#include <stddef.h>
```
rkimballn1: Is this include needed? I'm not sure

test/provenance.cpp
```diff
@@ -0,0 +1,297 @@
...
+    //      \                    / \
+    //       G{tag_a,tag_b,tag_c}   F{}
+    //
+    // Comment:
+    //   * G is the replacement root, and its insertion kills A, B, and C.
+    //   * D is post-dominated by G, but E and F are not. Therefore D should take on the subsumed
+    //     tags, but E and F should not.
+    //
+
+    /*
```
rkimballn1: Why is this part commented out? A comment saying why would be good, or remove it.

adstraw: this is a test we intend to enable with a separate commit; it can be removed for this PR

src/ngraph/graph_util.cpp
aprocter: I don't think `ProvenanceConfig` needs to be any more complex than a `bool` at this point.

aprocter: ~I think I would rather query a global/static here than pass this is an as an argument. I don't see much utility in being able to override the global setting. Hopefully we will eventually have a fast enough implementation that this disable flag will not be needed anyway. :)~

I think I see the use for passing the `bool` as an argument in testing, actually. Hopefully it will be temporary.

src/ngraph/provenance_config.hpp
```diff
@@ -0,0 +1,30 @@
+//*****************************************************************************
```
aprocter: I think this class is probably not needed. Just a `bool` is enough.

test/provenance.cpp
```diff
@@ -0,0 +1,297 @@
...
+using namespace std;
+using namespace ngraph;
+using ::testing::Return;
+
+using ProvSet = std::unordered_set<std::string>;
+
+TEST(provenance, provenance)
+{
+    auto provenance_config = std::make_shared<MockProvenanceConfig>();
+    EXPECT_CALL(*provenance_config, is_enabled()).WillRepeatedly(Return(true));
```
aprocter: Instead of using mocks here, let's just pass `true` to our `replace_node` calls.

src/ngraph/graph_util.cpp
aprocter: Actually... let's make it a global/static with get/set functions, and have the unit tests get/set that global.

test/provenance.cpp
tachyon77: That's be definitely simpler. But I believe, @diyessi would have an objection to it. He is suggesting to use either pass_config.hpp or global getter-setter. I think we need a offline quick discussion.

aprocter: I'd actually prefer a global getter/setter as well. (Sorry, should have updated this comment.)

src/ngraph/graph_util.cpp
tachyon77: > Actually... let's make it a global/static with get/set functions, and have the unit tests get/set that global.

This sounds like the popular view at this point. I will go ahead with it.

test/provenance.cpp
diyessi: I'd prefer the pass one was global, so global is okay here.

test/provenance.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+using namespace std;
+using namespace ngraph;
+using ::testing::Return;
+
+using ProvSet = std::unordered_set<std::string>;
+
+TEST(provenance, provenance)
+{
+    set_provenance_enable(true);
```
aprocter: It's not likely to matter, but let's store the current value from `get_provenance_enable()` and restore it at the bottom of the test.

diyessi: Or RAII it.

src/ngraph/provenance.hpp
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+//*****************************************************************************
+
+#pragma once
+
+#include <cstdlib>
+
+namespace ngraph
+{
+    static bool provenance_enabled = std::getenv("NGRAPH_PROVENANCE_ENABLE") != nullptr;
```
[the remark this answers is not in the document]
tachyon77: Not sure how set_provenance_enabled() will have access to it then.

tachyon77:
TESTNOW

tachyon77:
'TESTNOW'

tachyon77:
TESTNOW

src/ngraph/provenance.hpp
rkimballn1: static should be `s_provenance_enabled`

PR #2752, opened 2019-04-12
Use only convolution direct algo for non float convs

nbpatel:


src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -1250,6 +1250,13 @@ namespace ngraph
                     Strides window_dilation_strides_adjusted;
 
                     mkldnn::algorithm convolution_algo = mkldnn_utils::get_conv_algo();
+
+                    if (node->get_input_element_type(0) != element::f32 &&
```
diyessi: This would also be true for element:f64, element::boolean, etc. Better to check for [u]int8 explicitly.

nbpatel: right. We only want it to choose between auto and direct for f32 convolutions.

jbobba: @diyessi only fp32, u8 and s8 end up here. Rest go to reference kernels.

PR #2753, opened 2019-04-12
helpers for function and node to determine the dynamic shape 

pruthviIntel:


src/ngraph/graph_util.cpp
```diff
@@ -642,3 +642,16 @@ std::vector<Output<Node>> ngraph::get_outputs_to(Node& src, Node& dst)
 
     return result;
 }
+
+bool ngraph::is_function_dynamic(const ngraph::Function* func)
```
louisfeng: Can we use `const shared_ptr<ngraph::Function>&`. This seems to be the standard in ngraph.

aprocter: Any reason not to just move the logic here into `Function::is_dynamic`?

pruthviIntel: @aprocter I thought we could reuse the code if we put in ngraph/graph_utils, but your are right, since is_dynamic will be attribute to function we can just move it to `Function::is_dynamic` :)

src/ngraph/graph_util.hpp
```diff
@@ -218,6 +218,14 @@ namespace ngraph
         return result_list;
     }
 
+    template <typename T>
+    bool is_dynamic(const T& dimensions)
```
aprocter: It really only makes sense to define this on `PartialShape`. A `Shape` is never dynamic, by definition.

src/ngraph/node.cpp
```diff
@@ -522,3 +522,11 @@ void Node::validate_and_infer_elementwise_logical()
 
     set_output_type(0, element::boolean, args_pshape);
 }
+
+// default implementation for the node to check if it contains partial shape
+// we will override this method, for the Op's which depends on additional shape
+// attribute to determine if node contains partial shape or not
+bool Node::is_dynamic()
+{
+    return this->get_output_partial_shape(0).is_dynamic();
```
aprocter: This should return true if any input or any output has dynamic shape.

src/ngraph/op/reshape.cpp
```diff
@@ -143,3 +144,13 @@ void op::Reshape::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
 
     adjoints.add_delta(get_argument(0), reshape);
 }
+
+bool op::Reshape::is_dynamic()
```
aprocter: Override shouldn't be necessary here, but we will need it on `DynReshape` and the like.

src/ngraph/function.cpp
```diff
@@ -224,3 +224,19 @@ void Function::set_placement(size_t placement)
 {
     m_placement = placement;
 }
+
+// TODO(pthoreho) this will be expensive, since we will be traversing all the nodes in
+// the graph, figure out if their is a way to cache the result and invalidate/update
+// the result if the function is modified
+bool Function::is_dynamic()
```
louisfeng: Could you please make this method `const`, i.e. `bool Function::is_dynamic() const`

src/ngraph/node.hpp
```diff
@@ -182,6 +182,7 @@ namespace ngraph
         virtual bool is_null() const { return false; }
         virtual bool is_op() const { return false; }
         virtual bool is_commutative() { return false; }
+        virtual bool is_dynamic();
```
louisfeng: This should also be `const`.

src/ngraph/function.cpp
```diff
@@ -224,3 +224,19 @@ void Function::set_placement(size_t placement)
 {
     m_placement = placement;
 }
+
+// TODO(pthoreho) this will be expensive, since we will be traversing all the nodes in
+// the graph, figure out if their is a way to cache the result and invalidate/update
+// the result if the function is modified
+bool Function::is_dynamic() const
```
diyessi: Let's call this `update_dynamic` and make it non const, so it can cache the result in `Function`. Let `is_dynamic` return the cached value. We can have the `Function` constructor call `update_dynamic` and we can call it from passes if we do things that might make a function non-dynamic through an optimization.

src/ngraph/node.cpp
```diff
@@ -522,3 +522,18 @@ void Node::validate_and_infer_elementwise_logical()
 
     set_output_type(0, element::boolean, args_pshape);
 }
+
+// default implementation for the node to check if it contains partial shape
```
diyessi: This could be a flag propagated during `validate_and_infer_shapes`

src/ngraph/function.cpp
louisfeng: @diyessi I think in most situations any function that takes a `ngraph::Function` object as a parameter can not make the assumption that `is_dynamic()` has the most up to date information and will need to always call `update_dynamic()` then call `is_dynamic()`. So even if a pass that changes a `Function` object state calls `update_dynamic`, another pass that take that resulting `Function` still have to call `update_dynamic()` before it can be certain that `is_dynamic()" returns correct information. 

Giving `Function` the responsibility to cache the value but does not control the logic when to update and keep that state consistent can be error prone. I would recommend keeping it simple right now with `is_dynamic()` always return the correct result and let the user determine when to cache the result.

pruthviIntel: @diyessi i also agree with @louisfeng , we thought about caching the `is_dynamic` result. but since function state can be changed outside the graph pass. We thought of keeping the API simple for now to keep us going.  this approach even though have cost associated with it but less error prone for now i believe.

src/ngraph/pattern/matcher.cpp
```diff
@@ -190,6 +190,16 @@ namespace ngraph
             add_node(graph_node);
             size_t watermark = m_matched_list.size() - 1;
 
+            // we will abort the match if any of graph_node or the pattern_node contains
+            // dynamic shape
+            if (m_strict_mode && graph_node->get_outputs().size() == 1)
+            {
+                if (pattern_node->is_dynamic() || graph_node->is_dynamic())
+                {
+                    return abort_match(watermark, false);
```
jbobba: shouldn't this be configurable instead of aborting on dynamic/partial shapes always?

louisfeng: Actually after looking at this, I am not sure we should do this check here. In my other PR, it added property to `Matcher` as well. So a matcher should specify whether it can handle dynamic shape or not and the GraphRewrite pass should do that check against the `Function`.

@pruthviIntel Do you think this makes sense?

louisfeng: To be more specific, the Matcher does not have problem with dynamic shapes (beside the strict mode, which checks shape). It's the *callback* may have problems with the dynamic shape.

pruthviIntel: @louisfeng should we remove this check? or query `get_property()` and check if it has `REQUIRE_STATIC_SHAPE` &&  `pattern_node->is_dynamic()`  and abort?

PR #2754, opened 2019-04-13
Add ConstantToBroadcast pass

rkimballn1:
This rewrites a Constant of some shape where all of the values are the same bit-exact value into a Constant scalar and a Broadcast to some shape.
I also added an option to reserialize that will run the ConstantToBroadcast pass on the graph it is reserializing. This converted the PP R50 graph from 182MB to 1MB in size.

Converts this:
pre_constant_to_broadcast

To this:
post_constant_to_broadcast


src/ngraph/pass/constant_to_broadcast.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "ngraph/op/constant.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+template <typename T>
+static bool is_data_constant(shared_ptr<op::Constant> constant)
+{
+    const T* data = constant->get_data_ptr<T>();
+    const T compare = data[0];
```
diyessi: Should check that `size > 0` first.

rkimballn1: you are correct sir. done

PR #2761, opened 2019-04-15
Fix install path on Windows

jfdong:
On windows when ngraph is built within onnx runtime, the following error is reported during ngraph install.

 ```
   -- Install configuration: "Release"
    CMake Error at cmake_install.cmake:63 (file):
      file INSTALL cannot find
      "C:/work/ngraph/release/private-onnxruntime/build/Windows/Release/ngraph/src/project_ngraph-build/src/ngraph/$(Configuration)/tbb.lib".
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Microsoft.CppCommon.targets(138,5): error MSB3073: The command "setlocal [C:\work\ngraph\release\private-onnxruntime\build

```

It is because cmake macro CMAKE_CFG_INTDIR is only valid in "build" process. Here is the cmake document ( https://cmake.org/cmake/help/v3.12/variable/CMAKE_CFG_INTDIR.html)
"For native build systems supporting multiple configurations in the build tree (such as Visual Studio Generators and Xcode), the value is a reference to a **_build-time_** variable specifying the name of the per-configuration output subdirectory. "


silee2:
@jfdong I don't think this is what we want for windows. CMAKE_BUILD_TYPE is empty for msbuild. Reason for using CMAKE_CFG_INTDIR was to have separate directories for different configurations.
Can you try the branch silee2/windows_install_cfg and see if it works? It uses the undocumented CMAKE_INSTALL_CONFIG_NAME for Windows. I don't have a windows machine other than a VM on my laptop so it will take to long for me to test.

jfdong:
@silee2 Thanks. I will try it within onnx runtime build.

jfdong:
@silee2 it works well on Windows. Could you please PR your branch. I will abandon this one.

PR #2768, opened 2019-04-17
[Dynamic Shape] Add Properties to PassBase

louisfeng:
This is part of refactoring the core passes to support dynamic shapes, we are adding properties to passes where each pass can specify which features it supports.

* Added a new utility class EnumMask to make working with enum properties easier.

As the first pass, I'm not removing any old code yet. Want to make sure everyone is OK with how we set and query the pass property.


src/ngraph/pass/pass.cpp
```diff
@@ -29,3 +33,19 @@ void pass::PassBase::set_state(ManagerState& state)
 {
     m_state = &state;
 }
+
+bool pass::PassBase::get_property(const PassPropertyMask& prop) const
+{
+    return m_property.is_set(prop);
+}
+void pass::PassBase::set_property(const PassPropertyMask& prop, bool value)
```
aprocter: Insert a blank line above

src/ngraph/pass/pass.hpp
```diff
@@ -45,6 +46,17 @@ namespace ngraph
             FOP_FUSIONS = 0x4,
             ALL_FUSIONS = 0xFFFFFFFF
         };
+        enum class PassProperty : uint32_t
+        {
+            //`DIFFERENTIABLE_FUSIONS` produce ops that support autodiff
```
diyessi: Let's leave things out until they are actually being used by something.
I am a little worried that using the enum is micro-optimizing and that sets would work just as well here.

louisfeng: If you are referring to `DIFFERENTIABLE_FUSIONS`, etc. sure I can remove them for now till we change the passes and merge with the `FusionType` above. 

In terms of enums, their usage here is not really for performance or optimization. I think for this use case (a set of true/false flags), it does fit the bill quite well. For anything more complex, we should use something different. Sure we could use a `set` or some other container for them. But bit mask works really well with enums. For example, if we want to look up a combination of flags (A|B|C), look up using a `set` is not very natural. Using enum is really just trying to keep things simple.

diyessi: I was referring to the unused ones; we can fill them in as we know what we actually need. I suspect we won't need those.

pruthviIntel:
LGTM :)

louisfeng:
> @louisfeng let me know if I my understanding is right we can set passmanger property when we explicitly register the pass.
> `pass_manager.register_pass<runtime::cpu::pass::CPUFusion>(pass::REGULAR_FUSIONS | REQUIRE_STATIC_SHAPE);` then inside inside `match_node()` of the matcher we can put additional check whether `is_dynamic()` violates `get_property()` and abort if necessary ?

The way I plan it to be used is each pass's constructor will set the property flags based on its features/requirements. Then in the PassManager during pass execution, it will check `Function`'s dynamic state and compare against each pass's property to make sure the right passes are ran.

src/ngraph/pass/pass.hpp
louisfeng: Fixed.

src/ngraph/pass/pass.cpp
louisfeng: Fixed.

pruthviIntel:
additionally, each pass can have  many function registered. if we refactor the passes like
```
 if (fusions & ngraph::pass::DIFFERENTIABLE_FUSIONS)
        {
            construct_conv_bias(); // DEPRECATED - Use CoreFusion
            construct_sigmoid_multiply();
        }

        if (fusions & ngraph::pass::REGULAR_FUSIONS)
        {
            construct_matmul();
            construct_matmulbias();
            construct_fprop_bn();
            .......
```
here based on the pass manager state , we should be good

louisfeng:
> additionally, each pass can have many function registered. if we refactor the passes like
> 
> ```
>  if (fusions & ngraph::pass::DIFFERENTIABLE_FUSIONS)
>         {
>             construct_conv_bias(); // DEPRECATED - Use CoreFusion
>             construct_sigmoid_multiply();
>         }
> 
>         if (fusions & ngraph::pass::REGULAR_FUSIONS)
>         {
>             construct_matmul();
>             construct_matmulbias();
>             construct_fprop_bn();
>             .......
> ```
> here based on the pass manager state , we should be good

Yes, I have added similar property to `Matcher` class. During construction of `Matcher` we also set these properties. We also need to add logic in the matcher `run_on_function()` method.

PR #2769, opened 2019-04-17
[ONNX] Remove unused initializer code in ValueInfo class

tsocha:
For all input initializers: initializer.name() == input.name().
this code:
https://github.com/NervanaSystems/ngraph/blob/tsocha/onnx-dead-code/src/ngraph/frontend/onnx_import/core/graph.cpp#L67-L77
already register all initializers.
Because of that removed code is unnecesary and probably dead.


postrational:
I have a feeling we may need this code in the near future.

PR #2775, opened 2019-04-17
Report error for C++ style comment ending with a backslash like

silee2:
// ... some content ... \
On Clang or AppleClang
PaddlePaddle sets -Werror=comment


silee2:
#2777 needs to be merged first.

PR #2779, opened 2019-04-18
[ONNX] Split ONNX tests into smaller files in test/onnx subdirectory

postrational:


test/onnx/onnx_import.in.cpp
```diff
@@ -0,0 +1,1412 @@
...
+
+    std::size_t size = function->get_output_size();
+    for (std::size_t i{0}; i < size; ++i)
+    {
+        std::shared_ptr<Node> node = function->get_output_op(i);
+        EXPECT_EQ(node->get_friendly_name(), "output_" + std::to_string(i + 1));
+    }
+}
+
+NGRAPH_TEST(onnx_${BACKEND_NAME}, model_add_abc)
```
tsocha: Why add_abc, binary_add_abc and add_abc_initializers tests are in the core section but addmul_abc is in operators section?

test/onnx/onnx_import.in.cpp
```diff
@@ -0,0 +1,1412 @@
...
+            {{{{2, 3, 4, 5, 6}, {2, 3, 4, 5, 6}, {2, 3, 4, 5, 6}, {2, 3, 4, 5, 6}},
+              {{2, 3, 4, 5, 6}, {2, 3, 4, 5, 6}, {2, 3, 4, 5, 6}, {2, 3, 4, 5, 6}},
+              {{2, 3, 4, 5, 6}, {2, 3, 4, 5, 6}, {2, 3, 4, 5, 6}, {2, 3, 4, 5, 6}}}})
+            .get_vector()};
+
+    Outputs outputs{execute(function, inputs, "${BACKEND_NAME}")};
+    EXPECT_TRUE(test::all_close_f(expected_output.front(), outputs.front()));
+}
+
+NGRAPH_TEST(onnx_${BACKEND_NAME}, model_reduce_log_sum)
```
tsocha: Maybe reduce tests should be in another file?

test/onnx/onnx_import.in.cpp
postrational: These tests were actually designed to test core functionality, so I grouped them there.

test/onnx/onnx_import.in.cpp
postrational: Agreed, but I think we can do this in the future.

PR #2789, opened 2019-04-19
[Tests] Add NgraphTestCase test runner class

postrational:


test/util/test_case.hpp
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            void run();
+
+        protected:
+            std::shared_ptr<Function> m_function;
+            std::unique_ptr<runtime::Backend> m_backend;
+            std::vector<std::shared_ptr<ngraph::runtime::Tensor>> m_input_tensors;
+            std::vector<std::shared_ptr<ngraph::runtime::Tensor>> m_result_tensors;
+            std::vector<std::shared_ptr<ngraph::op::Constant>> m_expected_outputs;
```
tsocha: Why expected outputs are stored in `ngraph` constant instead of `vector`?

test/util/test_case.cpp
```diff
@@ -0,0 +1,57 @@
...
+    auto function_results = m_function->get_results();
+    NGRAPH_CHECK(m_expected_outputs.size() == function_results.size(),
+                 "Expected number of outputs is different from the function's number of results.");
+
+    auto handle = m_backend->compile(m_function);
+    handle->call_with_validate(m_result_tensors, m_input_tensors);
+
+    for (int i = 0; i < m_expected_outputs.size(); ++i)
+    {
+        auto result_tensor = m_result_tensors.at(i);
```
tomdol: You can use `const auto&` instead here to avoid new instances of `shared_ptr`.

test/util/test_case.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+namespace ngraph
+{
+    namespace test
+    {
+        class NgraphTestCase
+        {
+        public:
+            NgraphTestCase(std::shared_ptr<Function> function, const std::string& backend_name)
+                : m_function(std::move(function))
```
tomdol: In general moving shared pointers is not a common practice. You can change the first parameter of this constructor to `std::shared_ptr&` and remove the `std::move` instead.

test/util/test_case.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+                m_input_tensors.push_back(tensor);
+
+                ++m_input_index;
+            }
+
+            template <typename T>
+            void add_multiple_inputs(const std::vector<std::vector<T>>& vector_of_values)
+            {
+                for (auto const& value : vector_of_values)
```
tomdol: East const -> west const? :)

test/util/test_case.hpp
```diff
@@ -0,0 +1,95 @@
...
+                             "All function results already have expected outputs.");
+
+                auto function_output_type = results.at(m_output_index)->get_element_type();
+                auto function_output_shape = results.at(m_output_index)->get_shape();
+                auto tensor = m_backend->create_tensor(function_output_type, function_output_shape);
+                m_result_tensors.push_back(tensor);
+
+                auto constant_node = std::make_shared<ngraph::op::Constant>(
+                    function_output_type, function_output_shape, values);
+                m_expected_outputs.push_back(constant_node);
```
tomdol: You can use `emplace_back` instead and pass the whole `make_shared` call from above directly as its argument.

PR #2793, opened 2019-04-21
Create backend as shared_ptr

rkimballn1:
The problem:
Shared library exported functions must be in C and must only return pointers to c++ classes, never the class itself. If you want to use an object like a shared_ptr as the return value of an export "C" function you are out of luck.
Some backends like Hybrid would like to keep a shared_ptr<Backend> for internal use, but because the current API for creating a backend only returns a pointer we are out of luck.

The Solution:
This PR creates a BackendConstructor c++ object and passes a pointer to a static instance of this object to the BackendManager where it can be used to construct shared_ptr<Backend> or any other c++ method.
The BackendConstructor class can also be used to add other c++ methods allowing simpler communication between ngraph and the backends.

Since all instances of the BackendConstructor class don't have any data members it is OK to make them static.

Bonus Effect: After this change, changes to how backends are created should result in compile-time errors instead of the current run-time errors.


src/ngraph/runtime/backend.hpp
```diff
@@ -44,9 +44,9 @@ class ngraph::runtime::Backend
     /// \brief Create a new Backend object
     /// \param type The name of a registered backend, such as "CPU" or "GPU".
     ///   To select a subdevice use "GPU:N" where s`N` is the subdevice number.
-    /// \returns unique_ptr to a new Backend or nullptr if the named backend
+    /// \returns shared_ptr to a new Backend or nullptr if the named backend
```
avijit-nervana: Why do you need to change this from unique_ptr to shared_ptr? If Hybrid needs a shared ptr internally - then it can get one from the unique_ptr. See this for example: https://stackoverflow.com/a/37885232

rkimballn1: No, you can't get a shared_ptr from a unique_ptr unless you own the unique_ptr, and the backend does not own this pointer.

diyessi: If you have hold of a `unique_ptr` you own it and can switch it to a `shared_ptr`. If you have a `unique_ptr&` then you don't own it, and a simpler approach would be to change the thing that gives you a reference to something that gives you a `unique_ptr` if it can, and, if it can't then things need to be `shared_ptr`. So, which of these scenarios are we in?

rkimballn1: There is a pointer that I want to shared between the backend and the frontend. In order to share the pointer it needs to be a shared_ptr. The backend used to return a raw pointer which was converted to a unique_ptr in ngraph, not in the backend.

avijit-nervana: What's backend and what's frontend? Can you please provide a specific example?
The history of this interface is - first it was returning shared_pointer - which resulted in various issues - so it was fixed by moving to unique_ptr. Now it's moving back to shared_ptr - so I wonder whether the same issues will be back?

avijit-nervana: Ok - based on some disussion - I now understand the reason. Looks good to me.

PR #2797, opened 2019-04-23
[Fused Ops] Add fused version of Elu

postrational:
Includes refactoring of ONNX importer `prelu` function to use fused `PReLu` op


src/ngraph/op/fused/elu.cpp
```diff
@@ -0,0 +1,60 @@
...
+        data->get_element_type(), Shape{}, std::vector<double>{0});
+    zero_node = ngraph::op::make_broadcast_node(zero_node, data->get_shape());
+
+    return {std::make_shared<ngraph::op::Maximum>(data, zero_node) +
+            alpha_node * std::make_shared<ngraph::op::Exp>(
+                             std::make_shared<ngraph::op::Minimum>(data, zero_node)) -
+            alpha_node};
+}
+
+shared_ptr<Node> op::Elu::copy_with_new_args(const NodeVector& new_args) const
```
tomdol: Missing `std::` namespace here and before the `make_shared` call.

src/ngraph/op/fused/elu.cpp
```diff
@@ -0,0 +1,60 @@
...
+NodeVector op::Elu::decompose_op() const
+{
+    auto data = get_argument(0);
+    auto alpha_node = get_argument(1);
+
+    alpha_node = ngraph::op::make_broadcast_node(alpha_node, data->get_shape());
+
+    std::shared_ptr<ngraph::Node> zero_node = std::make_shared<ngraph::op::Constant>(
+        data->get_element_type(), Shape{}, std::vector<double>{0});
+    zero_node = ngraph::op::make_broadcast_node(zero_node, data->get_shape());
```
tsocha: Please, use `make_constant_node` instead of making `Constant` first and then broadcast it.

src/ngraph/op/fused/elu.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "ngraph/op/add.hpp"
+#include "ngraph/op/constant.hpp"
+#include "ngraph/op/exp.hpp"
+#include "ngraph/op/maximum.hpp"
+#include "ngraph/op/minimum.hpp"
+#include "ngraph/op/multiply.hpp"
+#include "ngraph/op/subtract.hpp"
+#include "ngraph/op/util/broadcasting.hpp"
+
+using namespace std;
```
tomdol: I'm not sure about the convention but please remove the `using namespace std` statement and use `std::` prefix for all calls to standard library functions below.

src/ngraph/frontend/onnx_import/op/elu.cpp
```diff
@@ -46,18 +38,8 @@ namespace ngraph
...
-                        std::make_shared<ngraph::op::Constant>(
-                            data->get_element_type(), Shape{}, std::vector<double>{0});
-                    zero_node = ngraph::op::make_broadcast_node(zero_node, data->get_shape());
-
-                    return {std::make_shared<ngraph::op::Maximum>(data, zero_node) +
-                            alpha_node *
-                                std::make_shared<ngraph::op::Exp>(
-                                    std::make_shared<ngraph::op::Minimum>(data, zero_node)) -
-                            alpha_node};
+                    return {std::make_shared<ngraph::op::Elu>(data, alpha_node)};
```
diyessi: Please remove the `{` and `}'

src/ngraph/op/fused/elu.cpp
postrational: OK, will do, but this will require moving `make_constant_node` from ONNX utils to common utils.

src/ngraph/op/fused/elu.cpp
diyessi: `std` and `ngraph` are okay for `using namespace` but using `using namespace std` and `std::` in the same file isn't except in rare cases.

rkimballn1: The vast majority of cpp files in ngraph have `using namespace std`

src/ngraph/frontend/onnx_import/op/prelu.cpp
```diff
@@ -43,44 +33,9 @@ namespace ngraph
                 NodeVector prelu(const Node& node)
```
rkimballn1: In cpp files in ngraph we do not use nested namespaces like this, we change `NodeVector prelu(...)` to `NodeVector onnx_import::op::set1::prelu(...)`. This removes all of the 16 extra spaces of indenting.

src/ngraph/op/fused/elu.cpp
rkimballn1: We do not prepend std:: in cpp files that have `using namespace std;`

src/ngraph/frontend/onnx_import/op/elu.cpp
postrational: These are needed to return the `NodeVector` with a single node.

src/ngraph/frontend/onnx_import/op/prelu.cpp
postrational: We have this convention in the entire `onnx_import` frontend, so I don't want to change it one file at a time. At some point we may decide to change it globally.

src/ngraph/op/fused/elu.cpp
postrational: Done.

src/ngraph/op/util/broadcasting.hpp
```diff
@@ -125,5 +125,38 @@ namespace ngraph
...
+        /// \param[in]  data   The data to initialize node with.
+        ///
+        /// \tparam     T      Input data value type.
+        ///
+        /// \return     The nGraph node representing Constant data.
+        ///
+        template <typename T>
+        std::shared_ptr<ngraph::Node> make_constant_node(const ngraph::element::Type& type,
+                                                         const ngraph::Shape& shape,
+                                                         const std::vector<T>& data)
```
jbobba: Can you add it here instead? https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/builder/make_constant.hpp

src/ngraph/frontend/onnx_import/op/elu.cpp
diyessi: I think it'd be clearer in this case to do it as `return NodeVector{std::make_shared<ngraph::op::Elu>(data, alpha_node)};`

postrational: Done.

src/ngraph/op/util/broadcasting.hpp
postrational: I'll do you one better: we can remove `make_constant_node` all together and just use `make_constant` instead.

jbobba:
@postrational a few more things to complete this PR
- [ ] Add ops to fused_op_tbl (https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/fused_op_tbl.hpp) That will then require you to add support for these ops in the serializer as well.
- [ ] Add at least one test to https://github.com/NervanaSystems/ngraph/blob/master/test/backend_fused_op.in.cpp 
- [ ] Add at least one test to https://github.com/NervanaSystems/ngraph/blob/master/test/type_prop.cpp

PR #2798, opened 2019-04-23
[ONNX] Refactor quantized ops tests

postrational:


test/util/test_case.hpp
```diff
@@ -75,11 +90,34 @@ namespace ngraph
...
-                    function_output_type, function_output_shape, values));
+                    function_output_type, expected_shape, values));
 
                 ++m_output_index;
             }
 
+            template <typename T>
+            void add_expected_output(const std::vector<T>& values)
+            {
+                auto shape = m_function->get_results().at(m_output_index)->get_shape();
```
tsocha: What if output shape is broken?
Default shape as 1D tensor can be better in this case.
Now expected shape is connected with ngraph which can be invalid.

test/util/test_case.hpp
```diff
@@ -75,11 +90,34 @@ namespace ngraph
...
+            void add_expected_output_from_file(ngraph::Shape expected_shape,
+                                               const std::string& basepath,
+                                               const std::string& filename)
+            {
+                auto filepath = ngraph::file_util::path_join(basepath, filename);
+                add_expected_output_from_file<T>(expected_shape, filepath);
+            }
+
+            template <typename T>
+            void add_expected_output_from_file(ngraph::Shape expected_shape,
```
tsocha: Why there is no signature for expected output from file without passing expected_shape?
Maybe we should differ function by types not names?

postrational: The signature without shape is just a backward-compatible workaround. In general, we should be providing expected shapes.

test/util/test_case.hpp
postrational: We cannot assume that the output shape will be flat.

PR #2799, opened 2019-04-23
Number typed unit tests

rkimballn1:
Start for making unit tests so that they can be generated for a variety of data types. Specifically designed to add bfloat16 to testing but can be used for int32_t and friends too.
This also checks out-of-bounds numbers and I will be adding +-inf and nan to the input data.


src/ngraph/runtime/reference/sin.hpp
```diff
@@ -31,6 +33,7 @@ namespace ngraph
                 for (size_t i = 0; i < count; i++)
                 {
                     out[i] = std::sin(arg[i]);
+                    NGRAPH_INFO << "sin[" << arg[i] << "] = " << out[i];
```
aprocter: Remove this?

gcwenger: 👍

PR #2801, opened 2019-04-23
[ONNX] Extended types support for NgraphTestCase

tomdol:


test/util/test_case.cpp
```diff
@@ -21,6 +21,45 @@
...
+        const auto result = read_vector<T>(results);
+        return ngraph::test::all_close_f(expected, result);
+    }
+
+    template <typename T>
+    typename std::enable_if<std::is_integral<T>::value, ::testing::AssertionResult>::type
+        compare_values(const std::shared_ptr<ngraph::op::Constant>& expected_results,
+                       const std::shared_ptr<ngraph::runtime::Tensor>& results)
+    {
+        const auto expected = expected_results->get_vector<T>();
```
postrational: At this stage, we may need to have access to `NgraphTestCase` member variables. Let's move this logic into the class.

test/util/test_case.cpp
```diff
@@ -21,6 +21,45 @@
...
+        const auto result = read_vector<T>(results);
+        return ngraph::test::all_close(expected, result);
```
postrational: For integral types we need to use a direct `==` comparison instead of `all_close`.

tomdol: As discussed - leaving as is, otherwise we're loosing information about particular elements that differ between `expected` and `result` vectors.

tomdol:
TESTNOW

PR #2804, opened 2019-04-23
dynamic shape support for algebraic_simplification pass 

pruthviIntel:
- support for algebraic_simplification to apply transformations for graph with dynamic shape
- refactored pass_manager to apply graph optimization based on dynamic state
- add unit test case to check pass attributes of algebraic_simplification graph pass


src/ngraph/pass/algebraic_simplification.hpp
```diff
@@ -29,10 +30,12 @@ namespace ngraph
 class ngraph::pass::AlgebraicSimplification : public FunctionPass
 {
 public:
-    AlgebraicSimplification()
+    AlgebraicSimplification(pass::PassPropertyMask property =
```
louisfeng: The pass properties are intrinsic to the pass or matcher. So it should only be set by the pass or matcher itself not passed in from constructor. I also made `set_property()` protected so that it can't be called externally.

src/ngraph/pass/manager.cpp
```diff
@@ -98,6 +98,13 @@ void pass::Manager::run_passes(shared_ptr<Function> func, bool transitive)
         {
             for (shared_ptr<Function> f : fs)
             {
+                // This checks is to skip the graph optimization when the graph pass relies on static shape
+                // but the function state is dynamic
+                if (function_pass->get_property(pass::PassProperty::REQUIRE_STATIC_SHAPE) &&
+                    f->is_dynamic())
```
louisfeng: I was thinking if we can store a cached `bool` value with the `fs` so we don't need to call `f->is_dynamic()` frequently, and update the cached `bool` value like you did for the `GraphRewrite` matchers.

src/ngraph/pass/algebraic_simplification.hpp
louisfeng: Sorry, to correct what I said, the `Matcher` class is a bit different than the pass. The Matcher constructor does allow the property to be passed in during construction. The reason is the callback function needs to have these pass properties. I'm thinking maybe it's better to make the pass property a parameter of the callback, I'll fix that in another PR.

src/ngraph/pass/algebraic_simplification.hpp
```diff
@@ -32,7 +33,9 @@ class ngraph::pass::AlgebraicSimplification : public FunctionPass
     AlgebraicSimplification()
         : FunctionPass()
     {
+        auto property = pass::PassPropertyMask(pass::PassProperty::REGULAR_FUSIONS) |
```
louisfeng: There's a list initializer constructor for EnumMask, so you could also do it this way: `pass::PassPropertyMask property{pass::PassProperty::REGULAR_FUSIONS, pass::PassProperty::REQUIRE_STATIC_SHAPE}`

src/ngraph/pass/manager.cpp
```diff
@@ -61,17 +61,26 @@ void pass::Manager::run_passes(shared_ptr<Function> func, bool transitive)
 {
     bool profile_enabled = getenv("NGRAPH_PROFILE_PASS_ENABLE") != nullptr;
 
-    vector<shared_ptr<Function>> fs;
+    vector<std::pair<shared_ptr<Function>, bool>> fs;
```
louisfeng: Sorry I missed the module pass in our discussion so I see why you need `f_array` for the module passes. But `fs` and `f_array` are sort of duplicate. Maybe it makes sense to keep the `bool` flags in a separate array. Do you think that works better?

PR #2806, opened 2019-04-23
Create mkldnn primitives at first iteration for codegen - part1.

ayzhuang:
1. Use a map of node and a tuple of string that is to be emitted to the generated file to build mkldnn primitives at first iteration, index, and deps.

2. part1 implements ops: add, batchnorm, concat, lrn, and lstm/rnn.

3. example of generated file without the header files:
[example_of_generated_file.txt](https://github.com/NervanaSystems/ngraph/files/3123159/example_of_generated_file.txt)


src/ngraph/runtime/cpu/cpu_external_function.hpp
```diff
@@ -125,6 +125,18 @@ namespace ngraph
                     return it->second;
                 }
 
+                // Return the tuple including the string to create mkldnn primitive, the deps and the index in CODEGEN
+                std::tuple<std::string, std::vector<size_t>, size_t>
+                    get_primitive_build_tuple(const Node* node) const
```
dcaballe: I'm a bit concerned about returning a copy of this relatively big tuple. Would it make sense to return a const reference instead and change the container to an `std::map` so that address is not invalidated? I'm not sure which approach would be better. I leave it up to you :)

src/ngraph/runtime/cpu/cpu_external_function.hpp
```diff
@@ -310,6 +322,12 @@ namespace ngraph
 
                 /// Map each node with mkldnn implementation to its mkldnn primitive index.
                 std::unordered_map<const Node*, size_t> m_node_primitive_idx_map;
+                /// Map each node with mkldnn implementation to its mkldnn primitive creating string and the deps.
+                std::unordered_map<const Node*,
+                                   std::tuple<std::string, std::vector<size_t>, size_t>>
+                    m_node_primitive_string_deps_index_map;
+                /// Name of the file to store descriptors for mkldnn_primitives
+                std::string m_desc_filename = "desc_file";
```
dcaballe: const?

src/ngraph/runtime/cpu/pregenerated_src/cpu_cg_runtime_context.hpp
```diff
@@ -70,4 +96,26 @@ extern "C" void destroy_cg_ctx(CPURuntimeContextCG* cg_ctx)
 {
     delete cg_ctx;
 }
+
+extern "C" void set_memory_ptr_cg(CPURuntimeContextCG* cg_ctx,
```
jbobba: dont think we need extern "C" on these functions since they are only going to get called from within the generated code.

src/ngraph/runtime/cpu/pregenerated_src/cpu_cg_runtime_context.hpp
```diff
@@ -70,4 +96,26 @@ extern "C" void destroy_cg_ctx(CPURuntimeContextCG* cg_ctx)
...
+                               size_t primitive_index,
+                               void* ptr)
+{
```
jbobba: CPURuntimeContextCG->set_memory_ptr() might be cleaner

src/ngraph/runtime/cpu/pass/cpu_mkldnn_primitive_build.cpp
```diff
@@ -798,6 +1419,50 @@ bool MKLDNNPrimitiveBuildPass::run_on_call_graph(const std::list<std::shared_ptr
...
+    for (const auto& shp_node : nodes)
+    {
+        if (!in_new_map(shp_node))
+        {
+            continue;
+        }
+
+        Node* node = shp_node.get();
+
+        // skip in place concat, relu, reshape, slice.
```
jbobba: shouldn't in-place elimination happen in cpu_emitter? E.g., https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/cpu_emitter.cpp#L1404

jbobba: we cant also skip relu. you can only skip non-destructive in-place.

ayzhuang: If we don't skip here, we write descriptors to the desc_file but don't read from the file for these nodes. The node after such node would read the wrong descriptors and get segfault.

src/ngraph/runtime/cpu/cpu_external_function.hpp
dcaballe: Thanks. You will have to replace `unordered_map` with `map` for `m_node_primitive_string_deps_index_map` since the address to the elements of an `unordered_map` might be invalidated.

PR #2808, opened 2019-04-24
[ONNX] Handle one element vector as scalars in quant ops.

tsocha:
This is workaround for situation when quant operators are expecting scalars but got a one element vector instead.
We can handle this situation by reshaping data from [1] to [] shape.


src/ngraph/frontend/onnx_import/op/quant_conv.cpp
```diff
@@ -166,8 +167,43 @@ namespace ngraph
                     int64_t groups{node.get_attribute_value<int64_t>("group", 1)};
 
                     auto data_scale = inputs.at(1);
```
postrational: Let's leave the list of inputs on top of function please, followed by reading attributes.

src/ngraph/frontend/onnx_import/op/dequantize_linear.cpp
```diff
@@ -81,6 +78,19 @@ namespace ngraph
                                                                            x->get_element_type());
                     }
 
+                    Shape zero_point_shape = zero_point->get_shape();
+                    if (zero_point_shape.size() == 1 && zero_point_shape[0] == 1)
```
postrational: We are repeating this block of code many times. Let's make a small helper function called `reshape_to_scalar_if_possible` and just call that many times.

postrational: You could even try to pass it a `NodeVector` to reduce the number of calls.

src/ngraph/frontend/onnx_import/op/quant_conv.cpp
```diff
@@ -160,15 +160,15 @@ namespace ngraph
 
                 NodeVector quant_conv(const Node& node)
                 {
-                    const NodeVector& inputs = reshape::interpret_as_scalar(node.get_ng_inputs());
+                    NodeVector inputs{node.get_ng_inputs()};
```
tsocha: Please make it `const`.

src/ngraph/frontend/onnx_import/op/quantize_linear.cpp
```diff
@@ -36,10 +36,12 @@ namespace ngraph
             {
                 NodeVector quantize_linear(const Node& node)
                 {
-                    NodeVector inputs{reshape::interpret_as_scalar(node.get_ng_inputs())};
+                    NodeVector inputs{node.get_ng_inputs()};
```
tsocha: Please make it `const`.

postrational:
Closing this for now. We'll reopen if needed.

PR #2814, opened 2019-04-24
Fix broken test in shape specialization pass

aprocter:
(Note that this pass is probably going away soon anyway in favor of something that does the job via constant folding, but I figure I might as well fix this while I'm looking at it.)


aprocter:
Removing @jbobba since he's out sick. Should be a low-risk change anyway (this pass isn't used yet).

PR #2815, opened 2019-04-24
add safety checks for mkldnn-assignement for quantized inner product

pruthviIntel:
- this PR address the MKLDNN exception due to u8u8 inner product which is not supported in MKLDNN


jbobba:
Can you also make the builder/emitter assert if the op is not supported by mkldnn

PR #2818, opened 2019-04-25
Update documentation section on graph visualization, fix ngraph bridge link, etc.

indie:


doc/sphinx/source/inspection/index.rst
```diff
@@ -0,0 +1,30 @@
+.. inspection/index: 
+
+Visualization Tools
+###################
+
+One option for visualizing the graph or node structure of a graph is to compile 
+with the :envvar:`NGRAPH_ENABLE_VISUALIZE_TRACING` flag. This generates a PDF,
```
rkimballn1: Not `NGRAPH_ENABLE_VISUALIZE_TRACING`. The flags `NGRAPH_ENABLE_*_TRACING` are for developers only because they generate intermediate graphs of the state between graph passes. These flags do not show the original source graph. We need to add a new flag to visualize the source graph. Twice in the last week I have gotten bug reports because users used these flags. They are evil. (the flags, not the users)
The output is png by default but can be controlled by setting `NGRAPH_VISUALIZE_TREE_OUTPUT_FORMAT`

doc/sphinx/source/inspection/index.rst
```diff
@@ -0,0 +1,30 @@
+.. inspection/index: 
+
+Visualization Tools
+###################
+
+One option for visualizing the graph or node structure of a graph is to compile 
+with the :envvar:`NGRAPH_ENABLE_VISUALIZE_TRACING` flag. This generates a PDF, 
+though it is not ideal for viewing large graphs.  
+
+To inspect larger graphs, it may be necessary to convert ``.json`` files
```
rkimballn1: When visualization is enabled a `dot` file is generated along with the png file. This dot file can be imported into these tools.

indie: PR #2821 has some changes that affect how this works, so I tried to add those changes (since they've already been merged).  

> We need to add a new flag to visualize the source graph. 

When it comes to explaining that primitive formats (`.pdf` / `.json` ) are terrible for viewing large graphs, it's probably better to not get into anything too specific until this gets more design work behind it.

doc/sphinx/source/inspection/index.rst
```diff
@@ -0,0 +1,37 @@
+.. inspection/index: 
+
+Visualization Tools
+###################
+
+One option for visualizing the graph or node structure of a graph is to compile
```
jbobba: NGRAPH_JSON_ENABLE will be enabled by default and available in our default wheels. So lets not mention it here. Instead, we could talk about how ngraph provides serialization and deserialization facilities along with the ability to create image formats.

contrib/tools/graphml/ngraph_json_to_graphml.py
```diff
@@ -0,0 +1,204 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+# Simple converter to convert nGraph JSON format to GraphML.
+#  
+# Original Author: Vasanth Tovinkere
+# Date: 2/5/2018
+#
```
diyessi: This is more than a year old and is working with a reverse-engineered undocumented internal json serialization format  which has changed since then. There are no tests for this program.

doc/sphinx/source/inspection/index.rst
```diff
@@ -0,0 +1,35 @@
...
+Visualization Tools
+###################
+
+nGraph provides serialization and deserialization facilities along with the 
+ability to create image formats. When visualization is enabled, a ``dot`` file 
+is generated, along with a ``png``. The default can be adjusted by setting the 
+``NGRAPH_VISUALIZE_TREE_OUTPUT_FORMAT`` flag to another format, like PDF. It
+can then be inspected with a third-party visualization tool such as:
+
+#. `Netron`_
```
aprocter: Netron is not supported, at least not yet, and it wouldn't be operating on the `dot` or `png`/`pdf` files.

doc/sphinx/source/inspection/index.rst
```diff
@@ -0,0 +1,35 @@
...
+
+#. `TFBoard`_
```
aprocter: TensorBoard (not TFBoard) cannot be used to visualize nGraph graphs directly.

doc/sphinx/source/inspection/index.rst
indie: I commented this out for now with the note that it is tentatively planned.

doc/sphinx/source/inspection/index.rst
indie: Done

doc/sphinx/source/inspection/index.rst
```diff
@@ -9,9 +9,9 @@ is generated, along with a ``png``. The default can be adjusted by setting the
 ``NGRAPH_VISUALIZE_TREE_OUTPUT_FORMAT`` flag to another format, like PDF. It
 can then be inspected with a third-party visualization tool such as:
 
-#. `Netron`_
+.. #. `Netron`_ support tentatively planned to come soon
 
-#. `TFBoard`_
+#. `TensorBoard`_
```
aprocter: I think we should remove this reference altogether. It cannot be used to visualize the graphs you speak about in the preceding paragraph.

indie: I see the word "cannot" there now.  Thanks.

diyessi: Agree

PR #2822, opened 2019-04-25
Build option to disable fast math

jbobba:


CMakeLists.txt
```diff
@@ -166,6 +166,7 @@ option(NGRAPH_LIB_VERSIONING_ENABLE "Enable shared library versioning" FALSE)
 option(NGRAPH_PYTHON_BUILD_ENABLE "Enable build nGraph python package wheel" FALSE)
 option(NGRAPH_PLAIDML_ENABLE "Enable the PlaidML backend" ${PLAIDML_FOUND})
 option(NGRAPH_DISTRIBUTED_ENABLE "Enable distributed training using MLSL/OpenMPI" OFF)
+option(NGRAPH_DISABLE_FAST_MATH "Disable fast math" OFF)
```
rkimballn1: How about `NGRAPH_FAST_MATH_ENABLE` with the default of ON. That way it follows the naming convention of the other flags.

baojun-nervana: would this say we are currently using eigen fast math already?

src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -165,6 +165,10 @@ if (NGRAPH_CPU_ENABLE)
         add_dependencies(cpu_backend ext_halide)
     endif()
 
+    if (NGRAPH_DISABLE_FAST_MATH)
+        target_compile_definitions(cpu_backend PRIVATE EIGEN_FAST_MATH=0)
```
baojun-nervana: I tried to set EIGEN_FAST_MATH=1. It doesn't change anything on the result somehow.

CMakeLists.txt
jbobba: yes, we are enabling by default.

PR #2823, opened 2019-04-26
Support constant folding for reshape and broadcast bf16 data

jbobba:
bf16 kernels for these ops will be identical to int16_t


src/ngraph/pass/constant_folding.cpp
```diff
@@ -234,6 +234,12 @@ void pass::ConstantFolding::construct_constant_reshape()
                          fold_constant_reshape<double>(constant_match, reshape_match, func));
             return true;
         }
+        else if (type == element::bf16)
+        {
+            replace_node(m.get_match_root(),
+                         fold_constant_reshape<int16_t>(constant_match, reshape_match, func));
```
aprocter: Should be okay since internally this is just moving 16-bit-whatevers around, but does it work to supply `<ngraph::bfloat16>` instead?

jbobba: wasn't aware of `<ngraph::bfloat16>`. I can take a closer look.

rkimballn1: I don't think it makes any difference but it would certainly read better if you used ngraph::bfloat16 here

PR #2825, opened 2019-04-26
[ONNX]  Add interpret_as_scalar helper

postrational:


src/ngraph/frontend/onnx_import/utils/reshape.cpp
```diff
@@ -266,6 +266,22 @@ namespace ngraph
                 return split(node, length_parts, axis_to_split);
             }
 
+            std::shared_ptr<ngraph::Node>
+                interpret_as_scalar(const std::shared_ptr<ngraph::Node>& node)
+            {
+                Shape node_shape = node->get_shape();
+                NGRAPH_CHECK((shape_size(node_shape) == 1),
```
tsocha: what will happen if `node_shape` will be `[]`?
This check will probably raise exception but `[]` shape is correct in this context.

src/ngraph/frontend/onnx_import/utils/reshape.cpp
```diff
@@ -266,6 +266,22 @@ namespace ngraph
...
+                             "Scalar value can't be derived from a node with ",
```
arogowie-intel: "(...) node with shape:" Will be more clear error message.

postrational: `node_shape` resolves to `Shape{1,2}`, so `node with shape:` will generate `node with shape: Shape{1,2}`

PR #2826, opened 2019-04-26
Change CPU backend to use PassConfig for codegen enable

rkimballn1:
PassConfig had a CPU specific constructor to pass CPU specific information around PassConfig. Changed to use PassConfig attributes to pass this information. Removed CPU specific code from ngraph::runtime namespace.
Fixed PassConfig to handle being passed as const.


src/ngraph/pass/manager.cpp
```diff
@@ -37,12 +37,12 @@ using namespace ngraph;
 
 pass::Manager::Manager()
 {
-    static const auto nevt = std::getenv("NGRAPH_ENABLE_VISUALIZE_TRACING");
+    static const auto nevt = std::getenv("NGRAPH_ENABLE_DEBUG_VISUALIZE_TRACING");
```
jbobba: why change the name here? Can we also follow a deprecation path here since they are bridges/scripts using this and docs that will need to be changed.

rkimballn1: reverted

PR #2827, opened 2019-04-26
[ONNX] Refactor LSTM tests to use NgraphTestCase

postrational:


test/util/test_case.hpp
```diff
@@ -38,6 +38,7 @@ namespace ngraph
             {
             }
 
+            void set_tolerance(int tolerance_bits) { m_tolerance_bits = tolerance_bits; }
```
arogowie-intel: I'm wondering why not to move this _tolerance bits_ value to `run` function as its input parameter?

postrational: The `run` function would end up with a lot of different signatures, as different comparator functions will have different parameters, all of them optional. Better to configure class and leave run without parameters.

PR #2828, opened 2019-04-26
Refactor distributed to be isolated to a few files.

diyessi:
This is the first step of changing CPU distributed support from something compiled into core into something supplied by the framework/bridge. This PR factors the distributed OP API implementation into an interface with one non-implementation and one each of OpenMPI and MLSL. Other implementations can be easily added now with no changes to ngraph.

It is expected that in the future the frameworks will set the distributed interface implementation to something appropriate for their environment; for backwards-compatibility, `distributed.cpp` currently does this according to build flags.


src/ngraph/distributed.cpp
```diff
@@ -14,90 +14,35 @@
...
-#ifdef NGRAPH_DISTRIBUTED_MLSL_ENABLE
-#include <mlsl.hpp>
-#elif NGRAPH_DISTRIBUTED_OMPI_ENABLE
-#include <mpi.h>
-#endif
-
 #include "ngraph/distributed.hpp"
+#include "ngraph/distributed/mlsl.hpp"
+#include "ngraph/distributed/null.hpp"
+#include "ngraph/distributed/open_mpi.hpp"
```
jianyinglang: One problem is that we cannot build the code with mlsl and mpi simultaneously because mlsl and mpi will use the same mpirun command which will cause runtime error for linking to two libraries.

diyessi: Both headers are conditionalized on mlsl/openmpi, only one of which can be enabled in the build configuration.

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -28,6 +28,7 @@
 #include "ngraph/op/add.hpp"
 #include "ngraph/op/all.hpp"
 #include "ngraph/op/allreduce.hpp"
+#include "ngraph/op/allreduce.hpp"
```
wenzhe-nrv: duplication with the line above?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -44,6 +44,7 @@
 #include "ngraph/op/add.hpp"
 #include "ngraph/op/all.hpp"
 #include "ngraph/op/allreduce.hpp"
+#include "ngraph/op/allreduce.hpp"
```
wenzhe-nrv: same duplication

src/ngraph/runtime/generic_cpu/gcpu_executable.hpp
```diff
@@ -441,11 +437,9 @@ class ngraph::runtime::gcpu::GCPUExecutable : public Executable
...
             break;
         }
-        case OP_TYPEID::BroadcastDistributed: {
-#ifdef NGRAPH_DISTRIBUTED_ENABLE
-            Distributed dist;
-            int Rank_ID;
-            Rank_ID = dist.get_rank();
+        case OP_TYPEID::BroadcastDistributed:
+        {
+            int rank_ID = get_distributed_interface()->get_rank();
```
wenzhe-nrv: `Rank_ID` is called below. I think the var names need to be consistent.

src/ngraph/distributed.cpp
```diff
@@ -14,90 +14,35 @@
...
-int ngraph::Distributed::get_size() const
-{
-#ifdef NGRAPH_DISTRIBUTED_MLSL_ENABLE
-    return static_cast<int>(MLSL::Environment::GetEnv().GetProcessCount());
-#elif NGRAPH_DISTRIBUTED_OMPI_ENABLE
-    int size;
-    MPI_Comm_size(MPI_COMM_WORLD, &size);
-    return size;
+#ifdef NGRAPH_DISTRIBUTED_OMPI_ENABLE
+        set_distributed_interface(std::unique_ptr<DistributedInterface>(
```
wenzhe-nrv: could you please put some log here and below to show when this is set?

src/ngraph/runtime/cpu/cpu_emitter.cpp
diyessi: Got it.

src/ngraph/runtime/cpu/cpu_external_function.cpp
diyessi: Got it again.

src/ngraph/runtime/generic_cpu/gcpu_executable.hpp
diyessi: Got it.

src/ngraph/distributed.cpp
diyessi: Added a settable name to the interfaces and logged setting of the interface.

src/ngraph/distributed/open_mpi.hpp
```diff
@@ -0,0 +1,114 @@
...
+                else if (element_type != element::Type_t::f32)
+                {
+                    throw std::runtime_error(
+                        "BroadcastDistributed op supports only f32 and f64 types");
+                }
+                MPI_Bcast(in, count, data_type, 0, MPI_COMM_WORLD);
+            }
+
+        protected:
+            std::string m_name;
```
jianyinglang: Should it string have name `"OpenMPI"`?

PR #2832, opened 2019-04-27
Free nodes in a non-recursive way so we don't blow out the stack for large graphs

rkimballn1:


src/ngraph/function.cpp
```diff
@@ -81,6 +81,25 @@ void Function::validate_nodes_and_infer_types()
...
+    // Nodes are linked in order from Results to Parameters or backwards through the graph.
+    // If you delete the nodes by simply releasing the results then the nodes recursively
+    // traverse the tree freeing each node's inputs. This will exceed a process's stack allocation
+    // for a sufficiently large graph.
+    // This deleter makes a list of nodes which can then be traversed in forward order to
+    // delete nodes. The only memory cost is the size of the list, but this is allocated on the
+    // heap.
+    std::list<Node*> list;
+
+    traverse_nodes(this, [&](shared_ptr<Node> n) { list.push_front(n.get()); });
```
aprocter: I think we should include the control deps here.

src/ngraph/graph_util.hpp
```diff
@@ -49,7 +49,7 @@ namespace ngraph
 
     void traverse_nodes(const Function* p,
                         std::function<void(std::shared_ptr<Node>)> f,
-                        bool include_control_deps);
+                        bool include_control_deps = false);
```
aprocter: If we're going to put a default here, I think `true` is more appropriate.

src/ngraph/function.cpp
diyessi: This only works if the last reference to the nodes is from the function.
Suppose we had a `Node::free_children(set<shared_ptr<Node>> nodes)` that adds each child node (inputs and control deps) to nodes and then releases them from the inputs. Then it goes through nodes. If `use_count` is 1, it calls free_children on the node, and it removes node from nodes no matter what `node_count` was. When nodes is empty, it returns. Make ~Node call free_children on itself with an empty set. I think that will take care of all the delete cases.

rkimballn1:
@diyessi Your description is exactly how this function works. It does what you want it to do except for the controldeps which were not in @aprocter's unit test.

src/ngraph/function.cpp
rkimballn1: @diyessi If by `works` you mean actually frees the nodes then you are correct. Since the nodes are shared_ptr we can't actually delete them as they may be used by others. This is by design. What we can do is unlink them so that each node is now standalone with no references to other nodes. At some later time when the last user of each node releases the last shared_ptr then the node will correctly be deleted. The Function does not own the nodes, it shares them with other.

src/ngraph/graph_util.hpp
rkimballn1: Reverted change

diyessi:
#2987 does this kind of deletion from `~Node` so it doesn't have problems with dangling non-function references to nodes (which occurred in some unit tests).

PR #2844, opened 2019-04-30
Dynamic-shape backend API change, with prototype implementation

aprocter:
Two main changes:

1. Update to the backend API to support dynamic shapes
   1. New `create_dynamic_tensor` API to create a tensor whose shape can change from iteration to iteration.
   2. Optional boolean parameter to `Backend::create` to require that the returned backend support `create_dynamic_tensor`. Default is `false`. If this is set to `true` and the backend does not natively support this---so far, none of them do---the requested backend will be wrapped (see below).
   3. `runtime::Tensor::get_shape()` and `runtime::Tensor::get_element_type()` are now `virtual`, because they have a slightly different meaning where dynamic tensors are concerned.
      1. On dynamic tensors, `get_shape()` can only be called if the tensor has actually been assigned a value with concrete shape. (In other words, after execution has written to output tensor.)
      2. `get_partial_shape()` is always allowed and will return the shape scheme specified at tensor creation time.
      3. For example, `get_partial_shape()` could return `{2,?,3}` while `get_shape()` returns `{2,3,3}`.

2. Backend wrapper for graph cloning-based support of dynamic shapes in backends that haven't yet been retrofitted to support them.
   1. This does _not_ yet cache the clones.
   2. This does _not_ yet support replacement of `Dyn*` ops.
   3. This _does_ support specialization of parameter shapes (see, e.g., the `abc` test).
   4. This does _not_ yet support instantiation of shape-relevant parameters (e.g., cases where the `output_shape` input of a `DynReshape` is being fed by an `int64`-typed vector `Parameter`).

Note that **none of these new features is activated** without the caller (bridge) explicitly requesting it. I would like to go ahead and merge this so I can start playing around with TF bridge integration.

EDIT: Note also that the wrapper is just for prototyping purposes, since I didn't want to go poking into each backend's internals just yet. I anticipate it will eventually go away.


src/ngraph/runtime/dynamic_wrapper/dynamic_wrapper_backend.cpp
```diff
@@ -0,0 +1,277 @@
...
+        else
+        {
+            NGRAPH_CHECK(false,
+                         "Internal error: Tensor is neither a WrappedStaticTensor nor a "
+                         "WrappedDynamicTensor");
+        }
+    }
+
+    // TODO: specialize_shapes needs to fill in values of shape-relevant params.
+    auto clone = specialize_shapes(m_wrapped_function, arg_element_types, arg_shapes);
```
nmostafa: Can this be invoked through pass manager ?

src/ngraph/runtime/dynamic_wrapper/dynamic_wrapper_backend.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace dynamic_wrapper
+        {
+            class DynamicWrapperBackend;
+            class WrappedExecutable;
+            class WrappedStaticTensor;
```
nmostafa: Not sure why we need a static tensor wrapper.

src/ngraph/runtime/dynamic_wrapper/dynamic_wrapper_backend.hpp
```diff
@@ -0,0 +1,112 @@
...
+            class WrappedDynamicTensor;
+        }
+    }
+}
+
+class ngraph::runtime::dynamic_wrapper::DynamicWrapperBackend : public Backend
```
nmostafa: Can you add some documentation about the purpose of each wrapper class ? For example, WrappedExecutable enables late compilation, code-caching, .. etc. 
I also not so fond of the class names. Maybe something like DynamicBackend, DynamicExecutable would be better.

src/ngraph/runtime/dynamic_wrapper/dynamic_wrapper_backend.cpp
aprocter: Should be doable. Let me poke on it and see how big a change it will be---I might want to save it for a later PR.

src/ngraph/runtime/dynamic_wrapper/dynamic_wrapper_backend.hpp
aprocter: I think you're right---my reason for adding this wound up being invalidated somewhere along the line. Let me take it out and see if anything breaks. :)

src/ngraph/runtime/dynamic_wrapper/dynamic_wrapper_backend.hpp
aprocter: Will add some documentation.

For the class names, I kind of like the idea of having `Wrapper` somewhere in the name because it emphasizes that dynamic shapes are not being implemented "directly". It doesn't overpromise, if you will. :) Let me think about it.

src/ngraph/runtime/dynamic_wrapper/dynamic_wrapper_backend.cpp
aprocter: @nmostafa I think I will actually punt on this for now, since it will need to wait until after #2875 goes in.

src/ngraph/runtime/dynamic_wrapper/dynamic_wrapper_backend.hpp
aprocter: Decided to go with `DynamicBackend`, `DynamicExecutable`, `DynamicTensor`.

PR #2845, opened 2019-05-01
Cyphers/fp16

diyessi:
Bare-bones fp16 support.


src/ngraph/CMakeLists.txt
```diff
@@ -406,6 +406,9 @@ set (SRC
     strides.cpp
     strides.hpp
     type/bfloat16.cpp
+    type/bfloat16.hpp
+    type/float16.cpp
+    type/float16.cpp
```
rkimballn1: Cut-and-paste error. I suspect you wanted to add float16.`hpp` here

diyessi: Fixed.

PR #2847, opened 2019-05-01
Klocwork issues.

diyessi:


src/ngraph/pass/core_fusion.cpp
```diff
@@ -767,10 +767,10 @@ void ngraph::pass::CoreFusion::construct_conv_bias()
                      << m.get_match_root()->get_name();
         auto pattern_map = m.get_pattern_map();
 
+        // Pattern matcher ensures these are the right classes
         auto conv_m =
-            std::dynamic_pointer_cast<ngraph::op::Convolution>(m.get_match_root()->get_argument(0));
-        auto bcast_m =
-            std::dynamic_pointer_cast<op::Broadcast>(m.get_match_root()->get_argument(1));
+            std::static_pointer_cast<ngraph::op::Convolution>(m.get_match_root()->get_argument(0));
```
jbobba: Since match_root() is a commutative op, we cannot assume that get_argument(0) is convolution. Why is KW complaining here?

diyessi: Pointer could be 0.

jbobba: Change to std::dynamic_pointer_cast

src/ngraph/pass/core_fusion.cpp
```diff
@@ -767,10 +767,10 @@ void ngraph::pass::CoreFusion::construct_conv_bias()
...
+        auto bcast_m = std::static_pointer_cast<op::Broadcast>(m.get_match_root()->get_argument(1));
```
jbobba: Change to std::dynamic_pointer_cast

PR #2851, opened 2019-05-01
mklml on linux has no soname set. This will cause linking against the…

silee2:
… full path for mklml when using prebuilt mkldnn and mklml.


silee2:
Should fix PaddlePaddle's wheel issue.

PR #2856, opened 2019-05-02
Fix convbias and 1dconv for CODEGEN.

ayzhuang:


src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -2495,8 +2495,6 @@ namespace ngraph
                            << args[1].get_name() << ");\n";
                     writer << "cg_ctx->set_memory_ptr(" << to_string(deps[2]) << ", "
                            << args[2].get_name() << ");\n";
```
nbpatel: out[0].get_name()

PR #2857, opened 2019-05-03
Generic Dot & builders for QLinearMatmul 

nbpatel:
This PR:
1.  Adds builders for QLinearMatmul op's in ONNX
2. Adds generic dot implementation (to support u8u8 matmul)
3. Supports non-zero zero point for QLinearMatmul by falling back to fp32 Dot

Note:
The builders will require changes in the ngraph-onnx bridge as to how they call into ngraph.


src/ngraph/builder/quantization/quantized_linear_matmul.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include "ngraph/coordinate_diff.hpp"
+#include "ngraph/node.hpp"
+
+namespace ngraph
+{
+    namespace builder
+    {
+        namespace quantization
+        {
+            std::shared_ptr<Node> QuantizedLinearMatmul(std::shared_ptr<Node> input,
```
adstraw: Let's drop this builder in favor of the one below w/ zero points.  The zero point API is the correct API.  Inside that builder we can check all zero points.  If they are constant with value 0 we can call QuantizedDot.  Else we fall back to fp32.

src/ngraph/builder/quantization/quantized_linear_matmul.hpp
```diff
@@ -0,0 +1,47 @@
...
+                                                        std::shared_ptr<Node> filter,
+                                                        std::shared_ptr<Node> input_scale,
+                                                        std::shared_ptr<Node> filter_scale,
+                                                        std::shared_ptr<Node> output_scale);
+
+            std::shared_ptr<Node> QuantizedLinearMatmul(std::shared_ptr<Node> input,
```
adstraw: I am trying to standardize across all the quant stuff between `std::shared_ptr<Node>` and `const std::shared_ptr<Node>&`.  Let's go with the latter, the reference.  Please make this change in all files.

src/ngraph/builder/quantization/quantized_linear_matmul.hpp
```diff
@@ -0,0 +1,47 @@
...
+                                                        std::shared_ptr<Node> filter,
```
adstraw: I don't think we want `input` and `filter` here.  This is more like `X` and `Y` or some such.

src/ngraph/op/experimental/quantized_dot.cpp
```diff
@@ -39,13 +39,29 @@ op::QuantizedDot::QuantizedDot(const shared_ptr<Node>& data,
...
     NODE_VALIDATION_CHECK(this,
-                          data_shape.size() == 2 && weights_shape.size() == 2 &&
-                              data_shape[1] == weights_shape[1],
+                          data_shape.size() == 2 && weights_shape.size() == 2,
                           "only valid tensors of rank 2 supported. data shape ",
                           data_shape,
                           " weights shape ",
                           weights_shape);
 
+    // Mkldnn requirement. So skip when it goes to reference::dot
```
adstraw: Can we handle this in CPU assignment?  Why are we encoding mkldnn requirements at the op level?

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
```diff
@@ -0,0 +1,103 @@
...
+                                                   shared_ptr<Node> output_scale)
+            {
+                auto requantization_scale = (input_scale * filter_scale) / output_scale;
+
+                // Since u8u8 goes to ref reshape the filter so that the ref dot can do a matmul
+                // which requires [m, n] * [n, k] order where as mkldnn requires [m, n] * [k, n]
+                if (input->get_element_type() == element::u8 &&
+                    filter->get_element_type() == element::u8)
+                {
+                    filter = make_shared<op::Reshape>(
```
adstraw: Can this reshape be done in the CPU backend?  We should not be encoding MKLDNN requirements at the op level.

src/ngraph/runtime/cpu/builder/dot.cpp
```diff
@@ -250,7 +250,8 @@ namespace ngraph
                            arg0_shape,
                            arg1_shape,
                            result_shape,
-                           reduction_axes_count);
+                           reduction_axes_count,
+                           1.0f);
```
adstraw: What is this magic number?

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
```diff
@@ -0,0 +1,103 @@
+//*****************************************************************************
+// Copyright 2017-2018 Intel Corporation
```
rkimballn1: Fix the date

src/ngraph/builder/quantization/quantized_linear_matmul.hpp
```diff
@@ -0,0 +1,47 @@
+//*****************************************************************************
+// Copyright 2017-2018 Intel Corporation
```
rkimballn1: Date

src/ngraph/op/experimental/quantized_dot.cpp
nbpatel: The checks/output shape set were already mkldnn specific. I added check for the output shape to be a conventional [m, k] for a [m,n]*[n,k].  This op seems to assume we only have a mkldnn implementation, which was true before this PR

src/ngraph/runtime/cpu/builder/dot.cpp
nbpatel: Requantization scale which is 1 for non quant dots.

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
nbpatel: this check is more for ref kernel. I am reshaping it to [m, n] * [n, k] which is what ref expects. The onnx bridge will pass [m, n]* [k, n] which is what mkldnn expects. Assuming that the builders are CPU specific for now, I dont see a issue with this

src/ngraph/op/experimental/quantized_dot.cpp
adstraw: this op should not assume we only have an mkldnn implementation

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
adstraw: builders should not be cpu specific

src/ngraph/builder/quantization/quantized_linear_matmul.hpp
adstraw: Maybe more like `A` and `B`.  `Y` is normally used to denote an output.

nbpatel:
> Two ONNX Runtime tests are still failing (also on `master`):
> 
> ```
> [  FAILED  ] QuantizeLinearMatmulOpTest.QLinearMatMul3D
> [  FAILED  ] QuantizeLinearMatmulOpTest.QLinearMatMul
> ```
> 
> It would be great to address those before we wrap this up.

Fixed it

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
```diff
@@ -0,0 +1,138 @@
...
+                auto reshaped_input1 = input1;
+                // Check if zero point is constant and zero
+                if (input0_zero != nullptr && input1_zero != nullptr && output_zero != nullptr &&
+                    check_zero_point(input0_zero) && check_zero_point(input1_zero) &&
+                    check_zero_point(output_zero))
+                {
+                    auto requantization_scale = (input0_scale * input1_scale) / output_scale;
+
+                    // Since u8u8 goes to ref reshape the input1 so that the ref dot can do a matmul
+                    // which requires [m, n] * [n, k] order where as mkldnn requires [m, n] * [k, n]
```
adstraw: This is the wrong fix, in my opinion.  We should document the API as [m, n] * [n, k] and have ONNX importer call it with those semantics.  ONNX should be doing this reshape, not the builder.  What if happens now if someone passes in [m, n] * [n, k]?  Input 1 would be reshaped and fail.

src/ngraph/frontend/onnx_import/op/matmul.cpp
```diff
@@ -114,7 +110,15 @@ namespace ngraph
                                 AxisVector{1, 0},
                                 Shape(right->get_shape().rbegin(), right->get_shape().rend()));
 
-                            return {std::make_shared<ngraph::op::QuantizedDot>(left, right, scale)};
+                            return {ngraph::builder::quantization::QuantizedLinearMatmul(
+                                left,
```
adstraw: I like the "left" and "right" naming convention better than "input_0" and "input_1".

src/ngraph/runtime/cpu/builder/dot.cpp
adstraw: A comment that denotes 1.0 as a requant scale would be good.

src/ngraph/op/experimental/quantized_dot.cpp
adstraw: I still think it would be best to not have mkldnn specifics in the ops but I will not hold up the PR over it.

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
nbpatel: For QuantizedDot we expect the bridges to pass [m, n] * [k, n] and thats how the output shape is set and all the checks are done based on that assumption. maybe @jbobba can comment more. We need to decide on one standard.

src/ngraph/frontend/onnx_import/op/matmul.cpp
nbpatel: input0 and input1 is more like ngraph naming. We've used that at many places including in dot kernels

src/ngraph/op/experimental/quantized_dot.cpp
rkimballn1: We really need to get rid of backend specific code in core ngraph. It belongs in the cpu backend, not here.

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
adstraw: Reshaping based on element type is simply not correct.  The builder should enforce a policy on the shape of its inputs.  In my opinion, the standard should be [m,n] * [n,k] as with the non quantized dot operator.  The caller then has the responsibility to conform it's inputs to that convention.  

Alternatively, the builder could have a more flexible policy wherein either [m,n] * [n,k] OR [m,n] * [k,n] are accepted.  The builder should then reshape inputs based on their shapes, not their element types.

There is a fundamental assumption here that callers of this builder want to run on either the CPU or on the ref kernel in the case where mkldnn has no support.  This may be the case now but it is not the case forever.  Builders and core ops need a more general implementation.

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
```diff
@@ -0,0 +1,139 @@
...
+using namespace ngraph;
+
+namespace ngraph
+{
+    namespace builder
+    {
+        namespace quantization
+        {
+            // Check if zero point is zero
+            bool check_zero_point(const shared_ptr<op::Constant>& input)
```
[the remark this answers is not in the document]
jbobba: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/graph_util.cpp#L434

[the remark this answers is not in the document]
jbobba: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/graph_util.cpp#L434

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -2519,3 +2521,69 @@ void ngraph::runtime::cpu::pass::CPUQuantFusion::construct_qconvb_add()
...
+        auto scale_new = pattern_map[scale];
+
+        if (input_0->get_element_type() == element::u8 &&
+            input_1->get_element_type() == element::u8)
+        {
+            return false;
+        }
+
+        std::shared_ptr<ngraph::op::Op> qmatmul;
+        if (input_1->get_shape()[0] != input_1->get_shape()[1])
```
jbobba: you will need to always reshape/transpose input_1 irrespective of its shape since QuantizedMatMul expects the reduction axis to be the first axis in input_1 while QuantizedDot has the reduction axis as the second axis.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -2519,3 +2521,69 @@ void ngraph::runtime::cpu::pass::CPUQuantFusion::construct_qconvb_add()
...
+        q_dot = std::make_shared<ngraph::op::QuantizedDot>(input0, input1, scale, false, true);
+    }
+    else if (requantize && !with_relu)
+    {
+        q_dot = std::make_shared<ngraph::op::QuantizedDot>(input0, input1, scale, true, false);
+    }
+    else
+    {
+        q_dot = std::make_shared<ngraph::op::QuantizedDot>(input0, input1, scale, true, true);
+    }
```
jbobba: you can use a single pattern for all the four variants of QuantizedDot since the pattern matcher only looks for op types and not the attributes.

src/ngraph/runtime/cpu/builder/quantized_dot.cpp
```diff
@@ -114,70 +115,50 @@ namespace ngraph
...
-                            ctx, deps[2], ctx->buffer_data[out0_buffer_index]);
-                        cpu::mkldnn_utils::mkldnn_invoke_primitive(ctx, ip_index);
-                    };
-                    functors.emplace_back(functor);
-                }
-                else
-                {
-                    throw ngraph_error("unsupported parameters for QuantizedDot via DEX");
-                }
+                auto scales_size = shape_size(args[2].get_shape());
```
jbobba: check that this is equal to '1'.  Even better, do this in QuantizeDot's validate_and_infer_types method

src/ngraph/runtime/cpu/builder/quantized_dot.cpp
```diff
@@ -114,70 +115,50 @@ namespace ngraph
...
+                                arg0_shape,
+                                arg1_shape,
+                                result_shape,
+                                arg0_buffer_index,
+                                arg1_buffer_index,
+                                arg2_buffer_index,
+                                out0_buffer_index,
+                                scales_size](CPURuntimeContext* ctx, CPUExecutionContext* ectx) {
+
+                    vector<float> dyn_scales;
```
jbobba: use a straight float instead of vector since scales_size is always going to be 1.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -989,8 +989,8 @@ static const runtime::cpu::pass::AssignOpMap s_dispatcher{
      &runtime::cpu::pass::CPUAssignment::assign<ngraph::op::Dequantize>},
     {TI(ngraph::op::QuantizedConcat),
      &runtime::cpu::pass::CPUAssignment::assign<ngraph::op::QuantizedConcat>},
-    {TI(ngraph::op::QuantizedDot),
-     &runtime::cpu::pass::CPUAssignment::assign<ngraph::op::QuantizedDot>},
+    {TI(ngraph::op::QuantizedMatmul),
+     &runtime::cpu::pass::CPUAssignment::assign<ngraph::op::QuantizedMatmul>},
     {TI(ngraph::op::QuantizedDotBias),
```
jbobba: We will need to move QuantizedDotBias to CPU backend, probably as a separate PR.

src/ngraph/runtime/cpu/builder/quantized_dot.cpp
adstraw: not sure, but we may need to allow for non-scalar scales in this op at some later date.  could we not just reject this op in cpu assignment is scale is a non-scalar.

src/ngraph/runtime/cpu/builder/quantized_dot.cpp
```diff
@@ -114,70 +115,51 @@ namespace ngraph
...
+                auto functor = [&,
+                                kernel,
+                                arg0_shape,
+                                arg1_shape,
+                                result_shape,
+                                arg0_buffer_index,
+                                arg1_buffer_index,
+                                arg2_buffer_index,
+                                out0_buffer_index,
+                                scales_size](CPURuntimeContext* ctx, CPUExecutionContext* ectx) {
```
jbobba: remove scales_size

src/ngraph/runtime/cpu/builder/quantized_dot.cpp
jbobba: Current ref kernel only supports a scalar scale though?

PR #2862, opened 2019-05-03
Added extra attributes to DynSlice

jbobba:
Adds masks for upper and lower bounds, new and shrink axis masks and ellipsis masks.


src/ngraph/op/constant.hpp
```diff
@@ -126,6 +126,9 @@ namespace ngraph
                 set_output_type(0, m_element_type, m_shape);
             }
 
+            Shape get_shape_val() const;
```
diyessi: Please add descriptions here of what these are for.

louisfeng: I'm not sure when we could get negative values. Do we know clamping them to 0 make sense? Should we throw an exception here?

PR #2873, opened 2019-05-06
Quantization API cleanup (non breaking)

adstraw:


adstraw:
* Change "offset" to "zero point" in doc and code
* Use `const shared_ptr<Node>&` instead of `shared_ptr<Node>` in all APIs
* Consistent APIs across quant builders and ops

PR #2875, opened 2019-05-06
Enable non-constructor use of shape inference

jbobba:
'validate_nodes_and_infer_types' now runs shape inference on all underlying nodes instead of requiring the IN_TRANSITION in node.hpp to be set.
Note that this is currently called from the pass manager after each pass.


src/ngraph/function.cpp
```diff
@@ -70,10 +70,6 @@ Function::Function(const std::shared_ptr<Node>& result,
                    const std::string& name)
     : Function(NodeVector{result}, parameters, name)
 {
-    // TODO this does not do anything while infer happens in the constructors
-    // and it will go away after we add shape during a clone; it is here now
-    // to assist development between those two stages.
-    validate_nodes_and_infer_types();
```
aprocter: @diyessi Thoughts on whether removing this makes sense? (I think it does.)

diyessi: Separate PR.

src/ngraph/op/result.cpp
```diff
@@ -28,15 +28,15 @@ op::Result::Result(const shared_ptr<Node>& arg)
     : Op("Result", check_single_output_args({arg}))
 {
     constructor_validate_and_infer_types();
+    // always borrow the placement conf even the default one
+    set_placement_index(get_argument(0)->get_placement_index());
```
jbobba: @rkimballn1 do we need this? DefaultPlacement pass seems to be doing the backend assignment anyway.

jbobba:
This could break existing backends since shape inference will now be re-run after every pass in the pass manager. Backends should be OK as long as shape-inference for backend specific ops is idempotent.

src/ngraph/op/result.cpp
diyessi: This should be replaced by a pass where/if it's needed, but can be a later PR.

rkimballn1: Leave that there for now

src/ngraph/op/experimental/dyn_reshape.cpp
```diff
@@ -46,7 +47,21 @@ void op::DynReshape::validate_and_infer_types()
...
-    set_output_type(0, get_input_element_type(0), PartialShape::dynamic(output_rank));
+    if (auto const_shape = dynamic_pointer_cast<op::Constant>(get_argument(1)))
+    {
+        // TODO: replace with const_shape->get_shapes_val()
+        auto out_shape = const_shape->get_vector<int64_t>();
+        Shape output_shape(shape_size(const_shape->get_shape()));
+        std::transform(out_shape.begin(),
+                       out_shape.end(),
+                       output_shape.begin(),
+                       [&](const int64_t& v) { return (v > 0) ? v : 0; });
```
ransford2011: This could just be `max(v,0)`.  You are clipping the values here correct?

jbobba:
> I'm not sure how this'll interact with the elementwise broadcasting PlaidML does, but I think we can work around that if needed.

https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/plaidml/plaidml_compiler.cpp#L115-L119
With this PR, we will run shape validation/inference after every pass.

PR #2880, opened 2019-05-07
Store Constant arrays where data is constant as a single value

rkimballn1:
This change serializes Constant data where all of the values are bit exact the same as a single constant value. This single value is then broadcast when reading to reconstitute the original Constant. This is done without changing the graph in any way.

I added a unit test to better exercise Constant serialization. I had to change all_close_f because it did not compare inf/nan in a way that I really think it should.

I also ran into issues where f16 (not bf16) was not supported properly in Constant so I switched over to using enum/switch so we can make it a compile-time error when something like this happens again.


test/util/all_close_f.cpp
```diff
@@ -82,8 +94,20 @@ uint32_t test::float_distance(float a, float b, float min_signal)
 
 uint64_t test::float_distance(double a, double b, double min_signal)
```
aprocter: @gcwenger Can you take a look at the changes to `close_f` and `float_distance`? (Rest of the PR looks fine to me.)

aprocter:
@rkimballn1 Note also that this is breaking the NGTF build: https://buildkite.com/ngraph/ngcore-cpu-tf-build/builds/932#6f1b6360-62b5-415c-b3bd-4a53e2f36d79/56-703

aprocter:
FWIW, it looks like `numpy.allclose` has an optional `equal_nan` flag (meaning "act as if nan is close to nan"), which defaults to `False`, and treats inf as close to inf:

https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html

PR #2881, opened 2019-05-07
Move ONNX lp-norm utility functions to core op/util.

arogowie-intel:
These functions will be needed for implementation of OpenVINO operators.

Additionally extends lp-norm functions with bias parameter.


postrational:
I added my comments to https://github.com/NervanaSystems/ngraph/pull/2888

src/ngraph/builder/norm.cpp
```diff
@@ -0,0 +1,153 @@
...
+            ///
+            /// \brief      Creates constant node with the given bias value.
+            ///
+            /// \param[in]  element_type  The node element type.
+            /// \param[in]  node_shape    The node shape.
+            /// \param[in]  bias          The value to fill node with.
+            ///
+            /// \return     The created Constant node.
+            ///
+            inline shared_ptr<Node> get_bias_node(const element::Type& element_type,
```
postrational: I'm still not convinced this is function is needed. Please consider removing it.

Since it creates a `Constant`, it should be renamed to reflect what it does, rather than what it's used for - change the name to `make_constant_node`

src/ngraph/frontend/onnx_import/op/reduce.hpp
```diff
@@ -98,8 +99,12 @@ namespace ngraph
                 ///
                 inline NodeVector reduce_l1(const Node& node)
                 {
+                    auto reduce_function = std::bind(ngraph::builder::l1_norm,
```
postrational: Please change the name from generic `reduce_function` to specific, like: `l1_norm_reduction`

src/ngraph/frontend/onnx_import/op/reduce.hpp
```diff
@@ -115,8 +120,12 @@ namespace ngraph
                 ///
                 inline NodeVector reduce_l2(const Node& node)
                 {
+                    auto reduce_function = std::bind(ngraph::builder::l2_norm,
```
postrational: Please change the name from generic `reduce_function` to specific, like: `l2_norm_reduction`

src/ngraph/frontend/onnx_import/op/reduce.hpp
arogowie-intel: Doesn't this come from the name of the function this is in?

src/ngraph/frontend/onnx_import/op/reduce.hpp
arogowie-intel: Doesn't this come from the name of the function this is in?

PR #2883, opened 2019-05-07
nbench special handle op name for "NNP_XXX"

fengleitian:
Since some op name in NNP is "NNP_XXX", it will only show "NNP" in performance result. Add code to special handle for "NNP_XXX" op name.


src/tools/nbench/nbench.cpp
```diff
@@ -71,14 +71,23 @@ vector<PerfShape> to_perf_shape(shared_ptr<Function> f,
...
+
 multimap<size_t, string> aggregate_timing_details(const vector<PerfShape>& perf_data)
 {
     unordered_map<string, size_t> timing;
     unordered_map<string, size_t> count;
     for (const PerfShape& p : perf_data)
     {
         auto node = p.get_node();
-        string op = node->get_name().substr(0, node->get_name().find('_'));
+        string op = get_common_name(node->get_name());
```
rkimballn1: I think we can just use node->description() here instead of trimming the node's name. get_name() might return something without an underscore if the user has set a friendly name.

fengleitian: much better. Thank you.

PR #2884, opened 2019-05-07
Langjian/mpi finalize

jianyinglang:
Add a simple fix to avoid crash in the final step of distributed TF training.


src/ngraph/distributed/open_mpi.hpp
```diff
@@ -37,20 +37,21 @@ namespace ngraph
...
             }
 
             ~OpenMPIDistributedInterface() override
             {
-                int flag = 0;
-                MPI_Initialized(&flag);
-
-                if (flag)
+                int is_mpi_finalized = 0;
+                MPI_Finalized(&is_mpi_finalized);
```
wenzhe-nrv: if this condition possible:
`is_mpi_finalized && m_initialized_mpi == True`

If yes, could we add a warning here? I don't expect this condition to be true but if it is, we want to know.

PR #2886, opened 2019-05-08
[FusedOps] Clamp

tomdol:
Adding fused op checklist:
- [x] Update serializer: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/serializer.cpp
- [x] Update https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/fused_op_tbl.hpp file.
- [x] Add UT here: https://github.com/NervanaSystems/ngraph/blob/master/test/backend_fused_op.in.cpp
- [x] Add simple UT here: https://github.com/NervanaSystems/ngraph/blob/master/test/type_prop.cpp#L13364
- [x] Add ops to aggregate nGraph header: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/ngraph.hpp#L98


src/ngraph/op/fused/clamp.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+using namespace std;
+using namespace ngraph;
+
+op::Clamp::Clamp(const shared_ptr<Node>& data, const double min, const double max)
+    : FusedOp("Clamp", {data})
+    , m_min{min}
+    , m_max{max}
+{
+    NODE_VALIDATION_CHECK(
```
aprocter: I think it's okay for now, but I feel like this check should be in `validate_and_infer_types`. It's hard to shoehorn it in there, though, because that function has to do the work of checking the decomposed graph as well, and that's handled by the base class.

jbobba: We have pre- and post-validation hooks for these kinds of checks. https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/util/fused_op.hpp#L40
Can we override and use those?

aprocter: Nice- didn’t realize that was there. Should be the right place for this check.

src/ngraph/op/fused/clamp.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "ngraph/node.hpp"
+#include "ngraph/op/op.hpp"
+#include "ngraph/op/util/fused_op.hpp"
+
+namespace ngraph
+{
+    namespace op
+    {
+        class Clamp : public ngraph::op::util::FusedOp
```
postrational: Please add documentation for class and constructor.

PR #2895, opened 2019-05-09
Add support for Fused Ops in Intel GPU backend

rkimballn1:


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -410,6 +410,7 @@ shared_ptr<runtime::Executable>
     {
         ngraph::pass::Manager pass_manager;
 
+        pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>();
```
dmyershov: @rkimballn1 our backend supports all types of Convolution-s from fused_op_tbl.hpp. In this case they will be decomposed. Therefore, we need to register FusedOpDecomposition pass with call back function which will return false for the non supported ops (Elu, Gemm, PRelu, DepthToSpace, SpaceToDepth).

Then,  FusedOpDecomposition  pass should be executed always. Otherwise our backend will fail. It should be moved outside of "if (!m_disable_backend_optimizations)"

rkimballn1: Should I just close this PR?

rkimballn1: There are many PRs which are adding decompose ops and I see that they just disable the unit tests for these new ops in the intel gpu backend. I hate seeing unit tests disabled.

dmyershov: Yes, please close this PR. Today I'll prepare my solution.

PR #2898, opened 2019-05-09
Fix clang compiler warnings

diyessi:


src/ngraph/op/util/unary_elementwise_arithmetic.cpp
```diff
@@ -18,6 +18,16 @@
 
 using namespace ngraph;
 
+op::util::UnaryElementwiseArithmetic::UnaryElementwiseArithmetic()
```
aprocter: I think this is extraneous to this PR, right?

src/tools/nbench/nbench.cpp
```diff
@@ -85,7 +85,7 @@ multimap<size_t, string> aggregate_timing_details(const vector<PerfShape>& perf_
     }
 
     multimap<size_t, string> rc;
-    for (const pair<string, size_t>& t : timing)
+    for (auto& t : timing)
```
rkimballn1: Can this be `const auto&`? I much prefer to keep the const

diyessi: `auto` can't make `const`ness go away.

src/ngraph/op/util/unary_elementwise_arithmetic.cpp
diyessi: Gone.

src/ngraph/op/util/unary_elementwise_arithmetic.hpp
```diff
@@ -44,6 +44,13 @@ namespace ngraph
             class UnaryElementwiseArithmetic : public Op
             {
             protected:
+                UnaryElementwiseArithmetic();
```
aprocter: I think this is another one that snuck in- sorry, should have flagged it on last review.

diyessi: Evicted. I don't see any more.

PR #2905, opened 2019-05-10
[Fused Op] GRN

arogowie-intel:
Global Response Normalization with L2 norm across channels only.


test/backend_fused_op.in.cpp
```diff
@@ -377,3 +377,27 @@ NGRAPH_TEST(${BACKEND_NAME}, gemm_broadcast_input_C)
     test_case.add_expected_output<double>(Shape{3, 4}, vector<double>(12, 7));
     test_case.run();
 }
+
+NGRAPH_TEST(${BACKEND_NAME}, grn_4d)
```
dmyershov: Will it give pass with INTELGPU backend? If not, please add it to intelgpu's unit_test.manifest

src/ngraph/builder/norm.hpp
```diff
@@ -21,6 +21,8 @@
 #include "ngraph/axis_set.hpp"
 #include "ngraph/node.hpp"
 
+using namespace std;
```
tomdol: I think we should not use it in header files.

src/ngraph/op/fused/grn.cpp
```diff
@@ -0,0 +1,90 @@
...
+                              ").");
+    }
+}
+
+NodeVector op::GRN::decompose_op() const
+{
+    const auto input_node{get_argument(0)};
+    const auto& input_shape{input_node->get_shape()};
+
+    auto data{input_node};
```
tomdol: Can you operate on the `input_node` directly? I think you're making an unnecessary copy here.

src/ngraph/op/fused/grn.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    return {data};
+}
+
+shared_ptr<Node> op::GRN::copy_with_new_args(const NodeVector& new_args) const
+{
+    if (new_args.size() != 1)
+    {
+        throw ngraph_error("Incorrect number of new arguments");
```
tomdol: I'm still not sure about it, but I think `NODE_VALIDATION_CHECK` should be used for validation to include the node info

test/backend_fused_op.in.cpp
```diff
@@ -715,3 +715,49 @@ NGRAPH_TEST(${BACKEND_NAME}, mvn_mean_variance_normalization_split_channels)
 
     test_case.run();
 }
+
+NGRAPH_TEST(${BACKEND_NAME}, grn_4d)
+{
+    Shape data_shape{1, 2, 3, 4};
```
tomdol: Not very important thing, but can you add `const` to everything you can in those tests?

src/ngraph/op/fused/grn.cpp
arogowie-intel: I think not necessarily, since this isn't validation of this node internal properties.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -2053,6 +2054,7 @@ shared_ptr<runtime::Executable>
         case OP_TYPEID::Gather:
         case OP_TYPEID::GatherND:
         case OP_TYPEID::GenerateMask:
+        case OP_TYPEID::GRN:
```
dmyershov: Please add `case OP_TYPEID::GRN:` line into is_supported_impl function switch (lines 2163 - 2172 of this file)

src/ngraph/runtime/intelgpu/unit_test.manifest
```diff
@@ -79,3 +79,5 @@ scatter_add_scalar_indices
 scatter_nd_add_batch_2d_to_3d
 scatter_nd_add_2d_to_3d
 zero_sized_erf
+grn_4d
```
dmyershov: These tests will work after decomposition. Please delete them from manifest file

arogowie-intel:
@dmyershov I see errors after fulfilling your request to add `case OP_TYPEID::GRN:` to the place you wanted. Now this case isn't handled in first place which causes errors. Should it be in both ?

PR #2912, opened 2019-05-11
IntelGPU backend: Enable FusedOpDecomposition pass

dmyershov:


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -2083,3 +2086,31 @@ bool runtime::intelgpu::IntelGPUBackend::is_supported_property(const Property pr
...
+    case OP_TYPEID::DepthToSpace:
+    case OP_TYPEID::Elu:
+    case OP_TYPEID::Gemm:
+    case OP_TYPEID::MVN:
+    case OP_TYPEID::Normalize:
+    case OP_TYPEID::PRelu:
+    case OP_TYPEID::SpaceToDepth:
+    {
+        return false;
+        break;
```
shssf: No need this line

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -409,10 +410,12 @@ shared_ptr<runtime::Executable>
         visualize_tree(func, "intelgpu_", "_orig");
     }
 
+    ngraph::pass::Manager pass_manager;
+    pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>(
+        runtime::intelgpu::IntelGPUBackend::is_supported_impl);
```
shssf: It might be shorter like "IntelGPUBackend::is_supported_impl"

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -2083,3 +2086,31 @@ bool runtime::intelgpu::IntelGPUBackend::is_supported_property(const Property pr
 
     return false;
 }
+
+bool runtime::intelgpu::IntelGPUBackend::is_supported(const Node& node) const
+{
+    return runtime::intelgpu::IntelGPUBackend::is_supported_impl(node);
```
shssf: It might be shorter like "IntelGPUBackend::is_supported_impl"

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -409,10 +410,12 @@ shared_ptr<runtime::Executable>
         visualize_tree(func, "intelgpu_", "_orig");
     }
 
+    ngraph::pass::Manager pass_manager;
```
shssf: If the manager moved out of the `m_disable_backend_optimizations` control you have no chance to guarantee that the backend works with absolutely the same graph as input graph. I think it is better to keep absolutely all optimizations and graph modifications under `m_disable_backend_optimizations` control

shssf:
I think that `pass::FusedOpDecomposition::op_query_t` type has wrong design. It should be redesigned but it might be done in several ways.
1. allow `this` in the type to avoid redundant static or delta functions to use
2. If look into future, more scalable solution might be passing the backend pointer to `ngraph::pass::Manager::ctor()` to allow usage of the backend functionality in all optimizations passes instead particular one. Also, optimizations might use other backend functionality like `Backend::is_supported_property()`.

PR #2913, opened 2019-05-11
Fix building on centos with gcc 4.8.5

rkimballn1:
These changes are required to make this work on Centos.
Centos is doing things differently with the following syntax
```auto scale_node{get_argument(1)};```
While gcc on ubuntu sets scale_node to a `shared_ptr<Node>`, centos makes an `initializer_list<shared_ptr<Node>>`. Yet another reason to avoid the `auto` keyword.


silee2:
I see the same issue with latest ICC. "auto" with C++ uniform initialization syntax does not seem to be well defined for C++11.

PR #2915, opened 2019-05-12
Address Klocwork issue. Fix bug?

rkimballn1:
I wasn't sure if you really wanted to only use the `results[0]` for everything but it sure looked like a bug to me and Klockwork was complaining.


aprocter:
Whoops!

PR #2926, opened 2019-05-14
Add support for more types in gather op.

tsocha:
Add suport for types:
-i8
-i16
-i32
-i64
-u8
-u16
-u32
-u64


src/ngraph/runtime/cpu/builder/gather.cpp
```diff
@@ -29,116 +29,126 @@ namespace ngraph
...
+                {
+                    functor = prepare_functor<uint32_t>(node, args, out, external_function);
+                }
+                else if (element_type == element::u64)
+                {
+                    functor = prepare_functor<uint64_t>(node, args, out, external_function);
+                }
+                else if (element_type == element::boolean)
+                {
+                    functor = prepare_functor<bool>(node, args, out, external_function);
```
rkimballn1: in ngraph element::boolean is not a c type bool, but is a `char` type.
Would be nice to have a unit test to check this out. Really does need a unit test for all types.

tsocha: OK. I will add these tests and change c type to char.

rkimballn1: It still looks like it is type bool

tsocha: Sorry, I missed that file in git add.

PR #2927, opened 2019-05-14
[FusedOps] ShuffleChannels

tomdol:


src/ngraph/op/fused/shuffle_channels.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+    size_t N = data_shape.at(0);
+    size_t C = data_shape.at(1);
+    size_t H = data_shape.at(2);
+    size_t W = data_shape.at(3);
+
+    // if the axis parameter is different than one, the C value must be taken from dimension 0
+    if (m_axis != 1)
+    {
+        std::swap(N, C);
```
tsocha: I'm not sure about this if and swap.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -2059,6 +2060,7 @@ shared_ptr<runtime::Executable>
         case OP_TYPEID::ScatterAdd:
         case OP_TYPEID::ScatterNDAdd:
         case OP_TYPEID::ShapeOf:
+        case OP_TYPEID::ShuffleChannels:
```
dmyershov: Please add "case OP_TYPEID::ShuffleChannels:" line into switch of runtime::intelgpu::IntelGPUBackend::is_supported_impl function (intelgpu_backend.cpp). By #2912 we implemented functionality to decompose unsupported fused ops.

src/ngraph/op/fused/shuffle_channels.cpp
arogowie-intel: This condition looks suspicious also to me. What's the logic behind this?

tomdol: I've removed it, it was a wrong approach

src/ngraph/op/fused/shuffle_channels.cpp
```diff
@@ -0,0 +1,93 @@
...
+    , m_axis{axis}
+    , m_groups{groups}
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::ShuffleChannels::pre_validate_and_infer_types()
+{
+    NODE_VALIDATION_CHECK(this,
+                          m_axis >= 0,
```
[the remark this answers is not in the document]
tomdol: I've changed it and now the constructor can also accept negative values meaning counting from the back. Can't use size_t in this case since the validation happens in a separate method.

src/ngraph/op/fused/shuffle_channels.cpp
```diff
@@ -0,0 +1,93 @@
...
+                          "Accepted values of the 'axis' parameter for ShuffleChannels are "
+                          "positive integers and zero.");
+
+    NODE_VALIDATION_CHECK(
+        this,
+        m_groups > 0,
```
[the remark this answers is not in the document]
tomdol: It's only enforced if `-Wnarrowing` compiler flag is on. Can we assume that?

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
dmyershov: No. This line should be it 2 places. In switch of compile function and in switch of is_supported_imp function

src/ngraph/op/fused/shuffle_channels.hpp
```diff
@@ -0,0 +1,60 @@
...
+            /// \param data - Node producing the input tensor
+            /// \param axis - channel dimension index in the data tensor
+            /// \param groups - a number of groups the channel dimension specified by axis should be split to
+            ShuffleChannels(const std::shared_ptr<ngraph::Node>& data,
+                            const int axis = 1,
+                            const size_t groups = 1UL);
+
+            void pre_validate_and_infer_types() override;
+
+            NodeVector decompose_op() const override;
```
arogowie-intel: Please add `virtual` keyword. It looks more starightforwad then that this virtual function.

src/ngraph/op/fused/shuffle_channels.hpp
```diff
@@ -0,0 +1,61 @@
...
+    {
+        /// \brief Permutes data in the channel dimension of the input
+        class ShuffleChannels : public ngraph::op::util::FusedOp
+        {
+        public:
+            /// \brief Constructs a ShuffleChannels node.
+            ///
+            /// \param data - Node producing the input tensor
+            /// \param axis - channel dimension index in the data tensor. A negative value means that the index should be calculated from the back of the input data shape.
+            /// \param groups - a number of groups the channel dimension specified by axis should be split to
```
postrational: `a number` -> `number`
`split to` -> `split into`

src/ngraph/op/fused/shuffle_channels.cpp
```diff
@@ -0,0 +1,107 @@
...
+    , m_groups{groups}
+{
+    if (axis < 0)
+    {
+        m_axis = axis + data->get_shape().size();
+    }
+    else
+    {
+        m_axis = axis;
+    }
```
jbobba: Move this to pre_validate_and_infer_types. We cannot assume here that 'data' will have a static shape.Can we save axis specification (including negative value) to m_axis and adjust on-the-fly when needed, possibly through a helper method 'get_shuffle_axis'.

src/ngraph/op/fused/shuffle_channels.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    constructor_validate_and_infer_types();
+}
+
+void op::ShuffleChannels::pre_validate_and_infer_types()
+{
+    const auto shape = get_argument(0)->get_shape();
```
jbobba: guard with get_input_partial_shape(0)->is_static()

src/ngraph/op/fused/shuffle_channels.cpp
```diff
@@ -0,0 +1,107 @@
...
+        "The channel dimension size has to be a multiple of the groups parameter value.");
+}
+
+NodeVector op::ShuffleChannels::decompose_op() const
+{
+    const auto data = get_argument(0);
+    const auto& data_shape = data->get_shape();
+
+    const auto reshaped = util::reshape(data, get_pre_shuffle_shape(data_shape));
+    const auto shuffled = util::reorder_axes(reshaped, {0, 2, 1, 3});
```
jbobba: will we need to add shape.size() == 4 to pre_validate_and_infer_types?

src/ngraph/op/fused/shuffle_channels.cpp
tomdol: Ok but in this case:
1. Should I validate the axis value against the input data shape in `pre_validate_and_infer_types` at all? Or should I do it only for static shapes?
2. Where should I adjust the axis with the proposed helper method? Would that be in `decompose_op`?

src/ngraph/op/fused/shuffle_channels.cpp
tomdol: I don't think so. I've tested it locally and it works for any tensor that is 1D or higher. An intermediate step for this operation is reshaping to 4D and shuffling data in this format. Then it's reshaped back to the original dimensions.

src/ngraph/op/fused/shuffle_channels.cpp
jbobba: > Ok but in this case:
> 
> 1. Should I validate the axis value against the input data shape in `pre_validate_and_infer_types` at all? Or should I do it only for static shapes?
Right. validate only if input shape is static.
> 2. Where should I adjust the axis with the proposed helper method? Would that be in `decompose_op`?
get_shuffle_axis could be called in decompose_op or from backends when they are figuring out how to execute the op.

tomdol: I've addressed your comment, please have a look again

PR #2928, opened 2019-05-14
[Fused] FakeQuantize operation.

arogowie-intel:


src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -2050,6 +2051,7 @@ shared_ptr<runtime::Executable>
         case OP_TYPEID::Elu:
         case OP_TYPEID::EmbeddingLookup:
         case OP_TYPEID::Erf:
+        case OP_TYPEID::FakeQuantize:
```
dmyershov: Please add "case OP_TYPEID::FakeQuantize:" line into switch of runtime::intelgpu::IntelGPUBackend::is_supported_impl function (intelgpu_backend.cpp). By #2912 we implemented functionality to decompose unsupported fused ops.

src/ngraph/runtime/intelgpu/unit_test.manifest
```diff
@@ -79,3 +79,6 @@ scatter_add_scalar_indices
 scatter_nd_add_batch_2d_to_3d
 scatter_nd_add_2d_to_3d
 zero_sized_erf
+fake_quantize
+fake_quantize_with_clip
+fake_quantize_with_clip_across_channels
```
dmyershov: These tests might work in case of decomposed FakeQuantize

arogowie-intel: At this moment this PR lacks operator implementation, thus I'll leave these uts skipped.

src/ngraph/op/fused/fake_quantize.cpp
```diff
@@ -0,0 +1,139 @@
...
+    const auto& data_pshape = get_input_partial_shape(0);
+    const auto& input_low_pshape = get_input_partial_shape(1);
+    const auto& input_high_pshape = get_input_partial_shape(2);
+    const auto& output_low_pshape = get_input_partial_shape(3);
+    const auto& output_high_pshape = get_input_partial_shape(4);
+
+    if (data_pshape.is_static() && input_low_pshape.is_static() && input_high_pshape.is_static() &&
+        output_low_pshape.is_static() && output_high_pshape.is_static())
+    {
+        const Shape& data_shape{data_pshape.to_shape()};
```
rkimballn1: These statements look odd to me. You are taking the reference of the temporary value returned by to_shape(). Where is the temporary stored? I would feed better if this were just a const Shape and not a reference. Does this code work for all of our compilers?

src/ngraph/op/fused/fake_quantize.cpp
```diff
@@ -0,0 +1,139 @@
...
+    //     output = output_low
+    // # round halfway cases away from zero
+    // round((x - input_low) / (input_high - input_low) * (levels-1)) /
+    //     (levels-1) * (output_high - output_low) + output_low
+    // elif x > input_high:
+    //     output = output_high
+    // else:
+    //     # input_low < x <= input_high
+    //
+    throw ngraph_error("Not yet implemented");
```
aprocter: @jbobba  Are we okay with `decompose_op` not being implemented yet?

test/backend_fused_op.in.cpp
```diff
@@ -757,3 +757,124 @@ NGRAPH_TEST(${BACKEND_NAME}, scale_shift)
     test_case.add_expected_output<double>(Shape{3, 6}, vector<double>(18, 6));
     test_case.run();
 }
+
+NGRAPH_TEST(${BACKEND_NAME}, fake_quantize)
```
jbobba: Can we disable the tests here instead of adding them individually to the various manifest files?

jbobba: And add a comment to enable them once FakeQuantize implementation is defined.

test/type_prop.cpp
```diff
@@ -14446,3 +14446,94 @@ TEST(type_prop, scale_shift)
     EXPECT_EQ(scale_shift_func->get_element_type(), element::f64);
     EXPECT_EQ(scale_shift_func->get_shape(), (Shape{3, 6}));
 }
+
+TEST(type_prop, DISABLED_fake_quantize)
```
jbobba: Lets implement these tests since they are independent of the decomposition for fake_quantize.

src/ngraph/op/fused/fake_quantize.cpp
jbobba: i think so. But we should override 'validate_and_infer_types' to provide shape inference while we figure out the decomposition.

src/ngraph/runtime/cpu/unit_test.manifest
```diff
@@ -4,9 +4,9 @@ shape_of_matrix
...
 
-# failing in CI build but passing on local machine
+#failing in CI build but passing on local machine
 max_3d_to_scalar_int32
 
-# Not implemented
-erf
-zero_sized_erf
+#Not implemented
+erf zero_sized_erf
```
postrational: I think this `erf zero_sized_erf` should be split into 2 lines:
```
erf
zero_sized_erf
```

tomdol: Good catch, fixed

src/ngraph/op/fused/fake_quantize.cpp
tomdol: I've changed those references to local objects.

src/ngraph/op/fused/fake_quantize.cpp
tomdol: I've added the implementation. Please have a look now.

test/backend_fused_op.in.cpp
tomdol: The tests are in place, one of them still disabled because it's failing.

src/ngraph/op/fused/fake_quantize.cpp
```diff
@@ -0,0 +1,163 @@
...
+    const auto quantized_data =
+        make_shared<op::Quantize>(data,
+                                  quant_scale,
+                                  zero_point,
+                                  element::i32,
+                                  axes,
+                                  op::Quantize::RoundMode::ROUND_NEAREST_TOWARD_INFINITY);
+
+    // the output element type of dequantization is set to the output_low element type
+    // because of the first op::Select usage below where last 2 tensors must have matching types
```
arogowie-intel: Please update comment.

PR #2936, opened 2019-05-15
Initial implementation of implicit broadcasting for eltwise ops

jbobba:
TODO

- [x] Figure out the autodiff story for implicit broadcast modes 
       Will be supported through reductions along broadcast axes in a future PR
- [x] Expand to other eltwise ops
- [x] Add more eltwise execution unit tests
- [x] Serializer support


src/ngraph/dimension.cpp
```diff
@@ -103,3 +103,32 @@ bool Dimension::merge(Dimension& dst, const Dimension d1, const Dimension d2)
         return true;
     }
 }
+
+bool Dimension::merge_bcast(Dimension& dst, const Dimension d1, const Dimension d2)
```
aprocter: Minor nit: I'd prefer to spell out `broadcast`.

aprocter: Not sure if there is anything numpy-specific in this behavior but we might think about it and name the function accordingly if so. (Can't imagine what other behaviors would make sense at the dimension level, though.)

src/ngraph/node.cpp
```diff
@@ -485,18 +486,28 @@ std::tuple<element::Type, PartialShape> Node::validate_and_infer_elementwise_arg
...
-                                  PartialShape::merge_into(pshape, get_input_partial_shape(i)),
-                                  "Argument shapes are inconsistent.");
+            if (autob == op::AutoBcastType::NUMPY)
+            {
+                NODE_VALIDATION_CHECK(
+                    this,
+                    PartialShape::bcast_merge_into(pshape, get_input_partial_shape(i), autob),
+                    "Argument shapes are inconsistent.");
+            }
+            else
```
aprocter: Might want to change this to `else if(autob==NONE)` and have an `else` that catches not-yet-implemented broadcast schemes.

src/ngraph/op/util/binary_elementwise_arithmetic.cpp
```diff
@@ -15,19 +15,30 @@
...
 
 void op::util::BinaryElementwiseArithmetic::validate_and_infer_types()
 {
-    validate_and_infer_elementwise_arithmetic();
+    validate_and_infer_elementwise_arithmetic(m_autob);
+}
+
+NodeVector op::util::BinaryElementwiseArithmetic::auto_broadcast()
+{
+    if (m_autob == op::AutoBcastType::NUMPY)
```
aprocter: Again, might want an `else` here.

src/ngraph/partial_shape.cpp
```diff
@@ -244,3 +245,33 @@ bool PartialShape::merge_into(PartialShape& dst, const PartialShape& src)
         return success;
     }
 }
+
+bool PartialShape::bcast_merge_into(PartialShape& dst,
```
aprocter: Again, would prefer to spell out `broadcast`.

src/ngraph/dimension.cpp
aprocter: Might rename to `broadcast_merge` for consistency with `merge_into_broadcast` below. (I dunno, maybe it's technically not inconsistent since one has the preposition.)

src/ngraph/op/util/binary_elementwise_arithmetic.cpp
```diff
@@ -15,19 +15,30 @@
...
 {
 }
 
 void op::util::BinaryElementwiseArithmetic::validate_and_infer_types()
 {
-    validate_and_infer_elementwise_arithmetic();
+    validate_and_infer_elementwise_arithmetic(m_autob);
+}
+
+NodeVector op::util::BinaryElementwiseArithmetic::auto_broadcast()
```
aprocter: This might be better as a non-member function.

src/ngraph/pass/auto_broadcast.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma once
+
+#include "ngraph/pass/pass.hpp"
+
+namespace ngraph
+{
+    namespace pass
+    {
+        class AutoBroadcast : public NodePass
```
aprocter: Maybe `ImplicitBroadcastElimination` is a better name?

aprocter: Also (no need to do for now, but) maybe we could have an option to specify which kinds of autobroadcast we do and don't want to eliminate. I think that some targets will support numpy broadcasting without elimination but might need Paddle-style to be converted to explicit.

test/type_prop.cpp
```diff
@@ -14446,3 +14446,22 @@ TEST(type_prop, scale_shift)
     EXPECT_EQ(scale_shift_func->get_element_type(), element::f64);
     EXPECT_EQ(scale_shift_func->get_shape(), (Shape{3, 6}));
 }
+
+TEST(type_prop, add_bcast)
```
aprocter: In addition to this kind of stuff I think we'll want to add some tests for `bcast_merge` and the like to `test/partial_shape.cpp`.

src/ngraph/dimension.cpp
diyessi: Can you make this return a result instead of via an arg? That or make `this` serve as `dst` rather than a static.  I think the static methods are leftovers from a different implementation of `Dimension`.

src/ngraph/node.cpp
diyessi: I was thinking along the lines of a broadcast resolver pure class that frameworks implement. Core could provide a NumPy implementation as an example, although I think none of our frameworks actually do NumPy.

src/ngraph/dimension.cpp
aprocter: @diyessi The statics are still in use. `merge` (like the new `merge_bcast`) can fail, and callers may want to recover gracefully from the failure, so I had made `merge` return `bool`. (Generally we are not using try/catch for situations like this in nGraph.)

src/ngraph/node.cpp
aprocter: Not sure I like the idea of an opaque resolver object. With that we wouldn't be able to deserialize graphs with implicit broadcast and run them in `nbench`, for example.

src/ngraph/dimension.cpp
jbobba: i dont think there is any numpy-specific behavior here. Most of that happens at the Shape level where there seem to be different approaches on how to line up axes.

src/ngraph/op/util/attr_types.hpp
```diff
@@ -49,5 +51,28 @@ namespace ngraph
             AUTO = SAME_UPPER,
             NOTSET = EXPLICIT
         };
+
+        enum class AutoBcastType
```
aprocter: Would prefer to change this to `AutoBroadcastType`

src/ngraph/op/util/attr_types.hpp
```diff
@@ -49,5 +51,28 @@ namespace ngraph
...
+        {
+            NONE = 0,
+            NUMPY
+        };
+
+        struct AutoBcastSpec
```
aprocter: `AutoBroadcastSpec`

src/ngraph/op/util/attr_types.hpp
jbobba: you really dont like Bcast :)

aprocter: I just really like typing.

jbobba:
> Good stuff! In the comments are a couple of cosmetic gripes. A few additional things I think we should do (ideally before merging, but I could be persuaded to take a raincheck):
> 
> 1. Expand support to all binops.
> 2. Add a few execution tests for binops with implicit broadcast. (Okay to disable in non-CPU/INTERPRETER BEs for now.)
> 3. (Likely consequence of 2) Update other backends to invoke.

Will remove WIP once i expand to all binops. We can run execution tests on other backends as well since they will invoke the ImplicitBroadcastElimination pass

aprocter:
> Will remove WIP once i expand to all binops. We can run execution tests on other backends as well since they will invoke the ImplicitBroadcastElimination pass

SGTM. I'd actually meant to backspace the comment about updating other BEs, since it occurred to me as I was typing it that the manifest exists. But it might be worth doing in this PR as well.

src/ngraph/dimension.cpp
```diff
@@ -103,3 +103,32 @@ bool Dimension::merge(Dimension& dst, const Dimension d1, const Dimension d2)
...
+    if (d1.is_dynamic() && d2.is_dynamic())
+    {
+        dst = d1;
+        return true;
+    }
+    else if (d1.is_dynamic() || d2.is_dynamic())
+    {
+        // One static. Set dst to static size if >1
+        auto ds = d1.is_dynamic() ? size_t(d2) : size_t(d1);
+        dst = (ds > 1) ? ds : Dimension::dynamic();
```
aprocter: I dunno why, but I genuinely love corner cases like this.

src/ngraph/op/util/attr_types.hpp
```diff
@@ -49,5 +51,33 @@ namespace ngraph
...
             NOTSET = EXPLICIT
         };
+
+        enum class AutoBroadcastType
+        {
+            NONE = 0,
+            NUMPY
+        };
+
+        struct AutoBroadcastSpec
```
rkimballn1: This could use some doc strings, also on ctors

src/ngraph/op/util/attr_types.hpp
```diff
@@ -49,5 +51,33 @@ namespace ngraph
...
+                , axis(0)
+            {
+            }
+            AutoBroadcastSpec(AutoBroadcastType t, size_t s)
+                : type(t)
+                , axis(s)
+            {
+            }
+
+            AutoBroadcastType type;
```
rkimballn1: I'm sure I am the only one who thinks this, but these should be m_type and there should be accessor functions.

src/ngraph/op/util/attr_types.hpp
```diff
@@ -49,5 +51,33 @@ namespace ngraph
             AUTO = SAME_UPPER,
             NOTSET = EXPLICIT
         };
+
+        enum class AutoBroadcastType
+        {
+            NONE = 0,
+            NUMPY
```
rkimballn1: It is a shortcut to just say that this works like numpy broadcast, but it would be nice to actually know the rules without having to go find numpy's documentation to find out how our code works.

src/ngraph/partial_shape.cpp
```diff
@@ -244,3 +245,33 @@ bool PartialShape::merge_into(PartialShape& dst, const PartialShape& src)
         return success;
     }
 }
+
+bool PartialShape::broadcast_merge_into(PartialShape& dst,
+                                        const PartialShape& src,
+                                        const op::AutoBroadcastSpec autob)
```
rkimballn1: pass by reference

src/ngraph/pass/implicit_broadcast_elimination.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace ngraph
+{
+    namespace pass
+    {
+        template <typename T>
+        NodeVector static explicit_broadcast(std::shared_ptr<T>& node)
+        {
+            if (node->get_autob().type == op::AutoBroadcastType::NONE)
+            {
+                return node->get_arguments();
```
rkimballn1: by having multiple returns you do not allow the compiler to use Return Value Optimization, forcing it to make a copy when returning NodeVector.

src/ngraph/serializer.cpp
```diff
@@ -1653,7 +1688,11 @@ static json write(const Node& n, bool binary_constant_data)
     }
     case OP_TYPEID::Acos: { break;
     }
-    case OP_TYPEID::Add: { break;
+    case OP_TYPEID::Add:
+    {
+        auto tmp = dynamic_cast<const op::Add*>(&n);
+        node["autob"] = write_auto_broadcast(tmp->get_autob());
```
rkimballn1: Would be better if you only wrote out autob if it is not the default value.

src/ngraph/op/and.cpp
```diff
@@ -19,14 +19,16 @@
 using namespace std;
 using namespace ngraph;
 
-op::And::And(const shared_ptr<Node>& arg0, const shared_ptr<Node>& arg1)
-    : BinaryElementwiseLogical("And", arg0, arg1)
+op::And::And(const shared_ptr<Node>& arg0,
+             const shared_ptr<Node>& arg1,
+             const AutoBroadcastSpec& autob)
+    : BinaryElementwiseLogical("And", arg0, arg1, autob)
```
nmostafa: Do we need to pass this to every constructor ? I assume autob mode should be fixed for all nodes during at least on ngraph invocation. Can't we instead implicitly read it from some configuration class/global  ?

src/ngraph/op/util/attr_types.hpp
```diff
@@ -49,5 +51,33 @@ namespace ngraph
...
+            {
+            }
+            AutoBroadcastSpec(AutoBroadcastType t, size_t s)
+                : type(t)
+                , axis(s)
+            {
+            }
+
+            AutoBroadcastType type;
+            size_t axis;
```
nmostafa: What is axis ? Couldn't find a use anywhere.

src/ngraph/pass/implicit_broadcast_elimination.hpp
jbobba: This is only used on nodes with two arguments. As such copying might not be too bad. Do you have an alternative in mind?

src/ngraph/op/and.cpp
jbobba: Provide a way to change the default constructor of AutoBroadcastSpec?

aprocter: It might make sense to have a global that influences the constructors' behavior for the sake of convenience, but I still think we'll need to store `autob` and respect that stored value, for things like deserialization.

src/ngraph/op/util/attr_types.hpp
jbobba: Used to indicate the start of alignment, I added a comment.

src/ngraph/pass/implicit_broadcast_elimination.hpp
rkimballn1: at the top define `NodeVector rc;`
instead of returns set `rc = <whatever>;`
then at the bottom `return rc;`
One return so RVO works. It is a good general rule to follow to only have one return per function if you  are returning something non trivial

src/ngraph/op/and.cpp
nmostafa: Basically, is it possible that nodes may have different auto-broadcast modes ? If not, then makes more sense to keep it as a global.

aprocter: Well, "possible" is up to us, right? ;)

Consider, for example, a builder (either one that's part of Core and meant to be reused across bridges, or one that's used internally by the bridge) that wants to use autobroadcast internally. How would that interact with the global?

PR #2942, opened 2019-05-15
Cleanup how compile flags are set and used by nGraph and external projects.

silee2:
Set C++11 through CMake and pass it down to external projects.
Prefer CMake variables such as CMAKE_POSITION_INDEPENDENT_CODE and
CMAKE_CXX_STANDARD instead of explicitly setting compiler dependent
flags.
Create json compilation database for external projects.
CMAKE_CXX_FLAGS is used as common global options for nGraph and external
projects.
add_compile_options() is used for local options for current and sub
directories.
add_definitions() is used for setting definitions for current and sub
directories.
Note: Global options are not passed down to some external projects.
Note: mkl-dnn resets CMAKE_CXX_FLAGS internally.
Note: MLSL is not CMake based.
Note: Eigen and json are header only library.


cmake/external_gtest.cmake
```diff
@@ -24,22 +24,9 @@ include(ExternalProject)
 SET(GTEST_GIT_REPO_URL https://github.com/google/googletest.git)
 SET(GTEST_GIT_LABEL release-1.8.1)
 
-if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    if (DEFINED NGRAPH_USE_CXX_ABI)
-        set(COMPILE_FLAGS "${COMPILE_FLAGS} -D_GLIBCXX_USE_CXX11_ABI=${NGRAPH_USE_CXX_ABI}")
```
rkimballn1: Don't we still need to set the ABI flag somewhere?

silee2: https://github.com/NervanaSystems/ngraph/blob/master/CMakeLists.txt#L318-L323 already sets it in CMAKE_CXX_FLAGS

silee2: And https://github.com/NervanaSystems/ngraph/pull/2942/files/056b0e2eeff308677e9242b4094a6e851f824b58#diff-2abb742095220ead92ffaef7701a3ec3R51 passes CMAKE_CXX_FLAGS down to external_gtest

silee2:
Found several issues. Adding WIP until they get resolved.

PR #2944, opened 2019-05-16
Workaround for #pragma GCC diagnostic error.

silee2:


src/ngraph/op/constant.cpp
```diff
@@ -53,9 +53,6 @@ vector<string> op::Constant::get_value_strings() const
 {
     vector<string> rc;
```
diyessi: Let's add a comment here and below to indicate that compiler flags are set via cmake so that three months from now it's not mysterious what's going on.

silee2:
closing since #2945 seems like a better solution

PR #2949, opened 2019-05-17
Add more infrastructure for specialization of cloned graphs

aprocter:
There are ~four~ five main changes in this PR:

1. Update `specialize_function` (renamed from `specialize_shapes`) to allow parameter _values_ to be substituted in the cloned function.
2. Add a `DynElimination` pass to convert `Dyn*` ops to static ops when their inputs are "constant enough" to do so. (This can usually only be done after `specialize_function` and a round of `ConstantFolding`, but the pass tries its best anyway.)
3. Update the wrapper backend to substitute _shape-relevant_ parameters in the cloned functions with constants, and run constant folding and `DynElimination` on the clone.
4. A handful of updates to `ConstantFolding` to handle `Convert` (which just happened to be the minimum I needed for my "hello world" TensorFlow example running on an experimental branch that uses the dynamic tensor APIs), and, as long as I'm here, `ShapeOf`.
5. Implements shape propagation through Transpose when the permutation arg is a `Constant`.

This gets us far enough that we can call the backend wrapper functionally complete (read: type 2 dynamic shapes should work), modulo additional smarts we will need to add to `DynElimination` and `ConstantFolding`.


aprocter:
Looks like I screwed something up in `DynSlice` shape prop, which is odd. Investigating.

src/ngraph/pass/dyn_elimination.cpp
```diff
@@ -0,0 +1,76 @@
...
+        if (perm_arg->get_element_type() != element::i64 ||
+            perm_arg->get_output_partial_shape(0).is_dynamic() ||
+            perm_arg->get_output_shape(0).size() != 1)
+        {
+            return false;
+        }
+
+        auto perm = perm_arg->get_axis_vector_val();
+
+        auto output_shape = ngraph::apply_permutation(data_shape, perm);
```
jbobba: Can we move this logic to `transpose_op->validate_and_infer_types()` so transpose can finalize its shape when the order input is a constant? 

We will also need a check in apply_permutation for valid permutations.

aprocter: Yes and yes.

aprocter: @jbobba I just pushed 08c7dc7 which should address these, please have a look

src/ngraph/pass/constant_folding.cpp
```diff
@@ -763,3 +766,216 @@ void pass::ConstantFolding::construct_constant_quantize()
...
+        constant->get_vector<TI>().data(), out_vec.data(), shape_size(out_shape));
+
+    return make_shared<op::Constant>(output_element_type, out_shape, out_vec);
+}
+
+template <typename TI>
+shared_ptr<op::Constant> fold_constant_convert_helper0(shared_ptr<op::Constant> constant,
+                                                       const element::Type& output_element_type)
+{
+    if (output_element_type == element::boolean)
```
rkimballn1: Use a switch statement here and add the #pragma stuff to make it a compile-time error if an enum is missing from the list.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -763,3 +766,216 @@ void pass::ConstantFolding::construct_constant_quantize()
...
+                                               const element::Type& output_element_type)
+{
+    auto& input_element_type = constant->get_output_element_type(0);
+
+    if (input_element_type == output_element_type)
+    {
+        return constant;
+    }
+
+    if (input_element_type == element::boolean)
```
rkimballn1: switch here too

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -79,30 +81,76 @@ bool runtime::dynamic::DynamicExecutable::call(
...
-        if (auto dynamic_tensor = std::dynamic_pointer_cast<runtime::dynamic::DynamicTensor>(input))
-        {
-            NGRAPH_CHECK(dynamic_tensor->has_storage());
-            arg_element_types.push_back(dynamic_tensor->get_wrapped_tensor()->get_element_type());
-            arg_shapes.push_back(dynamic_tensor->get_wrapped_tensor()->get_shape());
-            wrapped_inputs.push_back(dynamic_tensor->get_wrapped_tensor());
-        }
-        else
+        // We'll use AlignedBuffers to back the base pointers, storing them in this vector for RAII
+        // purposes. It seems we need to wrap them in unique_ptr rather than store them directly in
```
rkimballn1: Agreed, move should not be disabled. I will  fix.

aprocter: Adding a `TODO` to address once #2949 is merged.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -763,3 +766,176 @@ void pass::ConstantFolding::construct_constant_quantize()
...
+        constant->get_vector<TI>().data(), out_vec.data(), shape_size(out_shape));
+
+    return make_shared<op::Constant>(output_element_type, out_shape, out_vec);
+}
+
+template <typename TI>
+shared_ptr<op::Constant> fold_constant_convert_helper0(shared_ptr<op::Constant> constant,
+                                                       const element::Type& output_element_type)
+{
+#pragma GCC diagnostic push
```
silee2: GCC 4.8.X has some missed with handling some pragmas. Workaround shown in https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/constant.hpp#L262-L266 is needed.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -763,3 +766,176 @@ void pass::ConstantFolding::construct_constant_quantize()
...
+                                               const element::Type& output_element_type)
+{
+    auto& input_element_type = constant->get_output_element_type(0);
+
+    if (input_element_type == output_element_type)
+    {
+        return constant;
+    }
+
+#pragma GCC diagnostic push
```
silee2: GCC 4.8.X has some missed with handling some pragmas. Workaround shown in https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/constant.hpp#L262-L266 is needed.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -763,3 +766,176 @@ void pass::ConstantFolding::construct_constant_quantize()
...
+    case element::Type_t::u8:
+        return fold_constant_convert_helper0<uint8_t>(constant, output_element_type);
+    case element::Type_t::u16:
+        return fold_constant_convert_helper0<uint16_t>(constant, output_element_type);
+    case element::Type_t::u32:
+        return fold_constant_convert_helper0<uint32_t>(constant, output_element_type);
+    case element::Type_t::u64:
+        return fold_constant_convert_helper0<uint64_t>(constant, output_element_type);
+    }
+#pragma GCC diagnostic pop
```
silee2: GCC 4.8.X has some missed with handling some pragmas. Workaround shown in https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/constant.hpp#L311-L313 is needed.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -763,3 +766,176 @@ void pass::ConstantFolding::construct_constant_quantize()
...
+    case element::Type_t::u8:
+        return fold_constant_convert_helper1<TI, uint8_t>(constant, output_element_type);
+    case element::Type_t::u16:
+        return fold_constant_convert_helper1<TI, uint16_t>(constant, output_element_type);
+    case element::Type_t::u32:
+        return fold_constant_convert_helper1<TI, uint32_t>(constant, output_element_type);
+    case element::Type_t::u64:
+        return fold_constant_convert_helper1<TI, uint64_t>(constant, output_element_type);
+    }
+#pragma GCC diagnostic pop
```
silee2: GCC 4.8.X has some missed with handling some pragmas. Workaround shown in https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/constant.hpp#L311-L313 is needed.

jbobba:
@ayzhuang can you look at the ConstantFolding changes and if it would break the CPU constant folding pass.

ayzhuang:
> @ayzhuang can you look at the ConstantFolding changes and if it would break the CPU constant folding pass.

Just like constant folding for dequantize and quantize, these two constant folding are not using the map provided by CPU backend, it would not break the CPU constant folding. If we want, we can add those to the map later.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -763,3 +766,184 @@ void pass::ConstantFolding::construct_constant_quantize()
     this->add_matcher(
         quantize_matcher, constant_quantize_callback, PassProperty::REQUIRE_STATIC_SHAPE);
 }
+
+template <typename TI, typename TO>
```
diyessi: Short descriptions of 0 and 1 needed.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -763,3 +766,190 @@ void pass::ConstantFolding::construct_constant_quantize()
...
+        replace_node(
+            m.get_match_root(),
+            fold_constant_convert(constant_match, convert_match->get_output_element_type(0)));
+        return true;
+    };
+
+    auto convert_matcher =
+        make_shared<pattern::Matcher>(convert_op, "ConstantFolding.ConstantConvert");
+    this->add_matcher(
+        convert_matcher, constant_convert_callback, {PassProperty::CHANGE_DYNAMIC_STATE});
```
louisfeng: Does this callback actually change dynamic state? It seems only converting a constant to a particular type?

aprocter: Good point! Will fix.

aprocter: As I think about it, it looks like the one in `DynElimination` for `Transpose` was not necessary either. It will replace the `Transpose` only under circumstances where the shape was already statically known. So I've removed that one. I've also removed the `CHANGE_DYNAMIC_STATE` for `ShapeOf`, because the shape-of-shape will be static in all cases where we replace. If you could double-check my reasoning I would be grateful :)

src/ngraph/pass/constant_folding.cpp
```diff
@@ -763,3 +766,188 @@ void pass::ConstantFolding::construct_constant_quantize()
...
+
+        replace_node(
+            m.get_match_root(),
+            fold_constant_convert(constant_match, convert_match->get_output_element_type(0)));
+        return true;
+    };
+
+    auto convert_matcher =
+        make_shared<pattern::Matcher>(convert_op, "ConstantFolding.ConstantConvert");
+    this->add_matcher(convert_matcher, constant_convert_callback, {});
```
louisfeng: Could you use constant `all_pass_property_off` for the matcher property parameter? While `{}` works, I hope this give better clarity.

PR #2950, opened 2019-05-17
Fixes incorrect and extra provenance tags

tachyon77:
In ngraph::replace_node(), we propagate provenance tags from all the going-away nodes to the node that replaces them. There was a bug in computing the set of nodes that should to be considered for this tag propagation.

This fix finds a better terminating condition for the traversal that finds all the relevant nodes from where tags are collected.

The idea is, we should only go as far as the arguments which are common between two sub-graphs. The first sub-graph is the graph rooted at the replacement node and includes its arguments recursively. The other sub-graph is the graph rooted at target node and includes its arguments recursively as well. Somewhere up the argument hierarchy these two sub-graphs share a set of common ancestors. And that set contains the terminating nodes for the provenance traversal.


src/ngraph/graph_util.hpp
```diff
@@ -75,6 +75,8 @@ namespace ngraph
 
     void replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> replacement);
 
+    NodeVector find_common_args(std::shared_ptr<Node> target, std::shared_ptr<Node> replacement);
```
aprocter: Minor thing: these parameter names should be changed, unless I'm misunderstanding the purpose of the function (in which case the function should probably be renamed).

tachyon77: I think you are not misunderstanding. I agree the parameter names should be more generic.

PR #2953, opened 2019-05-19
Add ArgMin/Max to zero dim elimination

rkimballn1:
Jira NGCORE-239

Please check that I got the +-inf correct for the correct operation.


src/ngraph/op/argmax.hpp
```diff
@@ -42,6 +42,13 @@ namespace ngraph
 
             virtual std::shared_ptr<Node>
                 copy_with_new_args(const NodeVector& new_args) const override;
+
+            /// \return The default value for op. Used for zero dimension tensor elimination.
+            virtual std::shared_ptr<Node> get_default_value() const override
+            {
+                auto type = get_input_element_type(0);
+                return ngraph::make_constant_from_string("-INFINITY", type, get_shape());
```
aprocter: It looks like `make_constant_from_string` ultimately bottoms out in `strtod` so I think `INFINITY` and `-INFINITY` should be okay.

aprocter: Oh wait, that wasn't your question... uh... hang on

src/ngraph/op/argmax.hpp
```diff
@@ -42,6 +42,13 @@ namespace ngraph
 
             virtual std::shared_ptr<Node>
                 copy_with_new_args(const NodeVector& new_args) const override;
+
+            /// \return The default value for op. Used for zero dimension tensor elimination.
+            virtual std::shared_ptr<Node> get_default_value() const override
+            {
+                auto type = get_input_element_type(0);
```
aprocter: This should be `m_index_element_type`.

src/ngraph/op/argmax.hpp
aprocter: I guess there are two cases here:

1. The reduction axis does not have size zero. Then it doesn't matter what we put here, because the output tensor will have zero elements.
2. The reduction axis has size zero. Then I'm actually not sure what we should put! TensorFlow calls it an error.

For now let's fill it with zero, but we should probably add a check to `IndexReduction` to make sure the reduced axis does not have size zero.

src/ngraph/op/argmax.hpp
aprocter: Same for ArgMin, btw.

src/ngraph/op/argmax.hpp
aprocter: Same for ArgMin. (Whatever values we put here are meaningless anyway.)

diyessi:
#3022 fixes the problem.

PR #2954, opened 2019-05-19
update visualize tree file extenstions and output formats

rkimballn1:
The output format is determined the file extension, so test.pdf will generate a PDF file and test.svg will generate an SVG file.
Removed the environment variable `NGRAPH_VISUALIZE_TREE_OUTPUT_FORMAT` as it could lead to wacky results like test.pdf actually being a PNG file.


aprocter:
Could you double check to make sure that `NGRAPH_ENABLE_VISUALIZE_TRACING=1` only generates `.dot` files and does not try to render them? If not, we need some way to enable that behavior.

PR #2955, opened 2019-05-19
Add save/load API to runtime

rkimballn1:
Jira NGCORE-452

Added `save` to runtime::Executable
Added `load` to runtime::Backend

Added save/load support to INTERPRETER along with a unit test to exercise the new API.

The default is that a backend does not support save or load and will return error if the API is used.


src/ngraph/runtime/executable.cpp
```diff
@@ -118,3 +118,8 @@ vector<runtime::PerformanceCounter> runtime::Executable::get_performance_data()
 {
     return vector<PerformanceCounter>();
 }
+
+bool runtime::Executable::save(std::ostream& output_stream)
+{
+    return false;
```
avijit-nervana: Consider adding some error message or make it a pure virtual.

src/ngraph/runtime/backend.cpp
```diff
@@ -85,3 +85,8 @@ bool runtime::Backend::is_supported_property(const Property prop) const
 void runtime::Backend::remove_compiled_function(std::shared_ptr<Executable> exec)
 {
 }
+
+std::shared_ptr<runtime::Executable> runtime::Backend::load(istream& input_stream)
+{
```
avijit-nervana: Add some error message or better - make it a pure virtual. This will help get errors early as opposed to silently failing the function.

src/ngraph/runtime/backend.hpp
```diff
@@ -109,6 +109,11 @@ class ngraph::runtime::Backend
                                                 ngraph::pass::PassConfig& pass_config,
                                                 bool enable_performance_data = false);
 
+    /// \brief Loads a previously saved Executable object from a stream.
+    /// \param input_stream the opened input stream containing the saved Executable
+    /// \returns A compiled function or nullptr on failure
+    virtual std::shared_ptr<Executable> load(std::istream& input_stream);
```
avijit-nervana: So far nGraph error communication has been through exception - should we stick to that? That way backends can propagate the error messages to the end user (via the framework).

src/ngraph/runtime/executable.hpp
```diff
@@ -69,6 +69,11 @@ class ngraph::runtime::Executable
     /// \returns an ngraph::ResultVector of all input parameters
     const ngraph::ResultVector& get_results() const;
 
+    /// \brief Save this compiled Executable to an output stream.
+    ///    Saved stream may be read with Backend::load
+    /// \returns true if successful, false otherwise.
+    virtual bool save(std::ostream& output_stream);
```
avijit-nervana: Consider throwing exception for error condition?

src/ngraph/runtime/executable.cpp
rkimballn1: I don't want to require that backends implement save/load. The error message should be handled by the caller. This call correctly returns a false indicating a failed save.

src/ngraph/runtime/backend.cpp
rkimballn1: Same as above

PR #2957, opened 2019-05-20
Change FusionType to enum class and use EnumMask

louisfeng:


louisfeng:
> Add description of API changes to `changes.md` and leave existing API deprecated but still functional for two releases (add a JIRA for killing the old version so we don't forget).

Fixed.

changes.md
```diff
@@ -2,6 +2,7 @@
 
 ## Passes
 * `LikeReplacement` pass must be run by all transformers.
+* `ngraph::pass::FusionType` is now a enum class. Constant values defined by `FusionType` are created for backward compatibility and will be removed in future releases.
```
indie: Probably should be "an enum class", but up to you if you want to change or not.

louisfeng: Thanks, fixed.

PR #2958, opened 2019-05-20
Fix Convert for boolean output type in CODEGEN.

ayzhuang:


src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -1299,8 +1299,16 @@ namespace ngraph
                 writer << "#pragma omp parallel for\n";
                 writer << "for (size_t i = 0; i < " << out[0].get_size() << "; i++)\n";
                 writer.block_begin();
-                writer << out[0].get_name() << "[i] = (" << result_element_type.c_type_string()
-                       << ")(" << args[0].get_name() << "[i]);\n";
+                writer << out[0].get_name() << "[i] = (";
+                if (result_element_type == element::boolean)
+                {
+                    writer << "bool";
```
gaurides: This would create 2 double quotes. One from line 1302/1311 and one here. is that ok?

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -1299,8 +1299,16 @@ namespace ngraph
...
+                }
+                else
+                {
+                    writer << result_element_type.c_type_string();
```
gaurides: 1. Shouldn't we also check/change in other places too where we use `result_element_type.c_type_string()`?

2. Actually, instead of `if/else` can we change in element_type.cpp -> get_type_info_map() ->
`{element::Type_t::boolean, TypeInfo(8, false, true, false, "char", "boolean")},` to return `bool`?

ayzhuang: We don't want to change the type from "char" to "bool" for nGraph boolean type. This change is similar to this PR for DEX:  https://github.com/NervanaSystems/ngraph/pull/2552.

src/ngraph/runtime/cpu/cpu_emitter.cpp
ayzhuang: "" is used to pass string to writer. "" in line 1302/1311 is not related to "" here.

gaurides: Thanks for the clarifications.

PR #2959, opened 2019-05-21
Allow NGRAPH_VISUALIZE_TREE_OUTPUT_SHAPES to output partial shapes

aprocter:


src/ngraph/pass/visualize_tree.cpp
```diff
@@ -287,6 +287,41 @@ string pass::VisualizeTree::add_attributes(shared_ptr<Node> node)
     return rc;
 }
 
+static std::string pretty_partial_shape(const PartialShape& shape)
```
aprocter: NB: we do have `operator<<` for this but in the context of a dot file I think it would look ugly.

PR #2962, opened 2019-05-21
Remove functions from cpu which were moved to core

gaurides:


gaurides:
1 test failed on onnx for accuracy reason. But that seems unrelated to changes here as they are basically just removing unused code.
It seems to fail for other PRs too.

PR #2966, opened 2019-05-22
[Fused Op] LSTMCell

arogowie-intel:
This PR introduces `LSTMCell` fused operator. It objective is to perform _cell_ computation within `LSTM` layer.  The equations and interface is based on [ONNX LSTM operator](https://github.com/onnx/onnx/blob/master/docs/Operators.md#LSTM). `LSTMCell` supports clipping and input-forget gates coupling.

The operator uses multiple inheritance in order to move common  xxxCell functionality to base class, which is `RNNCellBase`. `RNNCellBase` stores all common parameters for `LSTM`, `GRU` and `RNN` operators, as well as provides functionality to choose appropriate activation function, which are wrapped by `ActivationFunction` class. At this moment _sigmoid_, _tanh_ and _hardsigmoid_ activation functions are supported.

While working on this PR, one issue has come up. The `FusedOpDecomposition` pass is at this moment run twice on CPU backend, however on INTERPRETER only once. This resulted in errors on INTERPRETER, since `LSTMCell` uses another fused operator. So the issue is nested fused operators. As a workaround I've added second pass of `FusedOpDecomposition` to INTERPRETER. In following PR there would be fix running `FusedOpDecomposition` until no changes were made.

Can someone advise me how about `copy_with_new_args` and serializer in case I have multiple constructors and some optional inputs?


src/ngraph/op/fused/lstm_cell.cpp
```diff
@@ -0,0 +1,352 @@
...
+    , m_R{R}
+    , m_H_t{H_t}
+    , m_C_t{C_t}
+    , m_activation_f{get_activation_function(0)}
+    , m_activation_g{get_activation_function(1)}
+    , m_activation_h{get_activation_function(2)}
+    , m_input_forget{input_forget}
+{
+    // Split B onto Wb and Rb and add them.
+    NODE_VALIDATION_CHECK(this,
```
aprocter: All checks should be done in `pre_validate_and_infer_types()`

src/ngraph/op/util/reshape.hpp
```diff
@@ -16,6 +16,9 @@
```
aprocter: Not in scope for this PR, but while I'm thinking about it: this stuff should be moved to `builder`

src/ngraph/op/fused/lstm_cell.cpp
arogowie-intel: Unfortunately I couldn't do that this time. I'm checking here since I'm using this input to produce other value. However this is not the biggest problem. But in lines https://github.com/NervanaSystems/ngraph/pull/2966/files#diff-cebfe1ce12ae7674ed2d17869b5893beR171 and https://github.com/NervanaSystems/ngraph/pull/2966/files#diff-cebfe1ce12ae7674ed2d17869b5893beR181  I have to check whether I have valid data and generate those values and couldn't do that after `constructor_validate_and_infer_types`, since it calls `decompose_op` which would try to use uninitialized values otherwise.

src/ngraph/op/util/reshape.hpp
arogowie-intel: Will do that in following PR.

src/ngraph/op/fused/lstm_cell.cpp
arogowie-intel: Other solution would be to move those lines (171, 181) directly to `decompose_op`, but thought that it is cleaner to initialize those members in constructor...

aprocter: I think we probably will need to do it in `decompose_op`, unfortunately. We're moving towards taking validation checks out of constructors, so that revalidation after passes can be done.

src/ngraph/op/fused/lstm_cell.cpp
```diff
@@ -0,0 +1,352 @@
...
+                                             std::vector<float>(shape_size(i_t->get_shape()), 1.f)),
+                i_t);
+    }
+    else
+    {
+        // f(Xt*(Wf^T) + Ht-1*(Rf^T) + Pf (.) Ct-1 + Wbf + Rbf)
+        f_t = m_activation_f(clip(add(f_t, mul(p_f, m_C_t)), get_clip()));
+    }
+    // ft (.) Ct-1 + it (.) ct
+    auto C = add(mul(f_t, m_C_t), mul(i_t, m_activation_g(clip(c_t, get_clip()))));
```
tsocha: Please consider fbroadcast data to the biggest one first and then use standard operators +-* for all of these equations.

src/ngraph/runtime/gpu/gpu_compiled_function.cpp
```diff
@@ -172,6 +172,7 @@ void runtime::gpu::GPUCompiledFunction::compile()
     pass_manager.register_pass<runtime::gpu::pass::BatchNormCache>();
     pass_manager.register_pass<ngraph::pass::LikeReplacement>();
     pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>();
+    pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>();
```
tsocha: Please comment why we need to double this line.

src/ngraph/runtime/intelgpu/intelgpu_backend.cpp
```diff
@@ -424,6 +425,8 @@ shared_ptr<runtime::Executable>
     {
         pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>(
             IntelGPUBackend::is_supported_impl);
+        pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>(
```
tsocha: Please comment why we need to double these lines.

src/ngraph/runtime/interpreter/int_executable.cpp
```diff
@@ -43,6 +43,9 @@ runtime::interpreter::INTExecutable::INTExecutable(const shared_ptr<Function>& f
     pass::Manager pass_manager;
     pass_manager.register_pass<pass::LikeReplacement>();
     pass_manager.register_pass<pass::FusedOpDecomposition>();
+    // Run this pass for the second time since, some fused operators like LSTMCell may use
+    // other fused operators inside.
```
tsocha: I know it is out of scope of this PR, but what if one of operators will be fused in 3 or more levels?
This should be resolved better than just copy lines of code(in 3 places for now).
It can be merged for now but someone should back to this problem later on.

src/ngraph/op/fused/lstm_cell.cpp
```diff
@@ -0,0 +1,352 @@
...
+    return {make_shared<op::Subtract>(args.at(0), args.at(1))};
+}
+
+static shared_ptr<Node> mul(const shared_ptr<Node>& lhs, const shared_ptr<Node>& rhs)
+{
+    auto args = op::numpy_style_broadcast({lhs, rhs});
+    return {make_shared<op::Multiply>(args.at(0), args.at(1))};
+}
+
+static shared_ptr<Node> clip(const shared_ptr<Node>& data, float threshold)
```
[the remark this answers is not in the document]
arogowie-intel: Will do it in following PR for RNN and GRU cells.

src/ngraph/op/fused/lstm_cell.cpp
arogowie-intel: I think I will wait until #2936 will be merged. When broadcasting will happen automatically I'll change it to operators.

src/ngraph/op/util/activation_functions.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+shared_ptr<Node> op::ActivationFunction::operator()(const shared_ptr<Node>& arg) const
+{
+    return m_function(arg, m_alpha, m_beta);
+}
+
+op::ActivationFunction op::get_activation_func_by_name(const string& func_name)
+{
+    using ActivationFunctionMap = unordered_map<string, op::ActivationFunction>;
+    using namespace placeholders;
```
tomdol: Seems that nothing from this namespace is used in the scope of this function.

src/ngraph/op/fused/lstm_cell.hpp
```diff
@@ -0,0 +1,199 @@
...
+            virtual NodeVector decompose_op() const override;
+            virtual std::shared_ptr<Node>
+                copy_with_new_args(const NodeVector& new_args) const override;
+
+            bool get_input_forget() const { return m_input_forget; }
+        private:
+            ///
+            /// \brief      The input data tensor. Shape: [batch_size, input_size].
+            ///
+            std::shared_ptr<Node> m_X;
```
aprocter: I don't think it will work to store the node pointers here. If we do a rewrite on the graph prior to fused op decomposition (which is allowed), replace_output will not know how to rewire these pointers. You'll need to refer to them according to input index.

aprocter: Same comment for m_W, m_R, m_H_t, m_C_t, m_p_iof, and m_bias.

src/ngraph/op/util/activation_functions.cpp
arogowie-intel: Eagle-eye :) Leftover after changes.

test/backend_fused_op.in.cpp
```diff
@@ -971,3 +971,315 @@ NGRAPH_TEST(${BACKEND_NAME}, split_var_len_parts)
...
+    const auto X = make_shared<op::Parameter>(element::f32, Shape{batch_size, input_size});
+    const auto W =
+        make_shared<op::Parameter>(element::f32, Shape{gates_count * hidden_size, input_size});
+    const auto R =
+        make_shared<op::Parameter>(element::f32, Shape{gates_count * hidden_size, hidden_size});
+    const auto H_t = make_shared<op::Parameter>(element::f32, Shape{batch_size, hidden_size});
+    const auto C_t = make_shared<op::Parameter>(element::f32, Shape{batch_size, hidden_size});
+
+    const auto lstm_cell = make_shared<op::LSTMCell>(X, W, R, H_t, C_t, hidden_size);
+    auto function =
```
postrational: Please don't call `decompose_op` in tests. Let's test the fused op, if the backend has direct support for it.

src/ngraph/op/fused/lstm_cell.hpp
```diff
@@ -0,0 +1,170 @@
...
+            /// \param[in]  R                 The recurrence weight tensor with shape:
+            ///                               [4*hidden_size, hidden_size].
+            /// \param[in]  H_t               The hidden state tensor at current time step with
+            ///                               shape: [batch_size, hidden_size].
+            /// \param[in]  C_t               The cell state tensor at current time step with
+            ///                               shape: [batch_size, hidden_size].
+            /// \param[in]  hidden_size       The number of hidden units for recurrent cell.
+            /// \param[in]  B                 The bias tensor for input gate with shape: [8*hidden_size].
+            /// \param[in]  P                 The weight tensor for peepholes with shape:
+            ///                               [3*hidde_size] - 3 equals to only iof gates.
```
diyessi: `hidden`

src/ngraph/op/fused/rnn_cell_base.hpp
```diff
@@ -0,0 +1,78 @@
+//*****************************************************************************
```
jbobba: Move this to ngraph/op/util/

src/ngraph/runtime/gpu/gpu_compiled_function.cpp
```diff
@@ -173,6 +173,9 @@ void runtime::gpu::GPUCompiledFunction::compile()
     pass_manager.register_pass<runtime::gpu::pass::BatchNormCache>();
     pass_manager.register_pass<ngraph::pass::LikeReplacement>();
     pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>();
+    // Run this pass for the second time since, some fused operators like LSTMCell may use
+    // other fused operators inside.
+    pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>();
```
jbobba: I would change `FusedOpDecomposition` to run repeatedly until there are no more unsupported `FusedOps` in the graph but that can happen in a separate PR.

arogowie-intel: Yeah, I've plans for this, but in a separate PR.

src/ngraph/op/fused/lstm_cell.hpp
```diff
@@ -0,0 +1,170 @@
...
+    {
+        ///
+        /// \brief      Class for lstm cell node.
+        ///
+        /// \note       It follows notation and equations defined as in ONNX standard:
+        ///             https://github.com/onnx/onnx/blob/master/docs/Operators.md#LSTM
+        ///
+        ///             Note this class represents only single *cell* and not whole LSTM *layer*.
+        ///
+        class LSTMCell : public util::FusedOp, public util::RNNCellBase
```
pruthviIntel: @arogowie-intel , how do we define the whole lstm layer?, isnt most of the framwork have timesetp `{t, batch_size, input_size}` while defining for the inputs for lstm layer?

arogowie-intel: From the nGraph perspective the whole LSTM layer isn't yet prepared (AFAIK). But we have defined ONNX version [here](https://github.com/NervanaSystems/ngraph/blob/arogowiec/fused_lstm_cell/src/ngraph/frontend/onnx_import/op/lstm.cpp). Please take a look at it's [inputs definitions](https://github.com/onnx/onnx/blob/master/docs/Operators.md#inputs-3---8). The first data input is indeed of shape: `[seq_length, batch_size, input_size]`.  And essentialy what we are doing [there](https://github.com/NervanaSystems/ngraph/blob/arogowiec/fused_lstm_cell/src/ngraph/frontend/onnx_import/op/lstm.cpp#L255-L358) is looping over `seq_length` axis of input and call `LSTMCell` on each batch.

pruthviIntel: posted more discussion on #ngraph-lstm channel in slack.

PR #2968, opened 2019-05-22
Switch some get_inputs uses to use the newer inputs

diyessi:
Started down the path of removing deprecated uses of `get_inputs()` before switching to another approach for what I was doing; didn't want to lose the work.


src/ngraph/runtime/cpu/builder/convolution.cpp
```diff
@@ -255,7 +255,7 @@ namespace ngraph
                 auto arg2_buffer_index = external_function->get_buffer_index(args[2].get_name());
                 auto arg3_buffer_index = external_function->get_buffer_index(args[3].get_name());
                 auto out_buffer_index = external_function->get_buffer_index(out[0].get_name());
-                size_t arg3_size = node->get_inputs()[3].get_tensor().size();
+                size_t arg3_size = node->inputs()[3].get_tensor().size();
```
aprocter:
```suggestion
                size_t arg3_size = node->input(3).get_tensor().size();
```

aprocter: Similar suggestion in many other spots

PR #2969, opened 2019-05-22
Move zero padded conv fusions from CPUFusion to CoreFusion.

ayzhuang:
Fix onnxruntime resnet50 performance degradation.


jbobba:
If possible, please move the zero pad tests from tests/cpu_fusion.cpp to core_fusion.cpp

PR #2977, opened 2019-05-24
Add CPU backend support for Tile op using Eigen kernel.

ayzhuang:


src/ngraph/op/experimental/tile.cpp
```diff
@@ -48,12 +48,22 @@ void op::Tile::validate_and_infer_types()
 
     if (arg_rank.is_static())
     {
-        // Repeats shapes should be of form {arg_rank} or dynamic
-        NODE_VALIDATION_CHECK(this,
-                              repeats_shape.compatible(PartialShape{arg_rank}),
-                              "Arg and padding below ranks mismatch");
-
-        output_rank = arg_rank;
+        if (size_t(arg_rank) == 0)
```
aprocter: Is there any reason for adding this capability? TF doesn't seem to allow this kind of usage. Not sure about others.

ayzhuang: No particular reason. Can remove if it's not needed.

aprocter: Let's go ahead and remove it for now.

PR #2979, opened 2019-05-26
Remove tensor offset from tensor read/write calls because it was never used

rkimballn1:
Tensor::read(void*, offset, size) is deprecated
Tensor::read(void*, size) added

Tensor::write(void*, offset, size) is deprecated
Tensor::write(void*, size) added


rkimballn1:
@avijit-nervana I added the old three-parameter version of the read/write calls back and deprecated.

src/ngraph/deprecated.hpp
```diff
@@ -31,6 +31,8 @@
 //
 #ifdef NGRAPH_DEPRECATED_ENABLE
 #define NGRAPH_DEPRECATED(msg) __attribute__((deprecated((msg))))
+#define NGRAPH_DEPRECATED_DOC /// \deprecated
```
diyessi: I think these crept in from another PR.

rkimballn1: Actually no, look down at runtime/tensor.hpp and you will see where they are used. It is needed because I added a doc \deprecated tag and that does not work if you don't use these changes.

PR #2983, opened 2019-05-27
[Fused] RNN and  GRU cells

arogowie-intel:
Add RNN and GRU cells as fused operators.


src/ngraph/op/fused/rnn_cell.cpp
```diff
@@ -0,0 +1,255 @@
...
+    //     bias = get_argument(4);
+    // }
+    // else
+    // {
+    //     // As default bias is all zeros, thus just initialize it with appropriate shape and zeros.
+    //     bias = op::Constant::create(input(0).get_element_type(),
+    //                                 Shape{2 * get_hidden_size()},
+    //                                 vector<float>(2 * get_hidden_size(), 0.f));
+    // }
+    // return bias;
```
tsocha: Please remove code which is commented out.

src/ngraph/op/fused/rnn_cell.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+            /// brief Add and initialize bias input to all zeros.
+            void add_default_bias_input();
+
+            ///
+            /// \brief The Activation function f.
+            ///
+            util::ActivationFunction m_activation_f;
+
+            static constexpr std::size_t m_gates_count{1};
```
aprocter: Proper prefix for a static var is `s_`, not `m_`.

src/ngraph/op/fused/gru_cell.cpp
```diff
@@ -0,0 +1,316 @@
...
+}
+
+void op::GRUCell::pre_validate_and_infer_types()
+{
+    const auto& x_pshape = get_input_partial_shape(0);
+    const auto& w_pshape = get_input_partial_shape(1);
+    const auto& r_pshape = get_input_partial_shape(2);
+    const auto& ht_pshape = get_input_partial_shape(3);
+
+    NODE_VALIDATION_CHECK(this,
```
aprocter: Instead of throwing an error here, can't we just set the output shapes to be dynamic?

src/ngraph/op/fused/gru_cell.hpp
```diff
@@ -0,0 +1,157 @@
...
+            ///
+            /// \brief The Activation function f.
+            ///
+            util::ActivationFunction m_activation_f;
+            ///
+            /// \brief The Activation function g.
+            ///
+            util::ActivationFunction m_activation_g;
+
+            static constexpr std::size_t m_gates_count{3};
```
aprocter: Should be `s_`.

src/ngraph/op/util/rnn_cell_base.hpp
```diff
@@ -71,10 +73,48 @@ namespace ngraph
...
+                /// \param[in]  data   The input tensor for clipping.
+                ///
+                /// \return     Node with element-wise clip operation.
+                ///
+                std::shared_ptr<Node> clip(const std::shared_ptr<Node>& data) const;
 
             private:
-                std::size_t m_hidden_size = 0.f;
-                float m_clip = 0.f;
+                const std::size_t m_hidden_size = 0.f;
```
aprocter: Is this used anywhere?

src/ngraph/op/util/rnn_cell_base.hpp
```diff
@@ -71,10 +73,48 @@ namespace ngraph
...
+                const float m_clip = 0.f;
```
aprocter: Should this be `const` or `static const`? (Is it possible for it to be initialized differently elsewhere?)

src/ngraph/op/util/rnn_cell_base.hpp
arogowie-intel: Yup, many times through this function [`get_hidden_size`](https://github.com/NervanaSystems/ngraph/pull/2983/files/0f8c8bf7430fd8ff2c09a74e993a8e307439c86b#diff-ca2772c2c3a02688d6215f75ff2da191R59). I.e [here](https://github.com/NervanaSystems/ngraph/pull/2983/files/0f8c8bf7430fd8ff2c09a74e993a8e307439c86b#diff-2ca43c9b3520dc9f4f25097d4351c299R219)

src/ngraph/op/util/rnn_cell_base.hpp
arogowie-intel: Sorry, maybe this default member initialization is confusing. I'll get rid of it. This should be only `const` - since it shouldn't change during op lifetime (actually we may let them change, however then we would have to ensure that the fused op subgraph is rebuild) and these are an individual object property. One `xxxCell` node instance may differ from the another. Since each `xxxCell` node shares these attributes I've decided to put them into common base class.

PR #2984, opened 2019-05-28
Move reshape functions from utils to builder.

arogowie-intel:


src/ngraph/builder/reshape.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include "ngraph/node.hpp"
+#include "ngraph/shape.hpp"
+
+namespace ngraph
+{
+    namespace builder
+    {
+        /// \brief      Change shape of input tensor.
+        ///
+        /// \param[in]  node   The node which shape will be used as input to Reshape.
```
indie: The node _containing_ the shape to be used as input to `Reshape`?

src/ngraph/builder/reshape.hpp
```diff
@@ -0,0 +1,63 @@
...
+        ///
+        /// \return: New node with reversed dimensions.
+        std::shared_ptr<Node> transpose(const std::shared_ptr<Node>& node);
+
+        /// \brief Flatten the input tensor into a 2D matrix.
+        ///
+        /// \param node The tensor to be flattened.
+        /// \param axis The axis dividing shape.
+        ///
+        /// \return The new node being a 2D matrix representing flattened input node.
```
indie: Might be better to keep tenses consistent, so this would be "The new node will be a 2D matrix representing the flattened input node" ... just a suggestion.

src/ngraph/builder/reshape.hpp
aprocter: Neither :) You're right, this phrasing is a bit awkward: something like `The node producing the tensor to be reshaped`?

src/ngraph/builder/reshape.hpp
aprocter: `The new node will have been going to be being`

PR #2986, opened 2019-05-28
Try to fix icc build problem

diyessi:


diyessi:
Not the cause of the problem.

PR #2992, opened 2019-05-29
Vasu/copyprop poc

vasuintel:
Removing the validation check requiring that elementwise args should have consistent shape. This is done as a temporary means to allow passing constants as arguments.


vasuintel:
I made a change in PaddlePaddle bridge to avoid broadcasting constant tensors before invoking elementwise ops. This condition causes the bridge change to fail. Hence I removed the check in ngraph.
I'll look at the failed unit tests and fix this change.

PR #2993, opened 2019-05-29
Much faster serialize/deserialize of broadcast constants

rkimballn1:
deserialize was 13 seconds, now 4
serialize was 170 seconds, now 1


src/ngraph/op/constant.cpp
```diff
@@ -49,6 +49,42 @@ op::Constant::~Constant()
 {
 }
 
+string op::Constant::convert_value_to_string(size_t index) const
```
aprocter: Can you refactor `get_value_strings()` to use this or did that not work out?

PR #2999, opened 2019-05-30
Handle 0 and -1 in shape inputs for dynamic reshape op

jbobba:


src/ngraph/op/experimental/dyn_reshape.cpp
```diff
@@ -47,9 +47,72 @@ void op::DynReshape::validate_and_infer_types()
...
+
+        if (!zero_dims && !negative_dims)
+        {
+            set_output_type(0, get_input_element_type(0), const_shape->get_shape_val());
+        }
+        else if (!get_input_partial_shape(0).is_static())
+        {
+            // We need input shape to determine output shape in the presence of
+            // zero and negative sized dims.
+            set_output_type(0, get_input_element_type(0), PartialShape::dynamic(output_rank));
```
aprocter: I think we can give some info by filling in `Dimension::dynamic()` for -1s, right? (We're not guaranteeing it, though, so we could do this some other time.)

src/ngraph/op/experimental/dyn_reshape.cpp
```diff
@@ -47,9 +47,72 @@ void op::DynReshape::validate_and_infer_types()
...
+            auto input_elements = shape_size(input_shape);
+
+            Shape output_shape(input_shape.size());
+            size_t output_elements = 1;
+            int negative_dim = -1;
+            for (size_t i = 0; i < input_shape.size(); i++)
+            {
+                if (out_shape_val[i] > 0)
+                {
+                    output_shape[i] = out_shape_val[i];
```
aprocter: This worries me a bit. In TF for example you can do this:

```
tf.reshape(np.ones(shape=(2,0,2)),shape=(0,2,2))
```

OTOH I believe PaddlePaddle wants exactly this behavior. Maybe we can make the zero-copying behavior optional?

aprocter: To be clear: by "optional" here I mean a boolean attribute on the node.

jbobba: ONNX seems to want the zero-copying behavior. "A dimension could also be 0, in which case the actual dimension value is unchanged (i.e. taken from the input tensor)." TF and PDPD dont seem to say anything about zero values. I'll add the flag but ONNX/Mxnet approach seems more useful.

aprocter: Thanks- it feels a bit ugly, but I do think it's necessary for TF.

PR #3003, opened 2019-05-31
Fix so you can build with NGRAPH_JSON_ENABLE=OFF

rkimballn1:


src/ngraph/serializer.hpp
```diff
@@ -62,3 +62,43 @@ namespace ngraph
...
     void set_serialize_output_shapes(bool enable);
 }
+
+#ifndef NGRAPH_JSON_ENABLE
+// Rather than making every reference to the serializer conditionally compile here we just
+// provide some null stubs to resolve link issues
+// The `inline` is so we don't get multiple definitions of function
+std::string inline ngraph::serialize(std::shared_ptr<ngraph::Function> func, size_t indent)
+{
+    return "";
```
jbobba: Should we throw an error instead of silent failure?

python/test/ngraph/test_basic.py
```diff
@@ -62,11 +62,14 @@ def test_serialization():
...
-
-    assert serial_json[0]['name'] != ''
-    assert 10 == len(serial_json[0]['ops'])
+    try:
+        serialized = computation.serialize(2)
+        serial_json = json.loads(serialized)
+
+        assert serial_json[0]['name'] != ''
+        assert 10 == len(serial_json[0]['ops'])
+    except Exception:
```
postrational: Let's narrow this exception down to the type we would expect when `NGRAPH_JSON_ENABLE=OFF`

diyessi: Let's leave that as a follow-up. I added a JIRA.

PR #3005, opened 2019-05-31
Bob/hybrid multi

rkimballn1:


src/ngraph/runtime/hybrid/hybrid_util.cpp
```diff
@@ -223,10 +223,8 @@ void runtime::hybrid::rewrite_function(const shared_ptr<Function>& f,
...
                 auto sub_function = make_shared<Function>(cluster_outputs, cluster_inputs);
                 sub_function->set_placement(placement);
-                auto fc = make_shared<runtime::hybrid::op::FunctionCall>(function_call_outputs,
-                                                                         function_call_inputs,
-                                                                         *sub_function,
-                                                                         backend_list[placement]);
+                auto fc = make_shared<runtime::hybrid::op::FunctionCall>(
+                    cluster_outputs, function_call_inputs, *sub_function, backend_list[placement]);
                 fc->set_placement_index(0);
                 for (size_t i = 0; i < function_call_outputs.size(); i++)
```
csullivan: nit: To me it looks strange to loop over `function_call_outputs` -- the consumers of the cluster -- here. I believe this loop could be simpler if we did,
```C++
                for (size_t i = 0; i < cluster_outputs.size(); i++)
                {
                    // First add a GetOutputElement to the ith output of the FunctionCall
                    auto goe = make_shared<ngraph::op::GetOutputElement>(fc, i);
                    goe->set_placement_index(0);
                    auto old_source = cluster_outputs[i];
                    ngraph::replace_node(old_source, goe);
                }
```

rkimballn1: agreed

PR #3006, opened 2019-05-31
Fuse Dropout

gaurides:
Optimizations to improve BERT performance
1. Fuses dropout
2. Uses uniform distribution for RNG: this is faster than the current implementation of GenerateMask. Mainly to improve perf for BERT. thoughts/comments welcome


gaurides:
@ashokei I pushed a commit to fix the order of declaration. Can you please try again? Thanks!

ashokei:
> @ashokei I pushed a commit to fix the order of declaration. Can you please try again? Thanks!

compile issue is resolved.

src/ngraph/runtime/cpu/kernel/dropout.hpp
```diff
@@ -0,0 +1,85 @@
...
+                                      bool training,
+                                      const double value)
+                {
+                    if (training)
+                    {
+                        double dropout_prob = 1 - value;
+                        size_t nthr = 28, nelems = count;
+                        size_t chunk_size = (nelems + nthr - 1) / nthr;
+
+#pragma omp parallel num_threads(nthr)
```
silee2: I don't think you get any randomness in the mask that gets generated.
Here is my thinking.
Random number sequence is seeded with omp thread number. So the same thread will get the same random sequence every kernel invocation. The kernel is trying to parallelize mask generation but if some openmp affinity scheduling will result in identical combined random sequence for mask generation every invocation. This sequence is then passed to uniform int distribution. The result would be the same 0, 1 mask generated every kernel invocation.

src/ngraph/runtime/cpu/kernel/dropout.hpp
```diff
@@ -0,0 +1,85 @@
...
+                                      T* out0,
+                                      T* out1_mask,
+                                      size_t count,
+                                      bool training,
+                                      const double value)
+                {
+                    if (training)
+                    {
+                        double dropout_prob = 1 - value;
+                        size_t nthr = 28, nelems = count;
```
silee2: Where is the magic number 28 coming from?

src/ngraph/runtime/cpu/kernel/dropout.hpp
gaurides: I see. let me fix it.

diyessi: The seed should be determined from the op. Maybe op seed plus i where i is the chunk the thread is working on. The idea is that if you set the seed to a value you should get the same random values each time you run so that you can reproduce results or debug.

src/ngraph/runtime/cpu/op/dropout.hpp
```diff
@@ -0,0 +1,47 @@
...
+    namespace op
+    {
+        class Dropout : public Op
+        {
+        public:
+            Dropout(const std::shared_ptr<Node>& input,
+                    const std::shared_ptr<Node>& gm_const,
+                    int seed,
+                    double value); // value = 1 - dropout_prob
+
```
diyessi: Please also add setters.

src/ngraph/runtime/cpu/kernel/dropout.hpp
gaurides: fixed the magic number in new commit

src/ngraph/runtime/cpu/kernel/dropout.hpp
gaurides: In new commit, getting the seed from the op. It will be independent of tid. Initializing each thread with same seed and then calling `msr.discard()` to advance the state to necessary state.
This way the parallel threads will generate same random numbers as though one would generate in serial.

src/ngraph/runtime/cpu/op/dropout.hpp
gaurides: I added the setters, but don't really see when we would need them as the values are initialized in ctor.

src/ngraph/runtime/cpu/op/dropout.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace ngraph
+{
+    namespace op
+    {
+        class Dropout : public Op
+        {
+        public:
+            Dropout(const std::shared_ptr<Node>& input,
+                    const std::shared_ptr<Node>& gm_const,
+                    int seed,
```
rkimballn1: Can we use `int32_t` for seed instead of `int`?

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -911,6 +917,76 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_add()
...
+void ngraph::runtime::cpu::pass::CPUFusion::construct_dropout()
+{
+    Shape shape{1, 1, 2, 2};
+    auto x = std::make_shared<pattern::op::Label>(element::f32, shape);
+    auto x_label = std::make_shared<pattern::op::Label>(x, nullptr, NodeVector{x});
+
+    int seed = 1234;
+    auto seed_label = std::make_shared<pattern::op::Label>(element::u32, Shape{0});
+
+    double value = 0.9; // some number
```
rkimballn1: This comment is not helpful.

src/ngraph/runtime/cpu/op/dropout.hpp
gaurides: yes, pushed in newer commit

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
gaurides: deleted.

src/ngraph/runtime/cpu/kernel/dropout.hpp
silee2: "This way the parallel threads will generate same random numbers as though one would generate in serial." -> This is true. But you would still get the same mask every time. vmsr[tid]'s state does not get updated. 
Check this mini example.
`
#include <vector>
#include <random>
#include <iostream>
#include <functional>

size_t nthr = 8;
size_t seed = 0;
size_t chunk_size = 10;
size_t rpt = 3;

std::function<void(size_t, size_t, size_t)> lrnd;

void use(const std::vector<std::minstd_rand> &vmsr, size_t i, size_t c, size_t nt)
//void use(std::vector<std::minstd_rand> vmsr, size_t i, size_t c, size_t nt)
{
        auto rnd = vmsr[i];
        for (size_t m = 0 ; m < c; m++)
        {
            std::cout << rnd() << " ";
        }
        //rnd.discard((nt - 1) * c);
        //vmsr[i] = rnd;
}

void setup()
{
    std::vector<std::minstd_rand> vmsr(nthr);
    for (size_t i = 0; i < nthr; i++)
    {
         vmsr[i].seed(seed);
         vmsr[i].discard(i * chunk_size);
    }
    auto myrnd = [&, vmsr] (size_t i, size_t c, size_t nt)
    {
        use(vmsr, i, c, nt);
    };
    lrnd = myrnd;
}

int main()
{
    setup();
    std::minstd_rand jmsr;
    jmsr.seed(0);
    for (size_t i = 0; i < rpt; i++)
    {
        std::cout << "Iter " << i << ": " << std::endl;
        for (size_t j = 0; j < chunk_size * nthr; j++)
        {
            std::cout << jmsr() << " ";
        }
        std::cout << std::endl;
    }
    std::cout << std::endl;
    for (size_t k = 0 ; k < rpt; k++)
    {
        std::cout << "Iter " << k << ": " << std::endl;
        for (size_t n = 0 ; n < nthr; n++)
             lrnd(n, chunk_size, nthr);
        std::cout << std::endl;
    }
    std::cout << std::endl;
    return 0;
}
`

gaurides: That's right. However, the frameworks, especially PDPD, are not maintaining a state either. So, if the framework passes the same seed, then we will get same mask.
In my implementation of dropout, I am trying to be close to this PDPD native implementation so it converges and performs like native PDPD BERT : https://github.com/NervanaSystems/ngraph-paddle/blob/14d88829b386c9f7601788c5539c08326dcbe2fe/paddle/fluid/operators/dropout_op.h#L58-L78
However, we also looked at other frameworks and they don't maintain state MxNet : https://github.com/apache/incubator-mxnet/blob/503c7509db67c43e0afe8aec0e006b0cacaf7128/src/operator/nn/dropout-inl.h#L90-L112
Tensorflow : https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/python/ops/nn_ops.py#L3044-L3048

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -3952,6 +3953,45 @@ namespace ngraph
...
+                }
+                size_t element_count = out[0].get_size();
+                size_t nthr = ngraph::runtime::cpu::executor::GetCPUExecutor().get_num_cores();
+                size_t chunk_size = (element_count + nthr - 1) / nthr;
+
+                writer.block_begin();
+                writer << "uint32_t seed = " << to_string(dropout->get_seed()) << ";\n";
+                writer << "double value = " << to_string(dropout->get_value()) << ";\n";
+
+                writer << "std::vector<std::minstd_rand> vmsr(" << to_string(nthr) << ");\n";
```
jbobba: this might lead to segfaults in codegen due to static variable initialization/destruction issues with MCJIT. Using states (similar to GenerateMask) and sticking this in there will solve the problem. For now, you could throw an unsupported exception here. Or change the kernel to make vmsr an optional input. If vmsr is provided, kernel uses it else it creates a new set of msrs internally and discards them.

gaurides: Ok, throwing an error for dropout codegen for now. Will work on support

PR #3011, opened 2019-06-03
[ONNX]  Add MatMulInteger op

postrational:


src/ngraph/builder/quantization/quantized_linear_matmul.hpp
```diff
@@ -37,6 +37,12 @@ namespace ngraph
 
             std::shared_ptr<Node> QuantizedLinearMatmulInteger(const std::shared_ptr<Node>& input0,
                                                                const std::shared_ptr<Node>& input1);
+
+            std::shared_ptr<Node>
```
adstraw: I still prefer a single builder with all inputs.

postrational: With the current solution I can avoid creating `Constant` nodes in the graph if ONNX doesn't provide these optional values.

src/ngraph/frontend/onnx_import/op/matmul_integer.cpp
```diff
@@ -0,0 +1,62 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include "op/matmul_integer.hpp"
+#include "ngraph/builder/make_constant.hpp"
+#include "ngraph/builder/quantization/quantized_linear_matmul.hpp"
+#include "ngraph/frontend/onnx_import/exceptions.hpp"
```
arogowie-intel: Unused header.

src/ngraph/frontend/onnx_import/op/matmul_integer.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+namespace ngraph
+{
+    namespace onnx_import
+    {
+        namespace op
+        {
+            namespace set_1
+            {
+                NodeVector matmul_integer(const Node& node)
```
arogowie-intel: This won't work for input tensors with rank > 2, since nGraph `Dot` doesn't work as NumPy's matmul. There should be a logic similar to this https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/frontend/onnx_import/op/matmul.cpp#L124 to handle higher order tensors.

test/onnx/onnx_import_quant.in.cpp
```diff
@@ -338,3 +338,102 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_qlinear_matmul_3d)
...
+    test_case.add_input(std::vector<uint8_t>{12}); // b_zero_point
+
+    test_case.add_expected_output({1, 1}, std::vector<int32_t>{-1}); // y
+    test_case.run();
+}
+
+NGRAPH_TEST(onnx_${BACKEND_NAME}, model_matmul_integer_4d)
+{
+    auto function = onnx_import::import_onnx_model(
+        file_util::path_join(SERIALIZED_ZOO, "onnx/matmul_integer_4d.prototxt"));
```
arogowie-intel: Please use input tensors with somewhat more complicated shape. This will let us to catch errors where we are messing multiplied axes.

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
```diff
@@ -94,6 +94,54 @@ namespace ngraph
...
+                                                                 input0_scale->get_element_type(),
+                                                                 axes);
+
+                    auto dq_input1 = make_shared<op::Dequantize>(input1,
+                                                                 input1_scale,
+                                                                 input1_zero_point,
+                                                                 input1_scale->get_element_type(),
+                                                                 axes);
+
+                    auto dot = make_shared<op::Dot>(dq_input0, dq_input1, 1);
```
arogowie-intel: We should be careful here. This `Dot` doesn't work as NumPy's matmul for higher order tensors (rank >2).

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
```diff
@@ -94,6 +94,54 @@ namespace ngraph
...
             }
+
+            shared_ptr<Node>
+                QuantizedLinearMatmulInteger(const std::shared_ptr<Node>& input0,
+                                             const std::shared_ptr<Node>& input1,
+                                             const std::shared_ptr<Node>& input0_zero_point,
+                                             const std::shared_ptr<Node>& input1_zero_point)
+            {
+                auto input0_zero = dynamic_pointer_cast<ngraph::op::Constant>(input0_zero_point);
+                auto input1_zero = dynamic_pointer_cast<ngraph::op::Constant>(input1_zero_point);
```
tsocha: There's no need to cast inputs to constants because `ngraph::is_zero()` does it internally.

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
```diff
@@ -94,6 +94,54 @@ namespace ngraph
...
+
+                    auto dot = make_shared<op::Dot>(dq_input0, dq_input1, 1);
+                    auto q_dot = make_shared<op::Quantize>(
+                        dot,
+                        output_scale,
+                        output_zero_point,
+                        output_zero_point->get_element_type(),
+                        axes,
+                        op::Quantize::RoundMode::ROUND_NEAREST_TOWARD_EVEN);
+                    return move(q_dot);
```
tomdol: In general you should avoid moving the return value, let the compiler choose the best way to optimize it. You can return the `make_shared` result directly in here.

src/ngraph/frontend/onnx_import/op/matmul_integer.cpp
postrational: Added `MatmulFactory` classes to address this issue.

test/onnx/onnx_import_quant.in.cpp
postrational: What shapes would you suggest? Can you give an example?

src/ngraph/builder/quantization/quantized_linear_matmul.cpp
postrational: Without `move` this raises a compiler warning: 
```
warning: prior to the resolution of a defect report against ISO C++11, 
    local variable 'q_dot' would have been copied despite being returned 
    by name, due to its not matching the function return type```

postrational: Done.

PR #3012, opened 2019-06-03
[ONNX] Add ConvInteger op

postrational:


src/ngraph/frontend/onnx_import/op/conv_integer.cpp
```diff
@@ -0,0 +1,87 @@
...
+            {
+                NodeVector conv_integer(const Node& node)
+                {
+                    const NodeVector& inputs = node.get_ng_inputs();
+                    auto num_inputs = inputs.size();
+                    auto input = inputs.at(0);
+                    auto filters = inputs.at(1);
+
+                    int64_t groups{node.get_attribute_value<int64_t>("group", 1)};
+                    ASSERT_VALID_ARGUMENT(node, (groups == 1))
```
arogowie-intel: What refrain us from doing sth similar to this: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/frontend/onnx_import/op/conv.cpp#L51-L93 ?

postrational: Nothing, I just don't think `groups` are actually used in real-world models any more.

src/ngraph/builder/quantization/quantized_linear_convolution.cpp
```diff
@@ -166,6 +166,68 @@ namespace ngraph
...
+                                                window_movement_strides,
+                                                window_dilation_strides,
+                                                padding_below,
+                                                padding_above,
+                                                data_dilation_strides);
+                }
+                else
+                {
+                    // Fall back to performing operation on dequantized floating-point values
+                    auto input_scale = make_constant(element::f32, Shape{}, 1);
```
tomdol: We have `make_constant` and `Constant::create`. Is any of them preferred?

src/ngraph/builder/quantization/quantized_linear_convolution.cpp
```diff
@@ -166,6 +166,68 @@ namespace ngraph
...
+
+                    auto quantized_output = make_shared<op::Quantize>(
+                        output,
+                        output_scale,
+                        output_zero_point,
+                        output_zero_point->get_element_type(),
+                        axes,
+                        op::Quantize::RoundMode::ROUND_NEAREST_TOWARD_EVEN);
+
+                    return move(quantized_output);
```
tomdol: Same as in the other PR - you should avoid moving the ret value explicitly and let the compiler choose an optimization method (RVO > move).

postrational: Done.

PR #3014, opened 2019-06-03
Use Eigen kernel for Gather when axis is 0.

ayzhuang:


test/backend_gather.in.cpp
```diff
@@ -36,6 +36,78 @@ using namespace ngraph;
...
+    copy_data(p, vector<float>{1.0f, 1.1f, 2.0f, 2.1f, 3.0f, 3.1f});
+    auto i = backend->create_tensor(element::i32, indices_shape);
+    copy_data(
+        i, vector<int32_t>{0, 1, 1, 2, 0, 1, 1, 2, 0, 1, 1, 2, 0, 1, 1, 2, 0, 1, 1, 2, 0, 1, 1, 2});
+    auto result = backend->create_tensor(element::f32, out_shape);
+
+    auto c = backend->compile(f);
+    c->call_with_validate({result}, {p, i});
+    EXPECT_TRUE(test::all_close_f(
+        (vector<float>{1.0f, 1.1f, 2.0f, 2.1f, 2.0f, 2.1f, 3.0f, 3.1f, 1.0f, 1.1f, 2.0f, 2.1f,
```
gaurides: does it always copy the whole row? based on the kernel code, i see you are calculating offset and extent. What indices would you need so that you don't copy the complete row?

test/backend_gather.in.cpp
```diff
@@ -36,6 +36,78 @@ using namespace ngraph;
 
 static string s_manifest = "${MANIFEST}";
 
+NGRAPH_TEST(${BACKEND_NAME}, gather_4d_indices_no_axis)
```
gaurides: does 4d here refer to rank of indices or the number 4 in indices?

src/ngraph/runtime/cpu/cpu_builder.hpp
```diff
@@ -204,6 +204,73 @@
...
+    }                                                                                              \
+    else if (R2 == 6)                                                                              \
+    {                                                                                              \
+        SELECT_RANK1(KV, ET, R1, 6, K);                                                            \
+    }                                                                                              \
+    else if (R2 == 7)                                                                              \
+    {                                                                                              \
+        SELECT_RANK1(KV, ET, R1, 7, K);                                                            \
+    }                                                                                              \
+    else                                                                                           \
```
gaurides: do you know why we expand these expressions instead of just 
```if (R2 > 7) throw error else pass R2 to SELECT_RANK1 ```

File: src/ngraph/runtime/cpu/cpu_builder.hpp [hunk 4]
```diff
@@ -204,6 +204,73 @@
...
+        KV = K<ET, 3, R2>;                                                                         \
+    else if (R1 == 4)                                                                              \
+        KV = K<ET, 4, R2>;                                                                         \
+    else if (R1 == 5)                                                                              \
+        KV = K<ET, 5, R2>;                                                                         \
+    else if (R1 == 6)                                                                              \
+        KV = K<ET, 6, R2>;                                                                         \
+    else if (R1 == 7)                                                                              \
+        KV = K<ET, 7, R2>;                                                                         \
+    else                                                                                           \
```
gaurides: similar comment as on line 254.

test/backend_gather.in.cpp
ayzhuang: When axis is 0 and the params has rank r,  gather always copies slices of rank r-1. In this unit test, the shape of params is {3, 2}. The offset for params slice would be {i, 0} and the extent would be {1, 3} with i be 0, 1, or 2.

test/backend_gather.in.cpp
ayzhuang: My mistake, should be 3d. It's the rank of indices.

src/ngraph/runtime/cpu/cpu_builder.hpp
ayzhuang: R1 and R2 are not constexpr thus can't be used in constant expression. We can't do KV= K<ET, R1, R2>.

gaurides: yes that's what i thought. I get it now. By doing it in levels SELECT_RANK2 -> SELECT_RANK1 and changing the R2 to constexpr R2. R2 or R1 in these are not the original variables passed.

PR #3018, opened 2019-06-04
Test messages improvements

postrational:
Includes:
* use unary `+` oprerator to cast for numeric display
* use `std::endl` instead of `\n` for cross-platform new lines

Currently the fail output message looks like this:
```
Actual: false (¤ is not close to ¥ at index 0§ is not close to ± at index 1² is not close to ¦ at index 2¨ is not close to ¤ at index 3)
```

With my proposed changes it looks like this:
```
Actual: false (164 is not close to 165 at index 0
167 is not close to 177 at index 1
178 is not close to 166 at index 2
168 is not close to 164 at index 3)
```


test/util/all_close.hpp
```diff
@@ -56,13 +56,14 @@ namespace ngraph
                     if (count < 5)
                     {
                         ar_fail << std::setprecision(std::numeric_limits<long double>::digits10 + 1)
-                                << a[i] << " is not close to " << b[i] << " at index " << i << "\n";
+                                << +a[i] << " is not close to " << +b[i] << " at index " << i
```
gcwenger: What advantage is conveyed by prefixing numbers with unary `+`?

test/util/all_close.hpp
```diff
@@ -56,13 +56,14 @@ namespace ngraph
...
+                                << std::endl;
```
gcwenger: Use of `\n` is often intentional because `std::endl` causes flushing, which can impact performance.
Why do you feel we need `std::endl` here?

diyessi: Performance in reporting test failures?

test/util/all_close.hpp
diyessi: The+ should go away.

postrational: The unary `+` prevents integer values (e.g. quantized model tests) from being displayed as `char`s and displays them as numeric values instead.

Currently the fail output message looks like this:
```
Actual: false (¤ is not close to ¥ at index 0§ is not close to ± at index 1² is not close to ¦ at index 2¨ is not close to ¤ at index 3)
```

With my proposed changes it looks like this:
```
Actual: false (164 is not close to 165 at index 0
167 is not close to 177 at index 1
178 is not close to 166 at index 2
168 is not close to 164 at index 3)
```
This is much more useful, as you can actually tell what the values are without trying to look up characters in an ASCII table.

I'm not sure what other solution, rather than `+`, to use here.
What would you suggest?

test/util/all_close.hpp
postrational: The reason for `endl` is to support output on platforms like Windows, which don't use `\n` as the endline character.

I don't think we need to worry about performance in reporting test failures too much.

test/util/all_close.hpp
gcwenger: Ok, given we're in a templated method I'm ok with the unary +:
https://isocpp.org/wiki/faq/input-output#print-char-or-ptr-as-number

diyessi: Just when I thought C++ couldn't get any uglier...

rkimballn1: You clearly need a comment as to why you did this. The fact that there is so much discussion means the code is not clear. Clarify please.

postrational:
Comments added.

PR #3019, opened 2019-06-04
Added getters for dynslice

jbobba:


jbobba:
I can add setters but is there a use-case for changing attributes of a node once it is created?

diyessi:
Maybe in a deserializer. We can hold off until such a use shows up.

PR #3020, opened 2019-06-04
Fix klocwork issues

diyessi:


src/ngraph/pass/core_fusion.cpp
```diff
@@ -1066,7 +1066,7 @@ void pass::CoreFusion::construct_conv_bias()
                      << m.get_match_root()->get_name();
         auto pattern_map = m.get_pattern_map();
 
-        auto conv_m = dynamic_pointer_cast<op::Convolution>(m.get_match_root()->get_argument(0));
+        auto conv_m = static_pointer_cast<op::Convolution>(m.get_match_root()->get_argument(0));
```
jbobba: this needs to be a dynamic_cast or we can switch to using get_description()

diyessi: You're right. I'll "not a problem" it in the next klocwork run.

PR #3021, opened 2019-06-05
[MLIR] Add ArgMin/ArgMax lowering support

dcaballe:
Adds support for ArgMin/Max index reduction Ops for any tensor rank.

**Few caveats:**
1. MLIR currently has no support for float point compare. So we only support integral types inputs. 
2. MLIR has no way of casting Index to/from Ints. That means our index tensor result during execution must be IndexType tensor (to be able to use its value to index into the input tensor).  Since nGraph expect Int outputs (Int32 or Int64), we must convert the IndexType to Integers. This is done in a call-back at the end (which is slow of course). Hopefully, this will all go away once we have explicit cast operation and can have the result in Integer tensor directly.

Co-authored-by: Diego Caballero <diego.caballero@intel.com>
Co-authored-by: nmostafa <nagy.h.mostafa@intel.com>


src/contrib/mlir/lowerer.cpp
```diff
@@ -398,8 +418,130 @@ namespace
...
+        ScopedContext scope(rewriter, loc);
+        Value* arg = operands[0];
+        auto arg_type = arg->getType().cast<MemRefType>();
+
+        Value* finalResult = m_pass.buildOutputDefs(op, rewriter)[0];
+        Type type = argmin.getResult()->getType();
+        NGTensorType resultTy = type.cast<NGTensorType>();
+        // MLIR doesn't support Index to/from Integer type-conversion
+        // We have to store our result in an IndexType tensor and call-back to a type-conversion routine in nGraph
+        // TODO: Fix this once MLIR provides explicit cast operations.
```
jbobba: Can we put in this workaround only when sizeof(intptr_t) != resultTy.getElementType()...getWidth()?

nmostafa: Gonna be ugly to avoid copying for argmin/max if IndexType matches IntType width. The problem is that func signature is built to say i64 is returned. Re-typing the output to IndexType will cause a mismatch with the ReturnOp. The MLIR canonicalizer fixes that mismatch silently, but that brings us back to square one. Ideally, I would have liked to just move the memref to the i64 result, but I don't think MLIR allows memref copy by pointer to avoid aliasing.
I am giving up on that for now (or could do a memcpy in the call-back instead), since upcoming casting Op should fix all this.

src/contrib/mlir/lowerer.cpp
```diff
@@ -436,6 +466,96 @@ namespace
...
+            NGRAPH_CHECK(resTy.isa<IntegerType>(),
+                         "Expected integer result type in index reduction");
+
+            // iterate over all argument dimensions
+            LoopNestBuilder(pAllIVs, argLbs, argUbs, steps)([&] {
+                // build a list of non-reduction IVs
+                for (auto i = 0; i < vArg.rank(); i++)
+                {
+                    if (i != axis)
+                        nonRedIVs.push_back(allIVs[i]);
```
dcaballe: Not sure I understand the magic here. If this loop is moved before LoopNestBuilder, allIVs elements are not captured properly. How does it work exactly? LoopNestBuilder goes over nonRedIVs elements just because it's used within LoopNestBuilder?

nmostafa: LoopNestBuilder will emit code and bind the Values to the IVs ValueHandles.

src/contrib/mlir/lowerer.cpp
```diff
@@ -59,6 +61,13 @@ namespace
 
 #include "op_lowerers.inc"
 
+    // Helpers
+    template <typename RedOp>
+    void lowerIndexReduction(Operation* op,
```
aprocter: (General comment, just picking on this line at random.) nGraph coding standards would call for `lower_index_reduction` here. Is this an MLIR style thing? If so, which style guidelines (MLIR or nGraph) should take precedence? (If not, can we switch things over to nGraph style?)

dcaballe: The idea was to keep nGraph style in the "entry point" files (i.e., compiler.hpp/.cpp) so that we keep nGraph style consistent in external APIs, and follow MLIR style on the more internal files (such us this one) that heavily use MLIR APIs. That would reduce the mix of nGraph/MLIR styles since we cannot control MLIR style. However, I think we are not being very consistent. I'll follow up with a commit to fix the style and add a comment about the style at the beginning of the file if there is no problem with this approach.

aprocter: I'd prefer to use nGraph style in both places, but as long as you're consistent about it I don't see it as a major issue.

dcaballe: Thanks, Adam. I leave it like this for now and open this question for discussion later on.

PR #3023, opened 2019-06-06
Use updated reduced-size clDNN

tomdol:
This PR updates clDNN to the newest version, which is stripped from boost. The updated version, when built, is only ~7MB, while the current - 350MB.


tomdol:
> Does this mean that boost is not used?

Yes, they got rid of boost completely. In an email thred where I was included, they suggested to remove those lines from our cmake file for clDNN.

PR #3026, opened 2019-06-06
Fix bug with serialization of zero-element constants

aprocter:
This is currently segfaulting when trying to serialize zero-element constants.


aprocter:
Note that #2972 also appears to have a fix for this.

PR #3027, opened 2019-06-06
Added support for ceil mode in AvgPool

jbobba:


src/ngraph/op/avg_pool.cpp
```diff
@@ -314,6 +329,11 @@ shared_ptr<Node> op::AvgPoolBackprop::copy_with_new_args(const NodeVector& new_a
 
 void op::AvgPool::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
 {
+    if (m_ceil_mode)
+    {
+        throw ngraph_error("Autodiff not supported on AvgPool with ceil_mode set");
```
aprocter: Can we open a JIRA for this? Shouldn't be too hard to do make it work.

test/type_prop.cpp
```diff
@@ -8647,6 +8647,27 @@ TEST(type_prop, avg_pool_3d_deduce_strided_padded_small)
     EXPECT_EQ(avg_pool->get_padding_above(), (Shape{6, 4, 5}));
 }
 
+TEST(type_prop, avg_pool_ceil_mode)
```
aprocter: We'll want some execution tests ASAP as well. Currently BEs will be silently ignoring this flag.

ilyachur:
Could you add support for ceil mode in MaxPool too?

test/type_prop.cpp
jbobba: i was thinking of modifying the interpreter kernel but wasn't sure how to handle the extra rows/columns (ignore/include) Will need to look at the frameworks.

PR #3028, opened 2019-06-06
Use convolution_direct algo 

nbpatel:


nbpatel:
Can you guys review it again? This fixes the failing test case ...also deleted that test case

PR #3029, opened 2019-06-07
GenerateMask correction

diyessi:
Add an attribute that controls if the seed should be set on each use
Convert to new virtual method for description implementation

When a seed is specified, frameworks set the seed on each use of Dropout. This adds an attribute to specify whether or not the seed should be set (for debugging) or random (for normal use). If not specified, the seed is not to be used.

Shape, seed, use_seed, probability are now exposed as arguments, but remain as attributes to simplify transition. Once transformers have switched to use the arguments, we can remove the attribute versions and bridges can make them dynamic.

Bridges will need to specify specify the proper value for `use_seed` and transformers will need to use it.


src/ngraph/op/experimental/generate_mask.hpp
```diff
@@ -53,8 +64,9 @@ namespace ngraph
             void validate_and_infer_types() override;
             Shape m_shape;
             element::Type m_element_type;
-            unsigned int m_seed;
-            double m_probability;
+            bool m_use_seed{false};
+            uint64_t m_seed{0};
+            double m_probability{0.0};
```
gaurides: Should the default value be 1.0 ?
I think, GenerateMask should have an argument to pass probability  instead of defaulting to 0 or 1

diyessi: The old constructor specifies everything. There's a second constructor now that takes no args and lets you set things after construction, to support other deserialization formats.

diyessi:
> Need test case for `use_seed=true`.
> Oh ok, i see, GenerateMask is only for CPU backend. So, i can add a test case with my other changes, but we would need the CPU implementation when true. Should it be part of this PR or a diff for CPU changes?

GenerateMask is used for other transformers on training devices.

src/ngraph/op/experimental/generate_mask.hpp
```diff
@@ -47,23 +59,29 @@ namespace ngraph
...
                 copy_with_new_args(const NodeVector& new_args) const override;
 
+            const element::Type& get_element_type() const { return m_element_type; }
+            void set_element_type(const element::Type& element_type)
+            {
+                m_element_type = element_type;
+            }
+
+            /// Deprecated accessor for transitional attributes
+            const Shape& get_shape() const { return m_shape; }
```
aprocter: This clashes with `Node::get_shape()`, which seems dangerous.

src/ngraph/runtime/reference/generate_mask.hpp
```diff
@@ -37,6 +37,30 @@ namespace ngraph
...
                 }
             }
+
+            template <typename T>
+            void generate_mask_no_state(
+                T* out, size_t count, bool training, uint32_t seed, double prob)
+            {
+                std::minstd_rand msr;
+                msr.seed(seed);
+                std::uniform_int_distribution<> dist(0, 1);
```
jbobba: shouldn't this be std::uniform_real_distribution? Alternatively, we could use bernoulli similar to the existing generate_mask

gaurides: In new commit, I changed similar to existing generate_mask without preserving state

src/ngraph/op/experimental/generate_mask.hpp
gaurides: renamed to get_mask_shape() in new commit.

PR #3033, opened 2019-06-07
Change for better user experience with json option

rkimballn1:
Code on master requires users that are in external repos to define the NGRAPH_JSON_ENABLE flag as a compiler definition in order to use json. This is not the desired behavior.

This PR changes it so that doing nothing allows you to use the json library, just as it was before the json disable feature was added. The cmake option is unchanged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -427,10 +427,6 @@ if(NGRAPH_DEPRECATED_ENABLE)
     add_definitions(-DNGRAPH_DEPRECATED_ENABLE)
 endif()
 
-if(NGRAPH_JSON_ENABLE)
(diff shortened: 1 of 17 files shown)


diyessi:
Wouldn't it be simpler to just switch the default for enable to be true, especially since this is temporary anyways?

rkimballn1:
The default *is* true. It is enabled. The problem is that in order to *not disable* json when compiling the user (think some other backend that is not in this repo) must define NGRAPH_JSON_ENABLE as a compiler define. If the user does not make that define then json is disabled. That is not good. The problem is not the cmake flag but is the compile-time define used to conditionally compile code. All I did is make it so that if you DON'T define the compile define then json is enabled.

PR #3034, opened 2019-06-07
Change behavior of elementwise divide for integral type to match Python.

silee2:
Addresses NGCORE-238
Done: 
- [o] Update UT
- [o] Update Reference Kernel
- [o] Update Interpreter backend
- [o] Update CPU Dex Eigen Kernel
- [o] Update CPU Codegen
- [o] Update Generic CPU backend
- [o] Add Op attribute to control C++11/Python behavior.
- [o] IntelGPU Disable Integer Divide UT that does Python style rounding- clDNN defines div and mod for integral types the c++11 way (divide rounds toward zero).
- [o] Update serializer to support legacy dump format (without rounding control.)


aprocter:
Will need to double check, but I think TensorFlow might actually need us to be able to handle both behaviors...

aprocter:
There is apparently a `TruncateDiv` op for TF (where `-1 / 2 == 0`), so I think we will need to find ways to support both behaviors (with a flag, or a new op, or whatever). I don't have strong feelings on which one should be the default.

diyessi:
We could borrow the appropriate quant rounding mode flags for an attribute.

silee2:
Tensorflow's doc is inaccurate for TruncateDiv: https://www.tensorflow.org/api_docs/python/tf/truncatediv
It says  x/y for integer type in the op description but then arg description for `x` list float types as well.
I think x and y need to be integer type.

aprocter:
The entry in ops.pbtxt allows integer and float types for both `x` and `y` (though, at least at the graph level, both have to be the same type), and playing with it at the REPL, `float32` seems to work. So yeah, I think the doc is inaccurate on that point.

aprocter:
The main point, though, is that it does have C-like semantics on integer types:

```
>>> minus_one = tf.constant(shape=(),value=-1,dtype='int32')
>>> two = tf.constant(shape=(),value=2,dtype='int32')
>>> tf.Session().run(tf.truediv(minus_one,two))
-0.5
>>> tf.Session().run(tf.floordiv(minus_one,two))
-1
>>> tf.Session().run(tf.div(minus_one,two))
-1
>>> tf.Session().run(tf.truncatediv(minus_one,two))
0
```

silee2:
Added a constructor arg that controls rounding for Integral Types.

test/backend_binary_elementwise.in.cpp
```diff
@@ -160,14 +160,36 @@ NGRAPH_TEST(${BACKEND_NAME}, divide_int32)
 
     // Create some tensors for input/output
     auto a = backend->create_tensor(element::i32, shape);
-    copy_data(a, vector<int32_t>{0x40000140, 0x40000001, 8, 16});
+    copy_data(a, vector<int32_t>{-10, -10, 10, 10});
```
rkimballn1: Those unusual integer inputs where put there in order to check that the integer is not converted to a bfloat16. Please put those values back. There is not much point in having 4 of the same value to test divide.

aprocter: Let's have both. The values @silee2 put in are intended to test the new behavior.

silee2: Restored modified test and added new test to test rounding mode.

PR #3036, opened 2019-06-07
Add DynElimination support for DynSlice

aprocter:
This PR adds support for `DynElimination` on the `DynSlice` op, including some execution tests running through the BE wrapper. It's rather more complicated than I expect most of these will be, because `DynSlice` has a lot of bells and whistles to support (TensorFlow's subset of) NumPy's slicing behavior.

Corner cases abound so much that I really thought the best way to test it was against a NumPy-based reference, which is one of the hairier things here. Could use some feedback on whether the tests in `test/ref_generators/generate_dyn_slice_ref.py` look complete enough.


aprocter:
On second thought I'm going to tag this [WIP] while I think about some more test cases to add. (In particular I *think* the handling of backwards slicing is correct, but it concerns me a bit.)

test/ref_generators/generate_dyn_slice_ref.py
```diff
@@ -0,0 +1,626 @@
...
+        t[3:0:-2]
+        t[0:3:2]
+        t[0:4:2]
+        t[0:5:2]
+        t[0:6:2]
+        t[0:100:2]
+        t[4:0:-2]
+        t[4:0:-3]
+        t[3:2:1]
+        t[4::-2]
```
sayantan-nervana: as far as indexing in numpy goes here are 2 possibilities:
1. stop is negative, stride is negative
`t[1:-5:-1] ---> (2,)`
`t[1:-1:-1] ---> (0,)` . (empty)

2. None
`t[1:None] ---> (3,)` . (None means end)
`x[1:None:-1] ---> (2,)`

aprocter: Thanks @sayantan-nervana. Added these, and they all pass :)

PR #3037, opened 2019-06-07
Added missing attribute to Result serialization

jbobba:


src/ngraph/serializer.cpp
```diff
@@ -1407,7 +1407,10 @@ static shared_ptr<ngraph::Function>
             }
             case OP_TYPEID::Result:
             {
-                node = make_shared<op::Result>(args[0]);
+                auto needs_default_layout = node_js["needs_default_layout"].empty()
```
rkimballn1: use `get_default_value<bool>(node_js, "needs_default_value", false)`

jbobba: do you want me to  implement get_default_value?

rkimballn1: Sorry, my mistake. The function is get_or_default, not get_default_value.

jbobba: oh, nice. Didn't notice it before.

PR #3038, opened 2019-06-08
TensorIterator

diyessi:
- [x] Test/example
- [x] Documentation
- [x] Shape inference
    - [x] SliceInput
    - [x] SliceOutput
    - [x] SequenceRepeat
    - [x] SequencePushFront
    - [x] TensorIterator
        - [x] Determination of iteration values
        - [x] Determination of sequence length
        - [x] Shape inference on iterator body
        - [x] TensorIterator outputs from body_outputs
        - [x] TensorIterator output shapes
- [x] Serialization
    - [x] SliceInput
    - [x] SliceOutput
    - [x] TensorIterator
    - [x] OutputVector
    - [x] Output
    - [x] SequenceRepeat
    - [x] SequencePushFront


test/serialize.cpp
```diff
@@ -340,3 +340,53 @@ TEST(serialize, non_zero_node_output)
...
+            Xseq,                                            // Xi values
+            make_shared<op::SequencePushFront>(Hinit, Hseq), // Hi values
+        },
+        ParameterVector{Xi, Hi}, // Body parameters
+        // Body outputs
+        OutputVector{Hseq, Yseq},
+        // TensorIterator outputs
+        OutputVector{// axis=1, start=0, stride=1, part_size=1, end=40
+                     make_shared<op::SliceOutput>(Yseq, 1, 0, 1, 1, 40)},
+        AxisSet{0, 1},
```
ilyachur: What are 0 and 1 in this case?

test/serialize.cpp
```diff
@@ -340,3 +340,53 @@ TEST(serialize, non_zero_node_output)
...
+        AxisSet{0});
```
ilyachur: What do you mean sequence_outputs?

test/serialize.cpp
```diff
@@ -340,3 +340,53 @@ TEST(serialize, non_zero_node_output)
...
+
+    auto result = make_shared<op::Result>(Output<Node>(tensor_iterator, 0));
+    auto f =
+        make_shared<Function>(ResultVector{result}, ParameterVector{X, Hinit, WH, WX, bH, WY, bY});
+    string s = serialize(f);
+    shared_ptr<Function> g = deserialize(s);
+}
```
ilyachur: As I understand this test checks nothing. Will it changed in the future?

test/serialize.cpp
```diff
@@ -340,3 +340,53 @@ TEST(serialize, non_zero_node_output)
...
+    auto Xseq = make_shared<op::SliceInput>(X, 1, 0, 1, 1, 40);
+    auto Hseq = make_shared<op::SequenceRepeat>(Ho);
+    auto Yseq = make_shared<op::SequenceRepeat>(Yo);
+    auto tensor_iterator = make_shared<op::TensorIterator>(
+        // Body inputs
+        OutputVector{
+            Xseq,                                            // Xi values
+            make_shared<op::SequencePushFront>(Hinit, Hseq), // Hi values
+        },
+        ParameterVector{Xi, Hi}, // Body parameters
```
ilyachur: Is this ParameterVector contains initial values?

src/ngraph/op/slice_input.cpp
```diff
@@ -0,0 +1,134 @@
...
+    , m_part_size(part_size)
+    , m_end(end)
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::SliceInput::validate_and_infer_types()
+{
+    // For now, keep things simple; no padding, forwards stepping, etc.
+    NODE_VALIDATION_CHECK(this, 0 <= m_axis, "Negative axis is not supported.");
```
ilyachur: Now we are not supporting m_axis, m_start, m_end and other parameters as a dynamic arguments for the Inference Engine TensorIterator.
Could we (together ngraph and dldt) think about how to make these parameters dynamic? It could add more flexibility in the future.

src/ngraph/op/slice_input.cpp
```diff
@@ -0,0 +1,134 @@
...
+    NODE_VALIDATION_CHECK(this, 0 <= m_start, "Negative start is not supported.");
+    NODE_VALIDATION_CHECK(this, 0 <= m_end, "Negative end is not supported.");
+    NODE_VALIDATION_CHECK(this, 1 <= m_part_size, "Part size must be positive.");
+    NODE_VALIDATION_CHECK(this, 1 <= m_stride, "Stride must be positive.");
```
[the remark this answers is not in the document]
diyessi: Negative stride in SliceInput.

src/ngraph/op/parameter.cpp
```diff
@@ -21,11 +21,12 @@
 using namespace std;
 using namespace ngraph;
 
+const string op::Parameter::type_name{"Parameter"};
```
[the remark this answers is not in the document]
diyessi: This shouldn't be in this PR. Thanks for catching it.

src/ngraph/op/sequence_push_front.cpp
```diff
@@ -0,0 +1,64 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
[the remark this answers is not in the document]
diyessi: The dates apply to all of nGraph, not just the file.

test/serialize.cpp
diyessi: Crazy thinking. I got rid of them.

test/serialize.cpp
diyessi: It's currently testing serialization and that shape inference is somewhat working.

test/serialize.cpp
diyessi: These are the body parameters.

test/serialize.cpp
diyessi: Some RNNs directly produce a sequence of outputs as they process their input sequence, others use just the last output to produce a new sequence as output.

src/ngraph/op/slice_input.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::SliceInput::validate_and_infer_types()
+{
+    // For now, keep things simple; no padding, forwards stepping, etc.
+    NODE_VALIDATION_CHECK(this, 0 <= m_axis, "Negative axis is not supported.");
+    NODE_VALIDATION_CHECK(this, 0 <= m_start, "Negative start is not supported.");
+    NODE_VALIDATION_CHECK(this, 0 <= m_end, "Negative end is not supported.");
```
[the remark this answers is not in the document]
diyessi: When we move beyond the "keep things simple" stage, we can do as in python; -1 means the end.

src/ngraph/op/slice_input.cpp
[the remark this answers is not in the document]
ilyachur: But as I can see from the line above "Stride must be positive."

src/ngraph/op/tensor_iterator.cpp
```diff
@@ -0,0 +1,128 @@
...
+        {
+            bool is_free = true;
+            for (auto input : node->inputs())
+            {
+                auto input_node = input.get_source_output().get_node();
+                if (bound_nodes.find(input_node) != bound_nodes.end())
+                {
+                    bound_nodes.insert(node.get());
+                    is_free = false;
+                    std::cerr << *this << " Bound: "
```
MariaKaglin: Why do you use `std::cerr` here? Is it really unsupported case or it is just debug information? 
If it is unsupported case, assert is better solution.

src/ngraph/op/tensor_iterator.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+const string op::TensorIterator::type_name{"TensorIterator"};
+
+void op::TensorIterator::validate_and_infer_types()
+{
+    NODE_VALIDATION_CHECK(this,
+                          get_input_size() == m_body_parameters.size(),
+                          "Number of inputs must be the same as number of body parameters");
+
+    // The number of iterations is determined by the shortest sequence input
```
MariaKaglin: Why we using the shortest sequence lengths? 
May be better to assume input sequences have equal lengths.

src/ngraph/op/tensor_iterator.cpp
```diff
@@ -0,0 +1,128 @@
...
+                    is_free = false;
+                    std::cerr << *this << " Bound: "
+                              << " : " << *node << std::endl;
+                    break;
+                }
+            }
+            if (is_free)
+            {
+                free_nodes.push_back(node.get());
+                std::cout << *this << " Free: " << *node << std::endl;
```
MariaKaglin: Is `free_nodes` and `bound_nodes` calculated just for the information or it will be used later somehow?

src/ngraph/op/tensor_iterator.hpp
```diff
@@ -0,0 +1,110 @@
...
+        };
+
+        /// \brief  Iterate a body over tensors, accumulating into tensors
+        class TensorIterator : public util::FusedOp
+        {
+        public:
+            NGRAPH_API
+            static const std::string type_name;
+            const std::string& description() const override { return type_name; }
+        public:
```
MariaKaglin: Why do you have two public sections here?

test/serialize.cpp
```diff
@@ -340,3 +340,49 @@ TEST(serialize, non_zero_node_output)
...
+    auto Xi = make_shared<op::Parameter>(element::f32, Shape{32, 1, 10});
+
+    // Body
+    auto Ho = make_shared<op::Relu>(
+        make_shared<op::Dot>(make_shared<op::Reshape>(Xi, AxisVector{0, 1, 2}, Shape{32, 10}), WX) +
+        make_shared<op::Dot>(Hi, WH) + make_shared<op::Broadcast>(bH, Shape{32, 20}, AxisSet{0}));
+    auto Yo = make_shared<op::Relu>(make_shared<op::Dot>(Ho, WY) +
+                                    make_shared<op::Broadcast>(bY, Shape{32, 5}, AxisSet{0}));
+
+    auto tensor_iterator = make_shared<op::TensorIterator>();
```
MariaKaglin: As far as I know, in order to create any nGraph operation you should specify all inputs. 
Why TwnsorIterator have another creation logic? It feels inconsistent with nGraph philosophy.

src/ngraph/op/slice_output.cpp
```diff
@@ -0,0 +1,137 @@
...
+        vector<Dimension> value_dimensions = static_cast<vector<Dimension>>(value_shape);
+        Dimension sequence_dimension = value_dimensions[0];
+        value_dimensions.erase(value_dimensions.begin());
+
+        NODE_VALIDATION_CHECK(this,
+                              m_axis < value_dimensions.size(),
+                              "Axis is not within the sequence element shape.");
+        Dimension value_dimension = value_dimensions.at(m_axis);
+
+        if (sequence_dimension.is_dynamic() || value_dimension.is_dynamic())
```
MariaKaglin: Please, answer the same questions like for `sliced_input`.

src/ngraph/op/sequence_push_front.hpp
```diff
@@ -0,0 +1,50 @@
...
+        class SequencePushFront : public Op
+        {
+        public:
+            NGRAPH_API
+            static const std::string type_name;
+            const std::string& description() const override { return type_name; }
+            /// \brief Empty op
+            SequencePushFront() = default;
+
+            /// \brief Prepend a value to a sequence
```
MariaKaglin: Please,  can you explain this operation and use cases?

src/ngraph/op/sequence_repeat.hpp
```diff
@@ -0,0 +1,42 @@
...
+    {
+        class SequenceRepeat : public Op
+        {
+        public:
+            static const std::string type_name;
+            const std::string& description() const override { return type_name; }
+            /// \brief Constructs an empty SequenceRepeat operation
+            SequenceRepeat() = default;
+
+            /// \brief Creates a repeating sequence of an op
```
MariaKaglin: Please,  can you explain why do we need this operation and how it will be used?

src/ngraph/op/tensor_iterator.cpp
diyessi: This is just debugging.

src/ngraph/op/tensor_iterator.cpp
diyessi: In this version of TensorIterator, some sequence lengths were unending. Not true in the current version.
What does IE do if the sequences are different lengths?

src/ngraph/op/tensor_iterator.cpp
diyessi: It will be used to determine which thins change from iteration to iteration and which inputs stay the same from iteration.

src/ngraph/op/tensor_iterator.hpp
diyessi: Still a work in progress; not cleaned up yet.

test/serialize.cpp
diyessi: I have been adding no-arg constructors so that you could create the nodes and then connect them together at some point.
But, in this case, we are trying to separate how TensorIterator is implemented from how it is set up so that we can more easily adjust the implementation when we have more experience with it.

src/ngraph/op/slice_output.cpp
diyessi: Sliced input goes away, hopefully tomorrow. It is folded into `set_sliced_input`.

src/ngraph/op/sequence_push_front.hpp
diyessi: This goes away because it was too confusing; it is folded into `set_initialized_input`

src/ngraph/op/tensor_iterator.cpp
```diff
@@ -0,0 +1,303 @@
...
+        input_for_value(initial_value).get_index(), body_parameter, successive_value));
+}
+
+Output<Node> op::TensorIterator::get_iter_value(const Output<Node>& body_value, int64_t iteration)
+{
+    auto output_index = get_output_size();
+    m_output_descriptions.push_back(
+        make_shared<BodyOutputDescription>(body_value, output_index, iteration));
+    set_output_size(output_index + 1);
+    return Output<Node>(shared_from_this(), output_index);
```
itikhono: Can we return the data as `op :: Result` instead of` `Output <Node>`? TensorIterator will look like a full ngraph function with input in the form of `Parameter` and output in the form of` op :: Result`.

src/ngraph/op/tensor_iterator.cpp
```diff
@@ -0,0 +1,303 @@
...
+{
+    return make_shared<ConcatOutputDescription>(
+        m_body_value, m_output_index, m_start, m_stride, m_part_size, m_end, m_axis);
+}
+
+op::TensorIterator::BodyOutputDescription::BodyOutputDescription(const Output<Node>& body_value,
+                                                                 uint64_t output_index,
+                                                                 int64_t iteration)
+    : OutputDescription(body_value, output_index)
+    , m_iteration(iteration)
```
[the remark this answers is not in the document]
diyessi: See the comment in the `.hpp` file. If we loop N times, it says which of those N iterations we should take the value from; normal would be -1 for the final value.
A backprop iteration will need a way for its body to reach into the corresponding fprop iteration body. Not sure just how that will work yet.

src/ngraph/op/tensor_iterator.cpp
```diff
@@ -0,0 +1,303 @@
...
+    return Output<Node>(shared_from_this(), output_index);
+}
+
+NodeVector op::TensorIterator::decompose_op() const
+{
+    // Stub
+    return NodeVector{};
+}
+
+#if 0
```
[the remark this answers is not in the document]
diyessi: Time.

src/ngraph/op/tensor_iterator.cpp
diyessi: It's not a function right now.  `Result` is more like a stable part of `Function` to hang a piece of graph on so that splicing is simple that it is an op.

itikhono: @diyessi, we would like to have Result op for two purposes: (1) as a clear indication of output in the IR v10 and (2) as an explicit point to connect back-edge and preserve intermediate results in a body to avoid any fusion that can remove this intermediate tensor. So we need them as explicit Result ops. And why do you think that this is not yet a function -- it is really a function, it is separated from an exterior function, it has parameters and results, but currently only Parameters are explicitly materialized, our proposal is to do for Results as well.

itikhono: @ayzhuang @diyessi TensorIterator is critical to us. Can we resolve this question this week?

src/ngraph/op/tensor_iterator.hpp
```diff
@@ -0,0 +1,300 @@
...
+            const NodeTypeInfo& get_type_info() const override { return type_info; }
+            // Forward declarations
+            class SliceInputDescription;
+            class BodyConnectionInputDescription;
+            class ConstantInputDescription;
+
+            TensorIterator() = default;
+            TensorIterator(const OutputVector& values);
+
+            class BodyLambda : public Lambda
```
itikhono: @ayzhuang @diyessi Thanks, it looks suitable for us, but could you explain the reason
Why did you define this class? It looks the same as `Lambda`. 
And the second question is: Why don't you want to use `ngraph::Function` for TensorIterator body?

itikhono: @diyessi thanks for the explanation. We will try to complete the review this week.
The helpers look really useful, thanks a lot!

src/ngraph/op/tensor_iterator.hpp
```diff
@@ -0,0 +1,300 @@
...
+                uint64_t m_body_parameter_index;
+            };
+
+            /// \brief Describes a body input formed from slices of an input to TensorIterator.
+            class SliceInputDescription : public InputDescription
+            {
+            public:
+                static constexpr DiscreteTypeInfo type_info{"SliceInputDescription", 0};
+                const DiscreteTypeInfo& get_type_info() const override { return type_info; }
+                /// \param input_index Position of the TensorIterator input
```
nmostafa: An example here of how a sliced input works would be helpful.

src/ngraph/op/tensor_iterator.hpp
```diff
@@ -0,0 +1,300 @@
...
+                int64_t m_start;
+                int64_t m_stride;
+                uint64_t m_part_size;
+                int64_t m_end;
+                int64_t m_axis;
+            };
+
+            /// \brief Describes a body input initialized from a TensorIterator input on the first
+            /// iteration, and then a body output thereafter.
+            class BodyConnectionInputDescription : public InputDescription
```
nmostafa: nit: What about calling it `MergedInputDescription` ?

src/ngraph/op/tensor_iterator.hpp
```diff
@@ -0,0 +1,300 @@
...
+                /// iterations.
+                BodyConnectionInputDescription(uint64_t input_index,
+                                               uint64_t body_parameter_index,
+                                               uint64_t body_value_index);
+                std::shared_ptr<InputDescription> copy() const override;
+
+                uint64_t m_body_value_index;
+            };
+
+            class ConstantInputDescription : public InputDescription
```
nmostafa: Rename to `InvariantInput` ? Constant is a bit misleading.

src/ngraph/op/tensor_iterator.hpp
```diff
@@ -0,0 +1,300 @@
...
+            /// \param start First index on axis of the slicing
+            /// \param stride Stepping of the slice
+            /// \param part_size Size of the slice on axis
+            /// \param end The last index on axis of the slicing
+            /// \param axis The axis to slice along
+            void set_sliced_input(const std::shared_ptr<Parameter>& parameter,
+                                  const Output<Node>& value,
+                                  int64_t start,
+                                  int64_t stride,
+                                  int64_t part_size,
```
nmostafa: `part_size` type here mismatches one on `SlicedInputDescription` above.

src/ngraph/op/tensor_iterator.hpp
```diff
@@ -0,0 +1,300 @@
...
+                                  int64_t end,
+                                  int64_t axis);
+            /// \brief Indicates that a body parameter has an initial value in the first iteration
+            /// and computed value thereafter
+            /// \param initial_value Value for the parameter in first iteration. This will be added
+            /// as an input to TensorIterator.
+            /// \param successive_value Value for the parameter in successive iterations. The
+            /// value is what is active in the most recent completed iteration.
+            void set_initialized_input(const std::shared_ptr<Parameter>& body_parameter,
```
nmostafa: rename to match class name above ?

src/ngraph/op/tensor_iterator.hpp
```diff
@@ -0,0 +1,300 @@
...
+                                       const Output<Node>& initial_value,
+                                       const Output<Node>& successive_value);
+            /// \brief Indicates that a body parameter has a constant value during iteration that
+            /// may depend on values computed outside of the iteration
+            /// \param body_parameter The body parameter
+            /// \param value The value supplied as an input to the block
+            void set_constant_input(const std::shared_ptr<Parameter>& body_parameter,
```
nmostafa: again `_invariant_` would be better.

PR #3040, opened 2019-06-10
[FUSED] Group Transpose Convolution

arogowie-intel:
Follow-on PR's:
- add support for more padding types (currently only EXPLICIT)


src/ngraph/op/fused/group_conv_transpose.cpp
```diff
@@ -0,0 +1,302 @@
...
+                {
+                    m_padding_begin[i] = total_padding / 2;
+                }
+            }
+            m_padding_end = m_padding_begin;
+        }
+    }
+}
+
+void op::GroupConvolutionTranspose::post_validate_and_infer_types()
```
tomdol: Looks like this method is obsolete

src/ngraph/frontend/onnx_import/op/conv_transpose.cpp
```diff
@@ -158,73 +79,31 @@ namespace ngraph
...
-                            padding_above[i] = padding_below[i];
-                            data_batch_shape[i + 2] = output_shape[i];
-                        }
+                        conv_node = std::make_shared<ngraph::op::GroupConvolutionTranspose>(
+                            data,
+                            filters,
+                            strides,
+                            dilations,
+                            output_padding,
+                            Shape(std::begin(output_shape), std::end(output_shape)),
```
tomdol: I think you can pass the output_shape directly in here

arogowie-intel: Unfortunately not :(  
```
no known conversion for argument 6 from ‘std::vector<long int>’ to ‘const ngraph::Shape&’
```

src/ngraph/op/fused/group_conv_transpose.cpp
```diff
@@ -0,0 +1,320 @@
...
+op::GroupConvolutionTranspose::GroupConvolutionTranspose(const shared_ptr<Node>& data,
+                                                         const shared_ptr<Node>& filters,
+                                                         const Strides& strides,
+                                                         const Strides& dilations,
+                                                         const CoordinateDiff& padding_begin,
+                                                         const CoordinateDiff& padding_end,
+                                                         const CoordinateDiff& output_padding,
+                                                         const size_t groups,
+                                                         const PadType& pad_type,
+                                                         const Shape& output_shape)
```
aprocter: Could you explain why `output_shape` needs to be a parameter here? Normally we would infer this.

arogowie-intel: That is actually ONNX specific. Since it enables to define `output_shape` and let infer required padding for this.

jbobba: It confused me as well but i assume ONNX had a reason for going down this path.

PR #3041, opened 2019-06-10
Friendlier graph visualization

rkimballn1:
Change graph visualize to use the friendly name instead of the unique name when labeling nodes in the visualized graph.

Change deserialize to set the friendly name to the "name" field from the serialized graph if there is no friendly name for a node. This way if a node is `Add_123` in the serialized graph then it will show up as `Add_123` in the visualized graph instead of some new `Add_XYZ` name.

Extended unit test to check that the visualized graph contains the friendly names.


aprocter:
@rkimballn1 Can you talk a bit about the motivation for this change? The "friendly names" for the most part don't seem to me to be very useful in visualization. You just wind up with a bunch of duplicate names and no indication of the node type.

rkimballn1:
This change was added because a user had complained that the names of the nodes change if you visualize a serialized graph as I said in the description. You are correct that the friendly names are not unique. I can add both friendly and unique names to the node so they will be uniquely identified.

aprocter:
I think friendly+unique should work.

PR #3043, opened 2019-06-10
Disable caching when node has state.

ayzhuang:


src/ngraph/graph_util.hpp
```diff
@@ -292,6 +292,9 @@ namespace ngraph
     // the output of this node with in-place kernels
     bool possibly_overwritten(Node* node);
 
+    // Return true if node has state.
```
diyessi: Let's make this a virtual with a default definition in node that returns false.

ayzhuang: If we add m_has_state and has_state() to Node, then we don't need has_state here. cpu_external_function could just do node->has_state().

src/ngraph/graph_util.cpp
```diff
@@ -27,6 +27,7 @@
 #include "ngraph/node.hpp"
 #include "ngraph/op/broadcast.hpp"
 #include "ngraph/op/constant.hpp"
+#include "ngraph/op/experimental/generate_mask.hpp"
```
aprocter: I don't think we need to add this #include.

ayzhuang: will remove. thanks.

PR #3044, opened 2019-06-10
Lower minimum CMake version to 3.1

silee2:
Some external projects require later version of CMake.
Building LLVM from source requires CMake 3.4.3
ONNX Import feature requires CMake 3.3
Halide requires CMake 3.2
Additional version checking will kick in if those features are enabled and proper error will be thrown if minimum requirement is not met.


silee2:
PDPD lists 3.0 as the minimum version and tried to go as low as possible. Closing the PR since that does not seem like a requirement.

PR #3045, opened 2019-06-11
CI Dockerfile for ngraph-onnx testing using iGPU

tomdol:


mchrusci:
A while back I changed CI flow to use docker image built in separate job and pished to repository, instead of building one each time CI is run. I'd need to modify CI to include new images.

There's important thing to consider if You want to enable CI on iGPU device as a step in a larger CI pipeline. Should iGPU and CPU tests run on different nodes? Example iGPU on NUC with Intel Iris Pro and CPU on SKX or BDW server. Or is it fine to run both CPU and iGPU tests on NUC? 
We need to either change CI flow to run different images on different nodes or have more NUCs dedicated to CI.

PR #3047, opened 2019-06-11
Add copy of friendly name to ngraph::copy_function

rkimballn1:
Unit test added too


src/ngraph/graph_util.cpp
```diff
@@ -249,11 +249,17 @@ std::list<std::shared_ptr<ngraph::Node>>
             }
             auto cloned_node = node->copy_with_new_args(cloned_args);
```
diyessi: This isn't the only place that called `copy_with_new_args` and various callers do various amounts of additional work that should be done, like copying over the control deps. I am thinking we should make `copy_with_new_args` protected and add a friend function that does the copy and the additional stuff like friendly name propagation and control deps and change the about 10 callers to use the function.

rkimballn1: I would agree, but this does solve the problem we are having. We can use further PRs to address the other issues.

PR #3048, opened 2019-06-11
Support arbitrary reduction op

ami-GS:


src/ngraph/op/allreduce.hpp
```diff
@@ -28,13 +28,26 @@ namespace ngraph
...
+            const Reduce_t reduce_type;
             const std::string& description() const override { return type_name; }
             AllReduce();
-            AllReduce(const std::shared_ptr<Node>& arg);
+            AllReduce(const std::shared_ptr<Node>& arg, const Reduce_t reduce_type = Reduce_t::sum);
 
             void validate_and_infer_types() override;
 
             std::shared_ptr<Node> copy_with_new_args(const NodeVector& new_args) const override;
+            std::string getReduceTypeString() const {
```
aprocter: nGraph house style would be `get_reduce_type_string`. In any case, I think it's better if this is a free function on `Reduce_t`, rather than a member of `AllReduce`. Or, better, define an `operator<<` overload for `Reduce_t`.

test/distributed.in.cpp
```diff
@@ -29,32 +29,67 @@
 using namespace std;
 using namespace ngraph;
 
-TEST(distributed_${BACKEND_NAME}, allreduce)
-{
+void test_allreduce_common_${BACKEND_NAME}(Reduce_t reduce_type) {
```
aprocter: There should be a `static` qualifier on this.

I don't think it's necessary to encode `${BACKEND_NAME}` in the function name, because `distributed.in.cpp` will be instantiated separately for each backend.

src/ngraph/distributed.hpp
```diff
@@ -24,6 +24,16 @@
 
 namespace ngraph
 {
+
+    enum class Reduce_t
```
aprocter: We usually don't use the `_t` suffix in nGraph.

src/ngraph/distributed.hpp
```diff
@@ -24,6 +24,16 @@
...
+    {
+        sum,
+        prod,
+        min,
+        max,
+        undefined
```
aprocter: What is the purpose of `undefined` here? If we have an `undefined` value, we should check for it and reject it inside `AllReduce::validate_and_infer_types`.

src/ngraph/distributed/mlsl.hpp
```diff
@@ -83,10 +83,23 @@ namespace ngraph
                     throw std::runtime_error("AllReduce op supports only f32 and f64 types");
                 }
 
+                auto allReduce_type = MLSL::RT_SUM;
```
aprocter: We generally do not use `camelCase` in nGraph. (Well, just in the name of the product, I guess. :))

src/ngraph/distributed.hpp
ami-GS: I create another `reducetion::Type`, like following `element::Type`, but both use `Type_t`

aprocter: I don't think `element::Type_t` should be called `element::Type_t`, but I have to admit it establishes a precedent. :)

src/ngraph/distributed/mlsl.hpp
```diff
@@ -83,10 +83,23 @@ namespace ngraph
                     throw std::runtime_error("AllReduce op supports only f32 and f64 types");
                 }
 
+                auto mlsl_reduce_type = MLSL::RT_SUM;
```
diyessi: Use `switch` and then you won't need the throw because the compiler will know you handled all the cases.

aprocter:
@ami-GS It looks like the PR is failing style checks in CI. Please run `make style-apply`, then commit and push.

src/ngraph/distributed/mlsl.hpp
ami-GS: You mean switch without `default`?

ami-GS: switch doesn't work easily. Can I use `if`? element_type is also using `if`

src/ngraph/distributed.cpp
```diff
@@ -22,6 +22,37 @@
...
 using namespace ngraph;
 
+NGRAPH_API const reduction::Type reduction::sum(reduction::Type_t::sum);
+NGRAPH_API const reduction::Type reduction::prod(reduction::Type_t::prod);
+NGRAPH_API const reduction::Type reduction::min(reduction::Type_t::min);
+NGRAPH_API const reduction::Type reduction::max(reduction::Type_t::max);
+
+std::ostream& reduction::operator<<(std::ostream& out, const reduction::Type& obj)
+{
+    if (obj.m_type == reduction::Type_t::sum)
```
rkimballn1: Change this to a switch statement and then put this around you switch
```
#if !(defined(__GNUC__) && (__GNUC__ == 4 && __GNUC_MINOR__ == 8))
#pragma GCC diagnostic push
#pragma GCC diagnostic error "-Wswitch"
#pragma GCC diagnostic error "-Wswitch-enum"
#endif
<switch statement here>
#if !(defined(__GNUC__) && __GNUC__ == 4 && __GNUC_MINOR__ == 8)
#pragma GCC diagnostic pop
#endif
```
This way if a new element is added to the enum then you will get a compile-time error

src/ngraph/distributed.hpp
```diff
@@ -24,6 +24,36 @@
...
+        {
+        public:
+            Type(const Type_t t)
+                : m_type(t)
+            {
+            }
+            friend std::ostream& operator<<(std::ostream&, const Type&);
+            bool operator==(const Type& other) const;
+            bool operator!=(const Type& other) const { return !(*this == other); }
+            Type_t m_type;
```
rkimballn1: make this private and add an accessor `Type_t get_type() const`

src/ngraph/distributed/mlsl.hpp
```diff
@@ -83,10 +86,19 @@ namespace ngraph
...
+                switch (reduce_type)
```
rkimballn1: Surround with the switch check code I showed above

src/ngraph/distributed/open_mpi.hpp
```diff
@@ -94,8 +97,16 @@ namespace ngraph
                 {
                     throw std::runtime_error("AllReduce op supports only f32 and f64 types");
                 }
+                auto mpi_reduce_type = MPI_SUM;
+                switch (reduce_type)
```
rkimballn1: surround with switch check

test/distributed.in.cpp
```diff
@@ -29,32 +29,79 @@
...
 
         auto v = vector<float>{1, 2, 3, 4};
         auto a = backend->create_tensor(element::f32, shape);
-        copy_data(a, vector<float>{1, 2, 3, 4});
-
         auto result = backend->create_tensor(element::f32, shape);
-
-        std::transform(
-            v.begin(), v.end(), v.begin(), std::bind1st(std::multiplies<float>(), comm_size));
+        if (reduce_type == reduction::sum)
```
rkimballn1: Change to switch and surround with check

ami-GS:
I cannot reproduce style error happens on CI.....

diyessi:
That's not a style error, that's a syntax error. You need a `->` before the `float`
https://en.cppreference.com/w/cpp/language/lambda

ami-GS:
> That's not a style error, that's a syntax error. You need a `->` before the `float`
> https://en.cppreference.com/w/cpp/language/lambda
I could build locally for some reason...

diyessi:
So can I (macos clang). C++ is really hard to parse; not all compilers catch everything.

PR #3049, opened 2019-06-12
IntelGPU backend: Switch to new clDNN: Fixed _GLIBCXX_USE_CXX11_ABI redifinition error during ngraph-bridge build

dmyershov:
In new clDNN conflicted -D_GLIBCXX_USE_CXX11_ABI=0 definition was removed from gcc command line.


avijit-nervana:
This PR needs to be merged to master before the Friday cutoff as this is `needed` for R0.22

PR #3051, opened 2019-06-12
Change hybrid to static library

rkimballn1:
This way it is not linked into ngraph. This will allow other non-ngraph-repo backends to more easily extend hybrid as part of their own build process.

This change still runs unit tests on hybrid library


test/CMakeLists.txt
```diff
@@ -230,6 +230,7 @@ endif()
 
 target_link_libraries(unit-test PRIVATE ngraph_test_util)
 target_link_libraries(unit-test PRIVATE ngraph libgtest)
+target_link_libraries(unit-test PRIVATE -Wl,--whole-archive hybrid -Wl,--no-whole-archive)
```
silee2: Will this work on Windows or macos? Looks gcc and gnu-ld specific.

rkimballn1: fixed to only run on linux

PR #3052, opened 2019-06-12
Support any root id for broadcast distributed

ami-GS:
originally this has supported 0 as root id.
This supports any root id by specifying argument. (default is 0)


src/ngraph/op/broadcast_distributed.hpp
```diff
@@ -27,12 +27,13 @@ namespace ngraph
...
         {
         public:
-            BroadcastDistributed(const std::shared_ptr<Node>& arg);
+            BroadcastDistributed(const std::shared_ptr<Node>& arg, int root_id = 0);
 
             void validate_and_infer_types() override;
 
             virtual std::shared_ptr<Node>
                 copy_with_new_args(const NodeVector& new_args) const override;
+            const int m_root_id;
```
aprocter: This should be private, with an accessor function `get_root_id`.

PR #3054, opened 2019-06-13
Add QuantizedConvolution op to ngraph core

nbpatel:


silee2:
`src/ngraph/frontend/onnx_import/op/quant_conv.cpp` still includes the removed `ngraph/op/experimental/quantized_conv.hpp` And this causes a compile error.

diyessi:
The TF bridge is probably the same way. Can you leave behind a header with an `using` so we can stay source compatible for two releases? Also, add a note about the experimental one being deprecated in the changes file.

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+        INPUT_SCALE = 2,
+        INPUT_ZERO_POINT,
+        FILTER_SCALE,
+        FILTER_ZERO_POINT,
+        OUTPUT_SCALE,
+        OUTPUT_ZERO_POINT
+    };
+
+    NODE_VALIDATION_CHECK(this,
+                          get_input_element_type(INPUT_SCALE).is_real() ||
```
aprocter: `get_input_element_type(INPUT_SCALE).is_dynamic() || get_input_element_type(INPUT_SCALE).is_real() || ...`

Same thing for `FILTER_SCALE` and `OUTPUT_SCALE`.

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+                              get_input_element_type(FILTER_SCALE).is_real() ||
+                              get_input_element_type(OUTPUT_SCALE).is_real(),
+                          "Scale must be a floating point number");
+
+    NODE_VALIDATION_CHECK(this,
+                          get_input_element_type(0) == get_input_element_type(INPUT_ZERO_POINT),
```
aprocter: `get_input_element_type(0).compatible(get_input_element_type(INPUT_ZERO_POINT))`

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+                          "Input Zero point element type (",
+                          get_input_element_type(INPUT_ZERO_POINT),
+                          ") must match input element type (",
+                          get_input_element_type(0),
+                          ")");
+
+    NODE_VALIDATION_CHECK(this,
+                          get_input_element_type(1) == get_input_element_type(FILTER_ZERO_POINT),
```
aprocter: `get_input_element_type(1).compatible(get_input_element_type(FILTER_ZERO_POINT))`

src/ngraph/op/quantized_convolution.cpp
aprocter: Are the types for `INPUT_SCALE`, `FILTER_SCALE`, and `OUTPUT_SCALE` allowed to be different from each other? If not, we should check them here.

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+                          "Filter Zero point element type (",
+                          get_input_element_type(FILTER_ZERO_POINT),
+                          ") must match filter element type (",
+                          get_input_element_type(1),
+                          ")");
+
+    // TODO Remove these checks once we support channelwise and vector of scales
+    NODE_VALIDATION_CHECK(this,
+                          shape_size(get_input_shape(2)) == 1 ||
```
aprocter: (This one will be a bit of a pain because a needed utility function is not defined yet.)

We'll need to define a `shape_size`-like function that works on `PartialShape`s. This should have a return type of `Dimension`, and the expression `partial_shape_size(pshape)` should return:

(1) `Dimension::dynamic()` if the rank of `pshape` PartialShape is dynamic
(2) `Dimension::dynamic()` if any dimension of `pshape` is unknown (except, I suppose, it could return 0 if there is a 0 anywhere)
(3) The same value as `shape_size(pshape.to_shape())`, if `pshape.is_static()` is true.

Then, this check would become `partial_shape_size(get_input_partial_shape(2)).compatible(1)`.

aprocter: Similar for all `shape_size` checks in the next few lines here.

aprocter: I'm a little confused about this check, actually... why is it `||`?

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+                          shape_size(get_input_shape(4)) == 1 ||
+                              shape_size(get_input_shape(5)) == 1,
+                          "Filter scale and filter zero point shape must be same and 1");
+
+    NODE_VALIDATION_CHECK(this,
+                          shape_size(get_input_shape(6)) == 1 ||
+                              shape_size(get_input_shape(7)) == 1,
+                          "Output scale and output zero point shape must be same and 1");
+
+    auto input_shape = get_input_shape(0);
```
aprocter: Generally we cannot use `get_input_shape` in validation anymore. We have to use `get_input_partial_shape`, which returns a `PartialShape`.

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    if (m_padding_above.size() == 0)
+    {
+        m_padding_above = conv_default_padding(this, input_shape, filters_shape);
+    }
+
+    set_output_type(0,
+                    m_output_type,
+                    util::infer_convolution_output_shape(this,
```
aprocter: Please use `infer_convolution_forward` from `validation_util.[ch]pp` instead.

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+                          shape_size(get_input_shape(6)) == 1 ||
+                              shape_size(get_input_shape(7)) == 1,
+                          "Output scale and output zero point shape must be same and 1");
+
+    auto input_shape = get_input_shape(0);
+    auto filters_shape = get_input_shape(1);
+
+    if (m_data_dilation_strides.size() == 0)
+    {
+        m_data_dilation_strides = conv_default_strides(this, input_shape, filters_shape);
```
aprocter: These defaults should not be stored in this way. If the input shapes change and we need to revalidate, we need to recompute the default values.

aprocter: Similar comment for the other attributes in the next few `if` statements.

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+                                                         m_data_dilation_strides,
+                                                         0, /* batch_axis_data,              */
+                                                         1, /* input_channel_axis_data,      */
+                                                         1, /* input_channel_axis_filters,   */
+                                                         0, /* output_channel_axis_filters,  */
+                                                         0, /* batch_axis_result,            */
+                                                         1  /* output_channel_axis_result,   */
+                                                         ));
+    NODE_VALIDATION_CHECK(this,
+                          get_output_element_type(0) == get_input_element_type(OUTPUT_ZERO_POINT),
```
aprocter: `get_output_element_type(0).compatible(get_input_element_type(OUTPUT_ZERO_POINT))`.

test/cpu_test.cpp
```diff
@@ -1893,3 +1894,59 @@ TEST(cpu_test, tensor_copy_from_different_layout)
 
     EXPECT_EQ((vector<uint8_t>{1, 4, 2, 5, 3, 6}), read_vector<uint8_t>(b));
 }
+
+// Adding this test case in cpu_test
+// because reference kernel isn't supporting intermediate
+// output types as of now
+TEST(cpu_test, quantized_conv_int32_output)
```
aprocter: If this op is being moved to core, we should put these tests in `backend_test.in.cpp` (or something like `backend_conv.in.cpp`) and add it to the manifest for backends that don't yet support it.

src/ngraph/runtime/cpu/builder/convolution.cpp
```diff
@@ -125,7 +125,12 @@ namespace ngraph
                                padding_below,
                                padding_above,
                                data_dilation_strides,
-                               1.0f);
+                               nullptr,
+                               nullptr,
+                               nullptr,
+                               nullptr,
+                               nullptr,
+                               nullptr);
```
jbobba: Aren't these optional arguments? Remove.

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
```diff
@@ -73,20 +78,22 @@ namespace ngraph
...
                         // Create MKLDNN convolution primitive during the first iteration.
                         // Assumes the scales dont change for the duration of the graph
                         if (ctx->first_iteration)
                         {
                             vector<float> dyn_scales;
-                            dyn_scales.assign(
-                                static_cast<float*>(ctx->buffer_data[arg2_buffer_index]),
-                                static_cast<float*>(ctx->buffer_data[arg2_buffer_index]) +
-                                    scales_size);
+                            // Calculate the requantization scale
```
jbobba: requantization scale could be a vector of values. You seem to be only picking up the first value.

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -191,10 +191,18 @@ namespace ngraph
                 auto scale_index = mkldnn_emitter->get_scale_index<OP>();
                 auto scales_size = shape_size(node->get_input_shape(scale_index));
                 writer << "std::vector<float> dyn_scales;\n";
-                writer << "dyn_scales.assign(" << args[scale_index].get_name() << ", "
-                       << args[scale_index].get_name() << " + " << std::to_string(scales_size)
-                       << ");\n";
-
+                if (is_same<OP, ngraph::op::QuantizedConvolution>())
+                {
+                    writer << "dyn_scales.push_back(*" << args[2].get_name() << " * "
```
jbobba: same comment about vector of scales.

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+    // TODO Remove these checks once we support channelwise and vector of scales
+    NODE_VALIDATION_CHECK(this,
+                          get_input_partial_shape(2).compatible(PartialShape{}) &&
+                              get_input_partial_shape(3).compatible(PartialShape{}),
+                          "Input scale and input zero point shape must be same and 1");
+
+    NODE_VALIDATION_CHECK(this,
+                          get_input_partial_shape(4).compatible(PartialShape{}) == 1 &&
+                              get_input_partial_shape(5).compatible(PartialShape{}) == 1,
+                          "Filter scale and filter zero point shape must be same and 1");
```
jbobba: Why "==1"?

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+    NODE_VALIDATION_CHECK(
+        this,
+        get_input_element_type(1).compatible(get_input_element_type(FILTER_ZERO_POINT)),
+        "Filter Zero point element type (",
+        get_input_element_type(FILTER_ZERO_POINT),
+        ") must match filter element type (",
+        get_input_element_type(1),
+        ")");
+
+    // TODO Remove these checks once we support channelwise and vector of scales
```
jbobba: mxnet bridge already uses vector of scales. Why are we breaking compatibility with that? We could let the op definition support a vector and its ok for reference kernel to not support it for now.

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
```diff
@@ -748,9 +748,25 @@ namespace ngraph
...
+                    auto output_zero_point =
+                        dynamic_pointer_cast<ngraph::op::Constant>(qconv->get_argument(7));
                     if (node->get_input_element_type(0) == element::u8 &&
                         node->get_input_element_type(1) == element::i8)
                     {
+                        // Mkldnn assumes zero point to be zero
+                        if (input_zero_point == nullptr || filter_zero_point == nullptr ||
+                            output_zero_point == nullptr || !(ngraph::is_zero(input_zero_point)) ||
+                            !(ngraph::is_zero(filter_zero_point)) ||
+                            !(ngraph::is_zero(output_zero_point)))
```
jbobba: is_zero checks for node type being a constant. you dont need to do the type check here.

src/ngraph/op/quantized_convolution.cpp
```diff
@@ -0,0 +1,196 @@
...
+                                               const shared_ptr<Node>& input_scale,
+                                               const shared_ptr<Node>& input_zero_point,
+                                               const shared_ptr<Node>& filter_scale,
+                                               const shared_ptr<Node>& filter_zero_point,
+                                               const shared_ptr<Node>& output_scale,
+                                               const shared_ptr<Node>& output_zero_point,
+                                               const element::Type& output_type,
+                                               const AxisSet& input_axes,
+                                               const AxisSet& filter_axes,
+                                               const AxisSet& output_axes)
```
jbobba: make the axes arguments optional since we dont expect most users to care about these.

src/ngraph/runtime/cpu/builder/quantized_conv.cpp
nbpatel: From https://intelpedia.intel.com/NGraphDesign/Quantization 
---Add axes to allow for non-scalar scale and zero point (**OK to force to scalar for now)**
didn't we agree to assume scale to be scalar for now?

ashokei: it still needs to take in as vector, but you can force/assume to support only vector[0] ? (ok to assume/force to single/scalar val). Easier to extend if we make it vector.

nbpatel: It is being stored in a vector and passed to mkldnn (because mkldnn expects a vector).

src/ngraph/runtime/cpu/pass/cpu_assignment.cpp
nbpatel: is_zero checks if a constant node is zero or not. I do dynamically cast it to Constant and check for if its value to be zero. Zero point should be zero and Constant for mkldnn kernel to kick in

silee2:
@nbpatel 
```
[ 58%] Building CXX object src/ngraph/runtime/cpu/CMakeFiles/cpu_backend.dir/pass/cpu_fusion.cpp.o
cd /home/dockuser/ngraph/BUILD-CLANG/src/ngraph/runtime/cpu && /usr/bin/clang++-3.9   -DCPU_BACKEND_DLL_EXPORTS -DIN_NGRAPH_LIBRARY -DNGRAPH_ENABLE_CPU_CONV_AUTO -DNGRAPH_VERSION=\"0.22.0-rc.0+4b5560f\" -DPROJECT_ROOT_DIR=\"/home/dockuser/ngraph\" -Dcpu_backend_EXPORTS -I/home/dockuser/ngraph/src/ngraph -isystem /home/dockuser/ngraph/src/ngraph/runtime/cpu/libmkldnn -I/home/dockuser/ngraph/src -isystem /home/dockuser/ngraph/BUILD-CLANG/mkldnn/include -isystem /home/dockuser/ngraph/BUILD-CLANG/eigen/src/ext_eigen -isystem /home/dockuser/ngraph/BUILD-CLANG/tbb/tbb-src/include -isystem /home/dockuser/ngraph/BUILD-CLANG/json/src/ext_json/include  -DEIGEN_MPL2_ONLY -DTBB_USE_THREADING_TOOLS -O2 -Wformat -Wformat-security -D_FORTIFY_SOURCE=2 -fstack-protector-all -Werror -march=native -DNGRAPH_CPU_ENABLE -O3 -DNDEBUG -fPIC   -Werror=return-type -Werror=inconsistent-missing-override -Werror=comment -pedantic-errors -Weverything -Wno-gnu-zero-variadic-macro-arguments -Wno-c++98-compat-pedantic -Wno-weak-vtables -Wno-global-constructors -Wno-exit-time-destructors -Wno-missing-prototypes -Wno-missing-noreturn -Wno-switch -Wno-switch-enum -Wno-covered-switch-default -Wno-undef -Wno-float-conversion -Wno-sign-conversion -Wno-padded -Wno-sign-compare -Wno-unused-parameter -Wno-conversion -Wno-double-promotion -Wno-undefined-func-template -std=c++11 -o CMakeFiles/cpu_backend.dir/pass/cpu_fusion.cpp.o -c /home/dockuser/ngraph/src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
/home/dockuser/ngraph/src/ngraph/runtime/cpu/pass/cpu_fusion.cpp:1968:18: error: declaration shadows a local variable [-Werror,-Wshadow]
            auto requantization_scale =
                 ^
/home/dockuser/ngraph/src/ngraph/runtime/cpu/pass/cpu_fusion.cpp:1873:10: note: previous declaration is here
    auto requantization_scale = std::make_shared<pattern::op::Label>(element::f32, Shape{});
         ^
1 error generated.
```

test/cpu_test.cpp
```diff
@@ -34,6 +34,7 @@
 #include "ngraph/op/fused/conv_fused.hpp"
 #include "ngraph/op/get_output_element.hpp"
 #include "ngraph/op/parameter.hpp"
+#include "ngraph/op/quantized_convolution.hpp"
```
aprocter: No longer necessary to add this #include.

src/ngraph/op/quantized_convolution.cpp
aprocter: I think this must be residue from the old check, which was `shape_size(foo) == 1`?

nbpatel: yes, it was :)

PR #3057, opened 2019-06-13
Backward compatibility for GenerateMask deserialize

gaurides:


src/ngraph/serializer.cpp
```diff
@@ -1006,11 +1006,15 @@ static shared_ptr<ngraph::Function>
                 auto type = read_element_type(node_js.at("type"));
                 auto seed = node_js.at("seed").get<unsigned int>();
                 auto probability = node_js.at("probability").get<double>();
-                auto use_seed_maybe = node_js.at("use_seed");
                 bool use_seed = false;
-                if (!use_seed_maybe.empty())
+                if (args.size() > 1)
```
rkimballn1: I think what you want here instead is `bool use_seed = get_or_default<bool>(node_js, "use_seed", false);`

PR #3068, opened 2019-06-14
[WIP] fix qconcat corner cases

ashokei:


src/ngraph/builder/quantization.cpp
```diff
@@ -101,8 +102,19 @@ namespace ngraph
...
 
             // output scale
-            auto min = make_shared<op::Min>(make_shared<op::Concat>(mins, 0), ngraph::AxisSet{0});
-            auto max = make_shared<op::Max>(make_shared<op::Concat>(maxs, 0), ngraph::AxisSet{0});
+            std::shared_ptr<Node> min =
+                make_shared<op::Min>(make_shared<op::Concat>(mins, 0), ngraph::AxisSet{0});
+            std::shared_ptr<Node> max =
+                make_shared<op::Max>(make_shared<op::Concat>(maxs, 0), ngraph::AxisSet{0});
+            auto zero = make_constant(min->get_element_type(), min->get_shape(), 0);
+            // ensure min is no more than zero.
```
diyessi: Why?

ashokei: @diyessi currently our's mismatches with TF

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/quantized_concat_op.cc#L93

With these changes, we should be similar. (calculate total out range first, then dequantize each input with input min/max, and quantize next using outut min/max).

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/quantized_concat_op.cc#L53

Still, this PR is WIP as the unit-test from TF bridge is not passing yet, need to check the actual tensor values / args we receive in core from tf-bidge (as the calculation seems same now).

mbrookhart:
@ashokei Is this PR dead? I just noticed it, went up while I was on bonding leave.

ashokei:
current qconcat in ngraph supports newer version, closing this PR due to older version being used in TF unittest (details were added in JIRA ticket).

PR #3071, opened 2019-06-14
Add DynReplaceSlice op; make Dimension support negatives

aprocter:
Adds a DynReplaceSlice op, suitable (among other things) for backprop of DynSlice (though I haven't implemented generate_adjoints yet because we have broader issues with dyn shapes in autodiff that haven't been tackled yet). Adds the op, type prop tests, DynElimination, and execution tests.

Another major change in this PR is that `Dimension` (and therefore `PartialShape`) has been changed to include negative values. This felt necessary in order to support the overlap of dynamicness with right-indexiness in `infer_slice_shape`, and I think it will come up elsewhere in the future.

Since validation logic generally will not have safeguards against negative dimensions in the shape, I've put a check in `set_tensor_type` to throw when dimensions are negative; indirectly, this will trigger if `validate_and_infer_types` tries to set a negative shape, which means any issues should be detected. 🤞


src/ngraph/validation_util.cpp
```diff
@@ -628,3 +628,174 @@ void ngraph::infer_auto_padding(const Shape& image_shape,
         padding_above.push_back(pad_type == op::PadType::SAME_UPPER ? padding_rhs : padding_lhs);
     }
 }
+
+PartialShape ngraph::infer_slice_shape(const Node* node,
```
aprocter: attn @jbobba: I moved this from DynSlice into validation_util (so that DynReplaceSlice can use it), and extended it to support partiality.

aprocter:
Adding `Changes API` because we have to remove the `ptrdiff_t` conversion from `Dimension`. I don't think this should affect anything outside of the ngraph repo itself, but can't say for sure.

aprocter:
Back to WIP because I just realized that *cough* a lot of the test logic was #if 0'd out.

src/ngraph/dimension.cpp
```diff
@@ -40,7 +40,7 @@ std::ostream& ngraph::operator<<(std::ostream& str, const Dimension& dimension)
 {
     if (dimension.is_static())
     {
-        return (str << size_t(dimension));
+        return (str << int64_t(dimension));
```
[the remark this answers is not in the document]
aprocter: Will likely get rid of these in a later PR, either by avoiding casts or by converting to `static_cast`.

src/ngraph/serializer.cpp
```diff
@@ -941,14 +942,46 @@ static shared_ptr<ngraph::Function>
...
+            }
             case OP_TYPEID::DynReshape:
             {
                 node = make_shared<op::DynReshape>(args[0], args[1]);
                 break;
             }
             case OP_TYPEID::DynSlice:
             {
-                node = make_shared<op::DynSlice>(args[0], args[1], args[2], args[3]);
+                auto lower_bounds_mask = node_js.at("lower_bounds_mask").get<set<size_t>>();
```
aprocter: attn @rkimballn1: While I was here, I noticed `DynSlice` was not serializing its attributes. NB, the update here is not backwards compatible. Since this is an `experimental` op I think that's probably okay, but let me know if you want me to fix that.

PR #3073, opened 2019-06-17
Recursive FusedOpDecomposition pass.

arogowie-intel:
Run recursively FusedOpDecomposition pass untill no more fused ops available.


src/ngraph/pass/fused_op_decomposition.hpp
```diff
@@ -25,12 +28,16 @@ namespace ngraph
...
+            ///         is supported by backend.
             using op_query_t = std::function<bool(const Node& node)>;
 
             FusedOpDecomposition(op_query_t callback = nullptr);
             bool run_on_node(std::shared_ptr<ngraph::Node> node) override;
 
         private:
+            /// \brief A function returning whether provided Node is supported by current backend.
+            ///        The returned value is used to control whether decompose operator or not.
             op_query_t m_callback = nullptr;
```
tomdol: Any chance we could rename this member so that at the call site it says what is checked? I'm thinking about something like: `is_node_supported_by_backend` or similar/shorter.

arogowie-intel: Done

PR #3077, opened 2019-06-17
[PRIVATE/MLIR] Support MLIR lowering of Relu

jbobba:


src/contrib/mlir/lowerer.cpp
```diff
@@ -336,6 +338,56 @@ namespace
         rewriter.replaceOp(op, {result});
     }
 
+    // ADD
```
nmostafa: Relu

src/contrib/mlir/lowerer.cpp
```diff
@@ -336,6 +338,56 @@ namespace
...
+        Type elemTy = lhs->getType().dyn_cast<MemRefType>().getElementType();
+        NGRAPH_CHECK(!elemTy.isa<FloatType>(),
+                     "NGReluOp with float element type should not be lowered until MLIR supports "
+                     "lowering !std.CmpF");
+
+        // clang-format off
+        LoopNestBuilder(pivs, lbs, ubs, steps)(
+            // single stmt body
+            [&] {
+                    if (auto floatTy = elemTy.dyn_cast<FloatType>()) {
```
nmostafa: I thought comparison intrinsics are there. Why not just do 
`iRes(ivs) = intrinsics::select(iLHS(ivs) < zero, iLHS(ivs), zero);`

jbobba: iLHS(ivs) is of type TemplatedIndexValue and we dont have overloaded comparisons for that. https://github.com/tensorflow/mlir/blob/master/include/mlir/EDSC/Helpers.h#L217-#L258

nmostafa: Comparison operators exist for `ValueHandle` though, and `TemplateIndexValue` has a cast operator to `ValueHandle`. You should be able to use the operator. 
```
ValueHandle mlir::edsc::op::operator<(ValueHandle lhs, ValueHandle rhs) {
  // TODO(ntv,zinenko): signed by default, how about unsigned?
  return createComparisonExpr(CmpIPredicate::SLT, lhs, rhs);
}
```

nmostafa:
Moved to https://github.com/NervanaSystems/ngraph/pull/3197

PR #3080, opened 2019-06-17
Address: https://github.com/intel/mkl-dnn/pull/483

silee2:
Fix Compilation error with GCC 9


silee2:
@diyessi: @indie reported a build failure with ClearLinux (GCC 9). It is a known issue with mkldnn and they already have a fix checked in. Should be update mkldnn to fix it in this release or bump it to next release? We use mkldnn tag v0.19 on branch rls-v0.19. rls-v0.19 has one more commit after v0.19 which is the patch for the issue.

PR #3087, opened 2019-06-18
[ONNX] Lp/InstanceNormalization operations.

arogowie-intel:
Add two new ONNX operators along with node validation macro based on ngraph NGRAPH_CHECK macro.


aslepko:
TESTNOW

PR #3088, opened 2019-06-18
First cut at async interface for backend tensor allocation and execution

rkimballn1:
Adding an async interface to the backend execute and tensor reads/writes. This is to enable double-buffering.


silee2:
Can you put a description? (What problem you are trying to solve, why this solution ...)

src/ngraph/runtime/backend.cpp
```diff
@@ -90,3 +97,150 @@ std::shared_ptr<runtime::Executable> runtime::Backend::load(istream& input_strea
...
+    case AsyncEvent::Type::EXECUTE:
+        event->get_executable()->call(event->get_outputs(), event->get_inputs());
+        event->signal_result();
+        break;
+    }
+}
+
+void runtime::Backend::async_thread_entry()
+{
+    unique_lock<std::mutex> lock(m_event_queue_mutex);
```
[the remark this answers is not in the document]
rkimballn1: It is not single threaded. The queue has one read thread and one write thread so mutex is needed.

src/ngraph/runtime/backend.cpp
```diff
@@ -90,3 +97,150 @@ std::shared_ptr<runtime::Executable> runtime::Backend::load(istream& input_strea
 {
     throw runtime_error("load opertion unimplemented.");
 }
+
+runtime::Backend::AsyncEvent::AsyncEvent(Type type,
```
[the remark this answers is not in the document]
rkimballn1: The reason it is here is to scope it properly. It is a backend async event so I thought it best to be a Backend::AsyncEvent. It is not a general-purpose async event and is tightly coupled with the Backend.

src/ngraph/runtime/tensor.cpp
```diff
@@ -95,3 +97,41 @@ void runtime::Tensor::copy_from(const ngraph::runtime::Tensor& source)
...
+    {
+        // auto f = m_promise.get_future();
+        return m_backend->post_async_write_event(
+            shared_from_this(), p, size_in_bytes, buffer_number);
+    }
+    else
+    {
+        throw runtime_error("Async operations not supported for this backend");
+    }
+    // using namespace std::placeholders;
```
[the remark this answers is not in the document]
rkimballn1: cleaned up.

src/ngraph/runtime/executable.hpp
```diff
@@ -51,6 +54,16 @@ class ngraph::runtime::Executable
     bool call_with_validate(const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
                             const std::vector<std::shared_ptr<runtime::Tensor>>& inputs);
 
+    /// \brief Asynchronously executes a single iteration of the Function. The `future` is used
+    ///     to monitor the execution.
+    /// \param outputs vector of runtime::Tensor used as outputs
+    /// \param inputs vector of runtime::Tensor used as inputs
+    /// \returns a valid std::future to monitor the execution. Use future.get() to get the results
+    ///     or future.wait*() to wait for completion.
+    virtual std::future<void>
```
earhart: This is the classic problem with `std::future<>` -- there's no `.then()` method, so you either have to poll or burn a thread waiting.  :-/

It might be good to return a different object, or take a callback.

src/ngraph/runtime/tensor.cpp
```diff
@@ -95,3 +97,41 @@ void runtime::Tensor::copy_from(const ngraph::runtime::Tensor& source)
...
+{
+    if (m_backend)
+    {
+        // auto f = m_promise.get_future();
+        return m_backend->post_async_write_event(
+            shared_from_this(), p, size_in_bytes, buffer_number);
+    }
+    else
+    {
+        throw runtime_error("Async operations not supported for this backend");
```
earhart: What's the caller supposed to do if async isn't supported?

test/async.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    // get() waits for the result to be ready
+    future.get();
+
+    auto future_r = r->begin_read(data_r.data(), data_r.size() * sizeof(float), 0);
+    ASSERT_TRUE(future_r.valid());
+
+    EXPECT_EQ(future_a.wait_for(ten_ms), future_status::ready);
+    EXPECT_EQ(future_b.wait_for(ten_ms), future_status::ready);
+    EXPECT_EQ(future_r.wait_for(ten_ms), future_status::ready);
```
earhart: The final one here should be `ASSERT_EQ()` -- no point going on if it fails.

sfvaroglu:
We had a sync up with @gcwenger and talked about tagging tensors for each iteration. The current API does not support tagging. The bridge plans to use multiple sets of tensors, e.g. set 0 for iteration 0, set 1 for iteration 1, etc. Ideally, the bridge should get a notification when it's okay to overwrite tensors. Would like this incorporated into the async API, @ashokei please follow up on this. Thanks

earhart:
I'm not sure why most of this is needed -- I might not be understanding the problem correctly?

It'd be useful to have a standard way to get a callback when operations complete.  But even without that, the backend is currently free to implement most things asynchronously if it chooses: when a caller invokes `Executable::call()`, the backend is free to return the thread to the caller immediately -- "Yup, we're done, what's next?"  The only time side effects become visible is when the caller goes to read data from an output tensor.

(This is how the PlaidML backend currently operates -- which is why nbench copies data back from the backend by default now; before that, we had some _really_ impressive benchmark results.  :-)

Also: there might be other long-running operations, like `Backend::compile()`.

Lastly: if we're double-buffering, I suspect it'd be useful to have a mechanism for pre-allocating the resource reservations used by the network (e.g. temporary memory).  I'm not quite sure what the right way is to think about that; perhaps we could pass options into `Executable` compilation -- "As you compile this, please assume that we're going to be running N instantiations at once, and make your memory allocation decisions appropriately"?

ashokei:
> The only time side effects become visible is when the caller goes to read data from an output tensor."

I notice other frameworks use this mechanism (tensor read/copy as a blocking/lock to indicate completion).  When the user tries to read output tensor , it either blocks for compute to be done or returns tensor data immediately. This kind of tensor tagging, can help in tracking linear-to-tile processing on them too. 

inputs -> linear-to-tile -> compute -> outputs

if tensors know when their linear-to-tile processing is done, frontend can copy next batch into inputs (without waiting for compute) ?

Or, a second (double buffer) set of tensors could be maintained, but tensor level tracking for task completion will help here too.

earhart:
You can also handle inputs by having the backend pin the inputs as long as it needs to (which might be a lot less than the network latency, depending on the network), or letting the backend handle the double-buffering (map-discard semantics, and the backend is free to use different memory).

src/ngraph/runtime/executable.hpp
rkimballn1: It seems to me that doing a callback only pushes the issue down one level. There is a new thread created which does the callback and that thread would be waiting on some future-like thing. The future is the stl way of doing this.

src/ngraph/runtime/tensor.cpp
rkimballn1: Contact the backend developer and get them to support it. In the future we can make this API required but for now it is optional.

rkimballn1:
@earhart Agree with you that this can be done more easily. I will rework it and let you know.

rkimballn1:
Big rework and simplification of this PR. Removed nbench changes as those can be added in a separate PR.

src/ngraph/runtime/executable.hpp
```diff
@@ -73,6 +73,36 @@ class ngraph::runtime::Executable
...
+
+    /// \brief Create a vector of output Tensors
+    /// \param output_index The index position in the output Result vector. This would be the same
+    /// order of Results passed into the outputs in the call() method.
+    /// \param pipeline_depth The number of stages in the output pipeline. For double-buffered output
+    /// you would specify pipeline_depth=2
+    /// \returns A vector of Tensors, one for each stage of the pipeline
+    virtual std::vector<std::shared_ptr<runtime::Tensor>> create_output_tensor(size_t output_index,
+                                                              size_t pipeline_depth);
+
```
shresthamalik: General questions: 
Going forward, will this be the recommended way to create input and output tensors?

Will the executable call() be able to take in inputs/outputs created by the backend->create_tensor() PI? 
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/backend.hpp#L68

Right now, theoretically it is possible to pass in the same input tensors to different executables (if the shape and size match) will that be still a possibility, say if we store the tensor?

gcwenger: @shresthamalik These new methods would be preferable because they tell us more.
Your question re tying together is a good one. That should still be ok, but we'll need to ensure it.
Let's chat in brief meeting next week.

rkimballn1: I would like to deprecate the Backend methods for creating tensors but that is a big issue and left for later.

csullivan: There are a number of optimizations we can do if we can guarantee the tensors which are passed in to multiple functions/executables are handled similarly. 

For example, if tensor0 (layout NHWC) is used by executableA and executableB, if both executables have the same boundary layout transformation (NHWC -> NCHW) then we can fuse the reshape into the DMA-write pipeline and skip the reshape on device for both executables A and B. But if executable A does the NHWC->NCHW conversion, but executable B does not, then we can't make the optimization.

We will handle this decision making in the backend, but the bridge teams should be aware of it so as to ensure optimal layout scheduling for the executable where possible.

PR #3097, opened 2019-06-18
Cyphers/mergeser

diyessi:
This is the first part of some serialization reorganization to later support:
* Serializing Output information (node+output number) for op inputs instead of nodes
* Support Output information in attributes
* Support functions sharing graph structure

This PR separates node [de]serialization from function [de]serialization, which changes indentation in the two giant switches leading to collisions when ops are added, so I would like to get it merged before moving on to Output information in attributes for TensorIterator.


src/ngraph/serializer.cpp
```diff
@@ -492,1228 +526,1246 @@ T get_value(nlohmann::json js, const string& key)
...
+    return m_node_map.at(name);
+}
 
+shared_ptr<Function> JSONDeserializer::deserialize_function(json& func_js)
+{
     string func_name = func_js.at("name").get<string>();
-    vector<string> func_parameters = func_js.at("parameters").get<vector<string>>();
-    vector<string> func_result = func_js.at("result").get<vector<string>>();
-    unordered_map<string, shared_ptr<Node>> node_map;
+    vector<json> func_parameters = func_js.at("parameters"); //.get<vector<json>>();
```
rkimballn1: Does get<vector<json>>() not work if you include it. Can you just include that code uncommented? If not then delete the comment.

src/ngraph/serializer.cpp
```diff
@@ -492,1228 +526,1246 @@ T get_value(nlohmann::json js, const string& key)
...
+        else
+        {
+            result.push_back(std::make_shared<op::Result>(fr));
+        }
+    }
+
+    if (results != 0 && results != func_result.size())
+    {
+        throw ngraph_error(
+            " Graph serialization is inconsistent. Some op::Results appear to be missing");
```
rkimballn1: Why the leading space before "Graph serialization"?

src/ngraph/serializer.cpp
```diff
@@ -492,1228 +526,1246 @@ T get_value(nlohmann::json js, const string& key)
...
+            break;
+        }
+        case OP_TYPEID::Cos:
+        {
+            node = make_shared<op::Cos>(args[0]);
+            break;
+        }
+        case OP_TYPEID::Cosh:
+        {
+            node = make_shared<op::Cos>(args[0]);
```
rkimballn1: This should be `Cosh`?

src/ngraph/serializer.cpp
```diff
@@ -492,1228 +526,1246 @@ T get_value(nlohmann::json js, const string& key)
...
-            }
-            case OP_TYPEID::QuantizedConvolutionBiasSignedAdd: { break;
-            }
-            case OP_TYPEID::QuantizedConvolutionRelu: { break;
-            }
-            case OP_TYPEID::QuantizedConvolution:
-            {
-                auto window_movement_strides =
-                    node_js.at("window_movement_strides").get<vector<size_t>>();
-                auto win
```
rkimballn1: Was this indented incorrectly? Seems a lot is just indenting.

src/ngraph/serializer.cpp
diyessi: Those work; I forgot to delete them. Will do.

src/ngraph/serializer.cpp
diyessi: Got it.

PR #3098, opened 2019-06-19
Add (dynamic) Range op

aprocter:
This is like `np.arange` or `tf.range`.

TODO:
- [x] Add execution tests


aprocter:
> Looks good to me. On a side note, type_prop is getting large and unwieldy. We need to start breaking it up.

@jbobba Agreed. (Not sure if you noticed, but in this PR I had to turn `/bigobj` on for the Windows build because of this. :))

PR #3099, opened 2019-06-19
Tracer of real data in ngraph-CPU for debug purpose

andreykhaykin:
Hi,

This PR adds a new tool for tracing kernels with real data from one iteration.

This tool helped a lot with different models in PP/TF for finding bugs in convergence.
We use it as a reference to "same" logs in lower layers of the full stack and also compare it with frameworks.
(AKA, side by side - "sbs" tool)

Activate this tool by set this env flag:
**NGRAPH_CPU_DEBUG_TRACER=1** 
it will dump trace_meta.log and trace_bin_data.log

or you can specify the names of logs with those flags:
NGRAPH_TRACER_LOG = "meta.log"
NGRAPH_BIN_TRACER_LOG = "bin.log"

the meta_log line looks : 
 kernel_name , serial_number_of_op, tensor_id, symbol_of_in_out, num_elements, shape, binary_data_offset, mean_of_tensor, variance_of_tensor.

line example from unit-test:
 K=Add   S=0   TID=0_0   >> size=4 Shape{2, 2} bin_data_offset=8 mean=1.5 var=1.25

the binary_log line looks:
tensor_id, binary data (tensor data)

I made questionable decisions in the implementation of this tool in ngraph, so any feedback is welcome.


andreykhaykin:
Hi, one test fails - can't open file/log from unit-test on Centos with docker.
In ubuntu from source ngraph repository, it passes. So what I should do? (how I fix/check it?)

src/ngraph/runtime/cpu/cpu_debug_tracer.cpp
```diff
@@ -0,0 +1,109 @@
...
+{
+}
+
+runtime::cpu::CPU_DebugTracer& runtime::cpu::CPU_DebugTracer::getInstance()
+{
+    static CPU_DebugTracer instance;
+    return instance;
+}
+
+void runtime::cpu::CPU_DebugTracer::init_streams(const char* trace_file_path,
```
rkimballn1: Make these `const string&`

src/ngraph/runtime/cpu/cpu_debug_tracer.hpp
```diff
@@ -0,0 +1,66 @@
...
+            private:
+                CPU_DebugTracer();
+
+                size_t m_serial_number;
+                std::fstream m_tracer_stream;
+                std::fstream m_tracer_bin_stream;
+            };
+        }
+    }
+}
```
rkimballn1: Missing CR at EOF

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1268,6 +1300,17 @@ void runtime::cpu::CPU_ExternalFunction::build(ngraph::pass::PassConfig& pass_co
     register_common_passes(pass_manager, pass_config);
     pass_manager.run_passes(m_function, false);
 
+    static const auto do_ngraph_trace = std::getenv("NGRAPH_TRACER_LOG");
+    auto& debug_tracer = runtime::cpu::CPU_DebugTracer::getInstance();
+
+    if (do_ngraph_trace != nullptr)
+    {
+        const char* trace_log_file = std::getenv("NGRAPH_TRACER_LOG");
```
rkimballn1: You already have this value in do_ngraph_trace, don't get it a second time.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1268,6 +1300,17 @@ void runtime::cpu::CPU_ExternalFunction::build(ngraph::pass::PassConfig& pass_co
...
+        const char* bin_log_file = std::getenv("NGRAPH_BIN_TRACER_LOG");
```
rkimballn1: make this static. You also need to make sure this is not null or you will segfault later.

test/cpu_debug_tracer.cpp
```diff
@@ -0,0 +1,118 @@
...
+    shared_ptr<runtime::Tensor> result = backend->create_tensor(element::f32, shape);
+
+    copy_data(a, vector<float>{0, 1, 2, 3});
+    copy_data(b, vector<float>{1, 2, 3, 4});
+
+    const char* trace_log_file = "trace_meta.log";
+    const char* bin_log_file = "trace_bin.log";
+
+    set_environment("NGRAPH_TRACER_LOG", trace_log_file, 1);
+    set_environment("NGRAPH_BIN_TRACER_LOG", bin_log_file, 1);
```
rkimballn1: You should probably have a unit test for when `NGRAPH_TRACER_LOG` is defined but `NGRAPH_BIN_TRACER_LOG` is not set. Currently your  code will likely segfault on that condition. You can't trust users to set both of these.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1268,6 +1300,17 @@ void runtime::cpu::CPU_ExternalFunction::build(ngraph::pass::PassConfig& pass_co
     register_common_passes(pass_manager, pass_config);
     pass_manager.run_passes(m_function, false);
 
+    static const auto do_ngraph_trace = std::getenv("NGRAPH_TRACER_LOG");
```
rkimballn1: Add support to allow this to be set programmatically, not only via environment variables

test/cpu_debug_tracer.cpp
```diff
@@ -0,0 +1,118 @@
...
+    shared_ptr<runtime::Tensor> b = backend->create_tensor(element::f32, shape);
+    shared_ptr<runtime::Tensor> result = backend->create_tensor(element::f32, shape);
+
+    copy_data(a, vector<float>{0, 1, 2, 3});
+    copy_data(b, vector<float>{1, 2, 3, 4});
+
+    const char* trace_log_file = "trace_meta.log";
+    const char* bin_log_file = "trace_bin.log";
+
+    set_environment("NGRAPH_TRACER_LOG", trace_log_file, 1);
```
rkimballn1: Use the set_tracer_log function you need to create

andreykhaykin: I didn't get you, can you explain what should I do and for what?

src/ngraph/runtime/cpu/cpu_external_function.cpp
andreykhaykin: do you mean via some variable? can you provide me some basic idea to this?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1433,6 +1476,12 @@ void runtime::cpu::CPU_ExternalFunction::build(ngraph::pass::PassConfig& pass_co
         op_names.push_back(node->get_name());
         handler->second(this, node.get(), in, out);
 
+        if (do_ngraph_trace != nullptr)
+        {
+            tv_inputs.push_back(in);
+            tv_outputs.push_back(out);
```
jbobba: Please extend OpAttributes to capture shape/size of these tensors in m_op_attrs.

src/ngraph/runtime/cpu/cpu_debug_tracer.cpp
```diff
@@ -0,0 +1,109 @@
...
+        num = (num - mean) * (num - mean);
+        float y = num - c;
+        float t = sum + y;
+        c = (t - sum) - y;
+        sum = t;
+    }
+    return sum / size;
+}
+
+void runtime::cpu::CPU_DebugTracer::dump_one_tensor(
```
jbobba: Can we templatize this? Would be great if we could at least dump floats and uint8/int8.

test/cpu_debug_tracer.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "ngraph/runtime/cpu/cpu_layout_descriptor.hpp"
+#include "ngraph/runtime/cpu/cpu_tensor_view.hpp"
+#include "ngraph/runtime/cpu/op/sigmoid_mul.hpp"
+#include "ngraph/util.hpp"
+#include "util/test_tools.hpp"
+
+using namespace ngraph;
+using namespace std;
+
+TEST(cpu_degub_tracer, check_flow_with_external_function)
```
jbobba: cpu_debug_tracer

test/cpu_debug_tracer.cpp
andreykhaykin: Hi, this tool supports this option. (only meta log dumped)

test/cpu_debug_tracer.cpp
andreykhaykin: ?

src/ngraph/runtime/cpu/cpu_external_function.cpp
andreykhaykin: Oh, I wanted to do this but was a little afraid to change the struct.

src/ngraph/runtime/cpu/cpu_debug_tracer.cpp
andreykhaykin: Done

andreykhaykin:
@jbobba, I can rename env flags to those, that you proposed. But as I remember ngraph-bridge have some TRACER flag for performance, will not it be confusing?

test/cpu_debug_tracer.cpp
jbobba: Please correct the typo.

jbobba:
> @jbobba, I can rename env flags to those, that you proposed. But as I remember ngraph-bridge have some TRACER flag for performance, will not it be confusing?

Right, we already have NGRAPH_CPU_TRACING. If you think it is too close, then i would recommend NGRAPH_CPU_DEBUG_TRACER.

src/ngraph/runtime/cpu/cpu_external_function.hpp
```diff
@@ -78,12 +80,19 @@ namespace ngraph
                 std::string Description;
                 std::vector<std::string> Outputs;
                 std::vector<std::string> Inputs;
+                std::vector<std::pair<size_t, ngraph::Shape>> Outputs_tensor_attrs;
```
jbobba: Last comment. You might want to capture element_type as well here so you could dump tensors only for float element types instead of the current implementation which seems to assume all tensors are float types.

andreykhaykin: @jbobba, I captured it. I have two options to solve this problem
(*my assumption that you are talking about the same type for one run and NOT mixed types for one iteration?!)
1) I can add switch/case with input "get_type_enum" in "dump_one_kernel".
2) I can template all the debugger class, but here I need to know from the beginning of built function what type tensors will be(same type for all) (if I know the type it will be better solution)

src/ngraph/runtime/cpu/cpu_external_function.cpp
rkimballn1: Here is an example of a set function https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/pass/manager.hpp#L72
That can programmatically set a variable that can get set via environment variable here
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/pass/manager.cpp#L40-L44
This allows unit tests or other code to specifically set the flag without resorting to environment variables, but environment variables still work too.

src/ngraph/runtime/cpu/cpu_external_function.hpp
jbobba: yes, same type for one run. (1) seems sufficient.

indie:
> to support this feature across multiple backends, we could go with 'NGRAPH_TRACER=1' 

`NGRAPH_TRACER=OFF` would be more consistent with what we have here: https://ngraph.nervanasys.com/docs/latest/buildlb.html#the-default-build, if it's going to be off by default and require manual enablement for a particular run?  Or does this change mean we'd delete that existing flag:  

`NGRAPH_DEBUG_ENABLE` since all logging would be captured in the tracelogs? 

































































+ @jbobba

jbobba:
> > to support this feature across multiple backends, we could go with 'NGRAPH_TRACER=1'
> 
> `NGRAPH_TRACER=OFF` would be more consistent with what we have here: https://ngraph.nervanasys.com/docs/latest/buildlb.html#the-default-build, if it's going to be off by default and require manual enablement for a particular run? Or does this change mean we'd delete that existing flag:
> 
> `NGRAPH_DEBUG_ENABLE` since all logging would be captured in the tracelogs? @jbobba

NGRAPH_DEBUG_ENABLE is targeted toward towards ngraph developers while this feature (NGRAPH_CPU_DEBUG_TRACER) is targeted more toward data scientists. NGRAPH_CPU_DEBUG_TRACER is also a runtime environment variable whereas NGRAPH_DEBUG_ENABLE is a build setting.

src/ngraph/runtime/cpu/cpu_debug_tracer.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    if (m_tracer_stream.is_open())
+    {
+        return;
+    }
+
+    static auto trace_file_path = std::getenv("NGRAPH_CPU_TRACER_LOG");
+    static auto trace_bin_file_path = std::getenv("NGRAPH_CPU_BIN_TRACER_LOG");
+    if (trace_file_path == nullptr)
+        trace_file_path = (char*)"trace_meta.log";
```
rkimballn1: Curly braces around single-line code blocks. Same for line 45.

indie:
generic_pyapiv  

Does this documentation capture the gist of what you wrote?  If I understand @jbobba correctly, it probably should be somewhere it won't be confused with  `NGRAPH_CPU_TRACING`

andreykhaykin:
@indie 
This doc seems good! 
And I think it's a good idea to add this to another PR.

silee2:
Clang caught this issue:
```

File: src/ngraph/runtime/cpu/cpu_debug_tracer.cpp [hunk 13]
```diff
@@ -0,0 +1,71 @@
...
+
+#include "ngraph/runtime/cpu/cpu_debug_tracer.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+runtime::cpu::CPU_DebugTracer::CPU_DebugTracer()
+    : m_serial_number(0)
+{
+    if (std::getenv("NGRAPH_CPU_DEBUG_TRACER") != nullptr)
```
diyessi: Cache this in a static member.

src/ngraph/runtime/cpu/cpu_debug_tracer.hpp
```diff
@@ -0,0 +1,123 @@
...
+                                     const std::string& in_out);
+
+            private:
+                void init_streams();
+
+                size_t m_serial_number;
+                std::fstream m_tracer_stream;
+                std::fstream m_tracer_bin_stream;
+
+                bool enable_tracing = false;
```
diyessi: `m_enable_tracing`

src/ngraph/runtime/cpu/cpu_debug_tracer.hpp
```diff
@@ -0,0 +1,123 @@
...
+            };
+        }
+    }
+}
+
+// use of kahan sum to reduce numeric error
+template <typename T>
+static float find_variance(const std::vector<T>& f_data, float mean, size_t size)
+{
```
diyessi: Don't use `static` here. Leads to warnings in every file that includes this header.

src/ngraph/runtime/cpu/cpu_external_function.hpp
```diff
@@ -72,18 +74,40 @@ namespace ngraph
 
             using OpMap = std::unordered_map<std::type_index, OpFunction>;
 #endif
+            struct TensorTracerAttributes
```
diyessi: Follow the coding standards.

src/ngraph/runtime/cpu/cpu_external_function.hpp
```diff
@@ -72,18 +74,40 @@ namespace ngraph
...
+                {
+                }
+            };
 
             struct OpAttributes
             {
                 std::string Description;
                 std::vector<std::string> Outputs;
                 std::vector<std::string> Inputs;
+                std::vector<TensorTracerAttributes> Outputs_tensor_attrs;
```
diyessi: Follow the coding standards.

src/ngraph/runtime/cpu/cpu_external_function.hpp
andreykhaykin: I don't understand, it's like OpAttributes there.
Define C++ class or type names with CamelCase.

src/ngraph/runtime/cpu/cpu_external_function.hpp
andreykhaykin: it should be lower case?(why other variable names here with uppercase)?

src/ngraph/runtime/cpu/cpu_external_function.hpp
diyessi: The member names should be `m_number_of_elements`, `m_shape` and `m_element_type`

aslepko:
TESTNOW

PR #3100, opened 2019-06-19
use lock_guard to protect the access to m_exec_map in cpu_backend

pruthviIntel:


pruthviIntel:
@pawelpiotrowicz just wondering if you were able to cross check against the model which used to segfault for multi threaded instance?

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -90,6 +90,7 @@ shared_ptr<runtime::Executable>
                                        ngraph::pass::PassConfig& pass_config,
                                        bool performance_counters_enabled)
 {
+    std::lock_guard<std::mutex> guard(m_exec_map_mutex);
```
jbobba: we only want to guard the access and modifications to m_exec_map. make_shared<CPU_Executable>(func, pass_config, performance_counters_enabled) should be outside the critical section.

pawelpiotrowicz:
> @pawelpiotrowicz just wondering if you were able to cross check against the Paddle-Paddle model which used to segfault for multi threaded instance?

@pruthviIntel  I'm not sure, could you clarify the question? an example ?

src/ngraph/runtime/cpu/cpu_backend.cpp
pruthviIntel: @jbobba i addressed the changes requested , can u check?

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -91,17 +91,23 @@ shared_ptr<runtime::Executable>
...
+    // m_exec_map_mutex will be released once the object `guard` goes out of scope
     {
-        rc = it->second;
+        std::lock_guard<std::mutex> guard(m_exec_map_mutex);
+        auto it = m_exec_map.find(func);
+        if (it != m_exec_map.end())
+        {
+            rc = it->second;
+        }
+        return rc;
```
nmostafa: That return should be inside the if, no ?

pruthviIntel: ah right!. thanks for catch it.

PR #3101, opened 2019-06-19
[MLIR] Add sub-graph extraction support

nmostafa:
Previously MLIR compilation was per nGraph function. It was all nodes or nothing. This change extracts sub-graphs of MLIR-supported nodes from the nGraph function. It is possible to have separate sub-graphs each can be JIT'ed and invoked as an MLIR module. 

Currently the algorithm is greedy: build the maximal sub-graph possible of support MLIR ops as long as we don't introduce a cycle in the graph. The sub-graph nodes are annotated with a CompiledKernel node that is intercepted for execution.


src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
```diff
@@ -28,52 +29,148 @@ using namespace ngraph::pass;
...
+        NodeVector inputs;
+        int graph_id = -1;
+
+        // unsupported ops, skip
         if (!is_supported_mlir_op(op))
         {
-            return false;
+            continue;
+        }
+        if (TI(Parameter) == TI(*op) || TI(Result) == TI(*op))
```
jbobba: How do we plan to treat 'constants'? supported or unsupported?

test/backend_binary_elementwise.in.cpp
```diff
@@ -87,6 +87,160 @@ NGRAPH_TEST(${BACKEND_NAME}, dot_add)
                                   vector<float>{35.f, 40.f, 45.f, 68.f, 82.f, 96.f}));
 }
 
+NGRAPH_TEST(${BACKEND_NAME}, dot_add_2_subgraphs)
```
jbobba: these might be better off in cpu_test for now

src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
ayzhuang: Right now constants are inputs to sub-graph since they are not supported?

nmostafa: They are inputs yes, and shouldn't be supported, since there is no operation required for them. So MLIR code will just read from the input tensor.

test/backend_binary_elementwise.in.cpp
nmostafa: Will move them.

src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
nmostafa: Thinking about it more, we should support constant ops in MLIR for perf reasons. We need to be able to do constant folding/prop, and have a constant tensor def op would enable that. Alternatively, we could leave it unsupported but those inputs as constant. I prefer the former.

test/backend_binary_elementwise.in.cpp
nmostafa: Maybe better to have mlir_test.cpp instead for mlir-specific tests. It will still be configured `${BACKEND_NAME}`. What do you think ?

nmostafa: I moved the tests to a separate file for mlir

src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
```diff
@@ -108,6 +270,30 @@ bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node
...
+    while (!work_list.empty())
+    {
+        auto node = work_list.back();
+        work_list.pop_back();
+        if (std::find(subgraph_ids.begin(), subgraph_ids.end(), get_subgraph(node)) !=
+            subgraph_ids.end())
+        {
+            // we hit one of the sub-graphs we want to extend. we have a cycle.
+            NGRAPH_DEBUG << "[CK Extract] Cycle found when trying to add node" << std::endl;
+            return true;
```
dcaballe: I think we shouldn't return here. This is just telling us that we found a cycle in one of the potential subgraphs we want to extend but we should try to find another subgraph we can extend instead of bailing out after the first cycle found, right?

nmostafa: we aren't bailing out, we are just saying a cycle is found. If you check the call-site we actually start a new sub-graph.

src/contrib/mlir/pass/mlir_subgraph_extraction.hpp
```diff
@@ -16,26 +16,114 @@
...
-
 namespace ngraph
 {
     namespace pass
     {
-        /// This pass creates CompiledKernel ops enclosing sub-graphs that will be compiled and
-        /// executed by MLIR.
-        // TODO: WIP. Currently we only create a single CompiledKernel op for the whole function
-        // body.
+        /// This pass creates CompiledKernel ops enclosing maximal sub-graphs of ops that are supported by MLIR
```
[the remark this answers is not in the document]
nmostafa: can be found in .cpp file `run_on_function`

test/backend_mlir.in.cpp
```diff
@@ -0,0 +1,250 @@
...
+#include "util/test_control.hpp"
+#include "util/test_tools.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+static string s_manifest = "${MANIFEST}";
+
+// Combined ops test
+NGRAPH_TEST(${BACKEND_NAME}, mlir_dot_add)
```
rkimballn1: I would remove the `mlir_` prefix since these tests are really more general than that. Also remove the `_mlir` from the file name.

nmostafa: These tests are tailored towards certain features in MLIR end-to-end compilation (e.g. sub-graph extraction, compiling multiple opcodes), and possibly to test MLIR specific optimizations in the future. I did that on purpose to be able to run all mlir tests `--gtest_filter=CPU.mlir_*` . Unless you have a strong opinion on this, I would rather keep it.

src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
```diff
@@ -28,52 +29,213 @@ using namespace ngraph::pass;
 
 #define TI(x) std::type_index(typeid(x))
 
+int MLIRSubgraphExtractionPass::MLIRSubgraph::m_curr_graph_id = 0;
+
+template <typename T>
+void MLIRSubgraphExtractionPass::MLIRSubgraph::add_inputs(T& inputs)
+{
+    // inputs list are not exclusive, avoid duplication
+    for (auto node : inputs)
```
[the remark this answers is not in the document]
dcaballe: My comment was more in the line of better using a std utility that could be optimized in many ways. Not a big deal, though.

src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
dcaballe: Yeah, that's what I meant by "bailing out": we are creating a new graph instead of trying to find an existing one in the remaining sub-graphs where the op doesn't create a cycle. It could be a future improvement but, please could you add a comment about this?

test/backend_mlir.in.cpp
dcaballe: Are they also running when MLIR is disabled?

test/backend_binary_elementwise.in.cpp
```diff
@@ -67,36 +67,6 @@ NGRAPH_TEST(${BACKEND_NAME}, add)
                                   (test::NDArray<float, 2>({{6, 8}, {10, 12}})).get_vector()));
 }
 
-NGRAPH_TEST(${BACKEND_NAME}, dot_add)
```
rkimballn1: This test should not be removed. It is still used. I would like to be able to run all dot* unit tests by typing BACKEND.dot*

test/backend_mlir.in.cpp
rkimballn1: It sounds like these tests are not named properly. This is not `mlir_dot_add` but is something more like `mlir_single_subgraph` or something like that. The fact that the test uses Dot and Add is irrelevant to the test. You don't need to rename this, I'm just saying the name does not seem to be correct.

nmostafa: @dcaballe The tests will run with and without MLIR. 
@rkimballn1 Dot and Add are the currently supported ops :) In the future, these tests will actually go away when we have more ops (so existing unit-tests will exercise MLIR pipeline). For now, I just want to make sure that these tests pass with MLIR enabled before any MLIR check-in.

test/backend_binary_elementwise.in.cpp
nmostafa: It was added by us to trigger MLIR compilation for a function, that's why I moved it to `backend_mlir.in.cpp`

test/backend_mlir.in.cpp
dcaballe: I'm just worried of adding more time to CI when we are not actually testing anything relevant if MLIR is not enabled. Also the confusion this might create: "why my (let's say) PlaidML change is breaking mlir_* tests if MLIR is disable?". However, I don't know about previous precedents/options in this regard so maybe @rkimballn1 or @diyessi can provide better feedback.

test/backend_binary_elementwise.in.cpp
rkimballn1: 👍

test/backend_mlir.in.cpp
rkimballn1: @nmostafa Add this to the test CMakeLists.txt file around line 175
```
if(NGRAPH_MLIR_ENABLE)
    list(APPEND MULTI_TEST_SRC backend_mlir.in.cpp)
endif()
```
This will only add your new test file if MLIR is enabled, making @dcaballe happy :)

nmostafa: Thanks for the hint ! I was wondering myself :) Will do

dcaballe: Nice, thanks!

PR #3106, opened 2019-06-20
Read provenance tags from serialized graph 

ransford2011:
This enables the ability to read provenance tags from the serialized graph and ensure that the values are preserved even when the graph is cloned.


src/ngraph/graph_util.cpp
```diff
@@ -233,6 +233,14 @@ std::list<std::shared_ptr<ngraph::Node>>
                 // There is a friendly name for this node so copy it
                 cloned_node->set_friendly_name(node->get_friendly_name());
             }
+
+            if (!node->get_provenance_tags().empty())
+            {
```
diyessi: No need to test for `empty` since the `for` is testing for it implicitly.

src/ngraph/serializer.cpp
```diff
@@ -1783,6 +1783,14 @@ shared_ptr<Node> JSONDeserializer::deserialize_node(json& node_js)
         {
             node->set_friendly_name(node_name);
         }
+        std::vector<json> prov_js = node_js.at("provenance_tags");
+        if (!prov_js.empty())
+        {
```
diyessi: Same comment about testing for empty.

aslepko:
TESTNOW

PR #3108, opened 2019-06-21
Fix unit-test failure in debug mode

gaurides:


gaurides:
This will be undone when we integrate MKLDNN v0.20 , as it has the official fix

cmake/mkldnn.patch
```diff
@@ -28,3 +28,16 @@ index f10feb20..05f47961 100644
...
  target_include_directories(${LIB_NAME} PUBLIC
+diff --git a/src/cpu/jit_avx512_common_conv_kernel.cpp b/src/cpu/jit_avx512_common_conv_kernel.cpp
+index 1bb98fa43..b8b54401f 100644
+--- a/src/cpu/jit_avx512_common_conv_kernel.cpp
++++ b/src/cpu/jit_avx512_common_conv_kernel.cpp
+@@ -3055,7 +3055,7 @@ void jit_avx512_common_conv_bwd_weights_kernel_f32::bias_kernel_3d() {
+ void jit_avx512_common_conv_bwd_weights_kernel_f32
+     ::compute_oh_loop_common()
+ {
+-    assert(jcp.harness == harness_mb_reduction);
```
diyessi: throw an ngraph_errror instead of asserting.

jbobba: this is mkldnn code.

PR #3109, opened 2019-06-21
Switch Output to use a shared_ptr to prevent nodes from disappearing early

diyessi:
Already have verbal approval from @aprocter


aslepko:
TESTNOW

PR #3115, opened 2019-06-24
[Py] Update documentation for utility test functions.

arogowie-intel:


python/test/ngraph/util.py
```diff
@@ -45,25 +48,24 @@ def run_op_node(input_data, op_fun, *args):
     comp_args = []
     op_fun_args = []
     comp_inputs = []
-    for idx, data in enumerate(input_data):
-        if np.isscalar(data):
-            op_fun_args.append(ng.constant(data, _get_numpy_dtype(data)))
-        else:
-            node = ng.parameter(data.shape, name=ascii_uppercase[idx], dtype=data.dtype)
```
tsocha: Why are you removed this if-else block?
At the first glance it looks like you changed functionality and removed ng.parameter creation.
Please explain.

arogowie-intel: After analysis of this utility function code and comparing it to what happens in Python API I came into conclusion that this `else` branch is simply unnecessary. Our Python API converts not only scalars but also `ndarray`'s into `Constant`s - take a look [here](https://github.com/NervanaSystems/ngraph/blob/05f2b9ef6b1a6a38bf892dade7b8f1b33444464e/python/ngraph/utils/types.py#L89-L104).

arogowie-intel: Glad you asked for this. I should have explained it in PR description.

python/test/ngraph/util.py
```diff
@@ -45,25 +48,24 @@ def run_op_node(input_data, op_fun, *args):
...
 
 def run_op_numeric_data(input_data, op_fun, *args):
+    # type: (NumericData, Callable, *Any) -> List[NumericData]
     """Run computation on node performing `op_fun`.
 
     `op_fun` has to accept a scalar or an array.
 
+    This function passess input data AS IS. This mean that in case they're a scalar (integral,
+    or floating point value) or a NumPy's ndarray object they will be automatically converted
+    to nGraph's Constant Nodes.
```
postrational: Please add a note what happens otherwise - with an example.
Please also explain why this function is needed.

python/test/ngraph/util.py
```diff
@@ -45,25 +48,24 @@ def run_op_node(input_data, op_fun, *args):
...
+    for data in input_data:
+        op_fun_args.append(ng.constant(data, _get_numpy_dtype(data)))
     op_fun_args.extend(args)
     node = op_fun(*op_fun_args)
     computation = runtime.computation(node, *comp_args)
     return computation(*comp_inputs)
 
 
 def run_op_numeric_data(input_data, op_fun, *args):
+    # type: (NumericData, Callable, *Any) -> List[NumericData]
```
postrational: Dows this function only accept `NumericData` for `input_data`?

arogowie-intel: Actually it may accept also nodes, but it is defined **only** to test the automatic raw data conversion path.

python/test/ngraph/util.py
arogowie-intel: Ad 2) This function is needed for unit testing purposes only - as you may see it is used only in our unit tests. 
Ad 1) Actually there is no _otherwise_ - it is used and defined to run on numeric data only.

diyessi:
#3279 should fix the CI problem.

PR #3120, opened 2019-06-25
Langjian/allreduce provenance tag

jianyinglang:
Print provenance_tag name for allreduce op when provenance_tag is available


jianyinglang:
Currently, the logging info only print out the `get_friendly_name`. But when there a graph pass rewrite, the `get_friendly_name` will not be passed. There is an option to set the op with `provenance_tag ` and we want to check this name in the backend when `get_friendly_name` is not available.

wenzhe-nrv:
> a graph pass rewrite
@diyessi just wonder how does the pass rewrite work? will the pass affect allreduce of if it doesn't rewrite allreduce op?

PR #3127, opened 2019-06-27
Change compiler action to EmitLLVMOnly.

ayzhuang:
1. Change compiler action from EmitCodegenOnly to EmitLLVMOnly and remove the workarounds. With EmitCodegenOnly, clang generates instructions like this "vmovaps %xmm0,0x8(%rsp)" where the memory address is not 16 bytes aligned. We had to use push_back to avoid such instructions. 

2. Do  not release function in compile() because we need to get the nodes from it when getting performance counter data for CODEGEN.


src/ngraph/codegen/compiler.cpp
```diff
@@ -343,7 +343,7 @@ std::unique_ptr<codegen::Module>
     preprocessor_options.RemappedFileBuffers.push_back({m_source_name, buffer.get()});
 
     // Create and execute action
-    m_compiler_action.reset(new EmitCodeGenOnlyAction());
+    m_compiler_action.reset(new EmitLLVMOnlyAction());
```
dcaballe: Did you figure out what is the difference between these two classes? If it's a bug maybe it's something we can report.

ayzhuang: EmitLLVMOnlyAction, EmitLLVMAction and EmitBCAction would not cause segfault.
EmitCodeGenOnlyAction, EmitAssemblyAction, and EmitObjAction would cause segfault.
In BackendUtil.cpp, 1)UsesCodeGen is set to true (so it will give error report if TargetMachine is not created) for second line three options before CreateTargetMachine(UsesCodeGen), 2)AddEmitPasses(CodeGenPasses, Action, *OS,
 DwoOS ? &DwoOS->os() : nullptr) is called for the second line three options. Inside of AddEmitPasses, some passes are added.

PR #3134, opened 2019-06-28
Failing test on INTERPRETER backend.

arogowie-intel:
This UT fails with following error **only** on `INTERPRETER` backend:
```
C++ exception with description "Output 0 shape {1,2,10,8} does not match Result shape {1,2,9,7}" thrown in the test body. 
```
Through my short investigation I've found that in `INTExecutable` [constructor](https://github.com/NervanaSystems/ngraph/blob/abdd9c0bc61ac536f56c2fc655900ca9c14633d5/src/ngraph/runtime/interpreter/int_executable.cpp#L46) input `function` object has correct result shapes **before** running passes, but incorrect **after** all passes are run.


silee2:
@diyessi You can find the original ONNX test case here: Goto https://github.com/onnx/onnx/blob/master/docs/Operators.md#ConvTranspose and look for the example named convtranspose_attributes The test has tree sub tests and the one in question is the first one.
This may be related: https://github.com/onnx/onnx/issues/102 I don't know if the bridge or importer needs to compute `adj` to help nGraph get the correct output size.

diyessi:
@arogowie-intel said the shape is only wrong after running passes; if so, this is a problem with some pass.

aprocter:
I'm not seeing any obvious suspects among the `INTERPRETER` passes... https://github.com/NervanaSystems/ngraph/blob/cfff3f1b39327c51fa046f4997bf6c81784fcd76/src/ngraph/runtime/interpreter/int_executable.cpp#L41-L63

aprocter:
The only passes `INTERPRETER` is running that `CPU` is not are `AssignLayout<DenseTensorLayout>` and `Liveness`. @arogowie-intel could you run with env var `NGRAPH_ENABLE_SERIALIZE_TRACING=1` and send us the json dumps?

silee2:
Found the problem. Clone function was incorrect. 
Just pushed the fix.

src/ngraph/op/fused/group_conv_transpose.cpp
```diff
@@ -203,7 +203,8 @@ shared_ptr<Node> op::GroupConvolutionTranspose::copy_with_new_args(const NodeVec
                                                       get_padding_end(),
                                                       get_output_padding(),
                                                       get_groups(),
-                                                      get_pad_type());
+                                                      get_pad_type(),
+                                                      get_output_shape());
```
aprocter: Oh, that'll do it :)

silee2:
This a case where output shape cannot be inferred correctly without extra information. In such case output shape was manually given but the clone function wasn't passing the manual output shape.

PR #3137, opened 2019-06-28
[MLIR] Bump MLIR repo to commit 82d5084 Wed Jun 26.

dcaballe:
Improvements and changes in dialect conversion APIs. Type conversion
and op rewriting are now a bit more independent mechanisms. Dialect
conversion now requires a target object with the legal transformations
to be applied. Legality checks to detect illegal operations after
dialect conversion have been strengthened significantly.

MLIR LLVM commit:
Author: Matt Arsenault <Matthew.Arsenault@amd.com>
Date:   Wed Jun 26 20:56:18 2019 +0000

    AMDGPU: Assert SPAdj is 0

    llvm-svn: 364473

MLIR commit:
commit 82d5084cf28bca7a0eebec283cdeae118dfa1d11 (HEAD)
Author: Jacques Pienaar <jpienaar@google.com>
Date:   Wed Jun 26 11:12:40 2019 -0700

    Split out TranslateClParser and add new parse method that reuses
    SourceMgr.


src/contrib/mlir/dialect/dialect.hpp
```diff
@@ -25,15 +25,17 @@
 #include "ngraph/check.hpp"
 namespace mlir
 {
-    class NGDialect : public mlir::Dialect
+    class NGraphOpsDialect : public mlir::Dialect
```
nmostafa: Is the rename needed ?

dcaballe: Just for alignment with MLIR changes. At the beginning I added the whole dialect as legal in the lowering. Then I saw they had renamed affine/standard dialects in the same way (StandardOpsDialect, AffineOpsDialect) and try to align ours with that.

PR #3140, opened 2019-06-28
PlaidML: Disable UT's that require RNN or GRU cell.

silee2:


silee2:
Noticed that some PlaidML tests are failing in General CI. This PR should fix the issue.

PR #3141, opened 2019-06-28
Separate validation from op definition

csullivan:
Change the design of ngraph operator validation to use registration by default instead of inheritance. This enables a backend to change the validation by registering a separate validation for a given op type without forcing the creation of a new inherited op type via polymorphism. The polymorphic interface for validation remains available for backwards compatibility. 

Registration implementation is added to ngraph/validation_util.hpp along with two macros that can be used for easy registration.

Notes: 
* The only deletion I made in this PR without reproducing it elsewhere is linked here: https://github.com/NervanaSystems/ngraph/compare/cs/op_validators?expand=1#diff-da85ef590a8755f206673fbfdc2a9a6aL38. Since op::Op does not have its own validation routine, this does not have a functional effect.
* In a follow on PR we could make `Node::validate_and_infer_types()` non-virtual and add handling for each of the backends if I can get buy-in. However, for now validation can still be overloaded via polymorphism or via the new registration semantics.


src/ngraph/runtime/cpu/op/bounded_relu.cpp
```diff
@@ -19,6 +19,16 @@
 using namespace std;
 using namespace ngraph;
 
+namespace ngraph
```
nmostafa: Why add support to only two cpu ops and not all of them ?

csullivan: These two are utilizing the validator for `UnaryElementwiseArithmetic` operators and so required the update. I did not touch backend specific op validations in this PR unless they were using validators defined in ngraph core.

csullivan:
@diyessi On the contrary, very little is changed functionally. The same validation functions are still called, just via type dispatch instead of polymorphism. That said, I'm quite open to any design that achieves the decoupling of validation and operator definition. If you would like to have a meeting to discuss changes the design you'd like to see, let's set it up.

diyessi:
There are many passes. Passes assume particular op semantics tied to the op class or description, depending on how they match. To add this type of polymorphism, you need to add something to node to say which set of semantics it is using and make all these passes and other things that are keyed off node description respect this indicator and change all the things that key off class switch to this other distinguisher (which needs to be done eventually anyways since dynamic class does not work across shared library boundaries on some systems).
Currently you need to change something like 17 files to add an op. I think that is the real problem. Two files should be all that is required for most ops.

csullivan:
Closing in preference of transactional validation design, NGraph/RFC/ValidationDesignUpdate.

PR #3146, opened 2019-07-01
[WIP] Fix unreachable warnings in constant_folding.cpp

dcaballe:
Just for discussion, I'm seeing a good bunch of warnings in constant_folding.cpp:

```
/localdisk/dcaballe/ngraph-mlir-pr/src/ngraph/pass/constant_folding.cpp: In function ‘std::shared_ptr<ngraph::op::Constant> fold_constant_convert_helper0(std::shared_ptr<ngraph::op::Constant>, const ngraph::element::Type&) [with TI = long unsigned int]’:
/localdisk/dcaballe/ngraph-mlir-pr/src/ngraph/pass/constant_folding.cpp:841:1: warning: control reaches end of non-void function [-Wreturn-type]
```

Do we have some special macro to throw exceptions for unreachable code? A better way to fix this?

Thanks!


aprocter:
This is a bit strange. Since the `switch` here is complete (which I think it must be, because we have `-Wswitch` and `-Wswitch-enum` warnings turned on locally), the end of the function is not really reachable, right? (I guess technically it is if the caller passes in an invalid value for the enum, but surely that's a "holding it wrong" kind of situation?)

If we have to insert something to mollify the compiler, I like the `NGRAPH_CHECK` here, but we should keep the pragma pop.

aprocter:
What compiler version are you using, by the way? I think we build with `-Wreturn-type` by default.

dcaballe:
I'm using GCC 8.1.0, which is needed for MLIR.

> if the caller passes in an invalid value

Yeah, that's the reason.

Ok, I can restore the pop but would it be necessary if we add the NGRAPH_CHECK?

I needed to assert on unreachable code several times in the past. Would it make sense to add a NGRAPH_UNREACHABLE that translates to NGRAPH_CHECK(false, ...) or do we have something like that already?

aprocter:
`NGRAPH_UNREACHABLE` as a simple shortcut makes sense to me.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -835,9 +835,8 @@ shared_ptr<op::Constant> fold_constant_convert_helper0(shared_ptr<op::Constant>
     case element::Type_t::u64:
         return fold_constant_convert_helper1<TI, uint64_t>(constant, output_element_type);
     }
-#if !(defined(__GNUC__) && (__GNUC__ == 4 && __GNUC_MINOR__ == 8))
-#pragma GCC diagnostic pop
```
aprocter: We should keep this pragma pop, and also do the `NGRAPH_CHECK`.

aprocter:
@rkimballn1 Do you remember why we turn off `-Wswitch` and `-Wswitch-enum`? Perhaps we could dispose of the pragmas altogether.

dcaballe:
Any comments or something missing here?

aprocter:
@dcaballe  Nope, LGTM

aprocter:
@dcaballe Is this still [WIP]? Please edit the title and remove the label if not.

PR #3132, opened 2019-06-28
Decomposition of fused ops in PlaidML backend

tomdol:
- The `FusedOpDecomposition` pass has been added to the PlaidML backend - models in OnnxBackendZooModelTest were failing without this pass
- I've ran ngrap's unit tests using PlaidML with the following device: `opencl_nvidia_geforce_gtx_960.0 : NVIDIA Corporation GeForce GTX 960 (OpenCL)` and excluded the failing tests in the PlaidML manifest

PR #3147, opened 2019-07-01
Decomposition of fused ops in PlaidML backend

tomdol:
Same as https://github.com/NervanaSystems/ngraph/pull/3132 but against the `r0.23` release branch


tomdol:
A bit too fast, I have to create another one without the master branch merges

PR #3149, opened 2019-07-01
Update mkl-dnn to v0.20

silee2:
Build will currently break. mkl-dnn release page needs to update assets. mklml is not uploaded yet.


aslepko:
TESTNOW

silee2:
@aslepko Can you check for any performance degradation before merging this PR?

aslepko:
I'll submit test runs. Keep in mind, that this will be a bit of a strange comparison for ngtf performance. Right now, with latest tensorflow, ngtf 0.16 performance is pretty bad for some models. Anyway, I'll hit off jobs before/after this branch.

aslepko:
Checked impact on performance in onnxrt jobs and ngraph-tf. Nothing standing out. Good to go.

PR #3155, opened 2019-07-02
[MLIR] Disable dump when NGRAPH_MLIR_DUMP_ALL is != 1

dcaballe:


src/contrib/mlir/compiler.cpp
```diff
@@ -447,7 +447,8 @@ mlir::StaticFloatMemRef* MLIRCompiler::allocate_memref_descriptor(mlir::Type typ
 
 void MLIRCompiler::dump_mlir_module(const std::string msg)
 {
-    if (std::getenv("NGRAPH_MLIR_DUMP_ALL") != nullptr)
+    char* env_var = std::getenv("NGRAPH_MLIR_DUMP_ALL");
```
nmostafa: I don't think checking a value is necessary. We don't do it for other env vars, why this one ? What if I do 'NGRAPH_MLIR_DUMP_ALL=ON' .. can be confusing.

dcaballe: I found confusing that NGRAPH_MLIR_DUMP_ALL=0 enabled the dump but, ok, if the rest of env vars have the same behavior.

PR #3158, opened 2019-07-02
Fix validate_and_infer_types protection so all classes are consistent

rkimballn1:
There were a few ops where validate_and_infer_types was protected instead of public like all of the rest of the ops.


aprocter:
Are we sure we don't want to resolve this in the opposite direction (make everything `protected`)? At least as it stands, `validate_and_infer_types` is almost always called by `Node::constructor_validate_and_infer_types` or `Node::delayed_validate_and_infer_types`. The other uses of it don't appear to be necessary (but also wouldn't be precluded by use of `protected`).

aprocter:
Also, `Node::validate_and_infer_types` is itself protected.

aprocter:
I think we should make another pass to make sure this is always `protected`, but this should be fine for now.

PR #3159, opened 2019-07-02
Add a TAG file to be used if workspace is not a git repo.

silee2:
nGraph release tar ball is not a git repo and current build system try to get version info from git tags. This PR adds a TAG file that will only be used if source code is not a git repo.

TODO:
- [ ] Add rc/release tagging script 
- [ ] Add release tarball creation script and Create TAG file from that script.


TAG
```diff
@@ -0,0 +1 @@
+v0.24.0-dev.0
```
rkimballn1: Look into github `post-checkout` hook to populate this file. Hooks look pretty cool and easy.

silee2:
Closing: #3337 replaces this PR

PR #3161, opened 2019-07-02
Add Batch Norm Inference Relu fusion.

ayzhuang:
This fusion reduces densenet121 latency for onnx runtime by about 30%.

unit test graph
before this fusion:
image

after this fusion:
image

densenet121 after this fusion:
image

densenet121 after constant folding:
image


ayzhuang:
> please add unit-test
> can you please rename "...with_multi_add" -> "...with_multiply_add" ? for readability (it's not multiple adds)
> oh.. didn't see it is WIP

Yes, will add unit test and rename the function. Thank you.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -650,6 +650,107 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_batch_norm_relu_global_sta
     this->add_matcher(m, callback);
 }
 
+void ngraph::runtime::cpu::pass::CPUFusion::construct_batch_norm_infer_relu_with_multiply_add()
```
nmostafa: Please add comment on what the transformation does here. An example would do.

ayzhuang: Added comments.

PR #3163, opened 2019-07-02
Make required numpy version for setup and install explicit.

silee2:
Trying to generalize #3162 and avoid building numpy from source
numpy 1.15 is the last binary wheel to support python 3.4
numpy 1.16 seems like the last binary wheel to support python 2.7
Also trying to avoid the issue mentioned here
https://github.com/pytest-dev/pytest-xdist/issues/136
This PR assumes nGraph supports python 2.7, 3.4, 3.5, 3.6, 3.7


python/requirements.txt
```diff
@@ -1,4 +1,4 @@
 six
 numpy==1.15.4; python_version == "3.4"
-numpy; python_version != "3.4"
+numpy==1.16.4; python_version != "3.4"
```
postrational: I think we should we hard-code only for Python 2.7 and 3.4.
Let the user use latest version (or version required by other packages) in other cases.

python/setup.py
```diff
@@ -363,6 +363,22 @@ def build_extensions(self):
         build_ext.build_extensions(self)
 
 
+setup_requires = [
+    'numpy==1.15.4; python_version == "3.4"',
+    'numpy==1.16.4; python_version != "3.4"',
```
postrational: Can we avoid replicating this code by parsing the `requirements.txt` file?
This should work:
```setup_requires = open('requirements.txt').read().splitlines()```

python/requirements.txt
silee2: Done

python/setup.py
silee2: Done

silee2:
@aslepko @postrational I have made all the requested changes. Should we merge this PR or #3164 to master? #3162 which is identical to #3164 was merged to r0.23

PR #3164, opened 2019-07-03
Update setup.py

aslepko:
same as https://github.com/NervanaSystems/ngraph/pull/3162 , just for master

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -381,7 +381,7 @@ setup(
     packages=packages,
     cmdclass={'build_ext': BuildExt},
     data_files=data_files,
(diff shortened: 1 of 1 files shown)


python/setup.py
```diff
@@ -381,7 +381,7 @@ def build_extensions(self):
     packages=packages,
     cmdclass={'build_ext': BuildExt},
     data_files=data_files,
-    setup_requires=['numpy'],
+    setup_requires=['numpy==1.16.4'],
```
postrational: Instead of specifying a version, we should specify something like `numpy<1.17.0`
Please test a solution like that.

PR #3175, opened 2019-07-04
[ONNX] Use nGraph auto padding in ONNX operators.

arogowie-intel:


src/ngraph/op/fused/group_conv_transpose.cpp
```diff
@@ -52,6 +52,7 @@ op::GroupConvolutionTranspose::GroupConvolutionTranspose(const shared_ptr<Node>&
 
 op::GroupConvolutionTranspose::GroupConvolutionTranspose(const std::shared_ptr<Node>& data,
                                                          const std::shared_ptr<Node>& filters,
+                                                         const PadType& pad_type,
```
aprocter: This changes an API signature. Can we retain the old constructor for two releases?

src/ngraph/op/util/attr_types.hpp
```diff
@@ -42,15 +42,17 @@ namespace ngraph
...
         /// VALID      - No padding
+        /// INVALID    - Used for validation of provided auto padding attribute value.
         ///
         enum class PadType
         {
             EXPLICIT = 0,
             SAME_LOWER,
             SAME_UPPER,
             VALID,
+            INVALID,
```
aprocter: AFAICT this is only used to signal failure of `get_ng_pad_type`. I would prefer a different mechanism for this.

PR #3176, opened 2019-07-04
[ONNX] Fix LSTM reversing input sequences.

arogowie-intel:
Use `ReverseSequence` in LSTM operator to correctly handle situations when user provides input `sequence_lens` to reverse only specified length for given input batch.


arogowie-intel:
Closing since #3194 which incorporated this changes is already merged.

PR #3177, opened 2019-07-05
Use libc++ for macOS core python wheel

earhart:
It turns out using libc++ is supported on Xcode >= 8, and required on Xcode >= 10.

Tested by building a wheel, installing it into an Anaconda environment, and running the basic.py example script.  Also installed the wheel as whl[plaidml], changed the backend to "PLAIDML" in the example, and verified that the script still worked (and that it didn't work with backend "PLAIDMLL" :-)


diyessi:
Waiting for confirmation that r0.23 changes are in this PR

PR #3180, opened 2019-07-08
Revert "memory allocator for CPU "

tomdol:
The changes introduced by this PR cause a segfault when using ngraph from python (either interactively or when running tests).

To reproduce:
- compile ngraph
- generate ngraph python bindings (`python setup.py develop`)
- run python tests from ngraph/python or ngraph-onnx repo
- run python, import ngraph, run inference, exit python (the segfault occurs when exitting python, during cleanup)


aslepko:
TESTNOW

aslepko:
TESTNOW

postrational:
This change can be closed in favor of the fix in this PR:
https://github.com/NervanaSystems/ngraph/pull/3182

PR #3181, opened 2019-07-08
Add Gelu OP

silee2:


src/ngraph/op/fused/gelu.cpp
```diff
@@ -0,0 +1,76 @@
...
+    }
+    return make_shared<Gelu>(new_args.at(0));
+}
+
+void op::Gelu::validate_and_infer_types()
+{
+    element::Type input_element_type = get_input_element_type(0);
+
+    NODE_VALIDATION_CHECK(
+        this, !input_element_type.is_dynamic(), "Argument element type must not be dynamic.");
```
aprocter: We should allow this, and set the output ET to `element::dynamic`.

src/ngraph/op/fused/gelu.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    NODE_VALIDATION_CHECK(this,
+                          input_element_type == element::f32 ||
+                              input_element_type == element::f64 ||
+                              input_element_type == element::f16,
+                          "Argument element type must be f16, f32 or f64 (got ",
+                          input_element_type,
+                          ").");
+}
```
aprocter: Don't we need to set the output ET/shape here? (Since we are overriding `validate_and_infer_types`, `FusedOp` won't do it for us.)

src/ngraph/op/fused/gelu.cpp
```diff
@@ -0,0 +1,76 @@
...
+    element::Type input_element_type = get_input_element_type(0);
+
+    NODE_VALIDATION_CHECK(
+        this, !input_element_type.is_dynamic(), "Argument element type must not be dynamic.");
+
+    NODE_VALIDATION_CHECK(this,
+                          input_element_type == element::f32 ||
+                              input_element_type == element::f64 ||
+                              input_element_type == element::f16,
+                          "Argument element type must be f16, f32 or f64 (got ",
```
aprocter: I think we should allow it for all floating types (including bf16), right?

src/ngraph/op/fused/gelu.cpp
silee2: I was overriding the wrong function. should have been `pre_validate_and_infer_types`

src/ngraph/op/fused/gelu.cpp
silee2: Fixed in new commit.

src/ngraph/op/fused/gelu.cpp
silee2: Fixed in new commit.

src/ngraph/op/fused/gelu.cpp
silee2: Fixed in new commit.

src/ngraph/op/fused/gelu.cpp
aprocter: I think this still won't quite work, because `decompose_op` is going to try to create constants with dynamic element type. (Not sure if `FusedOp::validate_and_infer_types` has some sort of check for dynamicnessicity.) That's probably fine for now, though.

PR #3184, opened 2019-07-08
Stabilize node sorting and fix some bugs.

diyessi:


src/ngraph/descriptor/output.cpp
```diff
@@ -31,12 +33,20 @@ descriptor::Output::Output(Node* node, size_t index, const shared_ptr<Tensor>& t
 // Add an input to the vector of inputs that use this output.
 void descriptor::Output::add_input(Input* input)
 {
-    m_inputs.insert(input);
+    // Keep the inputs in insertion order to keep sorts deterministic
+    if (find(m_inputs.begin(), m_inputs.end(), input) == m_inputs.end())
```
aprocter: I'm a little concerned about this, in cases where an output has a lot of target inputs (e.g., a learning rate constant being read many places). In practice, don't know how big a deal it will be.

diyessi: We could add a set if that becomes an issue.

PR #3187, opened 2019-07-09
Enable ConstantFolding for Product, Reverse; add a couple of dynamic tests

aprocter:
Adds constant folding support for the `Product` and `Reverse` ops, and a couple more toy dynamic tests.

Tagging @ayzhuang because I think adding new ops to `ConstantFolding` might be a bit delicate perf-wise, and she's dealt with this before.


aprocter:
@ayzhuang Sounds good, thanks! If you're not too concerned about performance issues, I'd propose saving the update to use the CPU `Product` kernel for a later PR.

ayzhuang:
> @ayzhuang Sounds good, thanks! If you're not too concerned about performance issues, I'd propose saving the update to use the CPU `Product` kernel for a later PR.

later PR works.

PR #3188, opened 2019-07-09
[ONNX] Fix shrink operator for uint* types

tsocha:


src/ngraph/frontend/onnx_import/op/shrink.cpp
```diff
@@ -43,14 +43,24 @@ namespace ngraph
...
+                    std::shared_ptr<ngraph::op::Constant> negative_lambd;
+                    const auto input_element_type = input->get_element_type();
+                    if (input_element_type.is_signed())
+                    {
+                        negative_lambd = ngraph::op::Constant::create(
+                            input_element_type, input->get_shape(), {-lambd});
+                    }
+                    else
+                    {
+                        negative_lambd = ngraph::op::Constant::create(
```
arogowie-intel: Please add some comment explaining why `negative_lambd` is actually equal to zero.

PR #3189, opened 2019-07-09
Fix crash on MacOS

gaurides:


src/ngraph/runtime/cpu/mkldnn_utils.cpp
```diff
@@ -692,8 +692,9 @@ bool runtime::cpu::mkldnn_utils::is_mkldnn_padded_layout(const mkldnn::memory::d
 
 bool runtime::cpu::mkldnn_utils::use_mkldnn_kernel(const ngraph::Node* node)
 {
-    if (auto* op_node = dynamic_cast<const ngraph::op::Op*>(node))
+    if (node->is_op())
```
nmostafa: Why is this problematic on Mac ?

gaurides: Not sure exactly why, but Mac has issues with dynamic cast. I remember similar problems in past where they used if check and then static_cast.

diyessi: RTTI across shared library boundaries is flakey.

PR #3196, opened 2019-07-09
Added rpath setting for PlaidML backend

avijit-nervana:
This is to solve the wrong rpath for the PlaidML backend library.


src/ngraph/runtime/plaidml/CMakeLists.txt
```diff
@@ -13,6 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ******************************************************************************
+set(CMAKE_MACOSX_RPATH 1)
```
silee2: Do we need to set this?
MACOSX_RPATH is already enabled here.
https://github.com/NervanaSystems/ngraph/blob/master/CMakeLists.txt#L63-L64

src/ngraph/runtime/plaidml/CMakeLists.txt
```diff
@@ -13,6 +13,14 @@
...
+if(APPLE)
+    set(CMAKE_INSTALL_RPATH "@loader_path/;@loader_path/../../..")
+elseif(DEFINED NGRAPH_RPATH)
+    set(CMAKE_INSTALL_RPATH "\$ORIGIN:\$ORIGIN/../../../:${NGRAPH_RPATH}")
```
silee2: Use semicolon to separate paths.
https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_RPATH.html

src/ngraph/runtime/plaidml/CMakeLists.txt
```diff
@@ -13,6 +13,14 @@
...
+else()
+    set(CMAKE_INSTALL_RPATH "\$ORIGIN:\$ORIGIN/../../../")
```
silee2: Use semicolon to separate paths.

PR #3197, opened 2019-07-09
[MLIR] Support MLIR lowering of Relu

nmostafa:
Only integer Relu until MLIR support F32 cmp


src/contrib/mlir/lowerer.cpp
```diff
@@ -136,8 +137,10 @@ namespace
     void DialectLoweringPass::populateNGraphToAffineConversionPatterns(
         OwningRewritePatternList& patterns)
     {
-        RewriteListBuilder<NGAddOpConversion, NGDotOpConversion, NGReturnOpConversion>::build(
-            patterns, &getContext(), *this);
+        RewriteListBuilder<NGAddOpConversion,
+                           NGReluOpConversion,
+                           NGDotOpConversion,
+                           NGReturnOpConversion>::build(patterns, &getContext(), *this);
```
dcaballe: Since this is going to grow a lot, maybe better follow alphabetic order? Same for visitors below?

src/contrib/mlir/lowerer.cpp
```diff
@@ -365,6 +368,61 @@ namespace
...
+        // Steps
+        auto steps = vLHS.getSteps();
+
+        NGRAPH_CHECK(lhs->getType().isa<MemRefType>());
+        Type elemTy = lhs->getType().dyn_cast<MemRefType>().getElementType();
+        NGRAPH_CHECK(!elemTy.isa<FloatType>(),
+                     "NGReluOp with float element type should not be lowered until MLIR supports "
+                     "lowering !std.CmpF");
+
+        // clang-format off
```
dcaballe: Please, remove clang format markets. It should look ok now.

src/contrib/mlir/compiler.cpp
```diff
@@ -279,7 +279,11 @@ namespace ngraph
             {
                 return compiler.create_binary_op<mlir::NGAddOp>(ng_node);
             }
-
+            template <>
+            mlir::Value* MLIRCompiler::COMPILE_OP_DECL(ngraph::op::Relu)
```
diyessi: Can you put blank lines between these template instantiations?

src/contrib/mlir/lowerer.cpp
nmostafa: I will instead use an .inc file. Let me see.

PR #3199, opened 2019-07-10
Cyphers/topmaster

diyessi:
- [x] Use "ordered set" for control dependencies
- [x] `get_ordered_ops` can call sort on root set without doing `get_ops`
- [x] Comment explaining that top sort is really generating a list of nodes from a root set
- [x] Test util should take a second arg of nodes that need to be in the list


src/ngraph/graph_util.hpp
```diff
@@ -81,154 +82,123 @@ namespace ngraph
...
-
-        std::list<std::shared_ptr<ngraph::Node>> result_list;
-        while (independent_nodes.size() > 0)
+        while (nodes_to_do.size() > 0)
         {
-            auto independent_node = independent_nodes.front();
-            result_list.push_back(node_map[independent_node]);
-            independent_nodes.pop_front();
-
-            for (const std::shared_ptr<Node>& user : independent_node->get_users())
```
rkimballn1: Thank you for getting rid of the use of get_users(). I do hate that method.

PR #3200, opened 2019-07-10
Added possibility to choose output format to serialize_onnx tool application.

mbencer:
- possible output types: json and cpio (-t|--t as program argument)
- default output type: json


src/ngraph/serializer.cpp
```diff
@@ -262,7 +262,8 @@ class JSONDeserializer
 };
 
 static string
-    serialize(shared_ptr<ngraph::Function> func, size_t indent, bool binary_constant_data);
+    serialize_to_json(shared_ptr<ngraph::Function> func, size_t indent, bool binary_constant_data);
```
tsocha: Do we need to keep `binary_constant_data` flag?
Maybe it is worth to remove it or add to `serialize_to_cpio` function as well?

src/ngraph/serializer.cpp
```diff
@@ -453,7 +462,7 @@ static string serialize(shared_ptr<Function> func, size_t indent, bool binary_co
 
 std::string ngraph::serialize(std::shared_ptr<ngraph::Function> func, size_t indent)
 {
-    return ::serialize(func, indent, false);
+    return ::serialize_to_json(func, indent, false);
```
tsocha: function name is `serialize` but in body we have only serialization to json is it intended?
Maybe we should change this function name.

src/ngraph/serializer.cpp
mbencer: For json default value of  ```binary_constant_data``` is false for cpio true. In current version to both   ```serialize_to_json``` and ```serialize_to_cpio``` were proper default arguments. As a result interface is consistent.

src/ngraph/serializer.cpp
mbencer: As a element of unification I added possibility to call serialize method with expected cpio output type for this overloaded ```serialize``` method

src/ngraph/serializer.hpp
```diff
@@ -23,30 +23,38 @@
...
 namespace ngraph
 {
-    /// \brief Serialize a Function to a json string
+    /// \brief Serialize a Function to a json string or cpio
     /// \param func The Function to serialize
     /// \param indent If 0 then there is no formatting applied and the resulting string is the
     ///    most compact representation. If non-zero then the json string is formatted with the
     ///    indent level specified.
-    std::string serialize(std::shared_ptr<ngraph::Function> func, size_t indent = 0);
+    std::string serialize(std::shared_ptr<ngraph::Function> func,
```
diyessi: Better is to add separate `serialize_json` and `serialize_cpio` functions and let serialize continue doing what is used to do. There could be other kinds of serialization.

arogowie-intel: Maybe pass some enum deciding on serialization kind then? Writing all of below three specialization overloads for each serialization kind seems to me over verbose.

mbencer: From my point of view the best solution will be creating common objective interface (abstract) to serialize and creation dedicated classes json_serializer and cpio_serializer which inherit the interface.
I've tried to implement this, but there are too many dependencies in ```serializer.cpp``` to do it without huge refactoring.

Implementation of separeted function for cpio and json cause 6 version of seralization method in ```serializer.hpp```

I've implemented version proposed by @arogowie-intel . What do you think about it?

src/tools/serialize_onnx/serialize_onnx.cpp
```diff
@@ -70,7 +76,9 @@ int main(int argc, char** argv)
 
         ngraph::stopwatch timer;
         timer.start();
-        ngraph::serialize(output, function, 2);
+        const auto output_format = output_type == "cpio" ? ngraph::SerializationOutputFormat::CPIO
```
postrational: User should get feedback if they type in `cpio` by mistake.

src/tools/serialize_onnx/serialize_onnx.cpp
```diff
@@ -33,18 +33,20 @@ DESCRIPTION
     Serialize an ONNX model
 
 SYNOPSIS
-        serialize_onnx [-i|--input <input file>] [-o|--output <output file>]
+        serialize_onnx [-i|--input <input file>] [-o|--output <output file>] [-t|--type (json|cpio)]
 
 OPTIONS
         -i or --input  input ONNX file
         -o or --output output serialized model
+        -t or --type type of output ('json' or 'cpio'), 'json' is default choice
```
postrational: `default choice` -> `the default`

python/pyngraph/serializer.cpp
```diff
@@ -25,8 +25,15 @@ namespace py = pybind11;
 
 void regclass_pyngraph_Serializer(py::module m)
 {
+    py::enum_<ngraph::SerializationOutputFormat>(m, "output_format")
```
postrational: Please add a Python unit test for this.

PR #3202, opened 2019-07-10
Provenance fix for new nodes during replace

tachyon77:
Fixes provenance issue for replace_node() when new nodes are added by the replacement.


test/provenance.cpp
```diff
@@ -258,5 +258,6 @@ TEST(provenance, provenance)
         replace_node(c, d);
 
         EXPECT_EQ(d->get_provenance_tags(), (ProvSet{"tag_c", "tag_d"}));
+        EXPECT_EQ(e->get_provenance_tags(), (ProvSet{"tag_c", "tag_d"}));
```
aprocter: Could you update the comment to reflect the fact that E is not tagged initially? Given that that is the case, I think this property makes sense.

PR #3203, opened 2019-07-10
Fix vgg16 perf regression

gaurides:
With TF 1.14, the input graph for VGG16 is as below 
tf_1 14_sub_graph
MaxPool before Relu prevents Relu to be fused into ConvolutionBias. 
This PR reverses the order to enable Convolution+Bias+Relu fusion .


src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -520,6 +520,51 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_bprop()
     this->add_matcher(m, callback);
 }
 
+void ngraph::runtime::cpu::pass::CPUPreFusion::construct_maxpool_relu_switch()
```
nmostafa: Can we make this more general ? Templatize it to take two ops types to reverse.

ayzhuang: If we want to make this more general, then we cannot create new op in the callback because different ops require different arguments. We need to keep both ops and change the inputs/outputs of these four nodes: node_before->op1->op2->node_after.

nmostafa:
Replaced with #3260

PR #3206, opened 2019-07-11
Add instructions for building the NGRAPH_PLAIDML backend,

indie:
I was able to successfully build the `NGRAPH_PLAIDML` backend on Linux.  Yay!  This has been added to what I think is a logical place in the doc, though I'm more than happy to move it if you have any ideas for where it should be moved. Also took a slight bit of editorial freedom in describing the libplaidml.so as a "hybrid" backend, though I'm not entirely sure that is what we want to call it. 

Feedback / comments from anybody is welcomed. Let's make this doc awesome.


doc/sphinx/source/buildlb.rst
```diff
@@ -180,6 +161,58 @@ The process documented here will work on CentOS 7.4.
...
+Building nGraph-PlaidML from source
+===================================
+
+The following instructions will create the ``~/ngraph_plaidml_dist`` 
+locally:
+
+#. Ensure you have installed the :ref:`prerequisites` for your OS.
+
+#. Install the prerequisites for the backend. Our hybrid ``NGRAPH_PLAIDML``
+   backend works best with Python3 versions. Feel free to use a virtual
```
dgkutnic: "Feel free to use a virtual environment" should probably be "We recommend that you use a virtual environment" due to some of the difficulties that users have seen when trying to install outside of a venv.

indie: Done

PR #3207, opened 2019-07-11
Re-enable -Wswitch and -Wswitch-enum

aprocter:
These warnings are currently enabled locally in a number of places via pragmas, but I think it'd be a good idea to enable them globally. Doing so actually exposed a couple of small problems, and doesn't lead to as much as you might expect in the way of boiler-plate case enumeration.


src/ngraph/op/max.cpp
```diff
@@ -52,6 +52,8 @@ shared_ptr<Node> op::Max::get_default_value() const
         return make_constant_from_string("0", get_element_type(), get_shape());
     case element::Type_t::bf16:
         return make_constant_from_string("-INFINITY", get_element_type(), get_shape());
+    case element::Type_t::f16:
+        return make_constant_from_string("-INFINITY", get_element_type(), get_shape());
```
dcaballe: nit: looks like all these bf/f cases can be simplified to fall into a single `return make_constant_from_string("-INFINITY", get_element_type(), get_shape());` statement?

src/ngraph/op/min.cpp
```diff
@@ -52,6 +52,8 @@ shared_ptr<Node> op::Min::get_default_value() const
         return make_constant_from_string("1", get_element_type(), get_shape());
     case element::Type_t::bf16:
         return make_constant_from_string("INFINITY", get_element_type(), get_shape());
+    case element::Type_t::f16:
+        return make_constant_from_string("INFINITY", get_element_type(), get_shape());
```
dcaballe: Same here

src/ngraph/op/max.cpp
aprocter: Fixéd.

src/ngraph/op/min.cpp
aprocter: Fixëd.

silee2:
I see two compile errors. One will be solved by #3229 and the other would need some change in CPU emitter.

Issue #706, opened 2018-03-21
Windows support


thefiddler:
Are there plans to add support for Windows? Is there any estimate about how long that could take?

jennifermyers:
We are thinking about Windows support, but haven't yet scoped out when we will be getting to it. We'd be happy to receive contributions in this space!

samolisov:
I've added a windows support for the CPU runtime and compilation (using MSVC.h from clang's toolchains to get a path to Windows SDK and Universal CRT and enabling C++14 for the compiler on Windows). So, the generated by NGraph code is able to be compiled on Windows... **BUT Intel MKLDNN doesn't work** under the LLVM JIT. I've found the root cause - there is some magic around `std::shared_ptr` (the C API works fine, but C++ from <mkldnn.hpp> doesn't), in the reality even this code won't work under LLVM JIT:

```cpp
#include <memory>
#include <stdio.h>

int main() {
    int i = 5;
    std::shared_ptr<int> sptr = std::make_shared<int>(10);
    printf("i = %d\n", *sptr);
}
```
but this code works fine:

```cpp
#include <memory>
#include <stdio.h>

int main() {
    int i = 5;
    std::unique_ptr<int> uptr = std::make_unique<int>(10);
    printf("i = %d\n", *uptr);
}
```
(I've tried under the `lli` tool, playing with all the options, even in the interpreter mode it doesn't work).

It seems as a bug in LLVM: if I get the generated by NGraph LLVM IR (.ll) and compile it with the `llc` tool and then link with `lld-link`, the generated .exe file works fine (I'm just adding a `main` function calling `init_cg_ctx()`). I'm planning to open an issue in LLVM.

The current status is the following: NGraph can work on Windows only if `NGRAPH_CODEGEN=1` and `NGRAPH_CPU_USE_TBB` is disabled and `CPURuntimeContextCG* cg_ctx` is not used in the generated code (so the *abc* and *abc_operator* examples will work but something more complicated - won't).

PR #3216, opened 2019-07-12
Get NGraph CPU Runtime working on MS Windows

samolisov:
To make NGraph and CPU Runtime in the codegen mode, the following
changes are implemented:

  - header rewriter: the limit for one multiline string is changed to
    16300 as the maximum available for MSVC, see this documentation for
    details:
    https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2026?view=vs-2019

  - Introduced 'MSVS' constant is changed to built-in 'MSVC' one in all
    CMakeList.txt files but CMAKE_BUILD_TYPE propagation.
    The 'external_llvm.cmake' and 'external_llvm_prebuilt.cmake' dependencies
    are available for Windows and MSVC.

  - The 'codegen/compiler' component is able to compile the code on
    Windows using Windows SDK. The following changes are introduced: find
    the path to Windows SDK using clang's MSVC toolchain (the
    toolchains/MSVC.h and toolchains/Cuda.h headers must be copied from
    the used clang source tree to get access to the private API); C++14
    is enabled for MSVC (since Windows SDK cannot be compiled with
    C++11 only); the compiler works with MSVC compatibility and
    extensions.

  - 'cpu_external_function.cpp' is changed a bit to be compiled with
     MSVC (external templates are introduced to break double code
     generation with MSVC) and some implementation dependencies are fixed
     as well.

  - The sized delete operator is added to 'cpu_cg_runtime_context.hpp'.
    Unfortunately, MSVC Runtime doesn't export operator
    void operator delete ( void* ptr, std::size_t sz ) noexcept; (since
    c++14) (mangled name: ??3@YAXPEAX_K@Z). It may be fixed later, but now
    we have to define a work around.

Issue: #706

Signed-off-by: Pavel Samolysov <samolisov@gmail.com>


samolisov:
I've got 'codegen' working on Windows but it fails on C++ std's file I/O during runtime while reading `desc_file`. I've developed a simple compiler based on the 'codegen' library and it works only if the library has been statically linked. Theoretically, it could be done to get a standalone C++ application linked statically with 'ngraph', 'cpu_backend' and 'codegen' but we certainly cannot get a 'python.exe', for example, built with 'ngraph' for the `ngraph-tf` bridge.

cmake/external_llvm_prebuilt.cmake
```diff
@@ -143,10 +143,12 @@ set(LLVM_LINK_LIBS
     ${INSTALL_DIR}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}LLVMDemangle${CMAKE_STATIC_LIBRARY_SUFFIX}
 )
 
-if(APPLE)
-    set(LLVM_LINK_LIBS ${LLVM_LINK_LIBS} curses z m)
-else()
-    set(LLVM_LINK_LIBS ${LLVM_LINK_LIBS} tinfo z m)
+if(UNIX OR APPLE)
+    if(APPLE)
```
diyessi: One case for UNIX, one for APPLE, no need to nest in a UNIX or APPLE.

src/ngraph/codegen/CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ add_library(codegen SHARED ${SRC})
 # This must be kept in sync with the LLVM + Clang version in use
 if(NOT WIN32)
```
diyessi: Have an `if (WIN32)` and an `if (APPLE or UNIX)` so we don't break things on RSX-11.

src/ngraph/codegen/toolchains/Cuda.h
```diff
@@ -0,0 +1,203 @@
+//===--- Cuda.h - Cuda ToolChain Implementations ----------------*- C++ -*-===//
+//
```
diyessi: This is going to need to go through protex.
@aslepko

src/ngraph/codegen/toolchains/MSVC.h
```diff
@@ -0,0 +1,161 @@
+//===--- MSVC.h - MSVC ToolChain Implementations ----------------*- C++ -*-===//
+//
```
diyessi: This is going to need to go through protex.
@aslepko

src/ngraph/codegen/compiler.cpp
```diff
@@ -190,6 +198,25 @@ void codegen::CompilerCore::initialize()
     // Prevent Eigen from using any LGPL3 code
     args.push_back("-DEIGEN_MPL2_ONLY");
 
+#ifdef _WIN32
+    // The Optimization Level (see bellow) parameter can be ignored on Windows
```
rkimballn1: spelling error `bellow`

src/ngraph/codegen/compiler.cpp
```diff
@@ -227,8 +254,13 @@ void codegen::CompilerCore::initialize()
     auto LO = m_compiler->getInvocation().getLangOpts();
     LO->CPlusPlus = 1;
     LO->CPlusPlus11 = 1;
+#ifndef _WIN32
     // Strange but need to manually disable c++14
```
rkimballn1: I wonder if this is still true? Can we just enable c++14 for everything? Testing this would be difficult because CI does not currently test codegen

src/ngraph/codegen/compiler.cpp
```diff
@@ -441,6 +473,67 @@ void codegen::CompilerCore::configure_search_path()
...
+
+    Driver TheDriver(Path, TO.Triple, m_compiler->getDiagnostics());
+    llvm::opt::InputArgList Args;
+    Triple TT(TO.Triple);
+
+    MSVCToolChain MSVC(TheDriver, TT, Args);
+
+    add_header_search_path(MSVC.getSubDirectoryPath(MSVCToolChain::SubDirectoryType::Include));
+
+    if (MSVC.useUniversalCRT()) {
```
rkimballn1: This file needs to conform to coding standard. If you  run `make style-apply` it will apply correct brace style, etc.

src/ngraph/codegen/toolchains/Cuda.h
rkimballn1: We also don't want clang-format style applied to these files. Not sure how we do this.

src/ngraph/codegen/toolchains/Cuda.h
```diff
@@ -0,0 +1,203 @@
+//===--- Cuda.h - Cuda ToolChain Implementations ----------------*- C++ -*-===//
```
rkimballn1: Is this the stock Cuda.h file? Can't we get it from the normal cuda install header location? Is this from LLVM headers? Can we use those?

src/ngraph/codegen/toolchains/MSVC.h
```diff
@@ -0,0 +1,161 @@
+//===--- MSVC.h - MSVC ToolChain Implementations ----------------*- C++ -*-===//
```
rkimballn1: Can we use the default header file for this one too?

src/ngraph/runtime/cpu/cpu_emitter.cpp
```diff
@@ -139,7 +139,9 @@
 using namespace std;
 using namespace ngraph;
 
-static bool s_use_ref_kernels = (std::getenv("NGRAPH_CPU_USE_REF_KERNELS") != nullptr);
+static const bool s_use_ref_kernels = (std::getenv("NGRAPH_CPU_USE_REF_KERNELS") != nullptr);
+
+static const size_t s_one = 1;
```
rkimballn1: Can we get a comment that this is needed for MSVC compiler or whatever the reason we need this?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -302,6 +309,133 @@ static StaticInitializers s_static_initializers(s_output_dir);
 
 #define TI(x) type_index(typeid(x))
 
+#ifdef _MSC_VER
+// MSVC (and clang on Windows via clang-cl) instantiates a template
+// even when we know that it will be instantiated in cpu_emitter.cpp.
+// It leads to linker error: 'error: duplicate symbol'.
+extern template void runtime::cpu::CPU_Emitter::EMITTER_DECL(op::Add);
```
rkimballn1: Can we alphabetize this list?

test/CMakeLists.txt
```diff
@@ -241,9 +241,9 @@ if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "^(Apple)?Clang$")
     target_compile_options(unit-test PRIVATE -Wno-undef -Wno-reserved-id-macro)
 endif()
 
-# So many type_prop tests these days that we need to set /bigobj flag for MSVS.
+# So many type_prop tests these days that we need to set /bigobj flag for MSVC.
```
rkimballn1: Is this specifically the type_prop test file? This file is being broken into individual files, one for each op, so we might be able to not set /bigobj. Is there any downside to setting this? If no downside then it really does not matter.

src/ngraph/runtime/cpu/cpu_external_function.cpp
nmostafa: And move to a header file. Too much clutter here.

test/CMakeLists.txt
samolisov: I put MSVC just to eliminate a custom MSVS option because I use NMake instead of  Visual Studio. I have no idea about disadvantages of `/bigobj`, sorry.

src/ngraph/runtime/cpu/cpu_emitter.cpp
samolisov: Thank you for the idea, done.

samolisov:
@rkimballn1 @diyessi About MSVC.h and CUDA.h (MSVC.h requires CUDA.h for something). I've imported these files from Clang to get the Visual Studio and Windows Kit installation finding logic. Unfortunately, in fact the logic is to find libs. not includes and some string manipulation is required. Another way is just to copy (borrow) the code of the (unfortunately marked as `static`) methods from the `clang\lib\Driver\ToolChains\MSVC.cpp` file.

src/ngraph/codegen/toolchains/MSVC.h
samolisov: We can add NGraph header in this file and a comment as it did for some files in ngraph-tf: https://github.com/tensorflow/ngraph-bridge/blob/master/src/tf_deadness_analysis.cc Shall I do it?

src/ngraph/codegen/compiler.cpp
samolisov: Done.

src/ngraph/runtime/cpu/cpu_external_function.cpp
samolisov: @rkimballn1 @nmostafa Well, does `cpu_emitter_templates.hpp` sound good?

src/ngraph/codegen/compiler.cpp
samolisov: @rkimballn1 I believe any changing for not Windows must be applied as a part of another commit and PR, in the current code the commit 'Strange but need to manually disable c++14' is already present. I think the reason could be the sized delete operator introduced in c++14, I had a lot of problems with its symbol on Windows (I have no idea why, but the symbol is not available for JITed code by default).

cmake/external_llvm_prebuilt.cmake
samolisov: Done.

src/ngraph/codegen/CMakeLists.txt
samolisov: Done.

samolisov:
Just for info, when the cpu runtime is built as a static library, codegen runs fine even on Windows. Unfortunately, by default the JITed code is not able to load any symbols from the main module (.exe file) and all cpu specific kernels must be exported using the '/export' directive for the linker or (it looks better) be built into a shared library (.dll).

samolisov:
The PR has been approved by @silee2. Can anyone take a look, please? The merge conflict has been fixed.

PR #3223, opened 2019-07-13
- MLIR Binary ElementWise Op

pruthviIntel:
- Template function for binary element wise Op's
- Support for the following op'
     1. Subtract
     2. Divide
     3. Multiply
     4. Greater
     5. Less
     6. Maximum
     7. Minimum

Test results can be found in the below attachment
[mlir_test_log.txt](https://github.com/NervanaSystems/ngraph/files/3393446/mlir_test_log.txt)


src/contrib/mlir/ops_supported.inc
```diff
@@ -4,6 +4,13 @@
 #endif
 
 MLIR_OP(Add)
+MLIR_OP(Subtract)
```
nmostafa: Let's keep them alphabetical.

src/contrib/mlir/lowerer.cpp
```diff
@@ -538,6 +552,76 @@ namespace
...
+                        iRes(ivs) = ValueHandle(iLHS(ivs)) < ValueHandle(iRHS(ivs));
+                    }
+                    else if (isa<NGMaxOp>(op))
+                    {
+                        iRes(ivs) = edsc::intrinsics::select(ValueHandle(iLHS(ivs)) > ValueHandle(iRHS(ivs)), ValueHandle(iLHS(ivs)) , ValueHandle(iRHS(ivs)));
+                    }
+                    else if (isa<NGMinOp>(op))
+                    {
+                        iRes(ivs) = edsc::intrinsics::select(ValueHandle(iLHS(ivs)) < ValueHandle(iRHS(ivs)), ValueHandle(iLHS(ivs)), ValueHandle(iRHS(ivs)));
+                    }
```
dcaballe: We should be able to implement this in a more visitor pattern fashion instead of using a cascade of if-elses. That would be more efficient since when the template parameter is instantiated, we already know the specific op so we wouldn't need to execute `isa` again.

src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
```diff
@@ -284,6 +291,20 @@ bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node
...
+        if (!node->input(0).get_element_type().is_integral())
+            return false;
+    }
+
+    if (TI(ngraph::op::Greater) == TI(*node) || TI(ngraph::op::Less) == TI(*node))
+    {
+        // TODO: Remove this when MLIR has float point cmp support
+        if (!node->input(0).get_element_type().is_integral())
+            return false;
+    }
```
dcaballe: if op is Greater or Less and the prev condition is not met, shouldn't we return true here since we aren't evaluating Greater/Less properties anywhere below? Same for the rest of ops?

src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
```diff
@@ -291,6 +334,10 @@ bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node
         {
             return false;
         }
+        else
+        {
+            return true;
+        }
```
dcaballe: super-nit: we wouldn't need `else` after `if` with `return`. Not sure if this is in nGraph guidelines or it's just an LLVM thing.

PR #3225, opened 2019-07-15
[MLIR] Concat

aprocter:
Adds support for `Concat` to MLIR compiler/lowerer.

All but two tests in `CPU.concat*` are passing. The two that are failing are triggering this check in `compiled_kernel.cpp`: https://github.com/NervanaSystems/ngraph/blob/3f6cda4c22636fb5c0b7d2038eb4676712504671/src/ngraph/op/experimental/compiled_kernel.cpp#L78
I think this failure is unrelated to this PR, other than the fact that this PR exposes these tests to the MLIR/compiled_kernel machinery.

Example dump thru affine dialect:

```
[ RUN      ] CPU.concat_5d
*** nGraph Dialect Dump: ***


func @main(%arg0: !ng.tensor<2x3x3x3x2xf32>, %arg1: !ng.tensor<2x3x2x3x2xf32>, %arg2: !ng.tensor<2x3x4x3x2xf32>) -> !ng.tensor<2x3x9x3x2xf32> {
  %0 = "ng.concat"(%arg2, %arg0, %arg1) {concatenation_axis = 2 : i64} : (!ng.tensor<2x3x4x3x2xf32>, !ng.tensor<2x3x3x3x2xf32>, !ng.tensor<2x3x2x3x2xf32>) -> !ng.tensor<2x3x9x3x2xf32>
  "ng.return"(%0) : (!ng.tensor<2x3x9x3x2xf32>) -> ()
}



*** Affine Dialect Dump: ***
#map0 = () -> (0)
#map1 = () -> (2)
#map2 = () -> (3)
#map3 = () -> (4)
#map4 = (d0) -> (d0 + 4)
#map5 = (d0) -> (d0 + 7)


func @main(%arg0: memref<2x3x3x3x2xf32>, %arg1: memref<2x3x2x3x2xf32>, %arg2: memref<2x3x4x3x2xf32>, %arg3: memref<2x3x9x3x2xf32>, %arg4: index) {
  affine.for %i0 = 0 to 2 {
    affine.for %i1 = 0 to 3 {
      affine.for %i2 = 0 to 4 {
        affine.for %i3 = 0 to 3 {
          affine.for %i4 = 0 to 2 {
            %0 = load %arg2[%i0, %i1, %i2, %i3, %i4] : memref<2x3x4x3x2xf32>
            store %0, %arg3[%i0, %i1, %i2, %i3, %i4] : memref<2x3x9x3x2xf32>
          }
        }
      }
    }
  }
  affine.for %i5 = 0 to 2 {
    affine.for %i6 = 0 to 3 {
      affine.for %i7 = 0 to 3 {
        %1 = affine.apply #map4(%i7)
        affine.for %i8 = 0 to 3 {
          affine.for %i9 = 0 to 2 {
            %2 = load %arg0[%i5, %i6, %i7, %i8, %i9] : memref<2x3x3x3x2xf32>
            store %2, %arg3[%i5, %i6, %1, %i8, %i9] : memref<2x3x9x3x2xf32>
          }
        }
      }
    }
  }
  affine.for %i10 = 0 to 2 {
    affine.for %i11 = 0 to 3 {
      affine.for %i12 = 0 to 2 {
        %3 = affine.apply #map5(%i12)
        affine.for %i13 = 0 to 3 {
          affine.for %i14 = 0 to 2 {
            %4 = load %arg1[%i10, %i11, %i12, %i13, %i14] : memref<2x3x2x3x2xf32>
            store %4, %arg3[%i10, %i11, %3, %i13, %i14] : memref<2x3x9x3x2xf32>
          }
        }
      }
    }
  }
  return
}
```


src/contrib/mlir/dialect/ops.td
```diff
@@ -182,6 +182,18 @@ def NGDotOp : NG_Binary_Op<"dot">
   let verifier = [{ return verifyOp(this); }];
 }
 
+// TODO(amprocte): Might be nice to rebase this on some sort of NG_Variadic_Op
+// class, but I'm not sure how to add concatenation_axis into the args if we
+// do that.
```
dcaballe: Maybe doing something similar to `list<OpTrait> traits = []` below would work?

src/contrib/mlir/lowerer.cpp
```diff
@@ -530,6 +530,102 @@ namespace
...
+                        axis++;
+                        make_copy_loop();
+                    });
+                }
+                else
+                {
+                    IndexedValue iv_operand(operand);
+                    iv_res(res_index_handles) = iv_operand(operand_index_handles);
+                }
+            };
```
dcaballe: Nice! I didn't know we could use LoopBuilder like this :)

src/contrib/mlir/lowerer.cpp
```diff
@@ -530,6 +530,102 @@ namespace
...
+            //     ...
+            //      for i_(r-2) := 0 to operand_dim[r-2]:
+            //       for i_(r-1) := 0 to operand_dim[r-1]:
+            //        result[i_0][i_1]...
+            //              [i_(A-1)][i_A + concatenation_axis_pos][i_(A+1)]...
+            //              [i_(r-2)][i_(r-1)]
+            //                  :=
+            //        operand[i_0][i_1]...[i_(r-2)][i_(r-1)]
+            std::vector<IndexHandle> res_index_handles;
+            std::vector<IndexHandle> operand_index_handles;
```
dcaballe: Could you please check if the consumer of these `std::vector`s actually expect an `llvm::SmallVector`? If so you could use those instead

src/contrib/mlir/lowerer.cpp
```diff
@@ -530,6 +530,102 @@ namespace
...
+
+            // Move up concatenation_axis_pos for the next operand.
+            concatenation_axis_pos = concatenation_axis_pos + v_operand.ub(concatenation_axis);
+        }
+
+        rewriter.replaceOp(op, {result});
+
+        return matchSuccess();
+    }
+
```
dcaballe: Would you mind using MLIR's camel naming convention in this file? I still have to fix this in this file and add an note with what he discussed about it.

src/contrib/mlir/dialect/ops.td
nmostafa: You can try to do dag concat in the leaf class. 
`let arguments = !con(arguments, (ins I64Attr:$concatenation_axis))`
https://llvm.org/docs/TableGen/LangIntro.html

src/contrib/mlir/lowerer.cpp
```diff
@@ -530,6 +530,102 @@ namespace
         return matchSuccess();
     }
 
+    REWRITER(NGConcatOp)
+    {
+        auto concat = cast<NGConcatOp>(op);
+        auto loc = concat.getLoc();
+
+        ScopedContext scope(rewriter, loc);
```
nmostafa: This is used by LoopBuilders, maybe better to move to right before it is needed below.

src/contrib/mlir/lowerer.cpp
```diff
@@ -530,6 +530,102 @@ namespace
...
+
+        for (auto& operand : operands)
+        {
+            // Retrieve and check some type info.
+            NGRAPH_CHECK(operand, "Unexpected null operand in ConcatOp");
+            auto operand_ty = operand->getType().dyn_cast<MemRefType>();
+            NGRAPH_CHECK(operand_ty, "Unexpected non-memref operand type");
+            NGRAPH_CHECK(elem_ty == operand_ty.getElementType(), "Types mismatch in ConcatOp");
+
+            // TODO(amprocte): Check consistency of operand shape (rank, dims other than at
```
nmostafa: That can be achieved via a call to op->verify(). I think it is safe to assume the code is already verified by calling Module->verify() beforehand.

src/contrib/mlir/lowerer.cpp
```diff
@@ -530,6 +530,102 @@ namespace
...
+            size_t axis = 0;
+            std::function<void()> make_copy_loop = [&] {
+                if (axis < rank)
+                {
+                    IndexHandle i;
+                    IndexHandle i_lb(v_operand.lb(axis));
+                    IndexHandle i_ub(v_operand.ub(axis));
+                    int64_t i_step = v_operand.step(axis);
+
+                    LoopBuilder(&i, i_lb, i_ub, i_step)([&] {
```
nmostafa: Is there a reason why we are doing it recursively ? MLIR already has `LoopNestBuilder` to build an entire loop nest which should be faster.

nmostafa:
You can remove the `compiled_kernel` assert. It is left-over from previous purpose of that node. Nodes obviously don't need to be of same type/shape.

src/contrib/mlir/lowerer.cpp
```diff
@@ -530,6 +530,102 @@ namespace
         return matchSuccess();
     }
 
+    REWRITER(NGConcatOp)
```
nmostafa: Can you please add a unit-test with Concat + other node (e.g. dot/add) ? Unless this is already tested with existing tests.

src/contrib/mlir/dialect/ops.td
aprocter: Will give this a try

src/contrib/mlir/lowerer.cpp
aprocter: Will check (thanks, still familiarizing myself with LLVM's parallel universe of types here  ;))

src/contrib/mlir/lowerer.cpp
aprocter: Will do

src/contrib/mlir/lowerer.cpp
aprocter: I think some of the `NGRAPH_CHECK`s above could probably go away as well, right?

src/contrib/mlir/lowerer.cpp
aprocter: Nice, should have looked for such a thing :)

src/contrib/mlir/lowerer.cpp
aprocter: Will do

src/contrib/mlir/lowerer.cpp
nmostafa: yes.

src/contrib/mlir/lowerer.cpp
```diff
@@ -530,6 +530,102 @@ namespace
...
+                if (axis < rank)
+                {
+                    IndexHandle i;
+                    IndexHandle i_lb(v_operand.lb(axis));
+                    IndexHandle i_ub(v_operand.ub(axis));
+                    int64_t i_step = v_operand.step(axis);
+
+                    LoopBuilder(&i, i_lb, i_ub, i_step)([&] {
+                        res_index_handles.push_back(axis == concatenation_axis
+                                                        ? IndexHandle(i + concatenation_axis_pos)
```
nmostafa: It's new to me how affine operations on IndexHandle auto-generate maps :)

aprocter: Yeah I was impressed  :)

src/contrib/mlir/lowerer.cpp
aprocter: Have rewritten to use this. Note that this winds up generating slightly different code at the affine level: the application of the affine map happens in the innermost loop, rather than in the loop for the concatenation-axis dimension. I figure this probably doesn't make any difference.

src/contrib/mlir/lowerer.cpp
aprocter: I've removed `NGRAPH_CHECK`s that I _think_ are unnecessary given the assumption that the module has been verified. You might take a second look to see if I'm overdoing it.

src/contrib/mlir/lowerer.cpp
aprocter: I can't seem to move this much lower without a segfault. Looks like something down in the bowels of MemRefView's ctor wants to refer to a ScopedContext.

src/contrib/mlir/dialect/ops.td
aprocter: Tablegen seems to interpret this as a recursive rebinding of a "arguments" or something... it complains about not being able to result `!con(!con(arguments, (ins I64Attr:$concatenation_axis)), (ins I64Attr:$concatenation_axis)))`. May be best to stick with this until someone who knows what they are doing with tablegen does a refactor. ;)

src/contrib/mlir/lowerer.cpp
aprocter: Removing the `compiled_kernel` assert enables two such tests---one of which seems to be failing! (The affine map is not being computed correctly.) Will investigate.

src/contrib/mlir/lowerer.cpp
aprocter: Have switched everything to `SmallVector` here.

aprocter:
@dcaballe Regarding tests:

```
$ NGRAPH_MLIR=1 test/unit-test --gtest_filter='CPU.*concat*'
Note: Google Test filter = CPU.*concat*
[==========] Running 16 tests from 1 test case.
[----------] 16 tests from CPU
[ RUN      ] CPU.concat_matrix_colwise
[       OK ] CPU.concat_matrix_colwise (12 ms)
[ RUN      ] CPU.concat_matrix_rowwise
[       OK ] CPU.concat_matrix_rowwise (9 ms)
[ RUN      ] CPU.concat_matrix_int64
[       OK ] CPU.concat_matrix_int64 (8 ms)
[ RUN      ] CPU.concat_vector
[       OK ] CPU.concat_vector (6 ms)
[ RUN      ] CPU.concat_4d_tensor
[       OK ] CPU.concat_4d_tensor (9 ms)
[ RUN      ] CPU.concat_2d_tensor
[       OK ] CPU.concat_2d_tensor (7 ms)
[ RUN      ] CPU.concat_in_place_2d_tensor
[       OK ] CPU.concat_in_place_2d_tensor (9 ms)
[ RUN      ] CPU.concat_in_place_propagate_2d_tensor
/localdisk/amprocte/Work/ngraph/build/test/backend_test_CPU.cpp:1031: Failure
Value of: test::all_close_f( (vector<float>{3, 7, 2}), read_vector<float>(result), MIN_FLOAT_TOLERANCE_BITS)
  Actual: false (7 is not close to 2 at index 1
2 is not close to 0 at index 2
diff count: 2 out of 3
passing criteria - mismatch allowed  @ mantissa bit: 24 or later (0 tolerance bits)
0 value(s) below min_signal: 0
tightest match   - mismatch occurred @ mantissa bit: 24 or next bit (3 vs 3 at [0])
loosest match    - mismatch occurred @ mantissa bit: 0 or next bit (2 vs 0 at [2])
median match     - mismatch occurred @ mantissa bit: 0 or next bit
)
Expected: true
[  FAILED  ] CPU.concat_in_place_propagate_2d_tensor (9 ms)
[ RUN      ] CPU.concat_5d
[       OK ] CPU.concat_5d (53 ms)
[ RUN      ] CPU.concat_zero_length_1d_last
[       OK ] CPU.concat_zero_length_1d_last (5 ms)
[ RUN      ] CPU.concat_zero_length_1d_middle
[       OK ] CPU.concat_zero_length_1d_middle (5 ms)
[ RUN      ] CPU.concat_zero_zero
[       OK ] CPU.concat_zero_zero (1 ms)
[ RUN      ] CPU.concat_zero_length_4d_middle
[       OK ] CPU.concat_zero_length_4d_middle (9 ms)
[ RUN      ] CPU.backwards_concat_vector
[       OK ] CPU.backwards_concat_vector (8 ms)
[ RUN      ] CPU.backwards_concat_axis_0
[       OK ] CPU.backwards_concat_axis_0 (11 ms)
[ RUN      ] CPU.backwards_concat_axis_1
[       OK ] CPU.backwards_concat_axis_1 (134 ms)
[----------] 16 tests from CPU (295 ms total)

[==========] 16 tests from 1 test case ran. (296 ms total)
[  PASSED  ] 15 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] CPU.concat_in_place_propagate_2d_tensor

 1 FAILED TEST
17:33:02.208622: [MAIN] Tests finished: Time: 297 ms Exit code: 1
```

With `NGRAPH_MLIR_DUMP_ALL=1`, all the above are dumping reasonable-looking affine code.

The failed test is probably due to the issue we've been discussing on Slack.

src/contrib/mlir/dialect/ops.td
nmostafa: There are subtleties about how the classes are expanded. I will address it later.

src/contrib/mlir/lowerer.cpp
aprocter: Alas, `LoopNestBuilder` seems like a better option, but this was fun to write  ;)

src/contrib/mlir/lowerer.cpp
dcaballe: We don't have to get crazy about checks but some checks would be needed. Module verification happens all at once before lowering and it might be disabled in some compilation modes in the future (same as in LLVM). We would also need to check for nullptrs (KW will complain) and the own lowerer may leave ops and types in an inconsistent way. I hit similar memref type asserts before. In general, I would say that checking for nullptrs and expected invariants, such as types, should be fine

src/contrib/mlir/lowerer.cpp
```diff
@@ -534,91 +534,72 @@ namespace
...
+        MemRefView vRes(result);
+        auto rank = vRes.rank();
 
         // For each operand, generate a separate loop to copy into the target slice of "result".
         // We'll keep track of the slice offsets via concatenation_axis_pos.
-        IndexHandle concatenation_axis_pos(index_t(0));
+        auto concatenationAxis = concat.concatenation_axis().getSExtValue();
+        IndexHandle concatenationAxisPos(index_t(0));
 
         for (auto& operand : operands)
```
dcaballe: nit: I think operand has pointer type, better use * instead of reference?

aprocter: The `auto&` here will give us references to the pointers, so I think this should be okay.

src/contrib/mlir/lowerer.cpp
aprocter: @dcaballe Makes sense, I'll put some checks in for nullptrs and type assumptions.

src/contrib/mlir/lowerer.cpp
aprocter: (After some offline chatter) looks like this is an unrelated bug, to be addressed by #3233.

aprocter:
@dcaballe @nmostafa I think comments are addressed at this point. Let me know.

src/contrib/mlir/lowerer.cpp
```diff
@@ -534,91 +534,77 @@ namespace
...
         auto concat = cast<NGConcatOp>(op);
         auto loc = concat.getLoc();
-
         ScopedContext scope(rewriter, loc);
 
         // Create Value for result, and extract type info.
         Value* result = m_pass.buildOutputDefs(op, rewriter)[0];
         NGRAPH_CHECK(result, "Unexpected null result in ConcatOp");
+        auto resultTy = result->getType().dyn_cast<MemRefType>();
+        NGRAPH_CHECK(resultTy, "Unexpected non-memref result type");
```
dcaballe: Sorry, my bad here. I missed the `dyn_cast` before. You can use `cast` instead of `dyn_cast` and remove the `NGRAPH_CHECK` since `cast` will assert if the casting doesn't succeed. That's the way it's usually done in LLVM.

PR #3227, opened 2019-07-15
Use mkl-dnn v1.0 or v0.x depending on compilation flag.

ayzhuang:
Modify CPU backend  to use mkldnn v1.0.
Support both v1.0 and v0.x for DEX and only v1.0 for CODEGEN.
Default is not mkldnn v1.0.
Use -NGRAPH_USE_LEGACY_MKLDNN=FALSE to build mkldnn v1.0.

Some notes:
1) mkldnn v1.0 does not depend on mkl. But we still build mkl because we use cblas_sgemm_batch from mkl.
2) mkldnn v1.0 splits hidden state and cell state, so Rnn/Lstm has one more input and one more output for v1.0.
3) mkldnn v1.0 does not keep format info like "nchw" in its memory descriptor any more. We get strides information from memory descriptor and put it in visualized graph.


src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -138,6 +138,1113 @@ std::vector<size_t>& MKLDNNEmitter::get_primitive_deps(size_t index)
     return m_primitive_deps.at(index);
 }
 
+mkldnn::memory::desc
+    MKLDNNEmitter::build_blocked_memory_descriptor(const mkldnn::memory::dims& dim,
+                                                   const mkldnn::memory::dims& strides,
+                                                   mkldnn::memory::data_type dtype) const
+{
+#if defined(USE_MKLDNN_V1)
```
nmostafa: I really don't like the scattered `#if` all-over the place. I would suggest creating two files `mkldnn_emitter_0_x.cpp.in` and `mkldnn_emitter_1_0.cpp.in` and #include them here based on the compile-time flag. 
What do you think ?

nmostafa: Or change the CMakeList to choose the right file.

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
```diff
@@ -138,6 +138,1113 @@ std::vector<size_t>& MKLDNNEmitter::get_primitive_deps(size_t index)
...
+    return dequantize_index;
+}
+
+size_t MKLDNNEmitter::build_reorder(const mkldnn::memory::desc& input_desc,
+                                    const mkldnn::memory::desc& result_desc)
+{
+    size_t input_index, result_index, primitive_index;
+
+#if defined(USE_MKLDNN_V1)
+    input_index = build_memory(input_desc);
```
nmostafa: identical code in #if and #else ?

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
ayzhuang: I can remove the #if inside of any function and write two functions instead. That way, only one #if is needed. There are a lot of other files need to be changes. Couple of #if would be fine in one file?

src/ngraph/runtime/cpu/mkldnn_emitter.cpp
ayzhuang: one is build_memory and another one is build_memory_primitive. In mkl_dnn v1.0, mkldnn::memory is no longer primitive.

src/ngraph/runtime/cpu/mkldnn_emitter.hpp
```diff
@@ -1400,8 +1159,575 @@ namespace ngraph
...
+                    {
+                        prim = new mkldnn::inner_product_forward({desc, attr, engine},
+                                                                 *mkldnn_primitives[input_idx],
+                                                                 *mkldnn_primitives[weights_idx],
+                                                                 *mkldnn_primitives[results_idx]);
+                    }
+
+                    mkldnn_primitives[ip_idx] = prim;
+                }
+#endif
```
dcaballe: This is actually a lot of code replication :). Would it be possible to use an approach similar to FORMAT macro in some cases to reduce part of the replication?

src/ngraph/runtime/cpu/mkldnn_utils.cpp
```diff
@@ -58,7 +58,26 @@ static const mkldnn_version_t* get_mkldnn_version()
 std::map<element::Type, const mkldnn::memory::data_type>&
     runtime::cpu::mkldnn_utils::get_mkldnn_data_type_map()
 {
-    // Mapping from POD types to MKLDNN data types
+// Mapping from POD types to MKLDNN data types
+#if defined(USE_MKLDNN_V1)
+    static std::map<element::Type, const mkldnn::memory::data_type> s_mkldnn_data_type_map = {
```
dcaballe: Would `unordered_map` be enough here? It would be more efficient. Same for other places where `map` and `set` are used.

ayzhuang:
> I share the same concern as Nagy wrt if/else and code replication (sorry, I didn't follow all the details of this discussion and if you reached a conclusion. Disregard my comments if they are not aligned with that). Is having support for multiple MKLDNN versions something temporal?

I will come back to mkldnn_emitter after finishing all the other files. Don't know for how long we will support both versions.

CMakeLists.txt
```diff
@@ -528,7 +536,12 @@ if(NGRAPH_CPU_ENABLE OR NGRAPH_GENERIC_CPU_ENABLE)
     include(cmake/external_eigen.cmake)
 endif()
 if(NGRAPH_CPU_ENABLE)
-    include(cmake/external_mkldnn.cmake)
+	if(NGRAPH_USE_MKLDNN_V1)
```
rkimballn1: fix the indenting on this

src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ set(SRC
     builder/avg_pool.cpp
     builder/argmin.cpp
     builder/argmax.cpp
-    builder/batch_norm.cpp
+ 	builder/batch_norm.cpp
```
rkimballn1: Why the whitespace change? It looks wrong now. Same comment on line 62

src/ngraph/runtime/cpu/pregenerated_src/cpu_cg_runtime_context.hpp
```diff
@@ -21,6 +21,57 @@
...
+                    QUANTIZEDAVGPOOL,
+                    QUANTIZEDMAXPOOL,
+                    QUANTIZEDCONCAT,
+                    QUANTIZEDDOTBIAS,
+                    QUANTIZEDMATMUL,
+                    QUANTIZEDCONVOLUTION,
+                    QUANTIZEDCONVOLUTIONBIAS,
+                    QUANTIZEDCONVOLUTIONBIASADD,
+                    QUANTIZEDCONVOLUTIONBIASSIGNEDADD,
+					QUANTIZEDCONVOLUTIONRELU,
```
rkimballn1: Are there tabs in this file? Please fix

src/ngraph/runtime/cpu/pregenerated_src/cpu_cg_runtime_context.hpp
```diff
@@ -29,32 +80,190 @@ struct CPURuntimeContextCG
...
+    std::vector<mkldnn::memory::desc*> mkldnn_scratchpad_mds;
+    AlignedBuffer* scratchpad_buffer;
     std::vector<char*> mkldnn_workspaces;
     std::vector<mkldnn::memory::desc*> mkldnn_descriptors;
 
-    mkldnn::engine global_cpu_engine = mkldnn::engine(mkldnn::engine::cpu, 0);
+    mkldnn::engine global_cpu_engine = mkldnn::engine(mkldnn::engine::kind::cpu, 0);
 
-    void set_memory_ptr(size_t primitive_index,
+	void set_memory_ptr(size_t index,
```
rkimballn1: I suspect tabs again. Indenting is wrong.

src/ngraph/runtime/cpu/builder/avg_pool.cpp
```diff
@@ -55,6 +55,10 @@ namespace ngraph
                     auto avg_pool_desc =
                         mkldnn_emitter->get_avg_pooling_forward_desc<ngraph::op::AvgPool>(node,
                                                                                           false);
+#if MKLDNN_VERSION_MAJOR >= 1
+                    mkldnn_emitter->query_scratchpad_pooling_forward(avg_pool_desc);
```
nmostafa: Instead we can define a MACRO/function to do this. If `MKLDNN_VERSION_MAJOR < 1` it resolves/does nothing. Same for all other uses with #if

src/ngraph/runtime/cpu/mkldnn_utils.hpp
```diff
@@ -24,6 +24,76 @@
...
+#define BN_FLAG_CLASS normalization_flags
+
+#define PADDING                                                                                    \
+        )
+
+#define SET_ROUND_MODE                                                                             \
+    {                                                                                              \
+    }
+
+#define QUERY_SCRATCHPAD_SUM(X) mkldnn_emitter->query_scratchpad_sum(X);
```
nmostafa: One macro that takes the func name + variadic args list would be better : `QUERY_SCRATCHPAD(sum, x)`
nit: wrap in do/while(0)

src/ngraph/runtime/cpu/mkldnn_utils.hpp
```diff
@@ -24,6 +24,76 @@
...
 
+#if MKLDNN_VERSION_MAJOR < 1
+#define FORMAT format
+#define FORMAT_KIND format
+#define FORMAT_KIND_UNDEF mkdnn_format_undef
+#define FORMAT_ANY mkldnn_any
+#define FORMAT_UNDEF mkldnn_undef
+#define DATA_UNDEF data_undef
+
+#define CHANGE_FORMAT                                                                              \
```
nmostafa: wrap in do/whole(0), so we can do:
`CHANGE_FORMAT();` Same for other macros.

src/ngraph/runtime/cpu/mkldnn_utils.hpp
```diff
@@ -24,6 +24,76 @@
...
+        weights_desc.data.format = mkldnn_oihw;                                                    \
+    }                                                                                              \
+    if (weights_desc.data.format == mkldnn_ncdhw)                                                  \
+    {                                                                                              \
+        weights_desc.data.format = mkldnn_oidhw;                                                   \
+    }
+
+#define BN_FLAG_CLASS batch_normalization_flag
+
+#define PADDING                                                                                    \
```
nmostafa: Why not leave the `)` out of the macro ?

src/ngraph/runtime/cpu/mkldnn_utils.hpp
```diff
@@ -24,6 +24,76 @@
...
+    , mkldnn::padding_kind::zero)
+
+#define SET_ROUND_MODE attr.set_int_output_round_mode(mkldnn::round_mode::round_nearest);
+
+#define QUERY_SCRATCHPAD_SUM(X)                                                                    \
+    {                                                                                              \
```
nmostafa: why not just empty def ?

src/ngraph/runtime/cpu/pregenerated_src/cpu_cg_runtime_context.hpp
```diff
@@ -128,23 +342,24 @@ extern "C" void destroy_cg_ctx(CPURuntimeContextCG* cg_ctx)
...
+	deserialize_memory_descs_and_build_memory(std::ifstream& desc_file,
+														 CPURuntimeContextCG* cg_ctx,
+														 size_t descs_count)
 {
     cg_ctx->mkldnn_descriptors = std::vector<mkldnn::memory::desc*>(descs_count);
     for (auto i = 0; i < descs_count; i++)
     {
-        size_t primitive_index;
-        desc_file >> primitive_index;
+    		size_t index;
```
nmostafa: nit: tabs ?

src/ngraph/runtime/cpu/mkldnn_utils.hpp
ayzhuang: why is do/while(0) better?

CMakeLists.txt
```diff
@@ -163,6 +163,7 @@ ngraph_var(NGRAPH_USE_PREBUILT_LLVM DEFAULT "FALSE")
 option(NGRAPH_UNIT_TEST_ENABLE "Control the building of unit tests" TRUE)
 option(NGRAPH_TOOLS_ENABLE "Control the building of tool" TRUE)
 option(NGRAPH_CPU_ENABLE "Control the building of the CPU backend" TRUE)
+option(NGRAPH_USE_MKLDNN_V1 "Use MKLDNN-v1.0" FALSE)
```
silee2: v1 will be the default at some point. shouldn't we have the opposite option like "NGRAPH_USE_LEGACY_MKLDNN" and set "TRUE" as default and phase it out at some point?

ayzhuang: will change the option. thanks.

src/ngraph/runtime/cpu/builder/avg_pool.cpp
ayzhuang: Changed to macro.

ayzhuang:
@nmostafa @rkimballn1 @aslepko Please review.

CMakeLists.txt
ayzhuang: have fixed indenting problems.

ayzhuang:
> Blocking for now for follow ups on performance impact in onnxrt and ngtf.

There are still some performance degradation in onnxrt, caused by mkldnn int8 convolution regression. Mkldnn team is working on that.

PR #3232, opened 2019-07-16
[MLIR] Add llvm.noalias attribute to nGraph func args

dcaballe:
Aliasing prevents vectorization and other optimizations in LLVM (dot2d,
for example). I found that LLVM dialect contemplates the possibility of
previous dialects adding llvm.noalias attribute to function arguments so
that the NoAlias attribute is generated in LLVM-IR. This should be good for
now although we should discuss how to model this in Standard dialect and co.


nmostafa:
I don't think this is something we need to add during dialect conversion, it should be somewhere during affine->std conversion within MLIR. If that attr is not inserted during affine lowering, then it is an MLIR issue and we should raise it. Thoughts ?

dcaballe:
Yeah, I agree that this is something that would need specific modeling in affine/standard instead of overusing the llvm.noalias attribute. I planned to send an email to the mailing list about this. I'm not sure, though, why this should be modeled as part of the affine->std lowering. Non-aliasing function parameters is something nGraph-specific that will not be true for other dialects so I think we should make that property explicit once we leave nGraph dialect, right?

PR #3233, opened 2019-07-16
[MLIR] Fix incorrect callback lookup

nmostafa:
Hit an MLIR snag with emitting call-backs for allocation directly in std dialect. We cannot return a type-erased memref and cast it to the requested type. This change uses `std.AllocOp` instead to allocate which calls `malloc` directly. This should be fine for static tensors since eventually we would like to pre-allocate a buffer and do buffer offset assignment on the temps where `std.AllocOp`s will be replaced with buffer+offset assignment to the memref underlying buffer. 

Some discussions on the topic:
https://groups.google.com/a/tensorflow.org/forum/#!topic/mlir/b3a4ZiX9kDU
https://groups.google.com/a/tensorflow.org/forum/#!topic/mlir/zqExhUHJHFk

This causes all concat tests to pass after #3225


src/contrib/mlir/lowerer.cpp
```diff
@@ -311,14 +315,17 @@ namespace
                                                      PatternRewriter& rewriter)
     {
         auto callBackFuncPtr = getModule().getNamedFunction(name);
-        if (callBackFuncPtr == nullptr)
+        if (callBackFuncPtr)
```
aprocter: The changes in this function were for `__mlir_allocate`. Do we still want them?

PR #3236, opened 2019-07-16
[Py] Added elu operator to Python API.

Ewa21:


python/ngraph/ops.py
```diff
@@ -202,6 +202,18 @@ def reshape(node, output_shape, input_order=None, name=None):
     return Reshape(node, AxisVector(input_order), Shape(output_shape))
 
 
+@nameable_op
```
postrational: Let's move this function to another place in the file, outside of the section marked as `# Unary ops`

python/ngraph/ops.py
```diff
@@ -202,6 +202,18 @@ def reshape(node, output_shape, input_order=None, name=None):
...
+def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
+    """Perform elu operation on data node element-wise with the specified alpha parameter.
+
+    :param data: One of: input node, array or scalar.
```
postrational: We now added information on supported types, but what is this param supposed to provide in terms of the function's logic?

python/ngraph/ops.py
```diff
@@ -202,6 +202,18 @@ def reshape(node, output_shape, input_order=None, name=None):
...
+    :param alpha: One of: input node or scalar.
```
postrational: What does this parameter do? Check the ONNX or nGraph docs for this information and add it here.

python/test/ngraph/test_ops_binary.py
```diff
@@ -184,3 +184,22 @@ def test_binary_operators_with_scalar(operator, numpy_function):
     result = computation(value_a)
     expected = numpy_function(value_a, value_b)
     assert np.allclose(result, expected)
+
+
+def test_elu_operator():
+    runtime = get_runtime()
```
postrational: Let's add one more test, where we pass in an `nd.array` and a scalar number as inputs instead of `ng.parameter`s

python/ngraph/ops.py
```diff
@@ -60,6 +60,23 @@ def constant(value, dtype=None, name=None):  # type: (NumericData, NumericType,
...
 
 
+@nameable_op
+def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
+    """Perform Exponential Linear Unit operation element-wise on data from input node.
+
+    Computes exponential linear: alpha * (exp(data) - 1) if < 0, data otherwise.
+
+    For more information refer to:
+    `Fast and Accurate Deep Network Learning by Exponential Linear Units (ELUs) <http://arxiv.org/abs/1511.07289>`_
```
arogowie-intel: Some artifact `_` at the end of line?

python/test/ngraph/test_ops_binary.py
```diff
@@ -184,3 +184,22 @@ def test_binary_operators_with_scalar(operator, numpy_function):
...
+    parameter_data = ng.parameter(data_shape, name='Data', dtype=np.float32)
+    parameter_alpha = ng.parameter(alpha_shape, name='Alpha', dtype=np.float32)
+
+    model = ng.elu(parameter_data, parameter_alpha)
+    computation = runtime.computation(model, parameter_data, parameter_alpha)
+
+    value_data = np.array([[-5, 1], [-2, 3]], dtype=np.float32)
+    value_alpha = np.array([3, 3], dtype=np.float32)
+
+    result = computation(value_data, value_alpha)
```
arogowie-intel: You may use our utility function in a similar way as here: https://github.com/NervanaSystems/ngraph/blob/etusien/elu/python/test/ngraph/test_reduction.py#L44. Then you just pass nGraph function `ng.elu` and its arguments as raw data (`value_data`, `value_alpha`) or parameters (`parameter_data`, `parameter_alpha`).

python/ngraph/ops.py
postrational: No, that's how ReStructuredText represents links.

python/ngraph/ops.py
```diff
@@ -60,6 +60,24 @@ def constant(value, dtype=None, name=None):  # type: (NumericData, NumericType,
...
 
+@nameable_op
+def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
+    """Perform Exponential Linear Unit operation element-wise on data from input node.
+
+    Computes exponential linear: alpha * (exp(data) - 1) if < 0, data otherwise.
+
+    For more information refer to:
+    `Fast and Accurate Deep Network Learning by Exponential Linear Units (ELUs)
+    <http://arxiv.org/abs/1511.07289>`
```
postrational: Unfortunately, removing `_` broke the RST link. We'll need to fix it.

PR #3251, opened 2019-07-17
pruthvi/cpu static backend

pruthviIntel:
1. This PR allows the cpu/interpreter to be build as the static library
2. `-DNGRAPH_INTERPRETER_STATIC_LIB_ENABLE=TRUE` && ` -DNGRAPH_CPU_STATIC_LIB_ENABLE=TRUE` can used as Cmake options to build `libinterpreter_backend.a` && `libcpu_backend.a` respectively


src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -50,6 +51,23 @@ extern "C" runtime::BackendConstructor* get_backend_constructor_pointer()
...
 }
 
+#ifndef CPU_BACKEND_STATIC
+extern "C" runtime::BackendConstructor* get_backend_constructor_pointer()
+{
+    return runtime::cpu::get_backend_constructor_pointer();
+}
+#endif
+
+void runtime::cpu::static_initialize()
```
nmostafa: Comment here about what this does and where it is called from ? How it is related to building static lib ?

samolisov:
The PR is very interesting since now getting the CPU backend as a static library looks like the only way to make the codegen (more pedantic, the generated code) works fine on Windows.

Unfortunately, I have the following problem: when the CPU backend is built as a static library and being linked with an .exe file, no static global variables are initialized; therefore, some builders from `runtime/cpu/builder` aren't registered. For the constant folding pass, for example, the following builder map is coming to the `ConstantFolding` constructor:

```
key: class ngraph::op::Negative, handler: 000000DBB70F9350
key: class ngraph::op::Add, handler: 000000DBB70F9350
key: class ngraph::op::Multiply, handler: 000000DBB70F9350
key: class ngraph::op::Subtract, handler: 000000DBB70F9350
key: class ngraph::op::Relu, handler: 000000DBB70F9350
key: class ngraph::op::Divide, handler: 000000DBB70F9350
key: class ngraph::op::Minimum, handler: 000000DBB70F9350
key: class ngraph::op::Maximum, handler: 000000DBB70F9350
key: class ngraph::op::Abs, handler: 000000DBB70F9350
key: class ngraph::op::Sqrt, handler: 000000DBB70F9350
```

When the backend is built as a DLL, the map contains tree additional members:

```
key: class ngraph::op::Broadcast, handler: 00000026158F8EF0
key: class ngraph::op::Pad, handler: 00000026158F8EF0
key: class ngraph::op::Reshape, handler: 00000026158F8EF0
```

This is because the linker just ignores object files that contains no used symbols at all (or maybe it is about the global variables only...). If I put a dummy function into the `runtime\cpu\builder\reshape.cpp` file and call this function from some place (from `cpu_external_function.cpp`), the builders from the object file are registered.

pruthviIntel:
@samolisov I am debugging the exact same issue...We have map to register all the Ops and their corresponding builder function .... The ops in the `..../runtime/cpu/builder/` are not registered in this Map.

samolisov:
@pruthviIntel I believe it could help: https://github.com/samolisov/ngraph/tree/static-lib-on-windows (please have a look at `cpu_builder_registry.cpp` (https://github.com/samolisov/ngraph/blob/static-lib-on-windows/src/ngraph/runtime/cpu/cpu_builder_registry.cpp) that just calls a dummy/empty function from some buider's translation units to make them be linked. The function `register_cf_builders()` is used in `cpu_external_function.cpp\register_common_passes`:  https://github.com/samolisov/ngraph/blob/static-lib-on-windows/src/ngraph/runtime/cpu/cpu_external_function.cpp#L1279

pruthviIntel:
@samolisov thanks for the pointers :) , I was trying to modify the Cmake 
```
--Wl,--whole-archive
  $<TARGET_PROPERTY:cpu_backend,INTERFACE_LINK_LIBRARIES>
  -Wl,--no-whole-archive
```
to include the whole archive , for some reason I am not able to get it working. But I guess for now your suggestion works. I modify to include all the builders ..now I can see the test passing on my local system

samolisov:
Just for info, when the cpu runtime is built as a static library, codegen runs fine even on Windows. Unfortunately, by default the JITed code is not able to load any symbols from the main module (.exe file) and all cpu specific kernels must be exported using the '/export' directive for the linker or (it looks better) be linked into a shared library (.dll).

pruthviIntel:
@avijit-nervana @wenzhe-nrv can u review this?

src/ngraph/runtime/backend.cpp
```diff
@@ -25,6 +31,28 @@
 using namespace std;
 using namespace ngraph;
 
+std::string runtime::Backend::s_backend_shared_library_search_directory;
+
+// This finds the full path of the containing shared library
+static string find_my_file()
```
avijit-nervana: May be the function should be renamed as: `find_my_pathname`?

src/ngraph/runtime/backend.cpp
```diff
@@ -25,6 +31,28 @@
...
+{
+#ifdef _WIN32
+    HMODULE hModule = GetModuleHandleW(L"ngraph.dll");
```
avijit-nervana: The Windows implementation looks for ngraph.dll whereas the Linux/Darwin doesn't. Is that expected?

src/ngraph/runtime/backend.hpp
```diff
@@ -142,6 +142,18 @@ class ngraph::runtime::Backend
...
+    /// \brief Allows sending backend specific configuration. The map contains key, value pairs
+    ///     specific to a particluar backend. The definition of these key, value pairs is
+    ///     defined by each backend.
+    /// \param config The configuration map sent to the backend
+    /// \param error An error string describing any error encountered
+    /// \returns true if the configuration is supported, false otherwise. On false the error
+    ///     parameter value is valid.
+    virtual bool set_config(const std::map<std::string, std::string>& config, std::string& error);
+
+    static void set_backend_shared_library_search_directory(const std::string& path);
```
avijit-nervana: Please provide comments. Also, shouldn't this be cumulative i.e., a list of paths as opposed to just one path. That would be more flexible (i.e., various backends could come from various directories - something we already run into during testing).

src/ngraph/runtime/backend.cpp
```diff
@@ -86,16 +114,30 @@ void runtime::Backend::remove_compiled_function(std::shared_ptr<Executable> exec
...
     // override this method for each supported backend to determine if the passed pointer is in
     // device pinned memory or not
     return false;
 }
 
-std::shared_ptr<runtime::Executable> runtime::Backend::load(istream& input_stream)
+void runtime::Backend::set_backend_shared_library_search_directory(const string& path)
 {
-    throw runtime_error("load opertion unimplemented.");
+    s_backend_shared_library_search_directory = path;
```
avijit-nervana: Please make this thread safe.

src/ngraph/runtime/cpu/builder/add.cpp
```diff
@@ -79,6 +79,10 @@ namespace ngraph
             }
 
             REGISTER_OP_BUILDER(Add);
+
+#ifdef NGRAPH_CPU_STATIC_LIB_ENABLE
```
avijit-nervana: This seems to be very fragile i.e., the same #ifdef repeated for every op. Instead consider moving this inside the REGISTER_OP_BUILDER macro or add some other macro?

src/ngraph/runtime/interpreter/CMakeLists.txt
```diff
@@ -14,15 +14,23 @@
...
+    add_library(interpreter_backend ${LIBRARY_TYPE} int_backend.cpp node_wrapper.cpp int_executable.cpp)
     if(NGRAPH_LIB_VERSIONING_ENABLE)
         set_target_properties(interpreter_backend PROPERTIES
             VERSION ${NGRAPH_VERSION}
             SOVERSION ${NGRAPH_API_VERSION})
     endif()
     target_link_libraries(interpreter_backend PUBLIC ngraph)
-
+    if(NGRAPH_INTERPRETER_STATIC_LIB_ENABLE)
+        target_compile_definitions(interpreter_backend PUBLIC NGRAPH_INTERPRETER_STATIC_LIB_ENABLE)
```
sindhu-nervana: Shouldn't this be `PRIVATE` or `${LIBRARY_TYPE}`

src/ngraph/runtime/backend_manager.cpp
```diff
@@ -68,6 +70,10 @@ vector<string> runtime::BackendManager::get_registered_backends()
 
 shared_ptr<runtime::Backend> runtime::BackendManager::create_backend(const std::string& config)
 {
+#ifdef NGRAPH_INTERPRETER_STATIC_LIB_ENABLE
```
nmostafa: @pruthviIntel , we need to register CPU backend as well if `NGRAPH_CPU_STATIC_LIB_ENABLE` is set. 
@rkimballn1 , do we still need to execute the logic below if the backend is already statically linked ?

pruthviIntel: https://github.com/NervanaSystems/ngraph/blob/6e019518b22633093e7b0577a2f503d35375b056/src/ngraph/runtime/cpu/cpu_backend.cpp#L63 -#L73 we are using static object to do that for both STATIC and SHARED cpu_backend

src/ngraph/runtime/cpu/builder/add.cpp
pruthviIntel: @avijit-nervana this is temporary workaround, i will make a follow up PR where we can build the backend with `--Wl,--whole-archive
  $<TARGET_PROPERTY:cpu_backend,INTERFACE_LINK_LIBRARIES>
  -Wl,--no-whole-archive` option. i tried this but have some issue with static initialization of data structure. I will debug and provide a fix for this in the follow up PR's

diyessi: Could just always register, static or dynamic. Future PR experiments.

nmostafa: Will fix later on. But yes, I want to remove global variable initializations.

PR #3259, opened 2019-07-18
Better error checking for Argmin/Argmax

diyessi:
Argmin and Argmax are not defined if the reduction axis has length 0.


src/ngraph/op/util/index_reduction.hpp
```diff
@@ -47,6 +47,8 @@ namespace ngraph
                                size_t axis,
                                const element::Type& index_element_type);
 
+                // Additional per-op validation
+                virtual void validate_reduction() const {}
```
aprocter: I'm not sure this needs to be virtual. The `IndexReduction` class is already baking in a lot of assumptions that seem to be tailor-made for `ArgMin` and `ArgMax`. Are we ever going to extend this class with other ops besides those?

src/ngraph/op/argmin.cpp
```diff
@@ -27,6 +27,23 @@ op::ArgMin::ArgMin(const Output<Node>& arg, size_t axis, const element::Type& in
     constructor_validate_and_infer_types();
 }
 
+void op::ArgMin::validate_reduction() const
+{
+    PartialShape input_shape = get_input_partial_shape(0);
+    Rank rank = input_shape.rank();
+    if (!rank.is_dynamic())
```
aprocter:
```suggestion
    if (rank.is_static())
```

aprocter: (Equivalent, just a style thing.)

test/type_prop/index_reduction.cpp
```diff
@@ -59,6 +59,50 @@ TEST(type_prop, index_reduction_invalid_rank)
...
 }
 
+TEST(type_prop, argmin_invalid_zero_reduction_axis)
+{
+    auto a = make_shared<op::Parameter>(element::f32, Shape{2, 0});
+
+    try
+    {
+        auto argmin = make_shared<op::ArgMin>(a, 1, element::i32);
+        FAIL() << "ArgMin c-tor should throw for zero-legnth reduction axis";
```
aprocter:
```suggestion
        FAIL() << "ArgMin c-tor should throw for zero-length reduction axis";
```

test/type_prop/index_reduction.cpp
```diff
@@ -59,6 +59,50 @@ TEST(type_prop, index_reduction_invalid_rank)
...
+    catch (...)
+    {
+        FAIL() << "Deduced type check failed for unexpected reason";
+    }
+}
+
+TEST(type_prop, sum_valid_zero_reduction_axis)
+{
+    auto a = make_shared<op::Parameter>(element::f32, Shape{2, 0});
+    auto sum = make_shared<op::Sum>(a, AxisSet{1});
```
aprocter: Might check the output shape here to make sure it is correct. Also, this doesn't really belong in `index_reduction.cpp`, because `Sum` is not an `IndexReduction` op.

test/type_prop/index_reduction.cpp
```diff
@@ -59,6 +59,50 @@ TEST(type_prop, index_reduction_invalid_rank)
...
+}
+
+TEST(type_prop, argmax_invalid_zero_reduction_axis)
+{
+    auto a = make_shared<op::Parameter>(element::f32, Shape{2, 0});
+
+    try
+    {
+        auto argmax = make_shared<op::ArgMax>(a, 1, element::i32);
+        FAIL() << "ArgMax c-tor should throw for zero-legnth reduction axis";
```
aprocter:
```suggestion
        FAIL() << "ArgMax c-tor should throw for zero-length reduction axis";
```

PR #3260, opened 2019-07-18
Fix perf regression in some models

gaurides:
In TF 1.14, for the sequence of Relu and Maxpool is switched. It shows as ConvolutionBias -> MaxPool -> Relu. This prevents ConvolutionBias + Relu fusion, and affects performance. This PR switches the 2 ops so that Relu can be fused.


src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -520,6 +520,68 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_bprop()
     this->add_matcher(m, callback);
 }
 
+static bool switch_nodes(std::shared_ptr<ngraph::Node> node1,
+                         std::shared_ptr<ngraph::Node> node2,
+                         size_t source_input_index = 0,
+                         size_t target_input_index = 0)
```
ayzhuang: target_input_index parameter is not needed.

src/ngraph/runtime/cpu/pass/cpu_fusion.cpp
```diff
@@ -520,6 +520,68 @@ void ngraph::runtime::cpu::pass::CPUFusion::construct_conv_bias_bprop()
...
+    auto arg = node1->get_argument(source_input_index);
+    node2->input(0).replace_source_output(arg);
+
+    node1->input(0).replace_source_output(node2->output(0));
+
+    // used implementation ref from replace_node
+    for (auto& input : target_inputs)
+    {
+        input.replace_source_output(node1->output(0));
+    }
```
ayzhuang: In replace_node, we have "replacement->add_node_control_dependents(target); target->clear_control_dependents();". Here we also need to take care of control dependency.

PR #3261, opened 2019-07-18
[MLIR] Quantize, Dequantize & QuantizedConvolution op defs

nbpatel:


src/contrib/mlir/dialect/ops.td
```diff
@@ -260,5 +260,46 @@ def NGAnyRedOp : NG_Axis_Reduction_Op<"any.red">
 // Terminator Ops
 def NGReturnOp : NG_Terminator_Op<"return">;
 
+// Quantized Ops
+def NGQuantizeOp :
+      NG_OneResult_Op<"quantize", [NoSideEffect]>,
+      Arguments<(ins NG_TensorType:$input, NG_TensorType:$scale, NG_TensorType:$zeroPoint,
+                 I64Attr:$outputType, I64ArrayAttr:$axes, I64Attr:$roundMode)>
```
nmostafa: `outputType` should be `TypeAttr` instead. We should also add Quantized Types as well.

nmostafa: Same for Dequantize.

PR #3263, opened 2019-07-19
[MLIR] Enable Gather Op

nmostafa:
Two changes here:
1. Refactor compiler.cpp to avoid code duplication. Added a helper `create_generic_ops` that will just map an nGraph node to an ng dialect instruction along with inputs/outputs. op specific attributes can be added after op creation. 
2. Enable GatherOp end-to-end. All CPU.gather_* pass. 
Sample code-gen for `CPU.gather_1d_indices_axis_2_4d_input`
```
`*** nGraph Dialect Dump: ***
func @main(%arg0: !ng.tensor<2x!ng.i32>, %arg1: !ng.tensor<2x2x3x3xf32>) -> !ng.tensor<2x2x2x3xf32> {
  %0 = "ng.gather"(%arg1, %arg0) {axis = 2 : i64} : (!ng.tensor<2x2x3x3xf32>, !ng.tensor<2x!ng.i32>) -> !ng.tensor<2x2x2x3xf32>
  "ng.return"(%0) : (!ng.tensor<2x2x2x3xf32>) -> ()
}
*** Affine Dialect Dump: ***
#map0 = () -> (0)
#map1 = () -> (2)
#map2 = () -> (3)


func @main(%arg0: memref<2xi32> {llvm.noalias = true}, %arg1: memref<2x2x3x3xf32> {llvm.noalias = true}, %arg2: memref<2x2x2x3xf32> {llvm.noalias = true}, %arg3: index) {
  affine.for %i0 = 0 to 2 {
    affine.for %i1 = 0 to 2 {
      affine.for %i2 = 0 to 3 {
        affine.for %i3 = 0 to 2 {
          %0 = load %arg0[%i3] : memref<2xi32>
          %1 = index_cast %0 : i32 to index
          %2 = load %arg1[%i0, %i1, %1, %i2] : memref<2x2x3x3xf32>
          store %2, %arg2[%i0, %i1, %i3, %i2] : memref<2x2x2x3xf32>
        }
      }
    }
  }
  return
}
```


src/contrib/mlir/compiler.cpp
```diff
@@ -301,131 +303,135 @@ namespace ngraph
             template <>
             mlir::Value* MLIRCompiler::COMPILE_OP_DECL(ngraph::op::Add)
             {
-                return compiler.create_binary_op<mlir::NGAddOp>(ng_node);
+                return compiler.create_generic_op<mlir::NGAddOp>(ng_node);
```
aprocter: I like this refactor. Is there anything other than `create_generic_op` now? If not, maybe a name like just `create_op` would be better.

src/contrib/mlir/compiler.cpp
```diff
@@ -301,131 +303,135 @@ namespace ngraph
...
+                mlir::Operation* op = result->getDefiningOp();
+                mlir::ArrayAttr red_axes_attr =
+                    compiler.m_builder->getI64ArrayAttr({(int64_t)idx_red->get_reduction_axis()});
+                op->setAttr("axes", red_axes_attr);
+                return result;
             }
 
             template <>
-            mlir::Value* MLIRCompiler::COMPILE_OP_DECL(ngraph::op::Dot)
+            mlir::Value* MLIRCompiler::COMPILE_OP_DECL(ngraph::op::ArgMin)
```
aprocter: There is still some shared boilerplate with `ArgMax` here to retrieve the reduction_axis attribute. Maybe it could be wrapped up in a function?

src/contrib/mlir/compiler.hpp
```diff
@@ -98,13 +98,18 @@ namespace ngraph
...
                                              "' in MLIR Compiler");
                 }
 
+                // Generic op lowerer to ng dialect.
+                // Simply maps ngraph tensors to values and generate an OP. No op-specific logic.
+                template <typename OP>
+                mlir::Value* create_generic_op(const ngraph::Node* ng_node);
+
                 template <typename UnaryOp>
                 mlir::Value* create_unary_op(const ngraph::Node* ng_node);
```
aprocter: I think we should remove this prototype now, along with that for `create_binary_op` etc.

src/contrib/mlir/lowerer.cpp
```diff
@@ -647,14 +647,131 @@ namespace
...
+        auto gatherOp = cast<NGGatherOp>(op);
+        auto loc = gatherOp.getLoc();
+        ScopedContext scope(rewriter, loc);
+
+        // Get operands
+        Value* result = m_pass.buildOutputDefs(op, rewriter)[0];
+        NGRAPH_CHECK(result, "Unexpected null result in GatherOp");
+        auto resultTy = result->getType().cast<MemRefType>();
+
+        Value* params = operands[0];
```
aprocter: Not sure I like the name `params`, but no big deal.

src/contrib/mlir/lowerer.cpp
```diff
@@ -647,14 +647,131 @@ namespace
...
+            paramsLbs.push_back(IndexHandle(vParams.lb(i)));
+            paramsUbs.push_back(IndexHandle(vParams.ub(i)));
+            paramsSteps.push_back(vParams.step(i));
+        }
+        NGRAPH_CHECK(paramsLbs.size() == vParams.rank() - 1 &&
+                         paramsUbs.size() == paramsLbs.size() &&
+                         paramsSteps.size() == paramsLbs.size(),
+                     "Incorrect loop nest bounds size for gather params");
+
+        paramsIVs = IndexHandle::makeIndexHandles(vParams.rank() - 1);
```
aprocter: Is this step necessary? I thought LoopBuilder would use the pointers to initialize these itself.

aprocter: (As opposed to just resizing the vector, I mean.)

src/contrib/mlir/compiler.cpp
```diff
@@ -301,131 +303,136 @@ namespace ngraph
...
             {
-                return compiler.create_binary_op<mlir::NGDotOp>(ng_node);
+                auto* idx_red = static_cast<const ngraph::op::util::IndexReduction*>(ng_node);
+                mlir::Value* result = compiler.create_generic_op<mlir::NGArgMinRedOp>(ng_node);
+                mlir::Operation* op = result->getDefiningOp();
+                mlir::ArrayAttr red_axes_attr = compiler.m_builder->getI64ArrayAttr({(int64_t)idx_red->get_reduction_axis()});
+                op->setAttr("axes", red_axes_attr);
+                return result;
```
dcaballe: I would suggest keeping `create_index_reduction` and make it call `create_generic_op` to avoid 360-365 and 371-376.

src/contrib/mlir/compiler.cpp
```diff
@@ -301,131 +303,136 @@ namespace ngraph
...
-            arg_values,
-            m_builder->getI64IntegerAttr(ng_node_concat->get_concatenation_axis()))
-        .getResult();
+        ->create<Op,
+                 ArrayRef<mlir::Type>,
+                 ArrayRef<mlir::Value *>,
+                 ArrayRef<mlir::NamedAttribute>>(
+                    mlir::UnknownLoc::get(&m_context),
+                    res_types,
+                    arg_values, {/* no attrs */}).getResult();
```
dcaballe: I think `create_generic_op` should return an op so that we can use it for ops returning multiple values. And the in client, we can get the op's single return or multiple returns.

src/contrib/mlir/lowerer.cpp
nmostafa: We are not initializing anything here, it just returns the right size vector.

src/contrib/mlir/compiler.cpp
nmostafa: I also changed `create_op` to return Operation instead and handle multi-value def at the dispatcher.

src/contrib/mlir/lowerer.cpp
```diff
@@ -771,7 +770,7 @@ namespace
     }
 
 #undef REWRITER
-/// End of pattern matchers
+    /// End of pattern matchers
     template <typename OP>
```
dcaballe: We probably need a white line between comment and template so that clang format keeps the comment wo indentation

PR #3269, opened 2019-07-19
remove unique_ptr from allocator static allocation

pruthviIntel:
- fixes paddle paddle test failures


src/ngraph/runtime/allocator.cpp
```diff
@@ -51,6 +51,6 @@ class ngraph::runtime::DefaultAllocator : public ngraph::runtime::Allocator
 
 ngraph::runtime::Allocator* ngraph::runtime::get_default_allocator()
 {
-    static std::unique_ptr<DefaultAllocator> allocator(new DefaultAllocator());
-    return allocator.get();
+    static DefaultAllocator* allocator = new DefaultAllocator();
```
s-barannikov: It is probably a wrong solution. It used to be a singleton (though it is usually implemented without `new`s) and now it looks like a memory leak. Every time this method is called, a new allocator is created and never destroyed.

Sorry for intervention. Please let me know if it is a member only section.

aprocter: Hi @s-barannikov. Definitely not a members-only section. :)

I think this should be fine, because of the `static` qualifier.

s-barannikov: @aprocter 
Hi,
> I think this should be fine, because of the static qualifier.

Yes, but who is gonna call the `delete`? Anyway, the class is only 8 bytes long, so... I can't just imagine how this could solve the problem :)

aprocter: Ah, sure, I misunderstood--thought you were saying that we'd leak a new instance with every call. @pruthviIntel would this not work instead? Not sure I have complete context on what we are trying to do here.

```
static DefaultAllocator allocator;
return &allocator;
```

s-barannikov: @aprocter 

> thought you were saying that we'd leak a new instance with every call

It is what I wrote initially. Then I realized that static is still static and edited the comment in hope no one will read email :) Sorry for misleading

rkimballn1: I would approve of the code that @aprocter suggested above.

pruthviIntel:
@aprocter @diyessi @rkimballn1 can u review it ?

aprocter:
@pruthviIntel For the record, my review wasn't needed for merge here (unless you wanted me to take a look), because I'm not a `CODEOWNER` and nobody actually requested my review. My icon just shows up in the "reviewers" box because I commented on the PR. GitHub's UI is weird in this way.

aprocter:
@baojun-nervana Do we know if the updated version of this PR still fixes the issue?

baojun-nervana:
> @baojun-nervana Do we know if the updated version of this PR still fixes the issue?

The last merge master failed CI. I just tried the 2nd to the last merge master and it still fix paddle.

pruthviIntel:
@aprocter thanks Adam for the suggestions. :)

diyessi:
@rkimballn1 hit an instance of this problem that is not fixed by the change in this PR. He's working on his own PR.

rkimballn1:
I am able to reliably recreate this segfault when running a single NNP unit test. I have a couple of PRs #3303 and #3304 (one for master and one for r0.24) which fix this without resorting to relying on order of static destructors.
I tried the fix in this PR but it does not address the segfault I get. With the change in this PR I still get the segfault.
I recommend closing this PR and use the two PRs I have above.

baojun-nervana:
Tested both #3303 and #3304, it won't fix paddle segfault unfortunately.

diyessi:
The transformer is going to need to work out what is really going on. I'll file a JIRA. I think this blocks 0.24.

gcwenger:
"Intentional allocation without explicit freeing" is actually the recommended technique to avoid issues with order of destruction:
https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use

We’ve typically used a variant of this:
https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2
`static` object as opposed to `static` pointer.
But this unfortunately has more potential for _last use_ issues. I've been ok with it in our code base because it tends not to freak people out about perceived “leaking”.
But if we're hitting last use issues and order of destruction we should switch to judicious use of "intentional allocation without explicit freeing" (note that heap teardown ensures we're not actually leaking). Note that we may want to look into whether we can tell leak detectors to ignore these.

aprocter:
@gcwenger So in other words @pruthviIntel was right to begin with? :)

pruthviIntel:
@aprocter I had same discussion before with @jbobba , he pointed out the same w.r.to heap tear down ..if their is a memory leak concern. So I went with it originally

pruthviIntel:
@gcwenger Thanks for pointing to references :)

src/ngraph/runtime/allocator.cpp
aprocter: (Comment for posterity: see the top-level PR thread. This is probably the right thing here.)

diyessi: Let's switch this to 0.24 as a hot fix and work out the ownership policy.

PR #3271, opened 2019-07-20
Convert G* operations to new op form

s-barannikov:
Also includes some quick-fixes for previously converted ops.

### DIFF
--- a/src/ngraph/op/add.cpp
+++ b/src/ngraph/op/add.cpp
@@ -49,7 +49,7 @@ void op::Add::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& 
     adjoints.add_delta(y, delta);
 }
 
-shared_ptr<Node> ngraph::operator+(const shared_ptr<Node>& arg0, const shared_ptr<Node>& arg1)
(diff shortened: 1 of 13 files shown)


aprocter:
@s-barannikov BTW, our internal CI system is not currently set up to handle pull requests from forks, so we will have to do some manual work to merge it when folks are back in the office on Monday.

s-barannikov:
@aprocter
No problem 😊 Should I create a branch in the upstream next time or continue with the fork?

aprocter:
@s-barannikov Nope, this is fine---this is our standard procedure for external contribs, at least for now.

src/ngraph/op/add.cpp
```diff
@@ -49,7 +49,7 @@ void op::Add::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector&
     adjoints.add_delta(y, delta);
 }
```
diyessi: I ran into problems the last time I tried this change; we might drop the operator+ part if they show up in merge CI.

diyessi:
Migrated to #3278

PR #3272, opened 2019-07-20
Make BatchNorm* constructors accept inputs by const reference

s-barannikov:
Other operations which have already been converted to new op form accept
`Output<Node>` by const reference. `BatchNorm*` ops were an exception.


diyessi:
Migrated to #3278

PR #3274, opened 2019-07-20
Make is_commutative method const and override it for some ops

s-barannikov:
* Qualify `is_commutative` method with `const`
* Override it for `Equal`, `Minumum`, `NotEqual` ops
* Ensure that all overrides of the method are `public`


diyessi:
Migrated to #3278

PR #3276, opened 2019-07-22
Python capsules support for ngraph::Function

tomdol:
This PR adds 2 static methods to `ngraph::Function` bindings. 

One method accepts a `PyCapsule` with a `shared_ptr<ngraph::Function>` in it, extracts it and returns the shared pointer to python so that the user can operate on a function object using its python API.

The second method takes a `shared_ptr<ngraph::Function>` from python, puts it in a pybind11 `PyCapsule` wrapper and returns it so that it can be passed (via python) to another library with python bindings.


python/pyngraph/function.cpp
```diff
@@ -49,4 +51,24 @@ void regclass_pyngraph_Function(py::module m)
             py::cast(self.get_output_shape(0)).attr("__str__")().cast<std::string>();
         return "<" + class_name + ": '" + self.get_friendly_name() + "' (" + shape + ")>";
     });
+    function.def_static("from_capsule", [](py::object* capsule) {
+        auto* pybind_capsule_ptr = capsule->ptr();
+        auto* capsule_ptr = PyCapsule_GetPointer(pybind_capsule_ptr, CAPSULE_NAME);
```
postrational: Please add a one-line comment to explain what `PyCapsule_GetPointer` does and where it comes from.
Shouldn't we include `Python.h` to use this?

python/pyngraph/function.cpp
```diff
@@ -49,4 +51,24 @@ void regclass_pyngraph_Function(py::module m)
...
+        auto* ngraph_function = static_cast<std::shared_ptr<ngraph::Function>*>(capsule_ptr);
+
+        return *ngraph_function;
+    });
+    function.def_static("to_capsule", [](std::shared_ptr<ngraph::Function>& ngraph_function) {
+        auto* sp_copy = new std::shared_ptr<ngraph::Function>(ngraph_function);
+        auto pybind_capsule = py::capsule(sp_copy, CAPSULE_NAME, [](PyObject* capsule) {
```
postrational: Please add a few 1-line comments explaining the calls made in this code.

postrational: I think it would be more readable to assign a name to this lambda, so we know what it does - `cleanup_callback` as in the PyBind docs seems like a good name.

python/pyngraph/function.cpp
tomdol: I does come from `Python.h` but this header is already included by `pybind11.h`

PR #3289, opened 2019-07-23
[MLIR] Deallocate all temp tensors before return

nmostafa:
Fix memory leaks for temp tensors. Dealloc all temps before return from a sub-graph function.


nmostafa:
I don't like this so much either. If all allocs are inserted in the entry block (which I wanted to do, but was limited by MLIR lowering at the time), then it is always safe to dealloc at each return. It also possible to unify the exit block in a pass before lowering to guarantee single exist block.

dcaballe:
Thanks. Just wondering if the function beginning/end allocation/de-allocation would be a problem if the graph consumes too much memory. As future work, maybe we could have a more local allocation/deallocation for those case.

PR #3292, opened 2019-07-23
[MLIR] Fix naming convention in MLIR files

dcaballe:
Add naming convention note per file to state which files should use
nGraph naming convention and which MLIR naming convention and align
naming convention in those files with such a note.


src/contrib/mlir/dialect/type.hpp
```diff
@@ -198,18 +202,18 @@ namespace mlir
...
 
-        Shape getShape() const { return m_shape; }
-        EltType getElementType() const { return m_eltType; }
+        Shape getShape() const { return mShape; }
+        EltType getElementType() const { return mEltType; }
     private:
         NGTensorTypeStorage(EltType eltType, Shape shape)
-            : m_eltType(eltType)
-            , m_shape(shape)
+            : mEltType(eltType)
```
nmostafa: MLIR/LLVM style doesn't have `m` prefix for members.

PR #3293, opened 2019-07-24
Finish moving unit tests into test/backend

aprocter:
Mostly this is just moving files wholesale, but `dynamic.in.cpp` was due for a splittin' as well. This completes the move of all `in.cpp` files into `test/backend`, except for the ones in `onnx`, which I don't want to mess with right now.

The only thing left that I think we might want to split but haven't is `autodiff.in.cpp`.


test/backend/dynamic.in.cpp
```diff
@@ -0,0 +1,294 @@
...
+#include "util/all_close_f.hpp"
+#include "util/test_control.hpp"
+#include "util/test_tools.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+static string s_manifest = "${MANIFEST}";
+
+NGRAPH_TEST(${BACKEND_NAME}, create_dynamic_backend)
```
aprocter: Note that I actually removed one test here, which checks that `supports_dynamic_tensors` returns _false_ when dynamic support has _not_ been requested. That test would only be valid as long as no backend directly implements dynamic support, which seems, well, pessimistic.

PR #3294, opened 2019-07-24
[Py] Added clamp operator to Python API

Ewa21:
Includes:
* New package for registration of fused ops in pyngraph.


python/ngraph/ops.py
```diff
@@ -555,6 +555,35 @@ def tanh(node, name=None):  # type: (Node, str) -> Node
...
+    .. code-block:: python
+
+        if data < min_value:
+            data=min_value
+        elif data > max_value:
+            data=max_value
+
+    :param data: Input tensor. One of: input node, array or scalar.
+    :param min_value: The lower bound of the <min_value;max_value> range. Scalar value.
+    :param max_value: The upper bound of the <min_value;max_value> range Scalar value.
```
tomdol: Missing dot after `range` :)

python/ngraph/ops.py
```diff
@@ -555,6 +555,35 @@ def tanh(node, name=None):  # type: (Node, str) -> Node
...
 
+@nameable_op
+def clamp(data, min_value, max_value, name=None):
+    # type: (NodeInput, ScalarData, ScalarData, str) -> NodeInput
+    """Perform clamp element-wise on data from input node.
+
+    Performs a clipping operation on an input value between a pair of boundary values.
+
+    If :code:`data` compares less than :code:`min_value`, sets :code:`min_value`;
+    else if :code:`max_value` compares less than :code:`data`, sets :code:`max_value`;
```
tsocha: `if :code:'max_value' compares less than :code:'data', sets :code:'max_value';`
Probably `less` isn't a good word here

python/pyngraph/ops/fused/clamp.cpp
```diff
@@ -17,14 +17,15 @@
 #include <pybind11/pybind11.h>
```
tsocha: It is weird that git is confused and threat `clamp.cpp` as new version of `elu.cpp`.
Is that possible that you removed `elu.cpp` by accident?

python/test/ngraph/test_ops_binary.py
```diff
@@ -1,5 +1,5 @@
 # ******************************************************************************
-# Copyright 2017-2019 Intel Corporation
+# Copyright 2018-2019 Intel Corporation
```
tsocha: Please revert back to 2017 here

python/pyngraph/ops/fused/clamp.cpp
tsocha: There is also a possibility that you forgot to add `elu.cpp` after moving this file to a new location.

tsocha: There is also a possibility that you forgot to add elu.cpp after moving this file to a new location.

python/pyngraph/ops/fused/clamp.cpp
```diff
@@ -17,14 +17,15 @@
...
 
-#include "ngraph/op/fused/elu.hpp"
-#include "pyngraph/ops/elu.hpp"
+#include "ngraph/op/fused/clamp.hpp"
+#include "pyngraph/ops/fused/clamp.hpp"
 
 namespace py = pybind11;
 
-void regclass_pyngraph_op_Elu(py::module m)
+void regclass_pyngraph_op_Clamp(py::module m)
```
tomdol: I'm not sure but it looks like the `Elu` bindings will be removed now. This file is required for `Clamp` but you should create a similar .cpp file for elu.

python/pyngraph/ops/fused/clamp.cpp
Ewa21: Accidentally, I converted the original file to `clamp.cpp` instead of its copy. However, the copy file is in right place and works well.

python/pyngraph/ops/fused/clamp.cpp
Ewa21: Accidentally, I converted the original file `elu.cpp` to `clamp.cpp` instead of its copy file. There is `elu.cpp` in the `fused` package.

python/ngraph/ops.py
```diff
@@ -555,6 +555,35 @@ def tanh(node, name=None):  # type: (Node, str) -> Node
     return Tanh(node)
 
 
+@nameable_op
+def clamp(data, min_value, max_value, name=None):
+    # type: (NodeInput, ScalarData, ScalarData, str) -> NodeInput
```
postrational: I believe this produces a `Node` not `NodeInput`

python/ngraph/ops.py
```diff
@@ -555,6 +555,35 @@ def tanh(node, name=None):  # type: (Node, str) -> Node
...
+    """Perform clamp element-wise on data from input node.
+
+    Performs a clipping operation on an input value between a pair of boundary values.
+
+    If :code:`data` compares less than :code:`min_value`, sets :code:`min_value`;
+    else if :code:`max_value` compares more than :code:`data`, sets :code:`max_value`;
+    otherwise remains unchanged :code:`data`.
+
+    Computes clamp:
```
postrational: `Computes clamp` -> `Clamp uses the following logic`:

python/ngraph/ops.py
```diff
@@ -555,6 +555,35 @@ def tanh(node, name=None):  # type: (Node, str) -> Node
...
 
 
+@nameable_op
+def clamp(data, min_value, max_value, name=None):
+    # type: (NodeInput, ScalarData, ScalarData, str) -> NodeInput
+    """Perform clamp element-wise on data from input node.
+
+    Performs a clipping operation on an input value between a pair of boundary values.
+
+    If :code:`data` compares less than :code:`min_value`, sets :code:`min_value`;
```
postrational: I would rewrite this paragraph as follows:
```
For each element in :code:`data`, if the element's value is lower than :code:`min_value`, it will be replaced with :code:`min_value`.  If the value is higher than :code:`max_value`, it will be replaced by :code:`max_value`.
Intermediate values of :code:`data` are returned without change.
```

PR #3296, opened 2019-07-24
Make validation a pass and add it after every pass by default

rkimballn1:
This is an alternative way to do validation between passes. If you don't like it we can close the PR.


src/ngraph/pass/manager.hpp
```diff
@@ -62,20 +61,30 @@ class ngraph::pass::Manager
             m_pass_names.push_back(typeid(T).name());
 #endif
         }
+        if (m_per_pass_validation && !std::is_same<T, Validate>::value)
```
aprocter: It's not a big deal, but I wonder if this is overkill.

src/ngraph/pass/validate.hpp
```diff
@@ -0,0 +1,35 @@
...
+    namespace pass
+    {
+        class Validate : public FunctionPass
+        {
+        public:
+            Validate()
+                : FunctionPass()
+            {
+            }
+            virtual bool run_on_function(std::shared_ptr<ngraph::Function> f) override;
```
earhart: `virtual override`?  (Why not just `final`?)

src/ngraph/pass/manager.hpp
rkimballn1: I was wondering the same thing too. I doubt that users will use the "auto insert validation" mode along with adding Validate passes into the graph. I will remove the `is_same`. Double validation never killed anyone.

src/ngraph/pass/validate.hpp
rkimballn1: Ugh. That's what I get for copy/paste. I will remove virtual.

src/ngraph/pass/manager.cpp
```diff
@@ -53,11 +53,7 @@ pass::Manager::~Manager()
...
 }
 
-void pass::Manager::initialize_default_passes()
-{
-}
-
-void pass::Manager::run_passes(shared_ptr<Function> func, bool transitive, bool revalidate)
+void pass::Manager::run_passes(shared_ptr<Function> func, bool transitive)
 {
     bool profile_enabled = getenv("NGRAPH_PROFILE_PASS_ENABLE") != nullptr;
```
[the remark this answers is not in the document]
rkimballn1: done

PR #3297, opened 2019-07-24
Add Output support for a number of builders

diyessi:
Preparatory work for starting to add Output<Node> support to fused ops.


src/ngraph/builder/autobroadcast.cpp
```diff
@@ -150,39 +150,38 @@ namespace ngraph
...
             }
 
-            return return_node;
+            return return_value;
         }
 
-        std::pair<std::shared_ptr<Node>, std::shared_ptr<Node>>
-            numpy_broadcast(const std::pair<std::shared_ptr<Node>, std::shared_ptr<Node>>& args)
+        std::pair<Output<Node>, Output<Node>>
+            numpy_broadcast_values(const std::pair<Output<Node>, Output<Node>>& args)
```
aprocter: Instead of adding these functions with the `_values` suffix, could you try keeping the existing ones and having them keep `Output<Node>` as arguments but return `shared_ptr<Node>` as results?

If that doesn't work: I'm a little worried that the `_values` suffix on the function names is hard to understand. Maybe `_from_outputs`?

postrational:
This is semantically a little confusing. We're now using the type `Output` for most inputs...

src/ngraph/op/util/rnn_cell_base.hpp
```diff
@@ -101,16 +99,15 @@ namespace ngraph
...
+                static std::shared_ptr<Node> mul(const Output<Node>& lhs, const Output<Node>& rhs);
                 ///
                 /// \brief      Creates node with element-wise clip operation with numpy broadcasting.
                 ///
                 /// \param[in]  data   The input tensor for clipping.
                 ///
                 /// \return     Node with element-wise clip operation.
                 ///
-                std::shared_ptr<Node> clip(const std::shared_ptr<Node>& data) const;
+                std::shared_ptr<Node> clip(const Output<Node>& data) const;
```
tsocha: `Output` type for input data is little confusing, isn't it?
Maybe it's worth adding an some kind of alias to it?

diyessi: An `Output` of an op provides a value for an `Input` of another op.

aprocter: @tsocha We've been chatting about this a little bit. I think the "Input/Output" nomenclature feels right when you are talking about the graph as a whole or doing rewrites post-construction, but I agree that when you are building the graph in an "applicative" style it sometimes reads a bit strangely. It might be a matter of getting people used to the new style, or maybe an alias is in order as you suggest.

Any thoughts on what a good alias/different name might be?

postrational: It feels like this represents an edge of the graph, or a connection between 2 nodes. 
Perhaps a name like `Connection` or `Link` could work?

aprocter: I don't think `Connection` or `Link` (or, say, `Edge`) is quite right, because the object in question really only corresponds to an output, and has no affinity with any particular target input. It's used in other contexts to say things like `n->output(i).get_partial_shape()` (where `n->output(i)` is an `Output<Node>`).

I wonder if `Value` as an alias for `Output<Node>` might make sense? Maybe something like `InputSource`?

tsocha: The easiest way is to use `Input` as an alias but is this name already used?
`InputSource` is seems good as well.

aprocter: `Input` is already taken, yeah. (It refers to an "input port" of a node, and the distinction is important at later phases when we are rewriting graphs.)

PR #3290, opened 2019-07-23
[MLIR] Enable affine dialect loop fusion

dcaballe:
Enable affine dialect loop fusion in nGraph pipeline. It also adds an
opt flag to enable/disable it when ngraph-opt is in place. Fusion seems
to work for simple cases. It wasn't able to fuse dot + add, though, at
least in my test case. One example that worked:

Input:
```
  %6 = alloc() : memref<2500x2500xf32>
  affine.for %i3 = 0 to 2500 {
    affine.for %i4 = 0 to 2500 {
      %7 = load %arg0[%i3, %i4] : memref<2500x2500xf32>
      %8 = load %0[%i3, %i4] : memref<2500x2500xf32>
      %9 = addf %8, %7 : f32
      store %9, %6[%i3, %i4] : memref<2500x2500xf32>
    }
  }
  %10 = alloc() : memref<2500x2500xf32>
  affine.for %i5 = 0 to 2500 {
    affine.for %i6 = 0 to 2500 {
      %11 = load %arg2[%i5, %i6] : memref<2500x2500xf32>
      %12 = load %0[%i5, %i6] : memref<2500x2500xf32>
      %13 = addf %12, %11 : f32
      store %13, %10[%i5, %i6] : memref<2500x2500xf32>
    }
  }
  %14 = alloc() : memref<2500x2500xf32>
  affine.for %i7 = 0 to 2500 {
    affine.for %i8 = 0 to 2500 {
      %15 = load %10[%i7, %i8] : memref<2500x2500xf32>
      %16 = load %6[%i7, %i8] : memref<2500x2500xf32>
      %17 = addf %16, %15 : f32
      store %17, %14[%i7, %i8] : memref<2500x2500xf32>
    }
  }
```
Output:
```
  %8 = alloc() : memref<2500x2500xf32>
  affine.for %i3 = 0 to 2500 {
    affine.for %i4 = 0 to 2500 {
      %9 = load %arg2[%i3, %i4] : memref<2500x2500xf32>
      %10 = load %2[%i3, %i4] : memref<2500x2500xf32>
      %11 = addf %10, %9 : f32
      %12 = affine.apply #map2(%i3, %i4, %i3, %i4)
      %13 = affine.apply #map3(%i3, %i4, %i3, %i4)
      store %11, %0[%12, %13] : memref<1x1xf32>
      %14 = load %arg0[%i3, %i4] : memref<2500x2500xf32>
      %15 = load %2[%i3, %i4] : memref<2500x2500xf32>
[...]

PR #3298, opened 2019-07-24
[MLIR] Re-compile sub-graph once on first invocation

nmostafa:
We currently compile lazily, but we did not cache the compiled module for next sub-graph invocation. This PR splits compilation from execution, and saves the compiler for next invocation. It assumes the sub-graph topology and ops has not changed, but its inputs tensors pointers might change. 

PR also removes disable (now invalid) CompiledKernel Halida tests.


aprocter:
Will try to get to this, but I'm a bit swamped right now, so please feel free to merge without my approval.

src/ngraph/runtime/cpu/builder/mlir_cpu_compiled_kernel.cpp
```diff
@@ -65,14 +65,26 @@ namespace ngraph
...
-                    // TODO: Decouple 'compile' and 'run' APIs. We want to be able to run the same
-                    // jitted code on different arguments.
-                    mlir_compiler.compile_and_run();
+                    MLIRCompiler& mlir_compiler = ctx->mlir_compilers.find(compiled_kernel)->second;
+                    if (!is_module_ready)
+                    {
+                        mlir_compiler.compile();
+                    }
+                    mlir_compiler.set_args(&ptr_args);
+                    mlir_compiler.run();
```
dcaballe: I think it would be better if we pass `ptr_args` as reference in `run`. We won't forget to set the args before running and we would avoid dealing with a nullptr most of the time since `m_external_args` will only be valid when we run. Does it make sense?

src/ngraph/runtime/cpu/cpu_runtime_context.hpp
```diff
@@ -66,6 +71,11 @@ namespace ngraph
                 State* const* states;
                 std::set<size_t> breakpoints;
                 size_t pc;
+#ifdef NGRAPH_MLIR_ENABLE
+                std::unordered_map<ngraph::op::CompiledKernel*,
```
dcaballe: description here would be nice

src/contrib/mlir/compiler.cpp
```diff
@@ -238,35 +239,70 @@ MLIRCompiler::TensorInfo MLIRCompiler::get_tensor_value(descriptor::Tensor* tens
...
+    // Lower Affine to Std Dialect
     {
-        NGRAPH_CHECK(false, "Incorrect module after dialect lowering");
+        mlir::PassManager pm;
+        // Lower affine ops
+        pm.addPass(mlir::createLowerAffinePass());
+        auto rr = pm.run(m_module.get());
+        NGRAPH_CHECK(succeeded(rr), "Affine loop lowering failed");
+
+        dump_mlir_module("Standard Dialect Dump:");
```
dcaballe: I'm not sure I understand why we need to change the structure of the pipeline, remove the `optimize` function and move everything into independent scopes within the same function. Is this needed to fix the recompilation issue?

dcaballe: I'm actually adding optimizations to the `optimize` function (https://github.com/NervanaSystems/ngraph/pull/3290/files#diff-47415dcfbd95a002c0c4486a98db979eR271 and more locally) so if there is no strong reason for this change wrt the re-compilation issue, I would prefer we restore the original version. We can revisit this later, if needed.

src/contrib/mlir/compiler.cpp
```diff
@@ -449,33 +476,39 @@ mlir::Operation* MLIRCompiler::create_index_reduction(const ngraph::Node* ng_nod
...
 
     mlir::Function* func = m_module->getNamedFunction("main");
     NGRAPH_CHECK(func && !func->getBlocks().empty(), "Function not found");
 
+    // Set external arguments
+    NGRAPH_CHECK(m_compiled_kernel, "No compiled kernel set for compiler");
+    NGRAPH_CHECK((m_compiled_kernel->get_arguments().size() +
+                  m_compiled_kernel->get_kernel_outputs().size()) == external_tensors.size(),
+                 "Number of arguments and outputs doesn't match number of tensors");
+    m_external_tensors = &external_tensors;
```
dcaballe: is `m_external_tensors` member needed anymore?

src/contrib/mlir/compiler.cpp
nmostafa: Aware of that. In your PR, you need to add `optimize_affine()` or something that will solely do the optimization (as opposed to lower + optimize)

src/contrib/mlir/compiler.cpp
nmostafa: `allocate_memrefs` still uses it. Dont want to pass it around.

src/contrib/mlir/compiler.cpp
```diff
@@ -238,9 +230,10 @@ MLIRCompiler::TensorInfo MLIRCompiler::get_tensor_value(descriptor::Tensor* tens
     return it->second;
 }
 
-// Lowers nGraph dialect to affine dialect.
+// Lowers nGraph dialect all the way to LLVM module.
```
dcaballe: Please, restore/update this comment to be aligned with the latest change.

PR #3300, opened 2019-07-24
Fix version number parsing when latest tag is not vxx.yy.zz

aprocter:
As of right now the newest tag is called `nbench_db`, which our CMake stuff can't parse when it tries to populate the version number info. This seems to work around it (it finds the most recent tag of the form `v*.*.*`).


silee2:
I don't see a tag called `nbench_db` in the Repo. All tags are v*.*.* Is CI auto creating a local tag?

aprocter:
@silee2 

Screen Shot 2019-07-24 at 4 30 27 PM

PR #3303, opened 2019-07-25
Fix for Allocator lifetime in CPU backend

rkimballn1:
*This PR is for r0.24 branch*

The static DefaultAllocator is being freed before it is done being used. I made the DefaultAllocator a member of CPU_Backend and it uses a pointer to this member to pass around. This way the DefaultAllocator's lifetime is the lifetime of the CPU_Backend and it fixes the segfault in the NNP backend when we use the CPU backend.
This is blocking the NNP backend from using ngraph 0.24


silee2:
@baojun-nervana Needs to confirm that the fix works for PDPD as well

rkimballn1:
This does not fix all problems with this

PR #3306, opened 2019-07-25
Some GetOutputElement changes to help with Output<Node>

diyessi:


src/ngraph/node.hpp
```diff
@@ -554,6 +556,13 @@ namespace ngraph
         ///
         /// TODO: Make a plan to deprecate this.
         std::shared_ptr<NodeType> get_node_shared_ptr() const { return m_node; }
+        /// \return A useable shared pointer to this output. If index 0, the node,
+        /// otherwise find or create a GOE.
+        std::shared_ptr<Node> as_node_shared_ptr() const NGRAPH_DEPRECATED("Transitional.")
```
aprocter: How about `as_single_output_node`?

src/ngraph/node.hpp
```diff
@@ -279,6 +279,8 @@ namespace ngraph
         /// Returns the partial shape for output i
         const PartialShape& get_output_partial_shape(size_t i) const;
 
+        std::shared_ptr<Node> get_output_as_shared_ptr(size_t i);
```
aprocter: How about `get_output_as_single_output_node`? (The long name is a hairshirt ;))

src/ngraph/node.cpp
```diff
@@ -86,6 +86,18 @@ std::shared_ptr<Node> Node::copy_with_new_inputs(const OutputVector& inputs) con
...
+std::shared_ptr<Node> Node::get_output_as_shared_ptr(size_t i)
+{
+    for (auto in : output(i).get_target_inputs())
+    {
+        if (in.get_node()->description() == op::GetOutputElement::type_name)
+        {
+            return in.get_node()->shared_from_this();
+        }
+    }
+    return get_output_element(output(i), true);
```
aprocter: Does `get_output_element` avoid returning an actualy GOE if the source node has only one output? (That's what I'd prefer to see here.)

src/ngraph/node.hpp
aprocter: BTW, I'm not sure we need to deprecate this necessarily. Even once GOE is gone, it still could make sense to have something that dangles out an `Identity` node hooked up to the _i_th input.

src/ngraph/node.cpp
diyessi: Yes. Things break if it doesn't.

PR #3313, opened 2019-07-26
Reshape sinking: fix issue with handling rank changing reshape.

silee2:


src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -247,32 +278,37 @@ static void materialize_shapes(shared_ptr<Node> n,
...
 }
 
 static void sink_reshape(shared_ptr<op::Reshape> reshape,
                          ReshapeMap& reorders,
                          set<shared_ptr<Node>>& reshapes_to_delete)
 {
+    NGRAPH_DEBUG << "Sinking Reshape :" << describe_reshape(reshape);
     auto orig_reshape = reorders.at(reshape->get_argument(0));
-    if (!reshape->get_is_transpose())
+    // 1) Not a Transpose or 2) Rank changing operation.
```
silee2: This PR has other other line changes but they are just there for helping debug printing. This IF condition change is the only functional change.

PR #3315, opened 2019-07-26
Some more changes of nodes to outputs

diyessi:


src/ngraph/op/util/broadcasting.cpp
```diff
@@ -153,16 +166,35 @@ static std::shared_ptr<ngraph::Node>
     }
 
     // Remove axes which have length of 1 from source shape
-    auto broadcasted_node = std::make_shared<ngraph::op::Reshape>(
-        node, ngraph::get_default_order(node->get_shape()), squeezed_shape);
+    ngraph::Output<ngraph::Node> broadcasted_value = std::make_shared<ngraph::op::Reshape>(
```
aprocter: Out of scope for this PR, but why are we not `using namespace ngraph` in this file?

src/ngraph/op/util/broadcasting.cpp
```diff
@@ -104,6 +104,19 @@ static std::pair<ngraph::Shape, std::vector<ngraph::Shape>>
     return get_numpy_broadcast_shapes(input_shapes);
 }
 
+static std::pair<ngraph::Shape, std::vector<ngraph::Shape>>
```
aprocter: The duplication in this file is a bit of a drag, but I don't have any bright ideas. (Nor even any dim ones.)

src/ngraph/op/util/broadcasting.cpp
diyessi: I wondered the same. It's not alone.

src/ngraph/op/util/broadcasting.cpp
diyessi: The vector ones aren't as friendly.

aprocter: FWIW, if `OutputVector` were its own class (is it not already?), we _could_ have an `as_single_output_nodes` akin to what we have for `Output`.

diyessi: It's not.

aprocter: I think this idea is probably a good one, but maybe in a later PR.

PR #3322, opened 2019-07-27
ConstantFolding for Equal, Greater, GreaterEq, Less, LessEq, NotEqual

aprocter:
Not feeling great about test coverage for this one, so putting [WIP] on it.


src/ngraph/pass/constant_folding.cpp
```diff
@@ -365,78 +381,227 @@ void pass::ConstantFolding::construct_constant_broadcast()
...
-            runtime::reference::subtract<T>(
-                a->get_data_ptr<T>(), b->get_data_ptr<T>(), out_vec.data(), shape_size(out_shape));
+            NGRAPH_CHECK(element::from<Tin>() == element::from<Tout>(),
+                         "Input/output types do not match");
+            vector<Tin> out_vec(shape_size(out_shape));
+            runtime::reference::add<Tin>(a->get_data_ptr<Tin>(),
+                                         b->get_data_ptr<Tin>(),
+                                         out_vec.data(),
+                                         shape_size(out_shape));
+            return make_shared<op::Constant>(binary->get_element_type(), out_shape, out_vec);
```
nmostafa: Not a strong request, but there is a lot of code duplication here. Can we factor out setting the out_vec and initializing the Constant Op ? Probably better to have a type_id->reference_kernel map also.

src/ngraph/pass/constant_folding.cpp
```diff
@@ -365,78 +381,227 @@ void pass::ConstantFolding::construct_constant_broadcast()
...
+    NGRAPH_UNREACHABLE("Unreachable switch case");
+}
 bool is_supported_binary_op(std::shared_ptr<Node> n)
 {
-    return (std::dynamic_pointer_cast<op::Add>(n) || std::dynamic_pointer_cast<op::Subtract>(n) ||
-            std::dynamic_pointer_cast<op::Multiply>(n) ||
-            std::dynamic_pointer_cast<op::Divide>(n) || std::dynamic_pointer_cast<op::Maximum>(n) ||
-            std::dynamic_pointer_cast<op::Minimum>(n));
+    return (
+        std::dynamic_pointer_cast<op::Add>(n) || std::dynamic_pointer_cast<op::And>(n) ||
```
nmostafa: Again, switching to table-based will save us all these checks.

src/ngraph/pass/constant_folding.cpp
aprocter: Agreed that the code duplication here is getting kind of hideous. Will see what I can factor out in a follow-on PR. I have some ideas for a broader refactor of the reference kernels, but would like to save that for an even later PR because it's a bit far-reaching.

PR #3331, opened 2019-07-30
[Py] Added gemm operator to Python API.

Ewa21:


python/ngraph/ops.py
```diff
@@ -520,6 +520,50 @@ def broadcast_to(node, new_shape, axis=None, name=None):
...
+                if :code:`transA` is :code:`True`, otherwise :code:`A` with shape (K, N).
+
+    :code:`B'`: The transpose of matrix :code:`B` with shape (K, N),
+                if :code:`transB` is :code:`True`, otherwise :code:`B` with shape (N, K).
+
+    :code:`C`: Matrix broadcastable to shape (M, N).
+
+    :code:`Y`: Matrix with shape (M, N).
+
+    For more information refer to:
```
postrational: GEMM, as the name suggests is a very general algorithm. I don't think we need to link to any publication for it.

Also, avoid deep-linking into PDF files, let's use links to Arxiv like this: https://arxiv.org/abs/1709.03395

PR #3332, opened 2019-07-30
[FUSED] Change alpha attribute in Elu from dynamic to static

tsocha:


tsocha:
> LGTM but could you comment on what motivated this change?

We're making this change to be compliant with our new operator specification.

src/ngraph/op/fused/elu.cpp
```diff
@@ -28,16 +28,18 @@
...
 {
     constructor_validate_and_infer_types();
 }
 
 NodeVector op::Elu::decompose_op() const
 {
     auto data = get_argument(0);
-    auto alpha_node = get_argument(1);
+    shared_ptr<Node> alpha_node =
+        make_shared<op::Constant>(data->get_element_type(), Shape{}, vector<double>{m_alpha});
```
GlebKazantaev: Why double constant depends on input data type? and what if data->get_element_type() returns something like FP16?

src/ngraph/serializer.cpp
```diff
@@ -2370,7 +2371,11 @@ json JSONSerializer::serialize_node(const Node& n)
         node["ellipsis_mask"] = tmp->get_ellipsis_mask();
         break;
     }
-    case OP_TYPEID::Elu: { break;
+    case OP_TYPEID::Elu:
+    {
+        auto tmp = dynamic_cast<const op::Elu*>(&n);
+        node["alpha"] = tmp->get_alpha();
```
GlebKazantaev: Should we check that dynamic_cast do not fails and returns not nullptr?

src/ngraph/op/fused/elu.hpp
```diff
@@ -35,13 +35,16 @@ namespace ngraph
             ///
             /// \param data Input tensor
             /// \param alpha Multiplier for negative values
-            Elu(const std::shared_ptr<ngraph::Node>& data,
-                const std::shared_ptr<ngraph::Node>& alpha);
+            Elu(const std::shared_ptr<ngraph::Node>& data, const double& alpha);
```
tomdol: PODs should not be passed by reference, please change to a simple `const double` for the `alpha` param.

python/pyngraph/ops/fused/elu.cpp
```diff
@@ -26,5 +26,5 @@ void regclass_pyngraph_op_Elu(py::module m)
 {
     py::class_<ngraph::op::Elu, std::shared_ptr<ngraph::op::Elu>, ngraph::op::Op> elu(m, "Elu");
     elu.doc() = "ngraph.impl.op.Elu wraps ngraph::op::Elu";
-    elu.def(py::init<const std::shared_ptr<ngraph::Node>&, const std::shared_ptr<ngraph::Node>&>());
+    elu.def(py::init<const std::shared_ptr<ngraph::Node>&, const double&>());
```
tomdol: This line will have to be updated when the constructor is changed to `const double`.

src/ngraph/op/fused/elu.cpp
```diff
@@ -52,9 +54,9 @@ NodeVector op::Elu::decompose_op() const
 
 shared_ptr<Node> op::Elu::copy_with_new_args(const NodeVector& new_args) const
 {
-    if (new_args.size() != 2)
+    if (new_args.size() != 1)
```
tomdol: Since you're already modifying this line, can you use `check_new_args_count` for consistency with other ops?

src/ngraph/op/fused/elu.cpp
tsocha: To perform operations each of the operands must be in the same type.
`op::Constant` perform a `static_cast` on input vector under the hood.

src/ngraph/serializer.cpp
tsocha: We don't check it in any operator.
I don't want to break file convention with it.
It is probably safe to assume that this dynamic_cast will be always successful.

PR #3334, opened 2019-07-30
[Py] Added depth_to_space operator to Python API.

Ewa21:


python/ngraph/ops.py
```diff
@@ -527,6 +527,27 @@ def convert(node, new_type, name=None):  # type: (Node, NumericType, str) -> Nod
...
+
+    Input tensor has shape [N,C,H,W], where N is the batch axis, C is the channel or depth,
+    H is the height and W is the width.
+
+    Output node produces a tensor with shape:
+
+    [N, C * :code:`block_size` * :code:`block_size`, H / :code:`block_size`, W / :code:`block_size`]
+
+    :param node: The node with input tensor data.
+    :param block_size: The size of the block of value to be moved.
```
indie: This sentence is hard to parse because "block of value" is maybe missing an article?

PR #3345, opened 2019-07-31
Convert CPU ops to new form

diyessi:
Remove obsolete sigmoid


src/ngraph/runtime/cpu/op/batch_mat_mul_transpose.hpp
```diff
@@ -34,14 +34,17 @@ namespace ngraph
         class BatchMatMulTranspose : public Op
         {
         public:
+            NGRAPH_API
+            static const std::string type_name;
```
nmostafa: Any reason this is puablic if we have `description()` ?

src/ngraph/runtime/cpu/op/batch_mat_mul_transpose.cpp
```diff
@@ -25,11 +25,13 @@
 using namespace std;
 using namespace ngraph;
 
-op::BatchMatMulTranspose::BatchMatMulTranspose(const shared_ptr<Node>& arg0,
-                                               const shared_ptr<Node>& arg1,
+const std::string op::BatchMatMulTranspose::type_name{"BatchMatMulTranspose"};
+
+op::BatchMatMulTranspose::BatchMatMulTranspose(const Output<Node>& arg0,
```
nmostafa: One minor thing, all tests now rely on implicit conversion between `node` and `Output<Node>`. Maybe we should add a test to use `Output<Node>` directly when constructing nodes, and perhaps use non-zero index ?

diyessi: Once all node class constructors have been upgraded, we can tighten things up some.

src/ngraph/runtime/cpu/op/batch_mat_mul_transpose.hpp
rkimballn1: With it public you can write something like `BatchMatMulTrasnpose::type_name` instead of typing the error prone literal string `"BatchMatMulTrasnpose"`. It is just a safer way to do it.

PR #3350, opened 2019-07-31
Add a compile time flag to support all ET's in cpu backend

nbpatel:
The libcpu_backend.so is reduced from 48M to 23M


src/ngraph/runtime/cpu/builder/reduction.hpp
```diff
@@ -41,10 +41,10 @@
         return;                                                                                    \
     }                                                                                              \
                                                                                                    \
-    if (reduction_axes.size() == arg_rank)                                                         \
+    if (reduction_axes.size() == arg_rank && is_fp_i64(args[0].get_element_type()))                \
```
rkimballn1: You could make an overload of `is_fp_i64` which takes an arg instead of an element type. Might make the code more readable.

src/ngraph/runtime/cpu/cpu_builder_optimized.hpp
```diff
@@ -0,0 +1,101 @@
+//*****************************************************************************
```
rkimballn1: No real need to change but just for consideration. The file name could be cpu_builder_size_optimized because optimized is a loaded word.

src/ngraph/runtime/cpu/cpu_builder_size_optimized.hpp
```diff
@@ -0,0 +1,101 @@
...
+#endif
+
+    if (et == ngraph::element::f32 || et == ngraph::element::i64)
+    {
+        return true;
+    }
+    return false;
+}
+
+#define SELECT_KERNEL_FOR_LIMITED_ET(KV, ET, K)                                                    \
```
jbobba: if NGRAPH_CPU_ALL_DATA_TYPE_ENABLE is set, this will redefine the macro to the smaller subset again!

nbpatel:
Ping for review :)

src/ngraph/runtime/cpu/kernel_selectors.hpp
```diff
@@ -0,0 +1,305 @@
...
+    else if (ET == element::u64)                                                                   \
+    {                                                                                              \
+        S(K, KV, ##__VA_ARGS__, uint64_t);                                                         \
+    }                                                                                              \
+    else                                                                                           \
+        throw ngraph_error("Unsupported element type " + ET.c_type_string() + " for kernel " #K);
+
+// Expand only selected datatypes. Named macros (e.g., F32_SELECT) are expanded based on build-flags
+#define EXPAND_ETS(K, KV, ET, S, ...)                                                              \
+    if (ET == element::boolean)                                                                    \
```
silee2: Shouldn't this be `if (BOOLEAN_EN && ET == element::boolean)` ? Otherwise we can have a segfault at runtime instead of going to the ngraph_error defined below. Same for all following if cases.

PR #3351, opened 2019-07-31
[MLIR] Fix issue with non-affine loads/stores in argmin/armax/gather ops

dcaballe:
Use std.loads and std.stores for indirect memory accesses in argmin/argmax/gather ops instead of the affine version.


src/contrib/mlir/lowerer.cpp
```diff
@@ -43,6 +43,8 @@ namespace
     using namespace mlir::edsc::op;
     using namespace ngraph::runtime;
     using namespace ngraph::runtime::ngmlir;
+    // Index notation to generate standard (i.e., non-affine) loads and stores.
+    using StdIndexedValue = TemplatedIndexedValue<intrinsics::std_load, intrinsics::std_store>;
```
dcaballe: I can try to push this to MLIR

PR #3353, opened 2019-07-31
[ONNX] Remove unused includes

postrational:


src/ngraph/frontend/onnx_import/op/lrn.cpp
```diff
@@ -17,7 +17,6 @@
 #include <memory>
 
 #include "core/node.hpp"
-#include "lrn.hpp"
```
tsocha: I'm not sure that this header should be removed.

src/ngraph/frontend/onnx_import/utils/convpool.cpp
```diff
@@ -15,10 +15,8 @@
 //*****************************************************************************
 
 #include <cmath>
-#include <string>
```
arogowie-intel: This one should be left here, but removed from header `convpool.hpp`

src/ngraph/frontend/onnx_import/utils/convpool.cpp
```diff
@@ -15,10 +15,8 @@
...
 #include <unordered_map>
 
-#include "convpool.hpp"
```
arogowie-intel: Why remove header file?

src/ngraph/frontend/onnx_import/utils/convpool.cpp
```diff
@@ -15,10 +15,8 @@
...
 #include "core/attribute.hpp"
 #include "core/node.hpp"
```
arogowie-intel: Both `attribute.hpp` and `node.hpp` are duplicated - they're already in header file.

Similarly `shape.hpp`, `coordinate_diff.hpp`

src/ngraph/frontend/onnx_import/utils/reshape.cpp
```diff
@@ -15,18 +15,15 @@
...
 #include <functional>
 #include <iterator>
 #include <numeric>
 #include <vector>
 
-#include "exceptions.hpp"
 #include "ngraph/builder/reshape.hpp"
 #include "ngraph/op/reshape.hpp"
 #include "utils/common.hpp"
-#include "utils/reshape.hpp"
```
arogowie-intel: Why removing header file of this source file?

postrational:
@arogowie-intel I'd like this to be a good start, I don't think I can do a complete clean up at this time. I addressed your comments, thanks.

PR #3354, opened 2019-07-31
Improve pragma usage.

silee2:
Limit compiler specific pragmas.
For example, MSVC should not have to deal with GCC or clang pragmas.


src/contrib/mlir/compiler.cpp
```diff
@@ -187,7 +187,7 @@ mlir::Type MLIRCompiler::get_mlir_type(const descriptor::Tensor* tensor)
 // Converts an nGraph element type into an MLIR type.
 mlir::Type MLIRCompiler::get_mlir_type(const element::Type& type)
 {
-#if !(defined(__GNUC__) && (__GNUC__ == 4 && __GNUC_MINOR__ == 8))
```
aprocter: I think most of these incantations can be eliminated, FWIW, since we now enable Wswitch and Wswitch-enum globally.

aprocter: Actually: I know we enable them globally for Clang, but not sure if we do so for other compilers, so these pragmas may still be having some effect. I suspect the CMake logic for the flags could be generalized to handle gcc as well, which would render these pragmas redundant.

silee2: where do we enable it? implicitly for clang?

aprocter: https://github.com/NervanaSystems/ngraph/blob/master/cmake/clang_4_0_flags.cmake

https://github.com/NervanaSystems/ngraph/blob/master/CMakeLists.txt#L322-L325

aprocter: (I think the `_4_0_` in the filename is incorrect.)

aprocter: Oh, and you're right---it comes from `-Weverything` and then we whitelist what we don't want. I dropped it from the whitelist in https://github.com/NervanaSystems/ngraph/pull/3207.

src/contrib/mlir/compiler.cpp
```diff
@@ -187,7 +187,7 @@ mlir::Type MLIRCompiler::get_mlir_type(const descriptor::Tensor* tensor)
...
+#if defined(__GNUC__) && !(__GNUC__ == 4 && __GNUC_MINOR__ == 8)
```
diyessi: Rather than repeat this mysterious-looking conditional a lot of places, can we define a more descriptive GNUC_BROKEN_ENUM_WARNING kind of thing and use it? Similarly for the CLANG ones?

PR #3365, opened 2019-08-01
 [Py] Added GRN operator to Python API.

Ewa21:


python/ngraph/ops.py
```diff
@@ -78,6 +78,22 @@ def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
     return Elu(as_node(data), as_node(alpha))
 
 
+@nameable_op
+def grn(data, bias, name=None):  # type: (Node, float, str) -> Node
+    r"""Perform Global Response Normalization with L2 norm (across channels only).
+
+    Computes GRN operation on channels for input blob:
```
postrational: I don't think we're using the word `blob` in our documentation so far. Let's keep it consistent.

python/test/ngraph/test_ops_fused.py
```diff
@@ -100,3 +100,26 @@ def test_clamp_operator_with_array():
     result = computation()
     expected = np.clip(data_value, min_value, max_value)
     assert np.allclose(result, expected)
+
+
+def test_grn_operator():
+    runtime = get_runtime()
+
+    data_value = np.arange(start=1.0, stop=25.0, dtype=np.float32).reshape(1, 2, 3, 4)
```
indie: `np.arange` 

Do you mean "arrange", like the arrangement of the reshape?  Or do you mean "a range" between values?

Ewa21: I mean "a range", array of evenly spaced values. `np.arange` returns `ndarray`.

postrational: https://docs.scipy.org/doc/numpy/reference/generated/numpy.arange.html

PR #3374, opened 2019-08-02
[Spec] Add axes input to LRN

mbencer:
Adding additional (optional) input to LRN to match new spec.


src/ngraph/op/lrn.hpp
```diff
@@ -35,7 +35,7 @@ namespace ngraph
         /// | Type                   | Description                                                                          |
         /// | ---------------------- | ------------------------------------------------------------------------------------ |
         /// | \f$N[n, c, d_1,\dots,d_n]\f$ | The tensor \f$T\f$, where \f$T[n, c, d_1,\dots,d_n] = \frac{N[n,i,d_1,\dots,d_n]}{ (bias + alpha * (\sum_{i=max(0,(nsize-1)/2)}^{min(C, (nsize-1)/2)+1} N[n,i,d_1,\dots,d_n]^{2}) ^ {2})}\f$ |
-        class LRN : public util::UnaryElementwiseArithmetic
```
aprocter: THANK YOU!

aprocter:
Need to undo the approval, actually... I thought it was one of the "layer" ops at first as opposed to Core, so wasn't looking quite closely enough. Will give it a full review later. Still, thanks for taking this out of UnaryElementwiseArithmetic. ;)

src/ngraph/op/lrn.cpp
```diff
@@ -34,7 +45,9 @@ op::LRN::LRN(const Output<Node>& arg, double alpha, double beta, double bias, si
 
 void op::LRN::validate_and_infer_types()
```
aprocter: If `axes` is a constant and `arg`'s rank is static, we need a check to make sure that all axes are in bounds. We'll need unit tests for that check, as well.

mbencer: - Checking shape of `axes` was added
- Unit tests to `arg` and `axes` was added

mbencer:
> * The `axes`, if constant, need to be checked in type propagation, and any added checks need tests under `test/type_prop`.
> * We need to update the reference implementation in `src/ngraph/runtime/reference/lrn.hpp`, and update `INTERPRETER` to use it.
> * We need tests for the new functionality in `test/backend/lrn.in.cpp`, and these need to pass on `INTERPRETER`.
> * We need to update the other backends (at least the CPU backend) to support the new flag, OR we need to update test backend manifests and inform the backend owners that the op's semantics have changed. (Current backends will be ignoring the `axes` input silently.)
> * Serializer (maybe just the deserializer, not sure) needs to be updated for the new input.
> * Ideally, we should document the new functionality. (This op does not have much existing documentation, unfortunately, so perhaps we can save this for another PR.)

Thank you very much for so precise explanation.
At the moment I've implemented changes related with test new input `axes` (implementation + unit test) and I updated deserialization.
Working on reference LRN implementation (and changes related to it) is in progress.

src/ngraph/op/lrn.cpp
```diff
@@ -34,16 +45,35 @@ op::LRN::LRN(const Output<Node>& arg, double alpha, double beta, double bias, si
...
+    const PartialShape& axes_shape = get_input_partial_shape(1);
 
     NODE_VALIDATION_CHECK(this,
                           input_shape.rank().is_dynamic() ||
                               static_cast<size_t>(input_shape.rank()) >= 3,
                           "Argument must have rank >= 3 (argument shape: ",
                           input_shape,
                           ").");
+
+    NODE_VALIDATION_CHECK(this, axes_shape.is_static(), "Input axes must be static.");
```
aprocter: (Going to go ahead and comment on this but I won't block on this in review. If you don't get a chance to update to address it, please open a JIRA.)

We should be able to handle all cases here:

1. If `axes_shape.is_static()` returns false, we just assume everything is okay. Else...
2. If `axes_shape` is not 1, we throw an exception (your PR is already doing this). Else...
3. If input 1 (the axes input) is not being driven by a constant, we just assume everything is okay. Else...
4. We get the value out of the constant node, and make sure that that value is in bounds.

aprocter:
@mbencer Is the work on the reference still pending?

mbencer:
> @mbencer Is the work on the reference still pending?

@aprocter I've tried to implement reference LRN implementation. At this moment I have some problems with handling case when there are more than one axis across which LRN is calculated. When I won't be able to implement it fast (until tomorrow inclusive) I will implement other review request changes and open JIRA task as you suggested.

src/ngraph/op/lrn.cpp
mbencer: Could you explain what do you mean by `Else...` statement. Should we handle all listed cases? For example adding support for dynamic shapes is required?

aprocter: Supporting dynamic shapes is required. So we have to handle all cases. (Shape of "axes" tensor is dynamic; shape of "axes" tensor is static and invalid; shape of "axes" tensor is static and valid but the value of the "axes" tensor is unknown; value of the "axes" tensor is known but invalid; and value of the "axes" tensor is valid.)

mbencer: Thank for explanation. Because the PR was merged I'll add all these requirements to the JIRA task (created for LRN reference implementation)

PR #3376, opened 2019-08-02
New method to handle numbers near zero

rkimballn1:
This allows some NNP unit tests to pass where the expected value is zero

Review is easier if you disable white-space differences.


test/util/all_close_f.cpp
```diff
@@ -58,6 +58,14 @@ uint32_t test::float_distance(float a, float b, float min_signal)
 
     FloatUnion a_fu{a};
     FloatUnion b_fu{b};
+    if (a_fu.i == 0 || b_fu.i == 0)
+    {
+        // Comparing to zero breaks this algorithm. If one of the numbers is zero then
+        // add a bias value to both values to get it away from zero.
+        constexpr float bias = 1.0f;
```
gcwenger: I'm not comfortable with adding `1.0f` to both numbers.
I think this would have the effect of returning a distance of `0` between any number < `0.0000001` (`1E-7`). I don't like that property.
Is there a reason why adding `FLT_MIN` as the bias doesn't work?
If adding `FLT_MIN` as the bias doesn't work, I worry that the approach is flawed.

test/util/all_close_f.cpp
```diff
@@ -113,6 +121,14 @@ uint64_t test::float_distance(double a, double b, double min_signal)
 
     DoubleUnion a_du{a};
     DoubleUnion b_du{b};
+    if (a_du.i == 0 || b_du.i == 0)
+    {
+        // Comparing to zero breaks this algorithm. If one of the numbers is zero then
+        // add a bias value to both values to get it away from zero.
+        constexpr float bias = 1.0f;
```
gcwenger: Same concerns as above, but with `DBL_MIN` preferred here.

PR #3379, opened 2019-08-02
private function to static initialize all the backend for static library builds

pruthviIntel:


src/ngraph/runtime/backend_manager.cpp
```diff
@@ -53,6 +53,7 @@ void runtime::BackendManager::register_backend(const string& name, BackendConstr
 
 vector<string> runtime::BackendManager::get_registered_backends()
 {
+    initilize_backends();
```
rkimballn1: It is always good to spell `initialize` correctly

PR #3380, opened 2019-08-02
Export get_backend_constructor_pointer as CPU backend API.

jfdong:
onnxruntime with nGraph EP run into run time error because of missing symbol “get_backend_constructor_pointer” in the CPU backend dll.


jfdong:
close this. 
https://github.com/NervanaSystems/ngraph/pull/3379 is the fix for master.
https://github.com/NervanaSystems/ngraph/pull/3381 is the fix for 0.22

PR #3382, opened 2019-08-02
Update reference kernels, constant folding to support auto-broadcast directly

aprocter:
The existing reference functions do not support autobroadcast functionality for binops. That means that ImplicitBroadcastElimination has to be run before ConstantFolding, which makes it unsuitable for some use cases related to dynamic shapes.

This PR extends the references to cover autobroadcasting for binops, removes the INTERPRETER dependency on ImplicitBroadcastElimination, and adds some tests for all of the above.

LIMITATION: The "CPU constant folders" still do not support autobroadcasting directly. ~I have added checks in this PR in case users try to do ConstantFolding with the CPU folders before ImplicitBroadcastElimination has been run.~ After some discussion with @nmostafa, changed this to a fallback to reference, rather than an assert.


src/ngraph/pass/constant_folding.cpp
```diff
@@ -870,29 +870,33 @@ shared_ptr<op::Constant> fold_constant_binary(shared_ptr<op::Constant> a,
     }
     else
     {
-        if (std::dynamic_pointer_cast<op::Add>(binary))
+        if (auto add_node = std::dynamic_pointer_cast<op::Add>(binary))
```
nmostafa: For CPU builders, can we instead check the auto-broadcast mode above instead ? assert it is not `BinaryElementwiseArithmetic` or `get_autob()` is None. I rather replace all the checks in the builders with just one check here.

aprocter: Ooh, great idea. Will do.

aprocter: @nmostafa Per this and our offline discussion, have moved the check into the `if` above and changed it from an assert to a fallback to reference. Note that the check is slightly different from what you describe: instead of checking `get_autob()` we check to make sure the input shapes match. This is because `ImplicitBroadcastElimination` does not actually change the autobroadcast spec, it just normalizes the shape so that downstream passes can safely ignore it.

PR #3385, opened 2019-08-03
Convert fused ops to value-based constructors

diyessi:


diyessi:
Separate PR -- need the constructor side of all ops done first.

PR #3386, opened 2019-08-05
Add method so that element::Type can be cast into enum

rkimballn1:
Old API is still there but deprecated


aprocter:
Removing review request from @avijit-nervana. He was tagged just for `cpu/builder/allreduce.*pp`, which is due to an outdated entry in `CODEOWNERS`.

PR #3390, opened 2019-08-05
Don't enable dl use in static builds

diyessi:
remove dladdr in static


src/ngraph/runtime/backend.cpp
```diff
@@ -37,6 +37,9 @@ std::string runtime::Backend::s_backend_shared_library_search_directory;
 // This finds the full path of the containing shared library
 static string find_my_pathname()
 {
+#ifdef NGRAPH_STATIC_LIB_ENABLE
```
nmostafa: So we cannot dyn load any other backends besides CPU and INTERPRETER if the flag is enabled ?

src/ngraph/runtime/backend_manager.cpp
```diff
@@ -32,12 +32,18 @@
 using namespace std;
 using namespace ngraph;
 
+#ifdef NGRAPH_STATIC_LIB_ENABLE
+#define DLERROR() ""
+#else
 #ifdef _WIN32
 #define CLOSE_LIBRARY(a) FreeLibrary(a)
 #define DLSYM(a, b) GetProcAddress(a, b)
+#define DLERROR() dlerror()
```
s-barannikov: This should be in the other branch of the `#if`.

diyessi: It's in all three branches; it's a NO-OP not on _WIN32.

s-barannikov: That's confusing. Didn't know there is `dlerror` function on Windows.

s-barannikov: Just to be clear: I'm talking about the implementation of the macro. `dlerror` sounds more posix-like.

src/ngraph/runtime/backend.cpp
rkimballn1: Agreed. This should be fixed to work correctly.

src/ngraph/runtime/backend_manager.cpp
rkimballn1: dlerror() is linux not windows. Linux has the empty macro and windows has dlerror()

src/ngraph/runtime/backend.cpp
diyessi: And what is "correctly"? There's no real documentation for what these are supposed to do.

diyessi:
Duplicated by #3394

PR #3391, opened 2019-08-05
[MLIR] Add unary op -- Negative 

nbpatel:
Just to get started with MLIR. Will create a different PR with all the remaining unary ops


src/contrib/mlir/lowerer.cpp
```diff
@@ -526,6 +526,58 @@ namespace
         return matchSuccess();
     }
```
pruthviIntel: You can add a utility function to lower all the Unary Op's , look at `lower_binary_elementwise` for example

src/contrib/mlir/lowerer.cpp
```diff
@@ -526,6 +526,58 @@ namespace
...
+        auto lbs = vLHS.getLbs();
+        auto ubs = vLHS.getUbs();
+        // Loop induction vars
+        auto ivs = IndexHandle::makeIndexHandles(vLHS.rank());
+        auto pivs = IndexHandle::makeIndexHandlePointers(ivs);
+        // Steps
+        auto steps = vLHS.getSteps();
+
+        NGRAPH_CHECK(lhs->getType().isa<MemRefType>());
+        Type elemTy = lhs->getType().dyn_cast<MemRefType>().getElementType();
```
dcaballe: Since you are checking the type with `isa` in 556, you should use `cast` instead of `dyn_cast`. `cast` will assert if the casting is wrong so we don't have to check that the return is a nullptr.

src/contrib/mlir/lowerer.cpp
```diff
@@ -526,6 +526,58 @@ namespace
...
+
+        LoopNestBuilder(pivs, lbs, ubs, steps)([&] {
+            ValueHandle val = iLHS(ivs);
+            if (auto floatTy = elemTy.dyn_cast<FloatType>())
+            {
+                ValueHandle two = intrinsics::constant_float(llvm::APFloat(2.0f), floatTy);
+                iRes(ivs) = val - val * two;
```
dcaballe: Interesting! Is it not possible to do just `- val`? I think the current implementation might not be accurate for floating point values due to intermediate rounding and floating point special values.

src/contrib/mlir/lowerer.cpp
nbpatel: Thats for another PR

src/contrib/mlir/lowerer.cpp
nbpatel: -val was giving errors

dcaballe: What kind of error? Maybe that op is not implemented in the EDSC and we can contribute it to MLIR. Also, maybe you could try to do something like `0 - val`. That should work and would lead to precision issues.

src/contrib/mlir/lowerer.cpp
nmostafa: Usually it is  `if(auto elemTy = dyn_cast<MemRefType>(..)) { /* do something with elemTy */ } else { assert(0); }`

src/contrib/mlir/lowerer.cpp
nmostafa: I suggested 0 - val before. Std doesn't have NegOp. But maybe we can add a negation operator to EDSC that does 0 - val. 
@nbpatel , you need to define the Constant value first

nmostafa: nit: no need for `instrinsics::`

src/contrib/mlir/dialect/ops.td
```diff
@@ -144,7 +144,7 @@ def NGCoshOp     : NG_Unary_Arith_Op<"cosh">;
 def NGExpOp      : NG_Unary_Arith_Op<"exp">;
 def NGFloorOp    : NG_Unary_Arith_Op<"floor">;
 def NGLogOp      : NG_Unary_Arith_Op<"log">;
-def NGNegOp      : NG_Unary_Arith_Op<"neg">;
+def NGNegativeOp : NG_Unary_Arith_Op<"neg">;
```
nmostafa: Why not `NGNegOp` ?

src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
```diff
@@ -342,6 +343,19 @@ bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node
             return true;
         }
     }
+
+    if (TI(ngraph::op::Negative) == TI(*node))
```
nmostafa: You should support both types for this PR. Anything limiting this ?

src/contrib/mlir/lowerer.cpp
dcaballe: I've never seen the `dyn_cast` + `assert(0)` idiom. `dyn_cast` is used when multiple kind of objects are expected. For example:

```
if (auto memRefTy = dyn_cast<MemRefType>() {
    do something with memRefTy
} else if (auto tensorTy = dyn_cast<TensorType>() {
    do something with tensorTy
} else if ...
```

If `memRefTy` and `tensorTy` are not needed (do something could be implemented using the base clase), `isa` is used instead, since it's more efficient.

If we expect a single type, such as in this case, `assert(isa<>)` + `cast` or just `cast` is preferred since `cast` has an assert inside.

src/contrib/mlir/lowerer.cpp
nbpatel: error: no match for ‘operator-’ (operand type is ‘mlir::edsc::ValueHandle’)  ...get the same error for -val and 0 - val

src/contrib/mlir/dialect/ops.td
nbpatel: no particular reason, just tried being consistent with ngraph

src/contrib/mlir/lowerer.cpp
nmostafa: The operator exists. What does your code look like. Should be
```
ValueHandel zero = constant_float(...);
iRes(ivs) = zero - vale;
```

src/contrib/mlir/dialect/ops.td
nmostafa: Let's change it then to be consistent with the op name :)

dcaballe: +1. I also think `NGNegOp` is clearer. I did some "renaming tuning" as well for the reduction Ops.

diyessi: It's `Neg` in ONNX.

PR #3392, opened 2019-08-06
Remove not-directly-used includes

s-barannikov:
Based on clang analysis.


diyessi:
Need to be careful on STL files; various compilers implicitly include things that others don't. For example, `iota` is one that is compiler-dependent.

s-barannikov:
The check I run only considered explicit includes. I.e. if a symbol from explicitly included header is used in current translation unit, the header it's not marked as unused.
But there is a potential problem if a symbol is defined in different headers by different STL implementations. One of the implementation should be non standard conformant in this case. Is that what you mean?

aprocter:
@s-barannikov Yup, it can happen that macOS and Linux (for example) disagree over whether including header “H” is sufficient to get a prototype for function “F”. So we might need to give this a try on Mac and Windows, at a minimum. I can give it a shot on our CI systems sometime in the next couple days, if someone else doesn’t get to it first.

s-barannikov:
@aprocter Thanks!

aprocter:
I ran a build on our internal CI systems, and everything seems to have _compiled_ successfully, but CI failed for clang-format issues. I'll run a clang-format fix and try again, but if everything compiles it's probably fine.

diyessi:
#3416

PR #3393, opened 2019-08-06
[Py] Added operators Shuffle Channels, Squared Difference and Squeeze to Python API.

Ewa21:


python/ngraph/ops.py
```diff
@@ -78,6 +78,94 @@ def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
...
+                           [[16., 17.], [18., 19.]],
+                           [[ 4.,  5.], [ 6.,  7.]],
+                           [[12., 13.], [14., 15.]],
+                           [[20., 21.], [22., 23.]]]]
+
+    :param data: The node with input tensor.
+    :param axis: Channel dimension index in the data tensor.
+                 A negative value means that the index should be calculated
+                 from the back of the input data shape.
+    :param group: Number of groups the channel dimension specified by axis should be split into.
```
indie: I had to read this 2-3 times to understand what it meant, but I think it make sense.  

Would another way of describing this be something like "The channel dimension specified by an axis should be split into `number` of groups"?

python/ngraph/ops.py
```diff
@@ -78,6 +78,94 @@ def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
...
+    """
+    return SquaredDifference(x1, x2)
+
+
+@nameable_op
+def squeeze(data, axes, name=None):  # type: (Node, NodeInput, str) -> Node
+    """Perform squeeze operation on input tensor.
+
+    Remove single-dimensional entries from the shape of a tensor.
+    Takes a parameter axes with a list of axes to squeeze.
```
indie: Grammatically correct would be either: "Takes parameter axes" (takes plural) or "takes a parameter axis" (takes a singular).  I think.  :)

python/ngraph/ops.py
```diff
@@ -78,6 +78,94 @@ def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
...
+    If axes is not provided, all the single dimensions will be removed from the shape.
```
indie: Similarly, "If axes are not provided ..." because to be grammatically correct, we say "If [`these things`] are not provided ...

python/ngraph/ops.py
postrational: The parameter is named `axes`.

> Takes a parameter (`axes`) ...
may be the right choice.

python/ngraph/ops.py
postrational: I think we should use a style to mark the name of the param, like so: 

> If `axes` are not provided

python/ngraph/ops.py
indie: Gotcha, yeah. The whole thing makes more sense w/the style adjustment.

python/ngraph/ops.py
```diff
@@ -78,6 +78,94 @@ def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
...
 
 
+@nameable_op
+def shuffle_channels(data, axis, groups, name=None):  # type: (Node, int, int, str) -> Node
+    """Perform permutation on data in the channel dimension of the input tensor.
+
+    The operation is the equivalent with the following transformation of the input tensor
+    :code:`data` of shape [N, C, H, W]:
+
+    :code:`data'` = reshape(:code:`data`, [N, group, C / group, H * W])
```
postrational: Let's change the name `data'` to `data_reshaped`.

python/ngraph/ops.py
```diff
@@ -78,6 +78,94 @@ def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
...
+
+    :code:`data''` = transpose(:code:`data'`, [0, 2, 1, 3])
```
postrational: Let's change the name `data''` to `data_transposed`.

python/ngraph/ops.py
postrational: ... specified by the `axis` parameter...

python/ngraph/ops.py
```diff
@@ -78,6 +78,94 @@ def elu(data, alpha, name=None):  # type: (NodeInput, NodeInput, str) -> Node
...
+                 A negative value means that the index should be calculated
+                 from the back of the input data shape.
+    :param group: Number of groups the channel dimension specified by axis should be split into.
+    :param name: Optional output node name.
+    :return: The new node performing a permutation on data in the channel dimension of the input tensor.
+    """
+    return ShuffleChannels(data, axis, groups)
+
+
+@nameable_op
```
postrational: Should this not use a `@binary_op` decorator?

postrational:
Please address my comments.

PR #3394, opened 2019-08-06
 Don't enable dl use in static builds for r0.25

pruthviIntel:


src/ngraph/runtime/backend_manager.cpp
```diff
@@ -107,13 +113,13 @@ shared_ptr<runtime::Backend> runtime::BackendManager::create_backend(const std::
             stringstream ss;
             ss << "Backend '" << type << "' not registered. Error:";
 #ifndef _WIN32
-            ss << dlerror();
+            ss << DLERROR();
 #endif
             throw runtime_error(ss.str());
         }
-
+#ifndef NGRAPH_STATIC_LIB_ENABLE
```
rkimballn1: I think you need to move this up a few lines. You are opening the shared library but then never using it. I think your intent is to not open the shared library, right?

src/ngraph/CMakeLists.txt
```diff
@@ -521,6 +521,14 @@ if(NOT NGRAPH_JSON_ENABLE)
     target_compile_definitions(ngraph PUBLIC NGRAPH_JSON_DISABLE)
 endif()
 
+if(NGRAPH_INTERPRETER_STATIC_LIB_ENABLE)
```
rkimballn1: If you are disabling the ability for a static ngraph to load dynamic libraries then you don't need the option to build a backend as a shared library. That makes no sense. You would builld a library that you can't use. 
There should be only one flag, `NGRAPH_STATIC_LIB_ENABLE` and if you try to build a backend which does not support static linkage it should report a cmake error. Remove these individual backend enable flags.

PR #3395, opened 2019-08-06
PlaidML Grouped Convolution Implementation

dgkutnic:
Implements the `GroupConvolution` operation for the PlaidML backend, and adds new unit tests to verify correctness of the op.


src/ngraph/runtime/plaidml/unit_test.manifest
```diff
@@ -178,7 +178,7 @@ conv_bias_2d
 conv_bias_3d
 conv_bias_bprop_2d
 conv_bias_add_2d
-group_conv
+#group_conv
```
earhart: You can just remove these lines

src/ngraph/runtime/plaidml/plaidml_compiler.cpp
```diff
@@ -87,7 +88,7 @@ std::shared_ptr<ngraph::runtime::plaidml::PlaidML_Executable>
     pass_manager.set_per_pass_validation(false);
 
     // We apply the same general-purposes passes as the CPU backend.
-    pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>();
+    pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>([] (const Node& node) -> bool {if (node.description() == "GroupConvolution") return true; return false;});
```
aprocter: I imagine this will fail the style check (along with a few other things throughout the PR), so run `make style-apply` if CI complains. Outside the clang-format rules, nGraph coding standards require single-statement `if` bodies to be enclosed in curly braces.

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.cpp
```diff
@@ -0,0 +1,109 @@
...
+    const auto& fd1 = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(filter_dilation[1])); 
+    const auto& pxb = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(padding_below[0]));
+    const auto& pyb = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(padding_below[1]));
+    const auto& pxa = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(padding_above[0]));
+    const auto& pya = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(padding_above[1]));
+    const auto& sx = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(strides[0]));
+    const auto& sy = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(strides[1]));
+
+    this->set_output(::vertexai::plaidml::function{R"(
+            function (I[N, CI, XI0, XI1], F[CO, FCI, XF0, XF1], DD0, DD1, FD0, FD1, G, PXB, PYB, PXA, PYA, SX, SY) -> (O) {
```
aprocter: For lines this long, you might need to wrap it in `//clang-format off` ... `//clang-format on`.

src/ngraph/runtime/plaidml/unit_test.manifest
dgkutnic: Done!

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace plaidml
+        {
+            namespace op
+            {
+                class GroupConvolution;
```
aprocter: How is this class being used? I don't see anything that actually instantiates it.

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.cpp
```diff
@@ -0,0 +1,109 @@
...
+                O[n, (CO/G) * g + co, x, y: N, CO, ((DD0 * (XI0 - 1) + PXA + PXB) - (FD0 * (XF0 - 1)) + SX) / SX, ((DD1 * (XI1 - 1) + PYA + PYB) - (FD1 * (XF1 - 1)) + SY) / SY] = 
+                    +(I[n, (CI/G) * g + ci, (x + FD0 * xf0 - PXB)/DD0, (y + FD1 * xf1 - PYB)/DD1] * F[(CO/G) * g + co, ci, xf0, xf1]), co < CO/G;
+            })"}(image, filter, dd0, dd1, fd0, fd1, grps, pxb, pyb, pxa, pya, sx, sy));
+
+}
```
silee2: I think you will get something like a no newline before EOF error.

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.cpp
aprocter: Just realized this is a multi-line literal, so `clang-format` might be okay with it. Anyway, it'll let you know.

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.hpp
dgkutnic: We don't have anything that instantiates it - this is just an artifact of the `Convolution` op (which I used as a template, and which actually has other passes that instantiate it). If we don't think anything will instantiate GroupConvolution, I can get rid of it.

aprocter: Gotcha- might be best to get rid of it in that case, but up to you.

dgkutnic: @tzerrell what do you think -- should we be keeping this or getting rid of it?

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.cpp
```diff
@@ -0,0 +1,109 @@
...
+    const auto& pxb = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(padding_below[0]));
+    const auto& pyb = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(padding_below[1]));
+    const auto& pxa = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(padding_above[0]));
+    const auto& pya = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(padding_above[1]));
+    const auto& sx = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(strides[0]));
+    const auto& sy = static_cast<::vertexai::plaidml::variable>(static_cast<const int64_t>(strides[1]));
+
+    this->set_output(::vertexai::plaidml::function{R"(
+            function (I[N, CI, XI0, XI1], F[CO, FCI, XF0, XF1], DD0, DD1, FD0, FD1, G, PXB, PYB, PXA, PYA, SX, SY) -> (O) {
+                O[n, (CO/G) * g + co, x, y: N, CO, ((DD0 * (XI0 - 1) + PXA + PXB) - (FD0 * (XF0 - 1)) + SX) / SX, ((DD1 * (XI1 - 1) + PYA + PYB) - (FD1 * (XF1 - 1)) + SY) / SY] =
```
tzerrell: Did you test that the output dimensions were right under slight variations of the input dimensions? Especially in cases where strides, dilations, data_dilations, or a combination thereof are not 1, it can be really easy to get off-by-one errors in output dimensions in limited cases.

Not sure whether such tests should go in permanently, that might add up to a lot of very similar tests.

dgkutnic: I have a suite of tests that check for correctness of shapes/values for various parameter values of strides/window dilations/data dilations/padding. Those are available in `test/backend/fused_op.in.cpp`. Let me know if you meant something else other than the ones in there. Thanks!

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.hpp
tzerrell: I don't particularly see a need for it, but perhaps @earhart would know differently?

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.cpp
tzerrell: I meant a more thorough suite, in particular one testing a variety of input shapes against various fixed collections of stride/dilation parameters. The current tests are all using the same input data size, and it might happen to be one that works (the off-by-one errors I'm referring to are often cases like "if input_shape % strides == 1 then fail else succeed" which a single input size wouldn't necessarily catch)

tzerrell: Also, it's really just the output dimensions I'm worried about for this, it seems unlikely that errors would crop up in the values in only these cases.

dgkutnic: Fair enough - I'll add those in.

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.hpp
earhart: We shouldn't need the .hpp file for this at all.  (Convolution has one since there's a pass that folds transposes into convolutions.)

dgkutnic: The consensus is that I'll remove this class. Thanks for the input!

src/ngraph/runtime/plaidml/plaidml_compiler.cpp
```diff
@@ -87,7 +88,11 @@ std::shared_ptr<ngraph::runtime::plaidml::PlaidML_Executable>
     pass_manager.set_per_pass_validation(false);
 
     // We apply the same general-purposes passes as the CPU backend.
-    pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>();
+    pass_manager.register_pass<ngraph::pass::FusedOpDecomposition>([](const Node& node) -> bool {
+        if (node.description() == "GroupConvolution")
```
diyessi: `(node.description() == GroupConvolution.type_name)`

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.cpp
dgkutnic: Fixed!

src/ngraph/runtime/plaidml/plaidml_compiler.cpp
dgkutnic: Fixed!

src/ngraph/runtime/plaidml/plaidml_ops_group_convolution.cpp
dgkutnic: Added in a couple of new tests, and the op is still acting as expected.

PR #3397, opened 2019-08-07
[MLIR] Enable affine loop tiling

dcaballe:
This PR enables loop tiling optimization in affine dialect. It
introduces the following flags for configuration.
  - affine-loop-tile: enables/disables the optimization. Disabled by
    default.
  - loop-tile-cache-level: provides the cache level to which apply loop
    tiling to. Cache level size is obtained from LLVM's TTI.
  - loop-tile-cache-size: provides a cache level size that overrides
    cache information from TTI.

The current use of TTI is a bit hacky since we have to pass a fake
LLVM's function to make it work. However, this should be enough to get
some basic target information until we have a target model in MLIR or
find a better approach.


src/contrib/mlir/compiler.cpp
```diff
@@ -300,15 +325,104 @@ void MLIRCompiler::lower_ng_dialect()
     m_engine = std::move(maybeEngine.get());
 }
 
+/// Creates target machine for current host.
+static llvm::Expected<std::unique_ptr<llvm::TargetMachine>> createDefaultTargetMachine()
```
dcaballe: This func will be simplified once we update the MLIR/LLVM version

src/contrib/mlir/compiler.cpp
```diff
@@ -300,15 +325,104 @@ void MLIRCompiler::lower_ng_dialect()
...
     if (clEnableAffineLoopFusion)
     {
         pm_opts.addPass(mlir::createLoopFusionPass());
     }
 
+    if (clEnableAffineLoopTiling)
+    {
+        unsigned cacheLevelSize =
+            getCacheLevelSize(targetInfo, clLoopTilingCacheLevel, clLoopTilingCacheSize);
+        LLVM_DEBUG(llvm::dbgs() << "Enabling Affine Loop Tiling for cache level "
```
dcaballe: I decided to use `LLVM_DEBUG` here so that we have a unified way of dumping the debug information related to MLIR (just by using the `-debug` flag). I struggle a bit to have all the debug information dumped without it. What do you think?

src/contrib/mlir/compiler.cpp
```diff
@@ -300,15 +325,104 @@ void MLIRCompiler::lower_ng_dialect()
...
+    llvm::StringMap<bool> featureMap;
+    llvm::sys::getHostCPUFeatures(featureMap);
+    for (auto& feature : featureMap)
+    {
+        subtargetFeatures.AddFeature(feature.first(), feature.second);
+    }
+
+    // Relocation model and code model are kept to default values.
+    machineBuilder->setCPU(llvm::sys::getHostCPUName());
+    machineBuilder->setCodeGenOptLevel(llvm::CodeGenOpt::Aggressive);
```
nmostafa: Will this cause -O3 level ? We do set opt level for the ORCJit else where.

src/contrib/mlir/compiler.cpp
```diff
@@ -65,17 +69,38 @@
...
+static llvm::cl::opt<unsigned>
+    clLoopTilingCacheLevel("loop-tile-cache-level",
+                           llvm::cl::init(2),
+                           llvm::cl::desc("Cache level to which to apply loop tiling"));
+
+static llvm::cl::opt<unsigned>
+    clLoopTilingCacheSize("loop-tile-cache-size",
+                          llvm::cl::init(0),
+                          llvm::cl::desc("Cache size to use in loop tiling. Use cache size "
+                                         "provided by TargetTransformInfo by default."));
```
nmostafa: Rename options to `affine-loop-tile-*` to be consistent. 
Also, refine that comment to mention that it overrides the cache size inferred from cache-level and TTI

src/contrib/mlir/compiler.cpp
```diff
@@ -300,15 +325,104 @@ void MLIRCompiler::lower_ng_dialect()
...
+    return optCacheLevelSize.getValue();
+}
+
+// Receives affine dialect as input and applies affine and standard dialect based optimizations.
+// Lowering from affine dialect to standard dialect happens along the way. Output consists of
+// standard dialect only ops.
 void MLIRCompiler::optimize()
 {
+    // Create target machine with all the current host features.
+    llvm::InitializeNativeTarget();
```
nmostafa: Outline all this to `initTarget()` ? And call it before `optimize()`

src/contrib/mlir/compiler.cpp
```diff
@@ -300,15 +325,104 @@ void MLIRCompiler::lower_ng_dialect()
...
+
+    // Create target transform info to obtain some target information to be used in MLIR
+    // optimizations. This is a temporary attempt to retrieve some target information by reusing
+    // LLVM TTI infra while MLIR does not have target model.
+    llvm::LLVMContext llvmContext;
+    auto module = make_unique<llvm::Module>("test", llvmContext);
+    module->setDataLayout(targetMachine->createDataLayout());
+    auto ttiSetupFunc = llvm::cast<llvm::Function>(
+        module
+            ->getOrInsertFunction("__ngraph_tti_setup",
```
nmostafa: What's the role of `__ngraph_tti_setup` ?

src/contrib/mlir/compiler.cpp
dcaballe: > Rename options to `affine-loop-tile-*` to be consistent.

I did this purposely, since we may have different loop tile optimizations (affine, stripe, LLVM, ...) and we could reuse the common configuration flags for all of them. I can change it, though, if you still think it doesn't make sense.

> Also, refine that comment to mention that it overrides the cache size inferred from cache-level and TTI

Sure

src/contrib/mlir/compiler.cpp
dcaballe: It applies only to LLVM's codegen but I'm not sure what actually does. The optimization level of the LLVM optimizer is set up differently, using `populate...` MLIR APIs.

src/contrib/mlir/compiler.cpp
dcaballe: It sounds reasonable. Please, let me take care of this after the MLIR update. We will have to pass a target machine to some MLIR APIs as well and that may lead to a different refactoring.

src/contrib/mlir/compiler.cpp
dcaballe: It's just a fake function that TTI uses to retrieve the LLVM context through `getTargetTransformInfo`. Nobody calls it.

src/contrib/mlir/compiler.cpp
nmostafa: Then rather keep this configurable based on `NGRAPH_MLIR_OPT_LEVEL`, otherwise debugging JIT'ed code will be hard.

src/contrib/mlir/compiler.cpp
nmostafa: Better have different knobs for each pass for debugging/performance tuning purposes.

src/contrib/mlir/compiler.cpp
nmostafa: Sure. We still need to do the general compiler refactoring as well.

src/contrib/mlir/compiler.cpp
dcaballe: Probably better use the ocl flag for that. We could have our own -O{0,1,2,3}. I'll open a JR for that since we would have to also change LLVM optimizer pipeline for that.

nmostafa: Not sure I follow. We already  have an env variable that controls the llvm pipeline. I just don't want confusion if I set OPT level to 0 and I still get optimized code while debugging. Can you verify that at least ? `NGRAPH_MLIR_OPT_LEVEL=0 NGRAPH_MLIR=1 NGRAPH_LLVM_OPTIONS="-print-after-all" ./unit-test --gtest-filter=CPU.mlir_*`

PR #3398, opened 2019-08-07
Fix version label

rkimballn1:
This fixes the version label creation. Before if you had a release like `v0.24.0` then the version label would be `0.24.0+hash` when it should have been just `0.24.0`
The separator for `git show-ref` is a space in my tests so I added support for any number of spaces or tabs between hash and label.


cmake/Modules/git_tags.cmake
```diff
@@ -42,10 +42,15 @@ function(NGRAPH_GET_TAG_OF_CURRENT_HASH)
     NGRAPH_GET_CURRENT_HASH()
 
     if (NOT ${TAG_LIST} STREQUAL "")
-        string(REGEX MATCH "${NGRAPH_CURRENT_HASH}\t[^\r\n]*" TAG ${TAG_LIST})
-        set(FINAL_TAG ${TAG})
+        # first look for vX.Y.Z release tag
+        string(REGEX MATCH "${NGRAPH_CURRENT_HASH}[\t ]+refs/tags/v([0-9?]+)\\.([0-9?]+)\\.([0-9?]+)" TAG ${TAG_LIST})
```
diyessi: This will match `v0.1.5xyz`. I think we should reject if it doesn't match a version exactly.

rkimballn1: I will add a unit test for that exact case. It currently `should` reject your example as invalid

rkimballn1: There are two regex tests to get a version tag in order of priority
1) Must match vX.Y.Z exactly
2) Must match vX.Y.Z-rc.N
Those are the only two matches. If neither of those match then it will find no version number

PR #3411, opened 2019-08-08
Fix sum reference to handle corner cases with +-inf

aprocter:
Got a report that ConstantFolding (which delegates directly to the reference) was returning nan on [-inf,-inf] and (I think) [+inf,+inf], which is wrong (should be -inf and +inf). I think this boils down to a corner case in our stable sum algorithm.


aprocter:
Going to originate this in r0.25 and cherry-pick back, actually.

PR #3412, opened 2019-08-08
Fix sum reference to handle corner cases with +-inf

aprocter:
Got a report that ConstantFolding (which delegates directly to the reference) was returning nan on [-inf,-inf] and (I think) [+inf,+inf], which is wrong (should be -inf and +inf). I think this boils down to a corner case in our stable sum algorithm.

### DIFF
--- a/test/backend/sum.in.cpp
+++ b/test/backend/sum.in.cpp
@@ -740,3 +740,91 @@ NGRAPH_TEST(${BACKEND_NAME}, sum_dynamic)
         ASSERT_TRUE(test::all_close_f(results, expected_results[i], MIN_FLOAT_TOLERANCE_BITS));
     }
 }
+
+NGRAPH_TEST(${BACKEND_NAME}, sum_neg_inf_neg_inf)
+{
+    Shape shape{2};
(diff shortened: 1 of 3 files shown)


test/backend/sum.in.cpp
```diff
@@ -740,3 +740,91 @@ NGRAPH_TEST(${BACKEND_NAME}, sum_dynamic)
         ASSERT_TRUE(test::all_close_f(results, expected_results[i], MIN_FLOAT_TOLERANCE_BITS));
     }
 }
+
+NGRAPH_TEST(${BACKEND_NAME}, sum_neg_inf_neg_inf)
+{
+    Shape shape{2};
+    auto A = make_shared<op::Parameter>(element::f32, shape);
+    auto f = make_shared<Function>(make_shared<op::Sum>(A, AxisSet{0}), ParameterVector{A});
+
```
diyessi: Better to try a bigger vector with a mix of finite and infinite and then a few permutations of it.

aprocter: We can do that, but these cases are the specific ones that were failing in the TF unit test that prompted this PR.

PR #3419, opened 2019-08-09
[Spec] Adjust op Proposal

mbencer:
New requirements:
- Expected elements type of `image_shape` should be float instead of u64
- Expected node type of `image_shape` should NOT be Constant
- Calculation shape of output type should be fixed


test/type_prop/proposal.cpp
```diff
@@ -0,0 +1,122 @@
...
+        auto proposal = make_shared<op::Proposal>(class_probs, class_logits, image_shape, attrs);
+        // Should have thrown, so fail if it didn't
+        FAIL() << "Invalid input tensor rank.";
+    }
+    catch (const NodeValidationFailure& error)
+    {
+        EXPECT_HAS_SUBSTRING(
+            error.what(),
+            std::string(
+                "Image_shape 1D tensor must have =>3 and <= 4 elements (image_shape_shape[0]"));
```
tsocha: Shouldn't be `>= 3` instead of `=>3`?

mbencer: Done

src/ngraph/op/experimental/layers/proposal.cpp
```diff
@@ -35,25 +35,67 @@ op::Proposal::Proposal(const std::shared_ptr<Node>& class_probs,
 
 void op::Proposal::validate_and_infer_types()
 {
-    // shape node should have integer data type. For now we only allow i64
-    auto image_shape_et = get_input_element_type(2);
+    const auto& class_prob_et = get_input_element_type(0);
+    const auto& class_logits_et = get_input_element_type(1);
+    const auto& image_shape_et = get_input_element_type(2);
+
+    NODE_VALIDATION_CHECK(this,
```
GlebKazantaev: As I understand we do not usually check type of input tensor, because it can be fp32 or f16 or smth else.

diyessi: I agree. Remove the these type checks.

mbencer: Done

PR #3420, opened 2019-08-09
Add support for operator sets and Softmax:1

postrational:


src/ngraph/node.cpp
```diff
@@ -298,6 +298,16 @@ void Node::set_friendly_name(const string& name)
     m_friendly_name = name;
 }
 
+size_t Node::get_opset_version() const
```
diyessi: Opset is a property of the node implementation, not its instantiation. This should be a virtual function.

src/ngraph/frontend/onnx_import/op/softmax.cpp
```diff
@@ -45,10 +45,7 @@ namespace ngraph
                         << "provided 'axis' value:" << axis
                         << " is out of input tensor dimensions range.";
 
-                    // create vector of capacity data_dimensions - axis_divider position
-                    std::vector<size_t> axes(data_shape.size() - axis);
-                    std::iota(std::begin(axes), std::end(axes), axis);
-                    return {std::make_shared<ngraph::op::Softmax>(data, axes)};
+                    return {std::make_shared<ngraph::op::set1::Softmax>(data, axis)};
```
postrational: Let's revert this change.

src/ngraph/node.hpp
```diff
@@ -430,6 +436,7 @@ namespace ngraph
         std::unordered_map<Node*, autodiff::Adjoints> m_adjoint_map;
         Placement m_placement = Placement::DEFAULT;
         size_t m_placement_index = placement_invalid;
+        size_t m_opset_version = 0;
```
aprocter: So is this the _earliest_, or the _latest_, opset version that this class belongs to? In the end I think this is the only issue I have that doesn't fall into the bucket of "implementation details". At a minimum, we probably need a way to represent that "this class represents `Reshape` as of opset versions `v3, v4, v5`".

src/ngraph/runtime/interpreter/int_executable.hpp
```diff
@@ -242,6 +242,16 @@ class ngraph::runtime::interpreter::INTExecutable : public Executable
     {
         const Node& node = *node_wrapper.get_node();
 
+        size_t opset_version = node.get_opset_version();
+        if (opset_version != 0)
```
aprocter: What happens if:

- INTERPRETER supports opset v5, so this check is changed to `!= 5`, but
- The op `Reshape` (for example) has not had its behavior changed since v4?

If the answer is "that's okay, because the class will have its version number bumped up to 5 even though the behavior has changed", what happens to things that only support v4? (Also, in that case, I think the checks in `Opset1Transformation` would become invalid.)

aprocter: BTW, let's use `NGRAPH_CHECK`.

src/ngraph/pass/opset1_transform.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+        auto replacement_node =
+            make_shared<op::set1::Softmax>(node->input(0).get_source_output(), axes.to_vector()[0]);
+        replace_node(node, replacement_node);
+        modified = true;
+        break;
+    }
+    default:
+    {
+        node->set_opset_version(1);
```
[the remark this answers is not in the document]
tomdol: I also thought about using the `friend` keyword and hiding the setter.

src/ngraph/pass/opset1_transform.cpp
```diff
@@ -0,0 +1,101 @@
...
+    size_t opset_version = node->get_opset_version();
+
+    if (opset_version == 1)
+    {
+        return modified;
+    }
+
+    if (opset_version != 0)
+    {
+        throw ngraph_error("Opset 1 transformation pass failed for " + node->get_name() +
```
tomdol: As discussed yesterday, I think we should have a unit test for this check to "guard" this behavior.

src/ngraph/op/softmax.cpp
```diff
@@ -91,3 +92,61 @@ void op::Softmax::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+}
+
+void op::set1::Softmax::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
+{
+    auto delta = deltas.at(0);
+
+    auto z = delta * shared_from_this();
+
+    std::vector<size_t> axes(get_shape().size() - m_axis);
+    std::iota(std::begin(axes), std::end(axes), m_axis);
```
tsocha: This algorithm looks little suspicious to me.
Do we have any unit test to verify it?

src/ngraph/op/softmax.hpp
```diff
@@ -53,5 +53,47 @@ namespace ngraph
...
+                Softmax()
+                    : m_axis(0)
+                {
+                    set_opset_version(1);
+                }
+                /// \brief Constructs a softmax operation.
+                ///
+                /// \param arg Node that produces the first input tensor.<br>
+                /// `[d0, ...]`
+                /// \param axis The axis positions (0-based) on which to calculate the softmax.
```
tsocha: `The axis position` instead of `The axis positions` sounds better IMO.
But I won't force this change.

src/ngraph/pass/opset1_transform.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    switch (get_typeid(node))
+    {
+    case OP_TYPEID::Softmax:
+    {
+        auto tmp = dynamic_cast<const op::set0::Softmax*>(node.get());
+        AxisSet axes = tmp->get_axes();
+
+        if (axes.size() != 1)
```
tsocha: Should we use `0` as default axis if `axes` is empty?

src/ngraph/pass/opset1_transform.cpp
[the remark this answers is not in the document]
mbencer: I am not sure about using `friend` mechanism in this case.
In order to have access to `set_opset_version` from level of `Opset1Transformation::run_on_node` it will be necessary to declare `Opset1Transformation::run_on_node` as `friend` in `Node` class. I think that adding `friend` for all opset transformation will be not good solution. 
What's more declaration `run_on_node' as `friend` in `Node` requires that `Opset1Transformation::run_on_node` is declared earlier. I am not sure how to ensure it (class forward declaration is not enough in this case).
One solution can be declaration **class** `Opset1Transformation` as `friend` in `Node`. Cons: access to all private members of `Node` and necessity to declare `friend` to all opset transformations (we can not declare `NodePass` as `friend class` because it does not work for inheritance).
Second: We can implement **function** with `friend` privileges to `Node` which will be changing opset version in other file which will be included only in transform `.cpp` files.
What do you think about it?

As I know declaring  **virtual** 'run_on_node' from

src/ngraph/frontend/onnx_import/op/softmax.cpp
mbencer: Done

src/ngraph/op/softmax.hpp
```diff
@@ -53,5 +53,47 @@ namespace ngraph
         private:
             AxisSet m_axes;
         };
+
+        namespace set0
```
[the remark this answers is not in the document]
mbencer: Done

src/ngraph/op/softmax.hpp
```diff
@@ -53,5 +53,47 @@ namespace ngraph
...
+        {
+            using ngraph::op::Softmax;
```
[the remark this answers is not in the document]
mbencer: I think it is good idea. Done.
1,2. It is interesting solution. I introduced `dev` namespace which in this case is the `v1` version. For "latest" I've not declared new namepsace. What do you think about it?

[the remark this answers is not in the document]
aprocter: I think that sounds good!

src/ngraph/pass/opset1_transform.cpp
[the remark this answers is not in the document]
aprocter: I suppose we could do something like this...

```
template <int VERSION>
friend class OpsetTransformation;
```

In any case, I think if we deal with the more fundamental issue (that a single opset version field attached to a node may not be sufficient), we may be able to avoid the issue altogether. I was chatting with @diyessi about this at our usual 1:1 today, and I think he has some good ideas here so let's ask him to chime in.

src/ngraph/op/softmax.cpp
mbencer: Can you clarify your doubts?
The place where `generate_adjoints` can be verified is `ngraph/test/backend/autodiff.in.cpp`, but any backend has not supported `v1` opset version yet.
Look at:
```
size_t opset_version = node.get_opset_version();
        if (opset_version != 0)
```
in `int_executable.hpp `
Have you idea to test it?

src/ngraph/pass/opset1_transform.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "ngraph/op/softmax.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+// @TODO: Shouldn't this be moved to a common utility class? This mapping to OP_TYPEID gets repeated many times.
+#define NGRAPH_OP(a, b) a,
+enum class OP_TYPEID
+{
+#include "ngraph/op/op_tbl.hpp"
```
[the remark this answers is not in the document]
mbencer: Do you think that we should have something like that:
```
#define NGRAPH_OP(a, b) a,
enum class OP_TYPEID
{
#include "ngraph/op/op_tbl.hpp"
};
#undef NGRAPH_OP

#define NGRAPH_OP(a, b) {#a, OP_TYPEID::a},
static unordered_map<string, OP_TYPEID> typeid_map{
#include "ngraph/op/op_tbl.hpp"
};
#undef NGRAPH_OP

#define NGRAPH_OP(a, b) a,
enum class OP_TYPEID_v0
{
#include "ngraph/op/op_tbl_v0.hpp"
};
#undef NGRAPH_OP

#define NGRAPH_OP(a, b) {#a, OP_TYPEID::a},
static unordered_map<string, OP_TYPEID> typeid_map_v0{
#include "ngraph/op/op_tbl_v0.hpp"
};
#undef NGRAPH_OP

#define NGRAPH_OP(a, b) a,
enum class OP_TYPEID_v1
{
#include "ngraph/op/op_tbl_v1.hpp"
};
#undef NGRAPH_OP

#define NGRAPH_OP(a, b) {#a, OP_TYPEID::a},
static unordered_map<string, OP_TYPEID> typeid_map_v1{
#include "ngraph/op/op_tbl_v1.hpp"
};
#undef NGRAPH_OP
```
and in `op_tbl_v0`, `op_tbl_v1` etc. should be list of supported ops in particular version?
What do you think about @postrational comment about moving `OP_TYPEID`, `typeid_map` to common file?

src/ngraph/runtime/interpreter/int_executable.hpp
mbencer: `NGRAPH_CHECK` was added.
I think that the question about offset support is key.
I understood that we update version of op: for instance from `v0` to `v1` only when specification has been changed. In such case I agree with you that the condition is incorrect.  
I aware that it is not very flexible but maybe each op in each version should have list of supported backends (or unsupported)?

src/ngraph/pass/opset1_transform.cpp
mbencer: According to specification default value of axis is `1`.
What do you think @postrational @aprocter @diyessi ? Can we use default value of `axis` when `AxisSet` is empty?

src/ngraph/op/softmax.hpp
mbencer: Done

src/ngraph/pass/opset1_transform.cpp
mbencer: The unit test was introduced

src/ngraph/pass/opset1_transform.cpp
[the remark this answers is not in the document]
aprocter: Yup, something more or less like that. Keep in mind that with the exception of upgrade passes, most things will not be playing with multiple versions.

I don't have a strong opinion on moving OP_TYPEID, etc., but also don't know much about how they are being used. @rkimballn1 should weigh in.

src/ngraph/pass/opset1_transform.cpp
aprocter: For conversion, I don't think so. The semantics is different.

tsocha: In `set0: // empty axes == all axes`.
So we can't convert it, can we?

mbencer: Done

src/ngraph/serializer.cpp
```diff
@@ -1832,8 +1835,16 @@ shared_ptr<Node> JSONDeserializer::deserialize_node(json node_js)
         }
         case OP_TYPEID::Softmax:
         {
-            auto softmax_axes = deserialize_axis_set(node_js.at("softmax_axes"));
-            node = make_shared<op::Softmax>(args[0], softmax_axes);
+            if (opset_version == 0)
```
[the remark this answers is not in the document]
tomdol: I was trying to figure this out with @mbencer today but we're still a bit confused. Do you mean that we should serialize Softmax in version 1 only and not serialize if a version 0 comes in?

Would the expectation here be that an opset upgrade pass is executed before the serialization happens?

src/ngraph/runtime/interpreter/int_executable.hpp
```diff
@@ -242,6 +242,19 @@ class ngraph::runtime::interpreter::INTExecutable : public Executable
     {
         const Node& node = *node_wrapper.get_node();
 
+        size_t opset_version = node.get_opset_version();
```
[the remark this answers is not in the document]
mbencer: In current version is used `op_version` instead of `opset_version` (`v0`, `v1` etc.) to determine if current backend supports the op.

 Can such version be merged to master? What is the status of @diyessi proposition? (https://github.com/NervanaSystems/ngraph/pull/3533/files)

src/ngraph/runtime/interpreter/int_executable.hpp
```diff
@@ -242,6 +242,19 @@ class ngraph::runtime::interpreter::INTExecutable : public Executable
...
+        bool is_opset_version_supported = opset_version == 0;
+        NGRAPH_CHECK(
+            is_opset_version_supported,
+            static_cast<std::ostringstream&>(
```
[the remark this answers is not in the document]
mbencer: Done

src/ngraph/op/softmax.cpp
mbencer: @diyessi What do you think about this part of algorithm? Have you maybe idea how to test `generate_adjoints`? ;)

src/ngraph/op/softmax.hpp
```diff
@@ -22,36 +22,79 @@ namespace ngraph
...
+            private:
+                size_t m_axis;
+            };
+        }
 
-        private:
-            AxisSet m_axes;
-        };
+        // default opset version
+        using v0::Softmax;
```
mbencer: @aprocter Can we make `op::v0::Softmax` visible in namespace `op`?

src/ngraph/pass/opset1_transform.cpp
```diff
@@ -0,0 +1,94 @@
...
+    switch (get_typeid(node))
+    {
+    case OP_TYPEID::Softmax:
+    {
+        auto tmp = dynamic_cast<const op::v0::Softmax*>(node.get());
+        AxisSet axes = tmp->get_axes();
+
+        NGRAPH_CHECK(
+            axes.size() == 1,
+            "Unable to convert Softmax:0 to Softmax:1 with zero or more than one axis. Node: ",
```
slyalin: This is not complete transformation if we react this way. To have reliable and complete transformation the result should be Transpose(to gather dimensions in case if axis set is sparse) -> Reshape(to flatten several axis to one) -> v1::Softmax -> Reshape -> Transpose.

src/ngraph/serializer.cpp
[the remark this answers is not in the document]
tomdol: Thanks for the clarification, I undestand your concerns now. They will need to be addressed eventually but I agree that it should be done separately.

src/ngraph/op/softmax.cpp
mbencer: Ok, thanks ;) @aprocter Should we throw `Not supported yet` exception in such case or is it acceptable to keep current version?

src/ngraph/op/softmax.hpp
```diff
@@ -22,36 +22,79 @@ namespace ngraph
...
+                                               const NodeVector& deltas) override;
+
+            private:
+                AxisSet m_axes;
+            };
+        }
+
+        namespace v1
+        {
+            class Softmax : public util::UnaryElementwiseArithmetic
```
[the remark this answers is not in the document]
mbencer: Done. I changed the base class to `Op`

src/ngraph/op/softmax.hpp
mbencer: From technical point of view no, but when we will have more `opset` supported by backends it can be probably confused to decide which version should be visible in `op` namespace

src/ngraph/pass/opset1_transform.cpp
```diff
@@ -0,0 +1,94 @@
...
+    bool modified = false;
+
+    size_t op_version = node->get_op_version();
+
+    if (op_version == 1)
+    {
+        return modified;
+    }
+
+    if (op_version != 0)
```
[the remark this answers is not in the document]
mbencer: Done

src/ngraph/pass/opset1_transform.cpp
slyalin: Actually it should be even more complicated if we consider the ability to be reshaped for a different shape. In this case we cannot deduce target shape for Reshape op as a constant. Instead it should be calculated as a sub-graph that depends on ShapeOf(input_tensor) and contains such operations as Split or StridedSlice, Multiply and Concat. And only then this target shape is consumed as the second input into Reshape.

mbencer: @slyalin, @aprocter I agree with @postrational The support for that kind of transformation has not been implemented yet so we want to add it in separate PR

aprocter: @postrational @mbencer Agreed---let's do it that way. For example, #3536 is waiting on this.

src/ngraph/pass/opset1_transform.cpp
```diff
@@ -68,6 +68,11 @@ bool pass::Opset1Transformation::run_on_node(shared_ptr<Node> node)
                  op_version,
                  " found.");
 
+// Not all enumeration values not explicitly handled in switch
```
postrational: Double negation in comment. Change to:
`// Not all enumeration values explicitly handled in switch`

src/ngraph/op/softmax.cpp
```diff
@@ -90,3 +97,70 @@ void op::Softmax::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+
+shared_ptr<Node> op::v1::Softmax::copy_with_new_args(const NodeVector& new_args) const
+{
+    check_new_args_count(this, new_args);
+    return make_shared<op::v1::Softmax>(new_args.at(0), m_axis);
+}
+
+void op::v1::Softmax::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
+{
+    throw ngraph_error("NYI");
```
postrational: I am opposed to any abbreviations. Please remove this New York Institute and replace with a meaningful sentence, such as: `op::v1::Softmax::generate_adjoints function is not implemented yet.`

mbencer: Done

src/ngraph/pass/opset1_transform.cpp
mbencer: Done

src/ngraph/pass/opset1_transform.cpp
slyalin: That's why v1::Reshape should be implemented first, then Softmax. There is a natural order in the new ops implementation process. But the tricky part is that we are discovering this order just in time when a particular op is being implemented. OK, let's postpone it right up to the moment when v1::Reshape is ready.

aprocter: Right- a lot of "oh, I see..." moments to come. Since subsequent PRs will not be blocking the underlying versioning infrastructure (in this case, Softmax was just chosen as a case study to try it out), doing things in a natural order should be less of a problem going forwards.

python/pyngraph/ops/softmax.cpp
```diff
@@ -24,10 +24,8 @@ namespace py = pybind11;
 
 void regclass_pyngraph_op_Softmax(py::module m)
 {
-    py::class_<ngraph::op::Softmax,
-               std::shared_ptr<ngraph::op::Softmax>,
-               ngraph::op::util::UnaryElementwiseArithmetic>
-        softmax(m, "Softmax");
+    py::class_<ngraph::op::Softmax, std::shared_ptr<ngraph::op::Softmax>, ngraph::op::Op> softmax(
+        m, "Softmax");
```
nmostafa: Format is correct here ?

src/ngraph/op/softmax.cpp
```diff
@@ -28,25 +28,32 @@
...
+                              axis < static_cast<size_t>(input_shape.rank()),
                               "Reduction axis (",
                               axis,
                               ") is out of bounds (argument shape: ",
-                              get_shape(),
+                              input_shape,
                               ").");
     }
 
+    if (input_shape.is_static())
```
nmostafa: nit: braces around single stmt bodies.

src/ngraph/op/softmax.cpp
```diff
@@ -28,25 +28,32 @@
...
     , m_axes(axes)
 {
     constructor_validate_and_infer_types();
 
+    const PartialShape& input_shape = get_input_partial_shape(0);
     for (auto axis : m_axes)
     {
         NODE_VALIDATION_CHECK(this,
-                              axis < get_shape().size(),
+                              axis < static_cast<size_t>(input_shape.rank()),
```
nmostafa: what about if rank is dynamic ?

src/ngraph/op/softmax.cpp
nmostafa: Same for V1 below

src/ngraph/op/softmax.cpp
```diff
@@ -90,3 +97,70 @@ void op::Softmax::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
     auto x = input_value(0);
     adjoints.add_delta(x, adjoint);
 }
+
+// *** SOFTMAX OP SET V1 ***
+const string op::v1::Softmax::type_name{"Softmax"};
```
nmostafa: Same `type_name` value for both versions ? We will possibly do something like  `op->type_name == op::Softmax::v0::type_name` to check for v0. Now we have to add a check for op version `&& op->get_op_version() == 0`

src/ngraph/pass/opset1_upgrade.hpp
```diff
@@ -0,0 +1,39 @@
...
+    namespace pass
+    {
+        class Opset1Upgrade : public NodePass
+        {
+        public:
+            ///
+            /// \brief    Constructor for the Opset 1 transformation pass.
+            ///
+            /// \details  This transformation pass iterates over all nodes in a graph and updates
+            /// opset version 0 ops
```
nmostafa: nit: clang-format probably messed up format here. Maybe break up the lines yourself.

src/ngraph/node.hpp
```diff
@@ -379,6 +379,8 @@ namespace ngraph
         /// Get all the nodes that uses the current node
         NodeVector get_users(bool check_is_used = false) const;
 
+        /// \return Version of this node
+        virtual size_t get_op_version() const { return 0; }
```
diyessi: `get_version` should be good enough since it's a method on the "op".

src/ngraph/op/softmax.cpp
diyessi: See #3533

python/pyngraph/ops/softmax.cpp
mbencer: I agree with you that previous version was more readable but clang forces such format ;)

src/ngraph/op/softmax.cpp
mbencer: Done

src/ngraph/pass/opset1_upgrade.hpp
mbencer: Fixed

src/ngraph/node.hpp
mbencer: Name was changed ;)

src/ngraph/op/softmax.cpp
mbencer: As I know `rank` method supports `dynamic` case

tsocha: If rank would be dynamic then static_cast will throw an exception.

mbencer: @tsocha Thanks, you are right, I added `NODE_VALIDATION_CHECK` to provide it

src/ngraph/op/softmax.cpp
```diff
@@ -28,24 +28,39 @@
...
-                              axis < get_shape().size(),
+                              axis >= 0 && axis < static_cast<size_t>(input_shape.rank()),
                               "Reduction axis (",
                               axis,
                               ") is out of bounds (argument shape: ",
-                              get_shape(),
+                              input_shape,
                               ").");
     }
+    if (input_shape.is_static())
```
aprocter: Not sure if I missed this on first review (or if the original code was just broken), but we need to allow the dynamic case here. Also, these checks belong in `validate_and_infer_types()`, not the ctor.

@mbencer If you like, please feel free to assign a JIRA to me for these issues. I'm working on some written guidelines for dynamic shapes; in the meantime it sometimes is faster if I just fix these things. :)

src/ngraph/op/softmax.cpp
```diff
@@ -28,24 +28,39 @@
...
-    : UnaryElementwiseArithmetic(arg)
+op::v0::Softmax::Softmax(const Output<Node>& arg, const AxisSet& axes)
+    : Op({arg})
     , m_axes(axes)
 {
     constructor_validate_and_infer_types();
 
+    const PartialShape& input_shape = get_input_partial_shape(0);
+    NODE_VALIDATION_CHECK(this,
+                          input_shape.rank().is_static(),
```
aprocter: Not sure if I missed this on first review (or if the original code was just broken), but we need to allow dynamic rank here. Also, these checks belong in `validate_and_infer_types()`, not the ctor.

@mbencer If you like, please feel free to assign a JIRA to me for these issues. I'm working on some written guidelines for dynamic shapes; in the meantime it sometimes is faster if I just fix these things. :)

PR #3421, opened 2019-08-09
Parameter attribute getters

diyessi:


src/ngraph/op/parameter.hpp
```diff
@@ -57,6 +58,19 @@ namespace ngraph
...
             void set_is_relevant_to_shapes(bool is_relevant);
 
+            const PartialShape& get_partial_shape() const { return m_partial_shape; }
+            PartialShape& get_partial_shape() { return m_partial_shape; }
+            void set_partial_shape(const PartialShape& partial_shape)
+            {
+                m_partial_shape = partial_shape;
+            }
+
+            const element::Type& get_element_type() const { return m_element_type; }
```
aprocter: I don't suppose it'll hurt anything, but this shadows `Node::get_element_type()`. But I guess that will eventually be going away.

src/ngraph/op/parameter.hpp
```diff
@@ -38,6 +38,7 @@ namespace ngraph
             NGRAPH_API
             static const std::string type_name;
             const std::string& description() const override { return type_name; }
+            /// \brief Constructions a tensor-typed parameter node.
```
aprocter: I don't think we need the phrase `tensor-typed` anymore.

PR #3423, opened 2019-08-09
[MLIR] Set optimization level for LLVM optimizer and codegen

dcaballe:
Now both LLVM optimizer and codegen are aligned with
"NGRAPH_MLIR_OPT_LEVEL" macro.

### DIFF
--- a/src/contrib/mlir/compiler.cpp
+++ b/src/contrib/mlir/compiler.cpp
@@ -107,6 +107,9 @@ static llvm::cl::opt<unsigned> clLoopTilingCacheSize(
 #define COMPILE_OP_DECL(op_name)                                                                   \
     create_op<op_name>(MLIRCompiler & compiler, const ngraph::Node* ng_node)
 
+// Default optimization level.
+unsigned MLIRCompiler::mlir_opt_level = 2;
+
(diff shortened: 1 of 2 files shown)


src/contrib/mlir/compiler.cpp
```diff
@@ -107,6 +107,9 @@ static llvm::cl::opt<unsigned> clLoopTilingCacheSize(
 #define COMPILE_OP_DECL(op_name)                                                                   \
     create_op<op_name>(MLIRCompiler & compiler, const ngraph::Node* ng_node)
 
+// Default optimization level.
+unsigned MLIRCompiler::mlir_opt_level = 2;
```
nmostafa: nit: Better use LLVM enum values ? `llvm::CodeGenOpt::Level::Default`  Here and throughout the change.

dcaballe: The problem is that llvm optimizer uses an unsigned, MLIR uses unsigned and cg uses this enum so there is no good one for all. I think it's better to keep the unsigned since MLIR APIs expect the same and there are more instances of those. Otherwise we'll have be doing castings every time.

PR #3430, opened 2019-08-13
[POC][WIP] Add ngraph::op::Merge that handles control flow merge of tensor value.

silee2:
This OP is to enable limited control flow support for certain if-else.
If the "if" and "else" block have same shape output(s), we execute both blocks and later use this new Merge OP to pick one of the output. This won't be great for performance but does not require any update to any existing passes to be control flow aware.
A scalar boolean input controls which value gets picked.


silee2:
"Merge" May not be the best name since the OP is actually like a "Mux"  But still not a general Mux since the input is limited to two. "Select" is another option but usually "Select" per element selection, not coarse grain mux.

silee2:
Code is fully functional but the PR is more like a POC for review. So I tagged it as WIP.

src/ngraph/op/merge.hpp
```diff
@@ -0,0 +1,41 @@
...
+    namespace op
+    {
+        class Merge : public Op
+        {
+        public:
+            NGRAPH_API
+            static const std::string type_name;
+            const std::string& description() const override { return type_name; }
+            Merge() = default;
+            Merge(const Output<Node>& cond, const Output<Node>& tval, const Output<Node>& fval);
```
nmostafa: We need a description here. How is this op different from `op::Select` ?

src/ngraph/op/merge.cpp
```diff
@@ -0,0 +1,67 @@
...
+        "Arguments do not have the same element type (tval element type: ",
+        get_input_element_type(1),
+        ", fval element type: ",
+        get_input_element_type(2),
+        ").");
+
+    NODE_VALIDATION_CHECK(
+        this, tval_shape.compatible(fval_shape), "tval and fval must have compatible shape");
+
+    PartialShape output_shape{tval_shape};
```
aprocter: I think rather than copying `tval_shape` we'll want to somehow combine info from both `tval_shape` and `fval_shape`. There are two cases I can think of, and we might pick either one:

Case 1. We require that `tval` and `fval` have the same shape. Then we can just use the existing "merge" function.

Case 2. We do _not_ require that `tval` and `fval` have the same shape. Then we would push forward a "least common denominator" shape. Can't remember if we already have a util function to compute this, but for example:

```
t:{2,3,3}, f:{2,3,3} -> out:{2,3,3}
t:{2,3,3}, f:{2,4,3} -> out:{2,?,3}
t:{2,3,3}, f:{2,?,3} -> out:{2,?,3}
t:{2,3,3}, f:{2,3} -> out:?
t:?, f:{2,3} -> out:?
```

I think it is probably the _second_ thing that we want. Option 1 winds up imposing restrictions on the runtime shape of one value (say, the t value if the cond evaluates to true) based on static information about the hypothetical runtime shape of a second value (the f value, if the cond evaluates to true) that is never actually computed, and which might not even actually exist. That seems like a weird (and in any case unnecessary) restriction.

src/ngraph/op/op_tbl.hpp
```diff
@@ -106,6 +106,7 @@ NGRAPH_OP(Max, ngraph::op)
 NGRAPH_OP(Maximum, ngraph::op)
 NGRAPH_OP(MaxPool, ngraph::op)
 NGRAPH_OP(MaxPoolBackprop, ngraph::op)
+NGRAPH_OP(Merge, ngraph::op)
```
dmyershov: Gerrit checks for INTELGPU backend are switched off. To be able to build INTELGPU backend w/o switch-enum error you need to add "case OP_TYPEID::Merge:" line into switch from src/ngraph/runtime/intelgpu/intelgpu_backend.cpp (between 2050 - 2106 lines)

PR #3431, opened 2019-08-13
GeluGrad24

diyessi:


test/backend/gelu_grad.in.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    std::transform(
+        input.begin(), input.end(), input.begin(), [](double x) -> double { return gelu_grad(x); });
+
+    auto handle = backend->compile(f);
+    handle->call_with_validate({result}, {a});
+    EXPECT_TRUE(test::all_close_f(input, read_vector<double>(result)));
+}
+
+NGRAPH_TEST(${BACKEND_NAME}, backwards_gelu_f32)
```
aprocter: I think these tests are probably better placed in `gelu.in.cpp`. (Not all ops have a corresponding `Grad` op.)

src/ngraph/op/fused/gelu_grad.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "ngraph/node.hpp"
+#include "ngraph/op/op.hpp"
+#include "ngraph/op/util/fused_op.hpp"
+
+namespace ngraph
+{
+    namespace op
+    {
+        class GeluGrad : public util::FusedOp
```
aprocter: Should we call ours `GeluBackprop`? This is more consistent with our current naming scheme. (Though actually if I had to start over I might rename everything else to `Grad`.)

silee2:
Don't forget to update the various `unit_test.manifest` files for not yet supported backends.

src/ngraph/op/fused/gelu_grad.hpp
silee2: I also like `Grad` better but we are already using `Backprop` in many places already (I guess thanks to how mkl-dnn names their primitives.)

src/ngraph/op/fused/gelu_grad.cpp
```diff
@@ -0,0 +1,75 @@
+//*****************************************************************************
```
silee2: Do we need separate gelu_grad.cpp and gelu_grad.hpp? Some Gelu is already a fused op, gelu and gelu_grad can be merged together

src/ngraph/op/fused/gelu.hpp
```diff
@@ -43,6 +43,9 @@ namespace ngraph
 
             virtual std::shared_ptr<Node>
                 copy_with_new_args(const NodeVector& new_args) const override;
+
+        protected:
+            void generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas) override;
```
swapna-intel: In the parent class, I see this. Is it ok to override a 'throw' ing unimplemented function? :  
```
void op::util::FusedOp::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
{
    // TODO
    throw ngraph_error("Autodiff on fused ops not supported yet");
}
```

test/backend/gelu_grad.in.cpp
diyessi: They could all go in the gelu file, but I'd rather do that for the master version.

src/ngraph/op/fused/gelu_grad.cpp
diyessi: The coding standards actually say they shouldn't be combined, but I think that would be reasonable.

silee2: I meant combining gelu_grad.cpp and gelu.cpp. And combining gelu_grad.hpp and gelu.hpp.

src/ngraph/op/fused/gelu_grad.hpp
swapna-intel: hmmm.. which one is more ngraph - user friendly? I like backprop, since it is more high level than grad, but is that what everyone 'gets' easily?

src/ngraph/op/fused/gelu.hpp
diyessi: Yes, particularly since it is implementing it.

swapna-intel: Then maybe the throw comment is misleading?

diyessi:
No earlier than 0.25

PR #3434, opened 2019-08-14
[MLIR] Disable support for boolean type in MLIR

dcaballe:
Ops on boolean types are currently not supported by MLIR Compiler until
the proper i1<->i8 type conversion mechanism is in place.


src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
```diff
@@ -266,6 +266,12 @@ bool MLIRSubgraphExtractionPass::run_on_function(std::shared_ptr<Function> func)
 
 bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node)
 {
+    // Disable any op using boolean type until we have support for i1<->i8 conversion in MLIR.
```
nmostafa: Could add that example you mentioned:
```
%0 = icmp %a, %b : i1
store %0, %c[arg1] : i8
```

PR #3435, opened 2019-08-14
[MLIR] Fix cycle detection during sub-graph construction

nmostafa:
Initial implementation wasn't catching a cycle in `backwards_softmax_*`. Before merging sub-graphs of predecessors of a node, we walk backwards along each path and check if existed the set of to-be-merged sub-graphs and entered again. It caps at 100 path length before bailing out on the merge.


src/contrib/mlir/pass/mlir_subgraph_extraction.hpp
```diff
@@ -115,7 +115,10 @@ namespace ngraph
             /// D
             /// we want to add D to sub-graph 1. C is an input to D. sugraph_ids are 1
             /// we traverse backwards C->A(1) and find 1, then we cannot add D since we will form a cycle
-            bool check_cycles(NodeVector& inputs, std::unordered_set<int>& subgraph_ids);
+            bool check_cycles(std::shared_ptr<Node> node,
+                                              std::unordered_set<int>& subgraph_ids,
+                                              bool inside_subgraphs = true,
+                                              unsigned depth = 0);
```
dcaballe: formatting?

nmostafa:
That would be nice. That property we are looking for is that no output of any of the set of subgraphs we want to merge reaches an external node that reaches back into another subgraph in the set. I don't see how that can be checked with any (post-)dominance property.

PR #3438, opened 2019-08-14
Add method to check if an Executable can create tensors

rkimballn1:


src/ngraph/runtime/backend.cpp
```diff
@@ -151,3 +151,20 @@ bool runtime::Backend::set_config(const map<string, string>& config, string& err
...
+
+bool runtime::Backend::executable_can_create_tensors()
+{
+    auto A = make_shared<op::Parameter>(element::f32, Shape());
+    auto function = make_shared<Function>(NodeVector{A}, ParameterVector{A});
+    auto exec = compile(function);
+    bool exec_can_create_tensors = false;
+    try
+    {
+        auto t = exec->create_input_tensor(0);
```
sayantan-nervana: can we add:

```
auto t0 = exec->create_input_tensor(0);
auto t1 = exec->create_input_tensor(0, 2);
auto t2 = exec->create_output_tensor(0);
auto t3 = exec->create_output_tensor(0, 2);
```

Then we explicitly check that all 4 variants are implemented instead before giving the green signal

rkimballn1: Done

PR #3442, opened 2019-08-14
[MLIR] Disable CPU fusion tests in MLIR

dcaballe:
CPU fusion is disabled in MLIR since fused ops are not supported in
nGraph dialect.


dcaballe:
> why only these few tests are disabled?

Because those are the only fusion ones failing for now with MLIR enabled. The rest seems to be passing.

dcaballe:
Added `cpu_debug_tracer, check_flow_with_external_function` as well since it's failing because there CompiledKernel ops generated for MLIR are not expected

PR #3443, opened 2019-08-14
[MLIR] Add unary ops - abs, sign & not

nbpatel:


src/contrib/mlir/lowerer.cpp
```diff
@@ -858,6 +876,51 @@ namespace
...
+                    iRes(ivs) = edsc::intrinsics::select((val < zero), zero - val, val);
+                }
+                else if (auto intTy = elemTy.dyn_cast<IntegerType>())
+                {
+                    ValueHandle zero = intrinsics::constant_int(0, intTy.getWidth());
+                    iRes(ivs) = edsc::intrinsics::select((val < zero), zero - val, val);
+                }
+                else
+                {
+                    NGRAPH_CHECK(false, "Unsupported type for Abs");
```
dcaballe: nit: we can use NGRAPH_UNREACHABLE instead when the condition is always false

src/contrib/mlir/lowerer.cpp
```diff
@@ -858,6 +876,51 @@ namespace
                     NGRAPH_CHECK(false, "Unsupported type for Negative");
                 }
             }
+            else if (isa<NGNotOp>(op))
+            {
+                iRes(ivs) = !val;
+            }
+            else if (isa<NGAbsOp>(op))
```
dcaballe: Could we try to create a function overloaded for each op so that it's more like a visitor implementation? Something like:
```
void generate_op_edsc(NGNotOp *op, ...) {}
void generate_op_edsc(NGAbsOp *op, ...) {}
... 
```
Would that work?

src/contrib/mlir/lowerer.cpp
```diff
@@ -858,6 +876,51 @@ namespace
...
+            {
+                if (auto floatTy = elemTy.dyn_cast<FloatType>())
+                {
+                    ValueHandle zero = intrinsics::constant_float(llvm::APFloat(0.0f), floatTy);
+                    iRes(ivs) = edsc::intrinsics::select((val < zero), zero - val, val);
```
dcaballe: Abs for floating point would also need to call the math function counterpart to do the right thing for special values (NaN, 0.0, etc.). I guess it's fine for now but please add a note about it so that it's addressed with the other math ops.

src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
```diff
@@ -344,7 +347,12 @@ bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node
         }
     }
 
-    if (TI(ngraph::op::Negative) == TI(*node))
+    if (TI(ngraph::op::Negative) == TI(*node) || TI(ngraph::op::Not) == TI(*node))
```
dcaballe: This code shouldn't be necessary. We are returning true by default at 360

PR #3444, opened 2019-08-14
GeluBackpropFactor for 0.25

diyessi:


src/ngraph/op/fused/gelu.cpp
```diff
@@ -74,3 +77,58 @@ void op::Gelu::pre_validate_and_infer_types()
...
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::GeluBackpropFactor::pre_validate_and_infer_types()
+{
+    element::Type input_element_type = get_input_element_type(0);
+
+    NODE_VALIDATION_CHECK(this,
+                          input_element_type.is_dynamic() || input_element_type == element::f32 ||
```
silee2: You can use `input_element_type.is_real()` instead of comparing agains f32, f64, f16 and bf16. Same can be done for `Gelu` as well

PR #3446, opened 2019-08-14
Copy outputs only if the use is outside the sub-graph

nmostafa:
We were copying edges of all output nodes of a sub-graph regardless of where the dst node lies. This fixes this by checking if the node is out of the sub-graph. 
Fixes:
```
638-[ FAILED ] CPU.multiple_result (8 ms)
900-[ FAILED ] CPU.lstm_cell_no_bias_no_peepholes (1077 ms)
904-[ FAILED ] CPU.lstm_cell_bias_peepholes (1047 ms)
908-[ FAILED ] CPU.lstm_cell_bias_peepholes_clip_input_forget (890 ms)
912-[ FAILED ] CPU.lstm_cell_activaction_functions (892 ms)
```


src/contrib/mlir/pass/mlir_subgraph_extraction.cpp
```diff
@@ -254,7 +254,10 @@ bool MLIRSubgraphExtractionPass::run_on_function(std::shared_ptr<Function> func)
 
             for (descriptor::Input* in_desc : input_descs)
             {
-                in_desc->replace_output(ck, i);
+                if (std::find(nodes.begin(), nodes.end(), in_desc->get_node()) == nodes.end())
```
dcaballe: I think `nodes` is an `unordered_set`. Could you please use nodes.find instead to use the hash implementation of find?

PR #3448, opened 2019-08-15
Dynamic "flatten" builder

aprocter:
Builder for ONNX-style "flatten-to-matrix" with dynamic axis selection.


baojun-nervana:
Thanks, works for Paddle well!

aprocter:
I have added to this PR a temporary workaround for another issue @baojun-nervana has observed with shape specialization (basically, it wasn't completely doing its job).

PR #3453, opened 2019-08-16
[MLIR] Use f64 APFloat

nmostafa:
Fixes:
```
CPU.gelu_f64
CPU.gelu_backprop_factor_f64
```


src/contrib/mlir/lowerer.cpp
```diff
@@ -1029,6 +1007,27 @@ namespace
...
+        }
+        else if (auto intTy = type.dyn_cast<IntegerType>())
+        {
+            return intrinsics::constant_int(0, intTy.getWidth());
+        }
+        else
+        {
+            NGRAPH_CHECK(false, "Unsupported type");
+        }
+    }
```
dcaballe: Nits:
- Please add { } for if-else statements based on coding standards.
- We can use NGRAPH_UNREACHABLE instead of NGRAPH_CHECK(false) 
- No need of `else` after `return`

src/contrib/mlir/lowerer.cpp
```diff
@@ -1029,6 +1007,27 @@ namespace
...
     }
+
+    ValueHandle createZeroConstant(mlir::Type type)
+    {
+        if (auto floatTy = type.dyn_cast<FloatType>())
+        {
+            if (floatTy.isF32())
+                return intrinsics::constant_float(llvm::APFloat(0.0f), floatTy);
+            else if (floatTy.isF64())
+                return intrinsics::constant_float(llvm::APFloat(0.0), floatTy);
```
dcaballe: What was the issue? We were creating a single precision constant for a double precision? Is APFloat constructor overloaded for float and double? Interesting...

nmostafa: Yes

PR #3454, opened 2019-08-16
[Spec] Implement support for axes input of LRN op in reference implementation

mbencer:
Expected results of unit tests were calculated using Python script. Here is version of LRN calculation across hw axes: 
```
import copy
import numpy as np
  
def LRN(input, size=3, bias=1.0, alpha=3.0, beta=0.5):
    output = copy.deepcopy(input)
    N = input.shape[0]
    C = input.shape[1]
    H = input.shape[2]
    W = input.shape[3]
    for n in range(N):
        for c in range(C):
            for h in range(H):
                    begin_h = max(0, h - (size-1)/2)
                    end_h = min(H, h + (size-1)/2 + 1)
                    for w in range(W):
                        begin_w = max(0, w - (size-1)/2)
                        end_w = min(W, w + (size-1)/2 + 1)
                        patch = input[n, c, begin_h:end_h, begin_w:end_w]
                        output[n, c, h, w] /= (
                            np.power(bias + (alpha/size) * np.sum(patch * patch), beta))
    return output

input = np.arange(0, 12, 1).reshape(2, 3, 2, 1).astype(np.float32)
result = LRN(input)
for elem in np.nditer(result):
    print(str(round(elem, 7)) + "f, ")
```
For other axes implementation is analogical.


src/ngraph/op/lrn.cpp
```diff
@@ -24,7 +25,7 @@ using namespace ngraph;
 const string op::LRN::type_name{"LRN"};
 
 op::LRN::LRN(const Output<Node>& arg, double alpha, double beta, double bias, size_t size)
-    : LRN(arg, op::Constant::create(element::i32, Shape{1}, {1}), alpha, beta, bias, size)
+    : LRN(arg, op::Constant::create(element::u64, Shape{1}, {1}), alpha, beta, bias, size)
```
aprocter: Can we make this `i64` instead? That's what we've generally standardized on for shape-like tensors, including those that represent axis sets (even though the internal representation is signed... it's complicated :))

src/ngraph/op/lrn.cpp
```diff
@@ -43,6 +44,15 @@ op::LRN::LRN(const Output<Node>& arg,
     constructor_validate_and_infer_types();
 }
 
+AxisSet op::LRN::get_reduction_axes() const
+{
+    auto axes_node = input(1).get_source_output().get_node_shared_ptr();
+    auto axes_constant = dynamic_pointer_cast<op::Constant>(axes_node);
+    auto axes_vector = axes_constant->get_vector<size_t>();
```
aprocter: I think there is some method like `get_axis_set_value()` you could use here.

test/backend/lrn.in.cpp
```diff
@@ -72,3 +75,221 @@ NGRAPH_TEST(${BACKEND_NAME}, lrn)
...
+
+    vector<float> args{0.0f, 1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f, 7.0f, 8.0f, 9.0f, 10.0f, 11.0f};
+    auto a = backend->create_tensor(element::f32, shape);
+    copy_data(a, args);
+
+    auto result = backend->create_tensor(element::f32, shape);
+    auto handle = backend->compile(f);
+    handle->call_with_validate({result}, {a});
+
+    vector<float> expected{0.0f,
```
aprocter: Could you add a comment in the C++ source with the numpy snippet you used to generate these values?

src/ngraph/op/lrn.cpp
mbencer: Done ;)

src/ngraph/op/lrn.cpp
mbencer: Done. It is `get_axis_set_val` ;)

test/backend/lrn.in.cpp
mbencer: Sure. Done

src/ngraph/op/lrn.cpp
```diff
@@ -43,14 +44,23 @@ op::LRN::LRN(const Output<Node>& arg,
     constructor_validate_and_infer_types();
 }
 
+AxisSet op::LRN::get_reduction_axes() const
+{
+    AxisSet axes{1}; // channel axis as default
+    if (auto const_op = dynamic_pointer_cast<op::Constant>(get_argument(1)))
```
diyessi: Use `input_value(1).get_node_shared_ptr()`

test/backend/lrn.in.cpp
diyessi: Better is `test/ref_generators`

src/ngraph/op/lrn.cpp
mbencer: Done

test/backend/lrn.in.cpp
mbencer: Done

src/ngraph/op/lrn.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include "ngraph/op/lrn.hpp"
 #include "ngraph/op/constant.hpp"
+#include "ngraph/op/constant.hpp"
```
rkimballn1: This header is already included on line 18

mbencer: Redundant include was removed

test/backend/lrn.in.cpp
```diff
@@ -37,14 +39,15 @@ using namespace ngraph;
 
 static string s_manifest = "${MANIFEST}";
 
-NGRAPH_TEST(${BACKEND_NAME}, lrn)
+NGRAPH_TEST(${BACKEND_NAME}, lrn_across_c)
```
tsocha:
```suggestion
NGRAPH_TEST(${BACKEND_NAME}, lrn_across_channels)
```

test/backend/lrn.in.cpp
```diff
@@ -72,3 +75,221 @@ NGRAPH_TEST(${BACKEND_NAME}, lrn)
...
+                           0.6469966f,
+                           0.7548294f,
+                           0.6620847f,
+                           0.7448453f,
+                           0.671156f,
+                           0.7382717f};
+    EXPECT_TRUE(test::all_close_f(expected, read_vector<float>(result)));
+}
+
+NGRAPH_TEST(${BACKEND_NAME}, lrn_across_nchw)
```
tsocha:
```suggestion
NGRAPH_TEST(${BACKEND_NAME}, lrn_across_all_dims)
```

PR #3456, opened 2019-08-16
Separate dynamic loading from static linking

diyessi:


silee2:
@diyessi I'm approving but not sure what we want to support. Shouldn't nGraph have just two disjoint configuration: All backend static vs all backend dynamic? I don't see a usage case where we want only some backends to be static.

PR #3457, opened 2019-08-16
[MLIR] Introduce flag -print-ngraph-ir-after-all.

dcaballe:
This PR is a stepping stone towards unifying nGraph MLIRCompiler printing
flags with those used in MLIR. It enables flag -print-ir-after-all
implemented in MLIR pass manager and adds flag -print-ngraph-ir-after-all
to MLIRCompiler so that we can use it to dump IR for those transformations
that we do directly in MLIRCompiler without using a proper pass. Eventually,
everything should be working as a pass and the nGraph variant of the
flag shouldn't be needed. NGRAPH_MLIR_DUMP_ALL macro is no longer
needed.

### DIFF
--- a/src/contrib/mlir/compiler.cpp
+++ b/src/contrib/mlir/compiler.cpp
@@ -81,6 +81,17 @@ using llvm::ArrayRef;
 using namespace ngraph;
 using namespace ngraph::runtime::ngmlir;
 
+// *** Debug flags ***
+
+static llvm::cl::opt<bool> clPrintIRAfterAll(
+    "print-ngraph-ir-after-all",
+    llvm::cl::init(false),
+    llvm::cl::desc(
+        "Print IR after transformation that are not implemented as passes in the MLIRCompiler. It "
+        "complements MLIR -print-ir-after-all and LLVM -print-after-all flags"));
+
+// *** Optimization flags ***
+
 static llvm::cl::opt<bool>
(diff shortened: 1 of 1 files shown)


src/contrib/mlir/compiler.cpp
```diff
@@ -322,15 +336,16 @@ void MLIRCompiler::lower_ng_dialect()
...
+    mlir::applyPassManagerCLOptions(pm);
+
     pm.run(m_module.get());
 
     if (failed(m_module->verify()))
     {
         NGRAPH_CHECK(false, "Incorrect module after dialect lowering");
     }
 
-    dump_mlir_module("Affine Dialect Dump (Pre-Optimizations):");
```
nmostafa: Why removing this one ?

dcaballe: Because it becomes redundant. It prints exactly the same as `-print-ir-after-all` for canonicalizer pass (line 323)

nmostafa: Having two flags at the MLIR level can be confusing especially that we will likely mix ngraph and mlir passes (e.g. ngraph pass , mlir CSE, ngraph pass ...etc.). I would prefer us to print with `-print-ir-after-all` is ON instead.

nmostafa:
Will merge after MLIR CI is up.

PR #3460, opened 2019-08-16
[MLIR] Enable clang-format on comments in contrib folder

dcaballe:
This PR starts the modular approach that we discussed to enable
clang-format on comments so that they are automatically formatted
instead of having to do that manually. It only covers the src/contrib
directory.


aprocter:
LGTM! I like the idea of parking these as high up the tree as we can, until eventually we can do it at the root.

PR #3461, opened 2019-08-17
Add .clang-format with comment wrap to src/ngraph

aprocter:
This adds an updated `.clang-format` to `src/ngraph` which requires comments to stay under 100 columns, and automatically reflows them if they are not.

Each immediate _subdirectory_ of `src/ngraph` has an override-to-the-override which keeps the current rules. This should allow gradual cleanup of the comments without the need for one giant PR, which is important because sometimes auto-reflow needs manual cleanup. In this PR, all the files that live directly in `src/ngraph` are updated. Future PRs can clean up `autodiff`, `builder`, `codegen`, etc.

See also #3460 from @dcaballe.

UPDATE: I noticed that there were a couple of subdirs where only one file was going to need to be changed, so I went ahead and folded those into this PR.


aprocter:
Reviewers: Generally you can ignore the new `.clang-format` files in your owned directories. The effect is to maintain the old formatting rules for comments within those directories. Once the file is _removed_ in a later PR, then you have to watch out. :) The actual number of existing files changed in this PR is only 7, and they are all in `src/ngraph` (not any subdir).

PR #3465, opened 2019-08-19
Fold Constant + ConvertLayout to reduce memory footprint.

ayzhuang:


src/ngraph/op/constant.hpp
```diff
@@ -28,11 +28,23 @@
...
+                class CPUConvertLayoutConstantFolding;
+            }
+        }
+    }
     namespace op
     {
         /// \brief Class for constants.
         class Constant : public Node
         {
+            friend class ngraph::runtime::cpu::pass::CPUConvertLayoutConstantFolding;
```
aprocter: Can you comment on the reason for this? Would prefer not to use `friend` here.

ayzhuang: Need to use get_data_ptr_nc() to get a void* for mkldnn::memory in{{input_desc, runtime::cpu::executor::global_cpu_engine}, input_ptr};

aprocter: If it's not too costly to copy the input data out to a temporary, let's do that. Otherwise, let's take `get_data_ptr_nc` public.

Do you know why MKL-DNN doesn't accept a const pointer here?

ayzhuang: MKL_DNN uses the same mkldnn::memory constructor for input and output. It would modify the output memory buffer. So it only accepts void*.
    mkldnn::memory in{{input_desc, runtime::cpu::executor::global_cpu_engine}, input_ptr};
    mkldnn::memory out{{result_desc, runtime::cpu::executor::global_cpu_engine}, result_vec.data()};

aprocter: But because it's only being supplied to an `input_desc` can we be sure it won't actually modify it? If so, maybe this is a job for `const_cast`.

ayzhuang: Changed to use const_cast.

PR #3468, opened 2019-08-19
Add replace_nodes function

aprocter:
~This adds a function to replace zero or more nodes in the graph according to their friendly names, via a key/value replacement map.~ Changed to a (node pointer->node pointer) map, since friendly_name is unreliable.


src/ngraph/graph_util.cpp
```diff
@@ -179,6 +180,45 @@ void ngraph::replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> re
     target->clear_control_dependents();
 }
 
+size_t
+    ngraph::replace_by_friendly_name(const shared_ptr<Function>& f,
+                                     const unordered_map<string, shared_ptr<Node>>& replacement_map)
+{
+    size_t num_replacements = 0;
+
+    for (auto& replacee : f->get_ops())
```
ilyachur: Could you also add a check that graph contains only one node for each name from the unordered_map?

aprocter:
@ilyachur One limitation of this function is that if you replace a `Parameter`, it does not actually update the parameter in the `Function`'s list. Is that the behavior you want? Please let me know if I need to change it.

ilyachur:
@aprocter Only one question if this function doesn't update the parameter in the Function's list how can I understand that parameter was changed?

aprocter:
Hi @ilyachur. I've updated it with a new signature that has two separate maps: one for parameter replacement, and another for replacement of "body" nodes.

Usage is a little bit subtle. In particular, note that it _is_ valid to replace a parameter in the body map, but the meaning is different.

- `x->y` in the _parameter_ replacement map means that `y` takes over `x`'s users, _and_ `y` replaces `x` in the parameter list.
- `x->y` in the _body_ replacement map means that `y` takes over `x`'s users, _but_ `y` does _not_ replace `x` in the parameter list.

Let me know your thoughts.

PR #3480, opened 2019-08-20
clang-format comments: /src/ngraph/runtime/interpreter etc.

aprocter:
Covers `/src/ngraph/runtime/reference/{dynamic,interpreter,generic_cpu,gpuh,interpreter,nop}`


aprocter:
Combined into #3488

PR #3489, opened 2019-08-21
Temporary megalomaniacal CODEOWNERS power grab

aprocter:
The final phase of the .clang-format bikeshedding will require every .clang-format file to be touched, which will probably slow down review. Think it's easiest if we put a global override to CODEOWNERS for the moment.


aprocter:
Eighty-sixing this per @diyessi. We'll just remove the auto-requests when the innocent `.clang-format` deletions happen later.

PR #3494, opened 2019-08-21
clang-format comments: /src/ngraph/runtime

aprocter:
Completes the updates for `/src/ngraph/runtime`, including files directly in `/src/ngraph/runtime`.


aprocter:
(Removing a bunch of auto-requested reviewers because all this does is remove now-redundant `.clang-format` overrides from subdirs.)

PR #3511, opened 2019-08-23
[MLIR] Fixes for cpu_fusion.validate_fuse_gru_inputs

nmostafa:
Fixes:
* Cycle checking wasn't considering if a node is part of a sub-graph and wasn't tracking the sub-graph inputs.
* Moving subgraph outputs to CK was happening during CKs creation (which itself adds new edges). That created inconsistent graph.
* Insert GOE if a CK has multiple outputs. 

Improvements:
* Do a sanity check in debug build on invariants : no cycles, inputs to sub-graph are inputs to CK, no outputs out of subgraph for output nodes
* Tabbed debug messages for better readability.

Misc Fixes:
* sub-graph topological sort traversed nodes out of the sub-graph. Was slowing down compilation. 
* Fix `as_single_output` to be able to always create a GOE


src/ngraph/graph_util.cpp
```diff
@@ -653,3 +653,47 @@ std::vector<Output<Node>> ngraph::get_outputs_to(Node& src, Node& dst)
...
+{
+    bool result = false;
+
+    for (auto param : func->get_parameters())
+    {
+        std::deque<std::shared_ptr<Node>> path;
+        // mirror of path stack for faster cycle check
+        std::unordered_set<std::shared_ptr<Node>> path_set;
+        result |= check_fwd_cycles(param, path, path_set, cycle_nodes);
+    }
```
diyessi: You could also have a reverse cycle from a result.

nmostafa: You mean for a result that is not reachable from the func params ? Backward cycles should be captured here for all reachable result nodes.

aprocter: Ah right... a Result might depend only on computations relative to Constants, with no Parameters involved.

nmostafa:
TESTNOW

PR #3512, opened 2019-08-23
[MLIR] Remove warnings (unused code + wrong neg op)

dcaballe:
I think the negation one is actually a bug. Please, make sure that's the expected logic.


src/contrib/mlir/dialect/ops.cpp
```diff
@@ -199,7 +199,7 @@ mlir::LogicalResult verifyOp(NGGatherOp* op)
     NGTensorType resType = r0.cast<NGTensorType>();
 
     // ensure result is compatible with input
-    if (!resType.getRank() == inputType.getRank() + indicesType.getRank() - 1)
+    if (resType.getRank() != inputType.getRank() + indicesType.getRank() - 1)
```
nmostafa: Good catch. Thanks for fixing it

PR #3513, opened 2019-08-23
Fix compiler warnings.

diyessi:


src/ngraph/runtime/cpu/pass/cpu_rnn_fusion.cpp
```diff
@@ -157,7 +157,7 @@ void ngraph::runtime::cpu::pass::LSTMFusion::construct_onnx_lstmcell_fprop()
 
             auto new_bias = std::make_shared<ngraph::op::Concat>(
                 NodeVector{gate_slices[0], gate_slices[2], gate_slices[3], gate_slices[1]}, 0);
-            return new_bias;
+            return std::move(new_bias);
```
nmostafa: What sort of warning did that throw ?

PR #3524, opened 2019-08-26
[MLIR] Enable LIT testing in CMAKE

dcaballe:
This PR enables LIT testing in nGraph for MLIR Compiler, introduced in
PR3523. We can now do `make check-mlir-lit` to run LIT tests.


test/mlir/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+configure_lit_site_cfg(
+    ${CMAKE_CURRENT_SOURCE_DIR}/lit.site.cfg.py.in
+    ${CMAKE_CURRENT_BINARY_DIR}/lit.site.cfg.py
+    MAIN_CONFIG
+    ${CMAKE_CURRENT_SOURCE_DIR}/lit.cfg.py
+    )
+
+#set(NGRAPH_MLIR_TEST_DEPENDS
+#    ngraph-opt
+#    )
```
dcaballe: To be enabled in the next commit

test/mlir/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+
+add_lit_testsuite(check-mlir-lit "Running the nGraph MLIR regression tests"
+    ${CMAKE_CURRENT_BINARY_DIR}
+    DEPENDS ${NGRAPH_MLIR_TEST_DEPENDS}
+    )
+set_target_properties(check-mlir-lit PROPERTIES FOLDER "Tests")
```
nmostafa: minor nit: are there going to be other types of check-mlir-* ? If not, then `check-mlir` might be better.

dcaballe: I planned to add `check-mlir` for both MLIR gtests + LIT tests

PR #3526, opened 2019-08-27
Make axis for TopK dynamic

aprocter:


src/ngraph/op/topk.cpp
```diff
@@ -76,6 +91,33 @@ void op::TopK::set_k(size_t k)
         op::Constant::create(element::i64, Shape{1}, {k})->output(0));
 }
 
+size_t op::TopK::get_top_k_axis() const
+{
+    auto d = get_top_k_axis_dynamic();
+    NGRAPH_CHECK(d.is_static(),
```
baojun-nervana: does this require top_k_axis to be a constant node? Otherwise it is not supported yet?

aprocter: `get_top_k_axis()` will not work if the value of `k` is unknown. This is generally how things work with these accessors for "formerly static attributes".

aprocter: In general, this won't cause issues for the moment---code that expects top_k_axis to be statically known isn't going to be invoked until after graph specialization.

src/ngraph/op/topk.cpp
```diff
@@ -32,8 +32,9 @@ op::TopK::TopK(const Output<Node>& arg,
                size_t k,
                bool compute_max,
                SortType sort)
-    : Op({arg, op::Constant::create(element::i64, Shape{1}, {k})->output(0)})
-    , m_top_k_axis(top_k_axis)
+    : Op({arg,
+          op::Constant::create(element::i64, Shape{1}, {k})->output(0),
+          op::Constant::create(element::i64, Shape{1}, {top_k_axis})->output(0)})
     , m_index_element_type(index_element_type)
```
diyessi: These constants (and in the other constructor) should be set up in `validate_and_infer_types`. 
All created constants need to be be added to the node's provenance group.

baojun-nervana: I updated provenance. Don't quite get "set up the constant in `validate_and infer_types`". Can you check?

diyessi: Don't create the constant in the constructor. Instead, validate_and_infer_types should create them if the inputs are missing. So here you would do `Op({arg})` and in validate_and_infer_types if the number of args is 1 then create the constants. You'll need to stick `k` in a member variable to get it from the constructor to validate.

PR #3527, opened 2019-08-27
[Py] Added Dequantize, Quantize, Quantized Convolution, Quantized Dot…

Ewa21:
… operators to Python API.


python/ngraph/ops.py
```diff
@@ -168,6 +170,190 @@ def mvn(data, axes, normalize_variance, eps, name=None):
...
 
 
+@nameable_op
+def quantize(data, scale, zero_point, new_type, axes, round_mode, name=None):
+    # type: (Node, Node, Node, NumericType, Set[int], int, str) -> Node
+    r"""Perform quantize operation on data from input node.
+
+    Computes quantize on the input tensor:
+
+    .. math:: output = ROUND((input /scale) + zero\_point)
```
tsocha:
```suggestion
    .. math:: output = ROUND((input / scale) + zero\_point)
```

python/ngraph/ops.py
```diff
@@ -168,6 +170,190 @@ def mvn(data, axes, normalize_variance, eps, name=None):
     return MVN(data, AxisSet(axes), normalize_variance, eps)
 
 
+@nameable_op
+def quantize(data, scale, zero_point, new_type, axes, round_mode, name=None):
+    # type: (Node, Node, Node, NumericType, Set[int], int, str) -> Node
```
postrational: The `round_mode` should not be marked as `int` type but rather the enum type you created.

python/ngraph/ops.py
```diff
@@ -168,6 +170,190 @@ def mvn(data, axes, normalize_variance, eps, name=None):
...
+
+    Computes quantize on the input tensor:
+
+    .. math:: output = ROUND((input / scale) + zero\_point)
+
+    :param data: The node with data tensor.
+    :param scale: Scale used for mapping.
+    :param zero_point: Zero point used for mapping.
+    :param new_type: Output element type.
+    :param axes: Number describes how to perform ROUND function.
```
postrational: This documentation should be using the names of the modes, not integer values.
Also, is this for `axes` or `round_mode`?

Ewa21: Ok. But shouldn't it provide informations about which number you should use to obtain a certain type of `ROUND`?

PR #3532, opened 2019-08-27
Add priority for calls

diyessi:
Adds a priority argument for execution that a backend can implement. Default is to ignore the priority.


src/ngraph/runtime/executable.hpp
```diff
@@ -44,6 +44,17 @@ class ngraph::runtime::Executable
     virtual bool call(const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
                       const std::vector<std::shared_ptr<runtime::Tensor>>& inputs) = 0;
 
+    /// \param outputs vector of runtime::Tensor used as outputs
+    /// \param inputs vector of runtime::Tensor used as inputs
+    /// \param priority backend-specific priority for this call
```
rkimballn1: I'm assuming that lower numbers are higher priority, but it would be nice for me not to need to assume. Add what priority means to the comment.

PR #3533, opened 2019-08-28
Remove static initializers from ops

diyessi:
* Aside from `node.hpp`, `node.cpp`, and `visualize_tree.cpp`, almost every change is the same.
* Replace the string static initializers for node types with a constexpr struct of name and version
to prevent static initialization issues with `string`. Technically, the version could be included in the string, but too many places currently use string literals to check node type.
* Add `has_type` and `as_type` for RTTI-free upcasting.


src/ngraph/node.hpp
```diff
@@ -79,11 +80,19 @@ namespace ngraph
     /// Alias useful for cloning
     using NodeMap = std::unordered_map<ngraph::Node*, std::shared_ptr<ngraph::Node>>;
 
+    struct NodeImplInfo
```
aprocter: I don't like this name. Maybe `NodeTypeInfo`?

src/ngraph/node.hpp
```diff
@@ -79,11 +80,19 @@ namespace ngraph
...
+    {
+        const char* type_name;
```
aprocter: Maybe just `name`?

src/ngraph/node.hpp
```diff
@@ -139,6 +148,41 @@ namespace ngraph
...
+
+        /// Casts a Node to a shared_ptr<T> if is of type T, nullptr otherwise;
+        template <typename NodeType>
+        std::shared_ptr<const NodeType> as_type() const
+        {
+            return has_type<NodeType>() ? shared_from_this() : std::shared_ptr<NodeType>();
+        }
+
+        /// Returns the NodeTypeInfo for the node's class.
+        /// During transition to type_info, return's a dummy type_info for Node if the class
```
indie: typo on returns

src/ngraph/op/add.hpp
```diff
@@ -30,8 +30,8 @@ namespace ngraph
         {
         public:
             NGRAPH_API
-            static const std::string type_name;
-            const std::string& description() const override { return type_name; }
+            static constexpr NodeTypeInfo type_info{"Add", 0};
+            const NodeTypeInfo& get_type_info() const override { return type_info; }
             /// \brief Constructs an unitialized addition operation
```
indie: Not part of your PR, but probably means to say "uninitialized" addition operation?

src/ngraph/op/subtract.hpp
```diff
@@ -27,8 +27,8 @@ namespace ngraph
         {
         public:
             NGRAPH_API
-            static const std::string type_name;
-            const std::string& description() const override { return type_name; }
+            static constexpr NodeTypeInfo type_info{"Subtract", 0};
+            const NodeTypeInfo& get_type_info() const override { return type_info; }
             Subtract() = default;
             /// \brief Constructs an subtraction operation.
```
indie: Constructs a subtraction op

src/ngraph/node.hpp
```diff
@@ -139,6 +148,41 @@ namespace ngraph
     public:
         virtual ~Node();
 
+        /// Tests if a node is of op type T
+        template <typename NodeType>
+        bool has_type() const
```
rkimballn1: I don't understand what `has_type` is. I do understand what `is_type` is but can you add a comment as to what this is. 
stl uses is_* all over the place

src/ngraph/node.hpp
```diff
@@ -79,11 +80,19 @@ namespace ngraph
...
+        const char* name;
+        uint64_t version;
+    };
+
     /// Nodes are the backbone of the graph of Value dataflow. Every node has
     /// zero or more nodes as arguments and one value, which is either a tensor
     /// or a (possibly empty) tuple of values.
     class Node : public std::enable_shared_from_this<Node>
     {
+        static constexpr NodeTypeInfo type_info{"Node_0", 0};
```
rkimballn1: I believe this should be "Node", not "Node_0"

src/ngraph/node.hpp
```diff
@@ -139,6 +148,41 @@ namespace ngraph
...
+        /// Tests if a node is of op type T
+        template <typename NodeType>
+        bool has_type() const
+        {
+            return &get_type_info() == &NodeType::type_info;
+        }
+
+        /// Casts a Node to a shared_ptr<T> if is of type T, nullptr otherwise;
+        template <typename NodeType>
+        std::shared_ptr<NodeType> as_type()
```
rkimballn1: I don't like that this further burns in the requirement that everything in every class in ngraph must be implemented as a shared_ptr. Can't you just return a NodeType? Or get rid of this method?

src/ngraph/node.hpp
```diff
@@ -139,6 +148,41 @@ namespace ngraph
...
+        std::shared_ptr<const NodeType> as_type() const
+        {
+            return has_type<NodeType>() ? shared_from_this() : std::shared_ptr<NodeType>();
+        }
+
+        /// Returns the NodeTypeInfo for the node's class.
+        /// During transition to type_info, return's a dummy type_info for Node if the class
+        /// has not been updated yet.
+        virtual const NodeTypeInfo& get_type_info() const { return type_info; }
+        virtual const char* get_type_name() const
```
rkimballn1: This should return a string to be more useful
With this I can't say `if (node.get_type_name() == "Abs")`

diyessi: You shouldn't be doing that anyways; there can be more than one version of Abs. It's also an accessor; `description` returns a string.

src/ngraph/node.hpp
diyessi: Got it.

src/ngraph/node.hpp
diyessi: Change to `is_type`

changes.md
```diff
@@ -1,5 +1,10 @@
 # API Changes
 
+## Op Definition
+* Every Op class must declare a `static constexpr NodeTypeInfo type_info{name, version}` in the class definition and define it in the .cpp file. See any op definition for an example.
+* The boolean node method `is_type<T>` is for testing if a node is the op `T`.
+* `T as_type_ptr<T>()` and `T as_typr<T>()` will upcast `Node` to an explicit op class if it is of class `T`, or `nullptr` if it is not.
```
nmostafa: typo `as_type<T>`*

aprocter: `as_typo<T>`

src/ngraph/node.hpp
```diff
@@ -79,11 +80,19 @@ namespace ngraph
     /// Alias useful for cloning
     using NodeMap = std::unordered_map<ngraph::Node*, std::shared_ptr<ngraph::Node>>;
 
+    struct NodeTypeInfo
+    {
+        const char* name;
```
ilyachur: Why do you use char* and not std::string?

diyessi: std::string cannot be a constexpr so initialization happens in unpredictable ways.

diyessi: Also, `description()` will return a string.

PR #3534, opened 2019-08-28
Pawepiot/pi vecstring2double

pawelpiotrowicz:
size of destination equal size of source, prepare destination and reserve memory for desired elements.
In real we have a lot of cases where result is re-allocated, we are working with 4K elements and more.


aprocter:
This is superseded by #3542, right?

pawelpiotrowicz:
@aprocter  yes, we can close it

PR #3535, opened 2019-08-28
Performance issue vector of string to double

pawelpiotrowicz:
size of destination equal size of source, prepare destination and reserve memory for desired elements.
In real we have a lot of cases where result is re-allocated, we are working with 4K elements and more.


diyessi:
Needs to be moved from fork into a branch.

diyessi:
Nothing left to merge.

PR #3536, opened 2019-08-28
[Spec] Implementation of Reverse:v1

tomdol:


postrational:
It would make sense to make this a PR agains the opset branch and later change the base branch, when that branch is merged.

postrational:
I changed the base branch, the PR is a bit cleaner, but there are conflicts due to changes in the opset branch.

src/ngraph/op/reverse.cpp
```diff
@@ -69,3 +69,84 @@ void op::Reverse::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+op::v1::Reverse::Reverse(const Output<Node>& data,
+                         const Output<Node>& reversed_axes,
+                         const std::string& mode)
+    : Op({data, reversed_axes})
+    , m_mode{mode}
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::v1::Reverse::validate_and_infer_types()
```
aprocter: If input 1 is fed by a `Constant` (and input 0 rank is static), we should still check to make sure that each element of the axis set is < rank. You can use `Node::is_constant()`, `dynamic_pointer_cast`, and `Constant::get_axis_set_val` for this.

aprocter: (I guess you'll need a separate codepath in case the "mask" behavior is selected, but I think we should still check it.)

src/ngraph/op/reverse.cpp
```diff
@@ -69,3 +69,84 @@ void op::Reverse::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+    const auto input_rank = input_shape.rank();
+
+    const auto rev_axes_shape = get_input_partial_shape(1);
+    const auto rev_axes_rank = rev_axes_shape.rank();
+
+    if (input_rank.is_static() && rev_axes_rank.is_static())
+    {
+        NODE_VALIDATION_CHECK(this,
+                              static_cast<size_t>(rev_axes_rank) == 1 &&
+                                  static_cast<size_t>(rev_axes_shape[0]) <=
```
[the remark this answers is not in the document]
aprocter: @slyalin Thoughts on this?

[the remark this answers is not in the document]
aprocter: (`mask` mode is a different story, BTW. There, we will need to check.)

src/ngraph/op/reverse.hpp
```diff
@@ -77,5 +77,44 @@ namespace ngraph
...
+            public:
+                NGRAPH_API
+                static const std::string type_name;
+                const std::string& description() const override { return type_name; }
+                Reverse() = default;
+                /// \brief Constructs a reverse operation.
+                ///
+                /// \param data The input tensor, some of whose axes are to be reversed.
+                /// \param reversed_axes The axes to reverse in a form of a set of indices or boolean mask.
+                /// \param mode The way reversed_axes should be interpreted - a set or a mask.
```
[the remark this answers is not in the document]
tomdol: That's what I would prefer too but according to the specs a string is expected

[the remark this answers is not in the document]
aprocter: Strings are inconsistent with the rest of the nGraph API and are somewhat less safe (we tend to rely on -Wswitch to check complete handling of enum cases, which is not possible with strings). @postrational @slyalin, what do you think?

src/ngraph/op/reverse.cpp
tomdol: I've done the validation for the `Constant` second input case and covered it with some tests. I've also added a separate path for the `mask` mode but it turned out it was validated by another check that already exists (checks if the number of values in the mask matches the input rank). It does not operate on the actual values of the constant but rather on shapes. Is this ok?

src/ngraph/op/reverse.cpp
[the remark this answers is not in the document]
tomdol: Done, but if there are any more comments - please feel free  to add them

src/ngraph/op/reverse.cpp
aprocter: Looks great! For `mask` I think that makes sense---any mask of the correct shape is valid by definition, so no need to check the values in the mask.

test/opset_pass/reverse_opset1upgrade_pass.cpp
```diff
@@ -0,0 +1,52 @@
+//*****************************************************************************
```
postrational: We may want to put upgrade and downgrade pass tests in the same file. I think naming this `reverse_opset_pass.cpp` should be sufficient.

src/ngraph/op/reverse.cpp
```diff
@@ -69,3 +70,116 @@ void op::Reverse::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+
+        if (m_mode == "mask")
+        {
+            NODE_VALIDATION_CHECK(
+                this,
+                static_cast<size_t>(rev_axes_rank) == 1,
+                "The reversed_axes input rank needs to be equal to 1 in 'mask' mode.");
+
+            const auto rev_axes_mask_elems_count = static_cast<size_t>(rev_axes_shape[0]);
+            NODE_VALIDATION_CHECK(this,
```
tsocha: `input_rank` is checked only in this `VALIDATION_CHECK` so maybe you should add here a new `if` and move `input_rank.is_static()` condition from line 106 to this `if`?

src/ngraph/op/reverse.cpp
```diff
@@ -69,3 +70,116 @@ void op::Reverse::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+{
+    check_new_args_count(this, new_args);
+    return make_shared<op::v1::Reverse>(new_args.at(0), new_args.at(1), m_mode);
+}
+
+void op::v1::Reverse::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
+{
+    const auto delta = deltas.at(0);
+
+    const auto x = input(0).get_source_output();
```
tsocha: You can use `input_value()` function here
```suggestion
    const auto x = input_value(0);
```

src/ngraph/op/reverse.cpp
```diff
@@ -69,3 +70,116 @@ void op::Reverse::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+    const auto reversed_axes = input(1).get_source_output();
```
tsocha:
```suggestion
    const auto reversed_axes = input_value(1);
```

src/ngraph/op/reverse.cpp
tomdol: Thanks, I've moved the the `input_rank` check deeper and realized that I also have to check if the first(or zero) dimension of the shape is static before I cast it to `size_t`.

src/ngraph/op/reverse.hpp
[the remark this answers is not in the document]
diyessi: I agree they should be enums here and in the spec.

src/ngraph/op/reverse.hpp
```diff
@@ -77,5 +77,59 @@ namespace ngraph
...
+            public:
+                enum class Mode
+                {
+                    INDEX,
+                    MASK
+                };
+
+                NGRAPH_API
+                static const std::string type_name;
+                const std::string& description() const override { return type_name; }
```
diyessi: These need to be switched to `type_info`
See `abs.hpp`/`abs.cpp`
They should work as-is for a short time.

src/ngraph/op/reverse.cpp
```diff
@@ -69,3 +70,132 @@ void op::Reverse::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+                              "The reversed_axes input must be a 1D tensor (got ",
+                              static_cast<size_t>(rev_axes_rank),
+                              ").");
+
+        if (m_mode == Mode::MASK)
+        {
+            NODE_VALIDATION_CHECK(
+                this,
+                static_cast<size_t>(rev_axes_rank) == 1,
+                "The reversed_axes input rank needs to be equal to 1 in 'mask' mode.");
```
slyalin: In 'index' mode it should be 1 as well. Why not to check outside m_mode == Mode::MASK condition?

src/ngraph/op/reverse.cpp
```diff
@@ -69,3 +70,132 @@ void op::Reverse::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+        if (rev_axes_node->is_constant())
+        {
+            const auto rev_axes_constant = dynamic_pointer_cast<op::Constant>(rev_axes_node);
+
+            if (m_mode == Mode::INDEX)
+            {
+                const AxisSet rev_axes = rev_axes_constant->get_axis_set_val();
+
+                NODE_VALIDATION_CHECK(this,
+                                      rev_axes.size() < rank,
```
slyalin: If all axes are set to be reversed, which is a valid case, this check fails. Should be `<=`.

src/ngraph/op/reverse.cpp
```diff
@@ -69,3 +70,132 @@ void op::Reverse::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVect
...
+                                      input_shape,
+                                      ").");
+
+                bool all_axes_in_range = all_of(rev_axes.begin(),
+                                                rev_axes.end(),
+                                                [&rank](const size_t axis) { return axis < rank; });
+
+                NODE_VALIDATION_CHECK(this,
+                                      all_axes_in_range,
+                                      "Some of the provided axes (",
```
slyalin: Minor suggestion. Can we replace "some" by specific numbers? We know them. Or print them all.

src/ngraph/op/reverse.hpp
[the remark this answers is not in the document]
slyalin: As I have already suggested in one of the reviews, let make them enums here, but in the same time provide handy functions to convert back and force. And this function should be separate from the main serialization/deserialization code to allow other serialization/deserealization routines (like OpenVINO IR reader) convert string to enum allowing different serialization format at higher level (xml vs. json).

src/ngraph/op/reverse.cpp
tomdol: They will all be printed together with the input rank. The `rev_axes` is included in the error message here.

PR #3541, opened 2019-08-29
[MLIR] Clean up MLIR related CMake files

dcaballe:
This PR prepares the introduction of src/contrib/mlir/tools and LIT testing by
refactoring and cleaning up the common part with src/contrib/mlir/compiler:

- Common code for src/contrib/mlir/tools and src/contrib/mlir/compiler is
moved to src/contrib/mlir/CMakeList.txt.
- More generic code, common for src/contrib/mlir/* and LIT configuration
files, is moved to cmake/external_mlir.cmake with pre-existing similar code.
- Redundant environment var LLVM_DIR is removed in favor of LLVM_CMAKE_PATH.
- Redundant cmake messages are removed.


cmake/external_mlir.cmake
```diff
@@ -42,10 +42,23 @@ execute_process(COMMAND "${CMAKE_COMMAND}" --build . --target ext_mlir_llvm
...
-# point find_package to the pre-built libs
-set(LLVM_DIR ${MLIR_LLVM_ROOT}/build/lib/cmake/llvm)
+# Enable modules for LLVM.
+set(LLVM_CMAKE_PATH "${MLIR_BUILD_DIR}/lib/cmake/llvm"
+    CACHE PATH "Path to LLVM cmake modules")
+list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_PATH}")
+include(AddLLVM)
+
+# Enable LLVM package, definitions and env vars.
+find_package(LLVM REQUIRED CONFIG)
```
nmostafa: We need `LLVM_DIR` set here, otherwise find_package will look else where and might use the system's LLVM instead, if present.

dcaballe: Ok, so can we rename `LLVM_CMAKE_PATH` to `LLVM_DIR` and use that one only? They were both the same and used indistinctly.

dcaballe: Everything work ok with and without `LLVM_DIR` with and without MLIR enabled but I replaced `LLVM_CMAKE_PATH` with `LLVM_DIR`. Please, let me know if I misunderstood something.

cmake/external_mlir.cmake
```diff
@@ -47,10 +47,23 @@ execute_process(COMMAND "${CMAKE_COMMAND}" --build . --target ext_mlir_llvm
 execute_process(COMMAND "${CMAKE_COMMAND}" --build . --target ext_mlir
                 WORKING_DIRECTORY "${MLIR_PROJECT_ROOT}")
 
-# point find_package to the pre-built libs
-set(LLVM_DIR ${MLIR_LLVM_ROOT}/build/lib/cmake/llvm)
+# Enable modules for LLVM.
+set(LLVM_DIR "${MLIR_BUILD_DIR}/lib/cmake/llvm"
+    CACHE PATH "Path to LLVM cmake modules")
+list(APPEND CMAKE_MODULE_PATH "${LLVM_DIR}")
```
nmostafa: This is necessary ? My understanding is `LLVM_DIR` will direct `find_package` to right location of `LLVM` lib

dcaballe: This is needed by the `include(AddLLVM)` below. I tried moving it after `find_package` to see if cmake module path was adjusted accordingly but it didn't work.

PR #3543, opened 2019-08-29
[Spec] Update fused operator GroupConvolution to new specification

tsocha:
This PR allow GroupConvolution operator to handle cases where number of groups are defined as first dimension of filters input.


src/ngraph/op/fused/group_conv.cpp
```diff
@@ -49,6 +51,26 @@ op::GroupConvolution::GroupConvolution(const Output<Node>& data_batch,
...
+                                       const CoordinateDiff& padding_above,
+                                       const Strides& data_dilation_strides,
+                                       const PadType& pad_type)
+    : FusedOp({data_batch, filters})
+    , m_window_movement_strides(window_movement_strides)
+    , m_window_dilation_strides(window_dilation_strides)
+    , m_padding_below(padding_below)
+    , m_padding_above(padding_above)
+    , m_data_dilation_strides(data_dilation_strides)
+    , m_groups(get_input_shape(1).at(0))
```
tomdol: Will this line work if the input shape is not static?

src/ngraph/op/fused/group_conv.hpp
```diff
@@ -72,6 +81,14 @@ namespace ngraph
             Strides m_data_dilation_strides;
             size_t m_groups{0};
             PadType m_pad_type{PadType::NOTSET};
+
+        private:
+            bool have_groups_in_filters() const
```
tomdol: This is a minor thing but it should probably be `has_groups_in_filters`.

src/ngraph/op/fused/group_conv.cpp
tsocha: Done

src/ngraph/op/fused/group_conv.cpp
```diff
@@ -49,10 +51,38 @@ op::GroupConvolution::GroupConvolution(const Output<Node>& data_batch,
...
+
 void op::GroupConvolution::pre_validate_and_infer_types()
 {
     auto data_shape = get_input_partial_shape(0);
     auto filters_shape = get_input_partial_shape(1);
+
+    if (has_groups_in_filters_shape())
+    {
+        NODE_VALIDATION_CHECK(
+            this, filters_shape[0].is_static(), "Groups stored in filters_shape must be static");
```
aprocter: We need to find a way to support this case. It would be better to allow validation to pass, but throw an exception in `get_groups` if there is not enough information. We could probably do this by internally storing `m_groups` as a `Dimension` type, and throwing an exception if that `Dimension` is dynamic.

tsocha: I modified `m_groups` to be internally stored as a dimension.

src/ngraph/op/fused/group_conv.cpp
```diff
@@ -49,23 +51,46 @@ op::GroupConvolution::GroupConvolution(const Output<Node>& data_batch,
...
 {
     auto data_shape = get_input_partial_shape(0);
     auto filters_shape = get_input_partial_shape(1);
+
     if (data_shape.is_static() && filters_shape.is_static())
     {
         // Data channels
         NODE_VALIDATION_CHECK(this,
-                              data_shape.to_shape()[1] % m_groups == 0,
+                              data_shape.to_shape()[1] % get_groups() == 0,
```
aprocter:
```suggestion
                              m_groups.is_dynamic() || data_shape.to_shape()[1] % static_cast<size_t>(m_groups) == 0,
```

aprocter: Similar change for all checks in this function.

src/ngraph/op/fused/group_conv.cpp
```diff
@@ -49,23 +51,46 @@ op::GroupConvolution::GroupConvolution(const Output<Node>& data_batch,
...
+                                       const CoordinateDiff& padding_above,
+                                       const Strides& data_dilation_strides,
+                                       const PadType& pad_type)
+    : FusedOp({data_batch, filters})
+    , m_window_movement_strides(window_movement_strides)
+    , m_window_dilation_strides(window_dilation_strides)
+    , m_padding_below(padding_below)
+    , m_padding_above(padding_above)
+    , m_data_dilation_strides(data_dilation_strides)
+    , m_groups(filters.get_partial_shape()[0])
```
aprocter:
```suggestion
    , m_groups(filters.get_partial_shape().rank().is_dynamic() ? Dimension::dynamic() : filters.get_partial_shape()[0])
```

src/ngraph/op/fused/group_conv.cpp
```diff
@@ -95,20 +120,33 @@ void op::GroupConvolution::post_validate_and_infer_types()
...
-    weights_shape_groups.at(IC) = get_input_shape(0).at(IC) / get_groups();
+    weights_shape_groups.at(IC) = data_shape.at(IC) / get_groups();
     // push_front the number of groups
     weights_shape_groups.insert(weights_shape_groups.begin(), get_groups());
     return weights_shape_groups;
 }
 
+size_t ngraph::op::GroupConvolution::get_groups() const
+{
+    NODE_VALIDATION_CHECK(this, m_groups.is_static(), "Number of groups must be static");
```
aprocter: Probably better to say `get_groups can only be called if the number of groups is static`, or something like that.

src/ngraph/op/fused/group_conv.cpp
```diff
@@ -130,39 +168,32 @@ shared_ptr<Node> op::GroupConvolution::copy_with_new_args(const NodeVector& new_
 NodeVector op::GroupConvolution::decompose_op() const
 {
     auto data = input_value(0);
+    auto data_shape = get_input_shape(0);
```
aprocter: In general, for now, we are not worrying about making `decompose_op` work with dynamic shapes, but be aware that in the future this will be a problem.

src/ngraph/op/fused/group_conv.cpp
tsocha: In this context `get_groups()` should return static value because `filters_shape` is static.
I wanted to avoid these `static_casts` so changed `m_groups` => `get_groups()`.
Does it make sense to you?

src/ngraph/op/fused/group_conv.cpp
tsocha: I removed this variable because it was an unused leftover from draft implementation.
Thanks for mentioning it.

src/ngraph/op/fused/group_conv.cpp
tsocha: Sounds good, done.

src/ngraph/op/fused/group_conv.cpp
tsocha: Done.

src/ngraph/op/fused/group_conv.cpp
aprocter: Ah, I wasn't following that line of reasoning! In general the line of reasoning that "x came from y, and y is all static, therefore x is static" makes sense, but in this case I think it will not _quite_ work, because `m_groups` was set once at construction time, but `filters_shape` may have changed due to graph rewriting.

Here's an idea:

- always set `m_groups` to `Dimension::dynamic()` in the initializer list;
- update `m_groups` at the top of `pre_validate_and_infer_types`.

That way, if the input shape has changed, `m_groups` will be brought up to date once the node is revalidated.

Or, we could change `get_groups()` to query the filter shape, and select the relevant dimension, and not store `m_groups` at all. (But then you would need accessors for `size_t` as well as `Dimension`.)

tsocha: We need to store `m_groups` because groups can still be provided via attribute.
I think that updating `m_groups` at the top of `pre_validate_and_infer_types` is a good solution.
But one question:
Is it possible that after rewriting graph pass which was messing up with filters could update `m_groups` via public setter?

aprocter: > We need to store m_groups because groups can still be provided via attribute.

I see. That seems like an awkward design, but I suppose it's not a big deal.

> Is it possible that after rewriting graph pass which was messing up with filters could update `m_groups` via public setter?

Yes, it is possible. But in general any time we are mutating an attribute, all bets are off with respect to graph validity.

aprocter: Yep!

src/ngraph/op/fused/group_conv.cpp
```diff
@@ -178,3 +217,10 @@ void op::GroupConvolution::generate_adjoints(autodiff::Adjoints& adjoints, const
 {
     throw ngraph_error("NYI");
 }
+
+bool ngraph::op::GroupConvolution::has_groups_in_filters_shape() const
+{
+    // If filters_rank is (data_rank + 1), then filters are divided by groups on first
+    // dim.
+    return ((get_input_shape(0).size() + 1) == get_input_shape(1).size());
```
slyalin: This function should always return true, so it is useless. Weights tensor always contains group dimension by the sepcification. It is the main (and only) difference from the main convolution. It was a question to specification: whether we need to introduce another operation for GroupConvolution in addition to regular Convolution? We decided to introduce another op to avoid sub-classing of original Convolution. I wanted to avoid depending on rank to understand whether we have grouped convolution or not. In the same time we have enough other operations where several `modes` exist. Now lets follow current specification and organize code somehow that Convolution and GroupConvolution can reuse the same base code.

PR #3546, opened 2019-08-29
Clean up and triage unit tests

dgkutnic:
With the latest version of nGraph, I've cleaned up what doesn't need to be on this list anymore, and added some notes on the items that do still need to be on this list.


src/ngraph/runtime/plaidml/unit_test.manifest
```diff
@@ -39,265 +39,24 @@ topk_2d_min_one                         # No plans to implement TopK
...
-
-# Need erf
 gelu_f32
 gelu_f64
 gelu_backprop_factor_f32
 gelu_backprop_factor_f64
 backwards_gelu_f32
 backwards_gelu_f64
-
-# From onnx tests
```
dgkutnic: Ack. My `gtest_filter` didn't catch these ones when I was testing locally. I'll add them back in.

PR #3548, opened 2019-08-30
Fixed NormalizeL2 decomposition for input shape length != 4

GlebKazantaev:


src/ngraph/op/fused/normalize_l2.cpp
```diff
@@ -71,9 +71,11 @@ NodeVector op::NormalizeL2::decompose_op() const
     const Shape input_shape{data.get_shape()};
 
     // Reshape to 4D tensor.
+    size_t axes_shift = 0;
```
postrational: Please add a comment explaining why we need to use this `axes_shift` variable.

diyessi: Better yet, remove it. The op should be rank-independent.

GlebKazantaev: I close this PR. The entire decompose method contains bugs and constraints. I've created ticket on @mbencer

PR #3552, opened 2019-08-31
Performance optimize reference TopK op

rkimballn1:
This runs about 4x faster with the tensor sizes and k values in ResNet50
Also add support for sorting output via value and index.
Add unit tests for sorting options
The default sort value was changed from NONE to VALUE to match TensorFlow's default

I believe we should removing the option to sort by index. Neither PaddlePaddle nor TensorFlow support this option.

I believe we should remove the `BottomK` option, or the `compute_max` argument to the TopK op. This is not used in any framework and I suspect it was added `because we can` not because it was required.


src/ngraph/op/topk.hpp
```diff
@@ -69,13 +69,13 @@ namespace ngraph
...
             /// \param compute_max Compute top k max or top k min?
-            /// \param sort SortType for sorting results, default - NONE
+            /// \param sort SortType for sorting results, default - SORT_VALUES
             TopK(const Output<Node>& arg,
                  const Output<Node>& k,
                  size_t top_k_axis,
                  const element::Type& index_element_type,
                  bool compute_max = true,
-                 SortType sort = SortType::NONE);
+                 SortType sort = SortType::SORT_VALUES);
```
diyessi: Let's make all the attributes have no defaults (i.e. required).

aprocter: I don't think we should change anything here (we can delete `Changes API` if we roll this back), and if we do change it I think we should remove the default. The bridge will have to make an explicit selection anyway, because it has to handle the non-default case.

rkimballn1: The old implementation actually produced a sorted output by default even though NONE was selected and the unit tests depend on the sorted order. The new implementation produces a non-sorted result by default, so if we revert this back to NONE then we need to have `Changes API` because the op behaves differently than it used to.

rkimballn1:
The current PR as it is does not change the API. It works just like it used to work. Changing the default value to SORT_VALUES makes TopK work exactly as it used to work. It used to produce sorted values if you did not specify a sort type and it still does that. The only difference now is that if you specify NONE as the sort type then it may or may not produce sorted results. The reference kernel will now not produce sorted results if you specify NONE. If you specify NONE it means that you don't care about the sort value.

src/ngraph/runtime/reference/topk.hpp
```diff
@@ -63,7 +77,8 @@ namespace ngraph
                       const Shape& out_shape,
                       size_t axis,
                       size_t k,
```
diyessi: As long as you're optimizing this, you might as well move the coordinate iterator loop inside the tests for sort, etc. instead of doing all those tests each time.

aprocter:
@rkimballn1 and I chatted about this offline. While it's true that the written definition of the op did not require sorting by default, the unit tests for TopK actually were insisting on sorted output. So... that's pretty cool. Changing the default here is probably the best solution.

PR #3555, opened 2019-09-03
[SPEC] Add limited support for optional broadcast_axes in DynBroadcast op.

tsocha:
Add support for broadcast_axes as optional input.
Restrictions:
- Input and target shape ranks must be static.
- Broadcasting dimension `1` to `N` is not supported by now.

### DIFF
--- a/src/ngraph/op/experimental/dyn_broadcast.cpp
+++ b/src/ngraph/op/experimental/dyn_broadcast.cpp
@@ -14,7 +14,7 @@
 // limitations under the License.
 //*****************************************************************************
 
-#include <algorithm>
+#include <numeric>
 
 #include "ngraph/op/experimental/dyn_broadcast.hpp"
 #include "ngraph/op/constant.hpp"
--- a/src/ngraph/op/experimental/dyn_broadcast.cpp
+++ b/src/ngraph/op/experimental/dyn_broadcast.cpp
@@ -22,6 +22,13 @@ using namespace ngraph;
 
 const string op::DynBroadcast::type_name{"DynBroadcast"};
 
+ngraph::op::DynBroadcast::DynBroadcast(const Output<Node>& arg, const Output<Node>& shape)
+    : Op({arg, shape})
+{
+    add_default_broadcast_axes();
+    constructor_validate_and_infer_types();
+}
+
 op::DynBroadcast::DynBroadcast(const Output<Node>& arg,
                                const Output<Node>& shape,
                                const Output<Node>& broadcast_axes)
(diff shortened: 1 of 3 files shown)


src/ngraph/op/experimental/dyn_broadcast.cpp
```diff
@@ -14,6 +14,8 @@
 // limitations under the License.
 //*****************************************************************************
 
+#include <numeric>
```
aprocter: (DRAFT NOTES)

- Optional input should be optional in the graph.
- Might as well move to v1 while we're here (but we can do in separate PR)

src/ngraph/op/experimental/dyn_broadcast.cpp
```diff
@@ -22,6 +24,13 @@ using namespace ngraph;
 
 const string op::DynBroadcast::type_name{"DynBroadcast"};
 
+ngraph::op::DynBroadcast::DynBroadcast(const Output<Node>& arg, const Output<Node>& shape)
+    : Op({arg, shape})
+{
+    add_default_broadcast_axes();
```
diyessi: This needs to be deferred until validation. Here set a flag that says no broadcast axes were provided and obey the flag in validate.

test/backend/dyn_broadcast.in.cpp
```diff
@@ -73,3 +73,46 @@ NGRAPH_TEST(${BACKEND_NAME}, dyn_broadcast)
...
+
+NGRAPH_TEST(${BACKEND_NAME}, dyn_broadcast_default_axes)
+{
+    // Create a graph for
+    //   f(x,shape:i32) = Broadcast(x,Convert<i64>(shape)).
+    auto x = make_shared<op::Parameter>(element::f32, PartialShape{Dimension::dynamic()});
+    auto shape = make_shared<op::Parameter>(element::i32, Shape{3});
+    auto shape_i64 = make_shared<op::Convert>(shape, element::i64);
+
+    auto bc = make_shared<op::DynBroadcast>(x, shape_i64);
```
diyessi: Only one constructor is being tested.

src/ngraph/op/experimental/dyn_broadcast.cpp
```diff
@@ -134,3 +143,17 @@ void op::DynBroadcast::generate_adjoints(autodiff::Adjoints& adjoints, const Nod
 {
     throw ngraph_error("generate_adjoints not implemented for DynBroadcast");
 }
+
+void ngraph::op::DynBroadcast::add_default_broadcast_axes()
+{
+    NGRAPH_CHECK(get_input_partial_shape(0).rank().is_static() &&
+                     get_input_partial_shape(1)[0].is_static(),
+                 "Optional parameter broadcast_axes is supported only if rank of input and target "
+                 "shape are static.");
```
slyalin: You cannot relay on dynamic/static shape here, because operation should remain reshapable regardless of the current input nodes. This code should work regardless of input shapes. That means that there should be a sub-graph with explicit ShapeOf and other ops are added to compute this thread input. Or, simpler and probably better alternative, absence of the third input should be treated as a separate case where Python rules of broadcasting is applied and 3rd port remains unconnected. Anyway, after introduction of versioning mechanism we have more freedom in this op modification and probably it is better to change 3rd input semantic to make it more simple and intuitive.

test/backend/dyn_broadcast.in.cpp
tsocha: Constructor with 3 input is tested in a test above.

tsocha:
Due to upcoming redesign of the Broadcast operation in the spec I'm closing this PR for now.

PR #3556, opened 2019-09-03
[SPEC] Implement ReduceSum:v1 and ReduceProduct:v1

mbencer:


src/ngraph/op/product.cpp
```diff
@@ -19,22 +19,48 @@
...
     : ArithmeticReduction(arg, reduction_axes)
 {
     constructor_validate_and_infer_types();
 }
 
-shared_ptr<Node> op::Product::copy_with_new_args(const NodeVector& new_args) const
+shared_ptr<Node> op::v0::Product::copy_with_new_args(const NodeVector& new_args) const
 {
     check_new_args_count(this, new_args);
     return make_shared<Product>(new_args.at(0), get_reduction_axes());
```
aprocter:
```suggestion
     return make_shared<op::v0::Product>(new_args.at(0), get_reduction_axes());
```

(Should be equivalent, but let's be safe.)

src/ngraph/op/product.cpp
```diff
@@ -19,22 +19,48 @@
...
 }
+
+const string op::v1::ReduceProd::type_name{"Product"};
+
+op::v1::ReduceProd::ReduceProd(const Output<Node>& arg,
+                               const AxisSet& reduction_axes,
+                               int keep_dims)
```
aprocter:
```suggestion
                               bool keep_dims)
```

src/ngraph/op/sum.cpp
```diff
@@ -20,27 +20,61 @@
...
+{
+    auto delta = deltas.at(0);
+
+    auto x = input_value(0);
+    auto& x_shape = x.get_shape();
+
+    adjoints.add_delta(x, make_shared<op::Broadcast>(delta, x_shape, get_reduction_axes()));
+}
+
+const string op::v1::ReduceSum::type_name{"Sum"};
```
aprocter:
```suggestion
const string op::v1::ReduceSum::type_name{"ReduceSum"};
```

src/ngraph/op/product.cpp
```diff
@@ -19,22 +19,48 @@
...
-shared_ptr<Node> op::Product::copy_with_new_args(const NodeVector& new_args) const
+shared_ptr<Node> op::v0::Product::copy_with_new_args(const NodeVector& new_args) const
 {
     check_new_args_count(this, new_args);
     return make_shared<Product>(new_args.at(0), get_reduction_axes());
 }
+
+const string op::v1::ReduceProd::type_name{"Product"};
+
+op::v1::ReduceProd::ReduceProd(const Output<Node>& arg,
```
nmostafa: Why the class name change if type_name is the same ? why not `Product` ? Same for `ReduceSum`

postrational: We are making this change to align with our new operator specification.

src/ngraph/op/product.cpp
mbencer: Done

src/ngraph/op/product.cpp
mbencer: I consider this case - for me also `bool` is obvious in such case, but I was based on the specification. Anyway I changed type to `bool` - probably it will be acceptable ;)

src/ngraph/op/sum.cpp
mbencer: Done. I think about it, the side effect is lack of support by `get_typeid` (in this moment it is not used). Probably is will be resolved during [Remove static initializers from ops](https://github.com/NervanaSystems/ngraph/pull/3533/files)

diyessi: Let's defer until later, maybe as part of a serializer cleanup.

src/ngraph/op/product.hpp
```diff
@@ -23,36 +23,90 @@ namespace ngraph
...
+                    copy_with_new_args(const NodeVector& new_args) const override;
+            };
+        }
+
+        namespace v1
+        {
+            /// \brief Product reduction operation.
+            ///
+            /// Reduces the tensor, eliminating the specified reduction axes by taking the product.
+            class ReduceProd : public util::ArithmeticReduction
```
ilyachur: From my point of view we should add a new cpp/hpp files with new name. It looks inconvenient that product.hpp contains ReduceProd class.

mbencer: I agree with you that it is a bit incompatible with correct code style but by now (as I know) it is only one case when name of ops has been changed between opsets versions. I have no clear attitude to your proposition. @aprocter, @diyessi What do you think about it?

aprocter: I agree with @ilyachur here, but I'd be open to fixing it in a later PR.

mbencer: Done. `ReduceProd` and `ReduceSum` were placed to separated files.

mbencer:
> I don't think this is updating validation to correctly handle `keep_dims`?

@aprocter Very important comment, thank you ;) I changed calculation output shape if `keep_dims==true` (according to specification in such case reduction axes should have dimensions set as 1).

test/opset_pass/product_opset1upgrade_pass.cpp
```diff
@@ -0,0 +1,48 @@
+//*****************************************************************************
```
postrational: We should simplify this filename.

test/opset_pass/sum_opset1upgrade_pass.cpp
```diff
@@ -0,0 +1,48 @@
+//*****************************************************************************
```
postrational: We should simplify this filename to match others.

src/ngraph/op/reduce_prod.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "ngraph/op/reduce_prod.hpp"
+#include "ngraph/graph_util.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+constexpr NodeTypeInfo op::v1::ReduceProd::type_info;
+
+op::v1::ReduceProd::ReduceProd(const Output<Node>& arg,
+                               const AxisSet& reduction_axes,
```
GlebKazantaev: Actually we don't need this constructor due to ReduceProd consumes two inputs according to spec.

src/ngraph/op/reduce_sum.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include "ngraph/op/reduce_sum.hpp"
+#include "ngraph/graph_util.hpp"
+#include "ngraph/op/broadcast.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+constexpr NodeTypeInfo op::v1::ReduceSum::type_info;
+
+op::v1::ReduceSum::ReduceSum(const Output<Node>& arg, const AxisSet& reduction_axes, bool keep_dims)
```
GlebKazantaev: The same thing as with ReduceProd.

src/ngraph/op/sum.hpp
```diff
@@ -72,34 +72,39 @@ namespace ngraph
...
-            ///
-            /// \param arg The tensor to be summed.
-            /// \param reduction_axes The axis positions (0-based) to be eliminated.
-            Sum(const Output<Node>& arg, const AxisSet& reduction_axes);
-            /// \brief Constructs a summation operation.
-            ///
-            /// \param arg The tensor to be summed.
-            /// \param reduction_axes The axis positions (0-based) to be eliminated.
-            Sum(const Output<Node>& arg, const Output<Node>& reduction_axes);
+            class Sum : public util::ArithmeticReduction
```
aprocter: Docstring for the class should be moved down as well.

src/ngraph/serializer.cpp
```diff
@@ -1608,7 +1610,15 @@ shared_ptr<Node> JSONDeserializer::deserialize_node(json node_js)
         case OP_TYPEID::Product:
         {
             auto reduction_axes = deserialize_axis_set(node_js.at("reduction_axes"));
-            node = make_shared<op::Product>(args[0], reduction_axes);
+            if (op_version == 0)
+            {
+                node = make_shared<op::Product>(args[0], reduction_axes);
```
aprocter: `v0` actually supports dynamic `reduction_axes` (meaning the old logic was broken :grimacing:). For v0, can we check if `reduction_axes` is missing, and call the two-input constructor on `args[0], args[1]` in that case? On the serializer side we will always want to treat `reduction_axes` as an input rather than an attribute.

aprocter: OTOH, `v1` should _always_ treat `reduction_axes` as an input, so please make that change too.

src/ngraph/serializer.cpp
```diff
@@ -2711,8 +2729,17 @@ json JSONSerializer::serialize_node(const Node& n)
...
-        node["reduction_axes"] = serialize_axis_set(tmp->get_reduction_axes());
+        if (op_version == 0)
+        {
+            auto tmp = dynamic_cast<const op::Product*>(&n);
+            node["reduction_axes"] = serialize_axis_set(tmp->get_reduction_axes());
+        }
+        if (op_version == 1)
+        {
+            auto tmp = dynamic_cast<const op::v1::ReduceProd*>(&n);
+            node["reduction_axes"] = serialize_axis_set(tmp->get_reduction_axes());
```
aprocter: Not an attribute in `v1`.

src/ngraph/serializer.cpp
```diff
@@ -2711,8 +2729,17 @@ json JSONSerializer::serialize_node(const Node& n)
     }
     case OP_TYPEID::Product:
     {
-        auto tmp = dynamic_cast<const op::Product*>(&n);
-        node["reduction_axes"] = serialize_axis_set(tmp->get_reduction_axes());
+        if (op_version == 0)
+        {
+            auto tmp = dynamic_cast<const op::Product*>(&n);
+            node["reduction_axes"] = serialize_axis_set(tmp->get_reduction_axes());
```
aprocter: Not an attribute in `v0`.

src/ngraph/serializer.cpp
```diff
@@ -1891,7 +1901,15 @@ shared_ptr<Node> JSONDeserializer::deserialize_node(json node_js)
         case OP_TYPEID::Sum:
```
aprocter: Ditto comments for `Product`.

src/ngraph/serializer.cpp
```diff
@@ -2904,8 +2931,17 @@ json JSONSerializer::serialize_node(const Node& n)
     }
     case OP_TYPEID::Sum:
```
aprocter: Ditto comments for `Product`.

src/ngraph/op/reduce_prod.cpp
mbencer: Done

src/ngraph/op/reduce_sum.cpp
mbencer: Done

src/ngraph/op/sum.hpp
mbencer: Done

src/ngraph/serializer.cpp
mbencer: Done
- pass `args[1]` when `reduction_axes` is empty (not serialized) for `v0::Product`
- pass `args[1]` always for `v1::ReduceProd`

src/ngraph/serializer.cpp
mbencer: Done
- pass `args[1]` when `reduction_axes` is empty (not serialized) for `v0::Sum`
- pass `args[1]` always for `v1::ReduceSum`

src/ngraph/serializer.cpp
mbencer: Done, removed

src/ngraph/serializer.cpp
mbencer: Done, removed

src/ngraph/serializer.cpp
mbencer: Done, removed

test/opset_pass/product_opset1upgrade_pass.cpp
mbencer: Done, name changed to `product_opset_pass.cpp`

test/opset_pass/sum_opset1upgrade_pass.cpp
mbencer: Done, name changed to `sum_opset_pass.cpp`

src/ngraph/op/product.hpp
postrational: Iliya's comment was addressed in https://github.com/NervanaSystems/ngraph/pull/3556/commits/82b6389c3f0a016d7963fa22e297e0a82dd097c2

PR #3559, opened 2019-09-04
[MLIR] Clear SubGraph construction info

nmostafa:
Don't hold to shared pointers after sub-graph extraction pass is done.
Fixes: `onnx_CPU.model_lstm_bdir_short_input_seq`


src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.cpp
```diff
@@ -550,6 +551,12 @@ bool MLIRSubgraphExtractionPass::check_cycles(std::shared_ptr<Node> node,
     return false;
 }
 
+void MLIRSubgraphExtractionPass::clean_up()
+{
+    m_id_to_graph.clear();
+    m_node_to_graph.clear();
+}
+
```
dcaballe: Sorry, could you please provide some context on the problem?

Is MLIRSubgraphExtrantionPass expected to be reused or persist after the CK creation? Is it not possible to destroy somehow MLIRSubgraphExtractionPass instead?

nmostafa: In nGraph, nodes keep shared_ptrs to their predecessors, so if a node feeds no result (output edges removed as in CSE), its shared_ptr ref count reaches zero and the dctr is called. The dctr also unlinks the node from the graph and remove predecessors output edges as well, which may call a pred dct, and so on. Thus the entire sub-graph is deleted and unlinked. If we hold a shared_ptr in the pass we prevent this deletion from happening. Apparently some passes expect that all nodes reach a result node.

PR #3560, opened 2019-09-04
[MLIR] Introduce ngraph-opt tool

dcaballe:
To be used, among other things, for LIT testing.
`elementwise_binary_ops.mlir` shows a simple LIT test
using ngraph-opt.

Dummy affine_lowering/lit_test.mlir test that served as initial testing
of LIT tool configuration is now removed.


src/contrib/mlir/tools/ngraph-opt/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ******************************************************************************
+
+set(LIB_LIBS
+    MLIRPass
+)
+
+add_library(ngraph_opt
```
nmostafa: Do you need the library ? Why not just the executable ? Also, if you must, place rename to something more clear `ngraph_opt_lib` , `ngraph_opt`

dcaballe: I did the rename. I'm following the same approach as for mlir-opt https://github.com/tensorflow/mlir/blob/master/tools/mlir-opt/CMakeLists.txt
I assumed there was some benefit in linking in two steps but I don't know.

src/contrib/mlir/tools/ngraph-opt/ngraph_opt.cpp
```diff
@@ -0,0 +1,81 @@
...
+                       llvm::cl::init(false));
+
+static llvm::cl::opt<bool>
+    verify_passes("verify-each",
+                  llvm::cl::desc("Run the verifier after each transformation pass"),
+                  llvm::cl::init(true));
+
+static std::vector<const mlir::PassRegistryEntry*>* pass_list;
+
+int main(int argc, char** argv)
```
postrational: Could you add some documentation for the tool and its CLI?

dcaballe: Sorry, I was focused on something else and the patch went in :). I'll address this in a separate commit.

PR #3562, opened 2019-09-04
Fixed NormalizeL2 decomposition for input shape length != 4

mbencer:


src/ngraph/op/fused/normalize_l2.cpp
```diff
@@ -92,12 +82,12 @@ NodeVector op::NormalizeL2::decompose_op() const
     auto builder_bias_mode =
         (m_eps_mode == EpsMode::MAX) ? builder::BiasMode::MAX : builder::BiasMode::ADD;
     Output<Node> norm = builder::l2_norm(data, reduction_axes, m_eps, builder_bias_mode);
-    norm = make_broadcast_node(norm, data.get_shape(), 0);
+    norm = numpy_style_broadcast(norm, data.get_shape());
```
mbencer: It is key change which should fix problems with incorrect shapes

postrational: Let's try to use `auto_broadcast` here.

test/type_prop/normalize.cpp
```diff
@@ -20,48 +20,6 @@
...
-    catch (const NodeValidationFailure& error)
-    {
-        EXPECT_HAS_SUBSTRING(error.what(),
-                             std::string("Input tensor rank must be 2, 3 or 4 dimensional"));
-    }
-    catch (...)
-    {
-        FAIL() << "Deduced type check failed for unexpected reason";
-    }
-}
```
GlebKazantaev: Could you please explain how this test works:
```
Shape data_shape{2, 3, 4};
auto data = make_shared<op::Parameter>(element::f32, data_shape);
const auto axes = make_shared<op::Constant>(element::u64, Shape{3}, vector<int64_t>{1, 2, 3});
```
How you can apply axes{1,2,3} to input_shape{2,3,4}?

src/ngraph/op/fused/normalize_l2.cpp
```diff
@@ -71,7 +61,7 @@ NodeVector op::NormalizeL2::decompose_op() const
     const Shape input_shape{data.get_shape()};
 
     // Reshape to 4D tensor.
-    if (input_shape.size() != 4)
```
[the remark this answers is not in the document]
mbencer: Yes you are right. I confirmed that it should be removed. It is some residue from previous version of specification (when arguments of normalize were `across_spatial` and `channel_shared` were arguments of normalize). After removing this code some unit tests (for example with 2D input across 3D axes) must be removed - but from my point of view such case doesn't make sense.

src/ngraph/op/fused/normalize_l2.cpp
mbencer: Thanks, Using `Divide` with param `AutoBroadcastType::NUMPY` gives expected results

test/type_prop/normalize.cpp
mbencer: I agree with with you. For me it doesn't make sense too. It is case connected with `Reshape` for `input_shape.size() == 4`. I didn't remove it in first version of this PR because I was not sure of somebody use it. Not it was removed.

src/ngraph/op/fused/normalize_l2.cpp
```diff
@@ -42,65 +42,66 @@ op::NormalizeL2::NormalizeL2(const Output<Node>& data,
...
+NodeVector op::NormalizeL2::decompose_op() const
+{
+    Output<Node> data{input_value(0)};
+    const Shape input_shape{data.get_shape()};
 
-    // Calculate norm over axes indicated by axes input param
-    auto axes_constant = dynamic_pointer_cast<op::Constant>(axes_node);
-    auto axes_vector = axes_constant->get_vector<size_t>();
-    AxisSet reduction_axes{axes_vector};
+    AxisSet reduction_axes = get_reduction_axes();
```
GlebKazantaev: Why you need axes set here and why not to pass 1 input (with axes) to the builder::l2_norm?  Otherwise it looks weird that if 1 input is not Constant then axes are empty set.

mbencer: - `l2_norm` takes as input `AxisSet` - we haven't access to `Output<Node>` data if it is not `Constant` in the `op` class. In order to support such case new version of op is need (NormalizeL2:v1) and new backends which can support it.
- In `pre_validate_and_infer_types` method (which is called always in constructor) is assert which provides that `axes_node->is_constant() == true`, so `get_reduction_axes` will be work correctly in any case

GlebKazantaev: Ok

GlebKazantaev:
@mbencer I've checked e2e tests that needs NormalizeL2 op with your changes, and it works! So I approve this changes.

src/ngraph/op/fused/normalize_l2.cpp
```diff
@@ -42,65 +42,66 @@ op::NormalizeL2::NormalizeL2(const Output<Node>& data,
...
-
-    // Input data must be 2, 3 or 4D tensor.
     NODE_VALIDATION_CHECK(this,
-                          (data_shape.size() >= 2 && data_shape.size() <= 4),
-                          "Input tensor rank must be 2, 3 or 4 dimensional (actual input "
-                          "shape: ",
-                          data_shape,
-                          ").");
+                          axes_node->is_constant(),
+                          "doesn't support 'axes' input of other type than a Constant.");
```
tomdol: This looks like an incomplete message/sentence

src/ngraph/op/fused/normalize_l2.cpp
```diff
@@ -42,65 +42,66 @@ op::NormalizeL2::NormalizeL2(const Output<Node>& data,
...
 
-    NODE_VALIDATION_CHECK(this,
-                          static_cast<size_t>(axes_pshape.rank()) == 1,
-                          "Input axes must have rank equals 1 (axes shape: ",
-                          axes_pshape,
-                          ").");
+    if (axes_rank.is_static())
+    {
+        NODE_VALIDATION_CHECK(this,
+                              static_cast<size_t>(axes_pshape.rank()) == 1,
```
tomdol: Perhaps you could use `axes_rank` instead of `axes_pshape.rank()` just for consistency

src/ngraph/op/fused/normalize_l2.cpp
```diff
@@ -42,65 +42,66 @@ op::NormalizeL2::NormalizeL2(const Output<Node>& data,
 
 void op::NormalizeL2::pre_validate_and_infer_types()
 {
-    const auto& data_pshape = get_input_partial_shape(0);
+    auto axes_node = input(1).get_source_output().get_node_shared_ptr();
```
tsocha:
```suggestion
    auto axes_node = input_value(1).get_node_shared_ptr();
```

src/ngraph/op/fused/normalize_l2.cpp
tsocha:
```suggestion
                              static_cast<size_t>(axes_rank) == 1,
```

src/ngraph/op/fused/normalize_l2.cpp
mbencer: Done

src/ngraph/op/fused/normalize_l2.cpp
mbencer: You are right, I copied previous description of this assert from `decompose_op` method. Fixed

src/ngraph/op/fused/normalize_l2.cpp
mbencer: Right, done

mbencer: Done

PR #3563, opened 2019-09-04
 [ONNX] Fix support for float16 in ONNX frontend.

dcaballe:
Initial support for float16 was done before the ngraph::float16 data
type was introduced in nGraph. This PR aligns ONNX frontend with the
latest float16 implementation.


src/ngraph/frontend/onnx_import/core/tensor.hpp
```diff
@@ -130,14 +153,14 @@ namespace ngraph
...
-                        return detail::__get_raw_data<double>(tensor.raw_data());
+                        return detail::__get_raw_data<double>(tensor.raw_data(),
+                                                              tensor.data_type());
                     }
                     if (tensor.data_type() == onnx::TensorProto_DataType_DOUBLE)
                     {
                         return detail::__get_data<double>(tensor.double_data());
                     }
-                    if ((tensor.data_type() == onnx::TensorProto_DataType_FLOAT) ||
-                        (tensor.data_type() == onnx::TensorProto_DataType_FLOAT16))
```
tsocha: If we would get a tensor with `FLOAT16` data type which is not stored in raw_data field we will get an exception.
You should probably move `(tensor.data_type() == onnx::TensorProto_DataType_FLOAT16))` check to INT32 section instead of just removing it.
ONNX doc:
```
// For int32, uint8, int8, uint16, int16, bool, and float16 values
// float16 values must be bit-wise converted to an uint16_t prior
// to writing to the buffer.
// When this field is present, the data_type field MUST be
// INT32, INT16, INT8, UINT16, UINT8, BOOL, or FLOAT16
repeated int32 int32_data = 5 [packed = true];
```
@postrational what do you think about it?

src/ngraph/frontend/onnx_import/core/tensor.hpp
```diff
@@ -109,11 +107,36 @@ namespace ngraph
...
+                            {
+                            case onnx::TensorProto_DataType_FLOAT: return sizeof(float);
+                            case onnx::TensorProto_DataType_UINT8: return sizeof(uint8_t);
+                            case onnx::TensorProto_DataType_INT8: return sizeof(int8_t);
+                            case onnx::TensorProto_DataType_UINT16: return sizeof(uint16_t);
+                            case onnx::TensorProto_DataType_INT16: return sizeof(int16_t);
+                            case onnx::TensorProto_DataType_INT32: return sizeof(int32_t);
+                            case onnx::TensorProto_DataType_INT64: return sizeof(int64_t);
+                            case onnx::TensorProto_DataType_BOOL: return sizeof(bool);
+                            case onnx::TensorProto_DataType_FLOAT16: return 2;
```
tsocha: Can we use `sizeof(ngraph::float16)` instead of hard-coding `2`?

diyessi: Technically, we should be verifying that the actual sizes in ONNX match the sizes of the types here. For example, `float` isn't guaranteed to be 32 bits or even IEEE format.

src/ngraph/frontend/onnx_import/core/tensor.hpp
dcaballe: Ok, I could do that but I don't have a way to test that case. Could someone provide one? Otherwise, maybe it's better to get an exception and implement it once we can test it.

tsocha: Ok, we can defer it for now I think.

PR #3564, opened 2019-09-04
[MLIR] Fix style in compiler and lowerer files

nmostafa:
For compiler.cpp/hpp, use nGraph naming convention for public methods only. LLVM naming convention for everything else. 

For lowerer.cpp, fix inconsistensies. Use LLVM naming convention.


src/contrib/mlir/compiler/compiler.cpp
```diff
@@ -14,8 +14,8 @@
 // limitations under the License.
 //*****************************************************************************
 
-// NOTE: This file follows nGraph format style and naming convention since it
-// exposes a public API to the rest of nGraph codebase.
+// NOTE: This file follows nGraph format style, and nGraph naming convention for public method names only, 
+// since they are exposed to the rest of nGraph code-base.
```
dcaballe: Something is missing here. We should say that we use MLIR format style and naming convention for the rest.

src/contrib/mlir/compiler/compiler.hpp
```diff
@@ -14,8 +14,8 @@
 // limitations under the License.
 //*****************************************************************************
 
-// NOTE: This file follows nGraph format style and naming convention since it
-// exposes a public API to the rest of nGraph codebase.
+// NOTE: This file follows nGraph format style, and nGraph naming convention for public method names only, 
+// since they are exposed to the rest of nGraph code-base.
```
dcaballe: same

nmostafa:
Fixed comment in two files and other files as well since we use MLIR code format and naming convention throughout.

PR #3568, opened 2019-09-05
[MLIR] Remove Fake Instructions

nmostafa:
Re-map function signature instead to add sub-graph result as a function parameter. Use output parameters as defined output symbol instead of inserting a fake def. 

Fake defs for MemMgr are also removed, in addition to code that changed the defs into a function argument.


src/contrib/mlir/compiler/lowerer.cpp
```diff
@@ -157,17 +198,15 @@ namespace
...
         OwningRewritePatternList patterns;
-        // Add default FuncOp type conversion. It replaces the incoming FuncOp with a *new* one
-        // with the converted types.
-        mlir::populateFuncOpTypeConversionPattern(patterns, &getContext(), typeConverter);
+
         populateNGraphToAffineConversionPatterns(patterns);
 
         // Create target that defines legal ops for nGraph dialect to be lowered to.
         ConversionTarget target(getContext());
         // TODO: Remove NGFakeInputOp. We need to set NGFakeInputOp as legal op because we generate
```
ayzhuang: Remove this comment.

PR #3572, opened 2019-09-05
Update Json to version 3.7.0

dcaballe:
This version doesn't crash while parsing large serialized models.


silee2:
Closing since #4261 is merged.

PR #3574, opened 2019-09-06
[Spec] Implement Pad:v1

mbencer:
Change to add a `v1` version of `Pad` op, which replaces static attributes `m_padding_below` and `m_padding_above` with dynamic inputs.


test/CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ set(SRC
     nop_elimination.cpp
     op.cpp
     opset_pass/softmax_opset_pass.cpp
+    opset_pass/pad_opset1upgrade_pass.cpp
```
postrational: I think we may want to keep all pad-related pass tests in one file. I would consider renaming this to `pad_opset_pass.cpp`

mbencer: Done

src/ngraph/op/pad.cpp
```diff
@@ -174,3 +174,97 @@ std::shared_ptr<Node> op::Pad::get_default_value() const
     }
     return std::make_shared<op::Broadcast>(input_value(1), get_shape(), axes);
 }
+
+const string op::v1::Pad::type_name{"Pad"};
+
+op::v1::Pad::Pad(const Output<Node>& arg,
+                 const Output<Node>& pads_begin,
+                 const Output<Node>& pads_end,
+                 const Output<Node>& arg_pad_value,
```
jbobba: Any reason for this node ordering? arg, arg_pad_value, pads_begin, pads_end would be more consistent with v0.

mbencer: I've thought about it. Such ordering is defined in `Pad` specification. I am not sure if we have to keep specification order or be consistent with `v:0`. @ilyachur @GlebKazantaev What do you think about it?

src/ngraph/serializer.cpp
```diff
@@ -2676,10 +2685,18 @@ json JSONSerializer::serialize_node(const Node& n)
     }
     case OP_TYPEID::Pad:
     {
-        auto tmp = dynamic_cast<const op::Pad*>(&n);
-        node["padding_below"] = tmp->get_padding_below();
-        node["padding_above"] = tmp->get_padding_above();
-        node["pad_mode"] = tmp->get_pad_mode();
+        if (op_version == 0)
+        {
+            auto tmp = dynamic_cast<const op::Pad*>(&n);
```
aprocter:
```suggestion
            auto tmp = dynamic_cast<const op::v0::Pad*>(&n);
```

(Best to be explicit when dealing with "version-aware" code, IMHO)

src/ngraph/op/pad.cpp
```diff
@@ -174,3 +174,97 @@ std::shared_ptr<Node> op::Pad::get_default_value() const
     }
     return std::make_shared<op::Broadcast>(input_value(1), get_shape(), axes);
 }
+
+const string op::v1::Pad::type_name{"Pad"};
+
+op::v1::Pad::Pad(const Output<Node>& arg,
```
aprocter: The spec doesn't say so, but for other ops things like `pads_begin` and `pads_end` are required to be `int64`. Should we do that here?

In any case, if `pads_begin` and `pads_end` are constant, we should still have the checks from the `v0` op for negative dim after padding, and for the input dimensions being too small in the EDGE or REFLECT modes.

test/type_prop/pad.cpp
```diff
@@ -402,3 +402,156 @@ TEST(type_prop, pad_partial_data_rank_dynamic_padding_static_attribs_rank_incons
...
+    auto pads_begin = make_shared<op::Parameter>(element::i64, Shape{1});
+    auto pads_end = make_shared<op::Parameter>(element::i64, Shape{1});
+    auto arg_pad_value = make_shared<op::Parameter>(element::f16, Shape{1});
+
+    try
+    {
+        auto pad = make_shared<op::v1::Pad>(arg, pads_begin, pads_end, arg_pad_value);
+
+        // Should have thrown, so fail if it didn't
+        FAIL() << "Too much negative padding not detected";
```
aprocter: Wrong message here.

src/ngraph/op/pad.cpp
GlebKazantaev: Arguments order the same as in spec and that's right.

src/ngraph/op/pad.cpp
```diff
@@ -174,3 +174,97 @@ std::shared_ptr<Node> op::Pad::get_default_value() const
...
+                 const Output<Node>& pads_begin,
```
GlebKazantaev: According to spec third input should be optional. And third input is necessary only when pad_mode='constant'.

src/ngraph/op/pad.cpp
```diff
@@ -174,3 +174,97 @@ std::shared_ptr<Node> op::Pad::get_default_value() const
...
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::v1::Pad::validate_and_infer_types()
+{
+    element::Type result_et;
+
+    const auto& arg_element_type = get_input_element_type(0);
+    const auto& arg_pad_element_type = get_input_element_type(3);
```
GlebKazantaev: How it will work with optional 3rd input ?

src/ngraph/op/pad.hpp
```diff
@@ -24,61 +24,112 @@ namespace ngraph
...
+                /// \param pads_begin The node which specifies the number of padding elements at the
+                /// beginning of each axis
+                /// \param pads_end The node which specifies the number of padding elements at the
+                /// beginning of each axis
+                /// \param pad_mode The padding mode: CONSTANT(default), EDGE, REFLECT or SYMMETRIC.
+                Pad(const Output<Node>& arg,
+                    const Output<Node>& pads_begin,
+                    const Output<Node>& pads_end,
+                    const Output<Node>& arg_pad_value,
+                    PadMode pad_mode = PadMode::CONSTANT);
```
GlebKazantaev: pad_mode is required and has no default value.

src/ngraph/op/pad.cpp
```diff
@@ -174,3 +175,211 @@ std::shared_ptr<Node> op::Pad::get_default_value() const
...
+
+    NODE_VALIDATION_CHECK(this,
+                          pads_end_element_type.compatible(element::Type_t::i64),
+                          "pads_end must be type i64 (axes type: ",
+                          pads_end_element_type,
+                          ").");
+
+    const auto& pads_begin_shape = get_input_partial_shape(1);
+    const auto& pads_begin_rank = pads_begin_shape.rank();
+    if (pads_begin_rank.is_static())
```
aprocter: You can collapse this and the check inside the `if` down to `pads_begin_rank.compatible(1)`, but no matter.

PR #3579, opened 2019-09-07
Attribute visitor

diyessi:
Visitor for node attributes, providing read/write access via standard types, as well as variant-like access via actual attribute type.
Can be extended for additional attribute types.


test/serialize.cpp
```diff
@@ -440,3 +440,85 @@ TEST(serialize, opset1_pad)
...
+enum class TuringModel
+{
+    XL400,
+    XL1200
+};
+
+namespace ngraph
+{
+    template <>
+    EnumNames<TuringModel>* EnumNames<TuringModel>::get()
```
ilyachur: Why do you use raw pointer and don't return a reference to a local static object?

diyessi: Too late in the day when I wrote it 😄

src/ngraph/type.hpp
```diff
@@ -71,4 +80,133 @@ namespace ngraph
...
+            return as_type<std::string>(TypeAdapter<Type>::m_value);
+        }
+        void set_string(const std::string& value) const override
+        {
+            TypeAdapter<Type>::m_value = as_type<Type>(value);
+        }
+    };
+
+    template <typename Type>
+    class ObjectAdapter : public TypeAdapter<Type>
```
ilyachur: What is the purpose of Object adapter? Will it serialise an abstract object?

src/ngraph/attribute_visitor.hpp
```diff
@@ -0,0 +1,46 @@
...
+    ///
+    /// Attributes are the values set when building a graph which are not
+    /// computed as the graph executes. Values computed from the graph topology and attributes
+    /// during compilation are not attributes.
+    class AttributeVisitor
+    {
+    public:
+        virtual ~AttributeVisitor() {}
+        virtual void on_attribute(const std::string& name, std::string& value) = 0;
+        virtual void on_attribute(const std::string& name, element::Type& value) = 0;
```
ilyachur: Why do you add `on_attribute` methods for `element::Type` and `PartialShape`? It is not basic types.

Why don't you want to use `VisitorAdapter` for all types? It looks like current solution has some specific adapters which are implemented in `on_attribute` methods (for `std::string`, `element::Type`, `bool` and etc.).

We can divide access to attribute and it serialisation/deserialisation, if visitor contains only one method `on_attribute` with  `VisitorAdapter` argument and we have default adapters for all basic types.

src/ngraph/attribute_visitor.hpp
```diff
@@ -0,0 +1,46 @@
...
+        virtual void on_attribute(const std::string& name, PartialShape& value) = 0;
+        virtual void on_attribute(const std::string& name, bool& value) = 0;
+        virtual void on_attribute(const std::string& name, int64_t& value) = 0;
+        virtual void on_attribute(const std::string& name, uint64_t& value) = 0;
+        virtual void on_attribute(const std::string& name,
```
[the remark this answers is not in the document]
diyessi: yes, as needed.

src/ngraph/type.hpp
diyessi: Yes. For Enums, we use the string of the enum element, for structures we need the type_info if it's variant, and the members, where the members can be done with the attribute visitor.

src/ngraph/attribute_visitor.hpp
diyessi: Mainly because I'm hooking into the current serializer right now and those two are serialized in ad-hoc ways. If we made switched to a new uniform format, we wouldn't need those special-cased.

src/ngraph/type.hpp
diyessi: Vectors are another special case for serialization; length and elements.

src/ngraph/attribute_visitor.hpp
ilyachur: Ok, I got it.

But why cannot you remove these special-cased methods and add a code on the serialiser part?
I understand that it can be WA but it will allow us to avoid some redundant methods in the new API.
Because users can be confused why these objects have privileges over other objects and can be used without adapters.

src/ngraph/attribute_visitor.hpp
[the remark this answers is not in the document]
ilyachur: But from my point of view we must implement these methods for all basic types at the same time or have only one method with VisitorAdapter for all type. Because in another case if we didn't implement (for example) uint8_t, but user wants to serialise/deserialise this type in own custom op, he will have to write a special adapter in order to work with it.

It is inconvenient.

src/ngraph/attribute_visitor.hpp
```diff
@@ -0,0 +1,52 @@
...
+    class AttributeVisitor
+    {
+    public:
+        virtual ~AttributeVisitor() {}
+        virtual void on_attribute(const std::string& name, std::string& value) = 0;
+        virtual void on_attribute(const std::string& name, element::Type& value) = 0;
+        virtual void on_attribute(const std::string& name, PartialShape& value) = 0;
+        virtual void on_attribute(const std::string& name, bool& value) = 0;
+        virtual void on_attribute(const std::string& name, size_t& value) = 0;
+        virtual void on_attribute(const std::string& name, std::vector<size_t>& value) = 0;
```
ilyachur: Why cannot we use adapter for all vectors? Vector is not basic type.

src/ngraph/attribute_adapter.hpp
```diff
@@ -0,0 +1,140 @@
...
+        virtual std::string get_string() const = 0;
+        ///
+        virtual void set_string(const std::string& value) const = 0;
+    };
+
+    class IntegerVectorAdapter
+    {
+    public:
+        virtual ~IntegerVectorAdapter(){};
+        static constexpr DiscreteTypeInfo type_info{"StringAdapter", 0};
```
ilyachur: Incorrect type info

src/ngraph/attribute_adapter.hpp
```diff
@@ -0,0 +1,140 @@
...
+        virtual ~StringAdapter() {}
+        static constexpr DiscreteTypeInfo type_info{"StringAdapter", 0};
+        virtual const DiscreteTypeInfo& get_type_info() const { return type_info; }
+        /// Returns the value as a string
+        virtual std::string get_string() const = 0;
+        ///
+        virtual void set_string(const std::string& value) const = 0;
+    };
+
+    class IntegerVectorAdapter
```
ilyachur: Maybe we can add template VectorAdapter which will allow to work with any vectors.
Or VectorAdapter will work together other adapters for specific types.

ilyachur:
@wenzhe-nrv @jbobba @silee2  Could you please take a look to this PR?

ilyachur:
@silee2 We are waiting only for your review.

PR #3581, opened 2019-09-08
New method for unit test control

rkimballn1:
The PR adds a new method for controlling unit test disable behavior. The old method is to define a unit test manifest file and that file is parsed for disabled tests. The new method calls a user-defined `ngraph::unit_test_control_handler` which can check certain runtime options such as which version of a device is being tested.
This allows for a single unit test executable to select one of a number of manifest files depending on which version of a device is being tested. It also allows for some non manifest file method for disabling gtest unit tests.

~~Usage:~~
~~If you do nothing you get the old behavior.~~

~~If you add the compile define `NGRAPH_UNIT_TEST_CONTROL_FUNCTION_DEFINED` then you must define a new function with the signature~~
```
string unit_test_control_function(const string& backend_name,
                                  const string& test_name,
                                  const string& manifest)
```
~~which is called for each unit test~~


rkimballn1:
@diyessi Unfortunately your suggestion does not work. gtest configures all of tests as part of static initialization so there is no opportunity for the user to override the default handler. By the time user code runs in main() gtest is already done configuring tests.
An alternative that does not require conditional compile is for each user of the tests be required to implement a callback function to handle unit test config.

rkimballn1:
Blocked. I need to verify this actually works with a backend in a separate repository.

PR #3582, opened 2019-09-09
[SPEC] Adjust Concat to specification

mbencer:
According to specification `m_concatenation_axis` should be changed to `m_axis`.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -4,7 +4,7 @@
 # Default rule: @diyessi gets everything.
 *                                   @diyessi
 
-/cmake/                             @rkimballn1 @silee2 @aslepko
+/cmake/                             @silee2 @aslepko
 
 /.ci/                               @aslepko
 /.ci/travis/                        @aslepko @postrational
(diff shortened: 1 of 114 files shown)


jbobba:
@mbencer @postrational why isn't this change creating v1::concat?

mbencer:
> @mbencer @postrational why isn't this change creating v1::concat?

@jbobba 
I've considered such solution for `Concat` but the only change is changing `m_concatenation_axis` by `m_axis`. In such case we have to copy almost all `Concat` code without any changes.
On the other hand as @diyessi mentioned solution without introduction `v1` forces change of other transformers which use `Concat`.

Probably the compromise can be introduction `Concat:v1` which inherits from `Concat:v0` and only override `m_concatenation_axis` - I prepared draft of such solution: https://github.com/NervanaSystems/ngraph/pull/3616/files
@postrational @diyessi @jbobba What do you think about it?

postrational:
I think we can simplify this PR by leaving `get_concatenation_axis` as an alias for `get_axis`.

mbencer:
> I think we can simplify this PR by leaving `get_concatenation_axis` as an alias for `get_axis`.

For me it is good idea ;) It is worth to emphasize that during serialization `axis` name is used - so it seems to be consistent. I implemented the change.

mbencer:
@aslepko, @indie, @silee2 and @wenzhe-nrv were added as reviewers by incorrect commits, so I removed them from review ;)

PR #3583, opened 2019-09-09
[MLIR] Encapsulate internal nodes of CompiledKernel Op to prevent lat…

ayzhuang:
…er passes from accessing them.


src/ngraph/runtime/cpu/builder/mlir_cpu_compiled_kernel.cpp
```diff
@@ -38,6 +38,17 @@ namespace ngraph
             {
                 auto& functors = external_function->get_functors();
 
+                CompiledKernel* compiled_kernel =
```
nmostafa: I don't think fixing up the graph here is a good idea. Why not just leave it as-is and let compiler do the write look up to resolve the tensor ptr.

src/ngraph/op/experimental/compiled_kernel.hpp
```diff
@@ -47,9 +47,18 @@ namespace ngraph
 
             const NodeVector& get_node_list() const { return m_node_list; }
             const NodeVector& get_kernel_outputs() const { return m_output_nodes; }
+            const std::vector<std::tuple<std::shared_ptr<Node>, size_t, size_t>>& get_vector() const
```
nmostafa:
```suggestion
            const std::vector<std::tuple<std::shared_ptr<Node>, size_t, size_t>>& get_input_map() const
```

src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.hpp
```diff
@@ -127,6 +127,10 @@ namespace ngraph
             NodeVector build_ck_nodes(std::shared_ptr<Function> func);
 
             void sanity_check(std::shared_ptr<Function> func, NodeVector& ck_nodes);
+            // For node B inside CompiledKernel ck such that A->B and A is outside of ck:
+            // replace input to B with a Label Op and add an entry to ck's
+            // m_node_arg_index_ck_arg_indx.
+            void encapsulate_nodes(NodeVector& ck_nodes);
```
nmostafa: You can move this to be part of CK node itself and add Label nodes and map updates during ck construction instead.

src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.cpp
```diff
@@ -39,6 +39,7 @@
 #include "ngraph/op/negative.hpp"
 #include "ngraph/op/relu.hpp"
 #include "ngraph/op/subtract.hpp"
+#include "ngraph/pattern/op/label.hpp"
```
nmostafa: I rather not abuse label nodes and maybe add a different node for that purpose. TensorIterator should have the same issue. Let's discuss it more.

src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.cpp
```diff
@@ -384,6 +386,45 @@ void MLIRSubgraphExtractionPass::sanity_check(std::shared_ptr<Function> func, No
     }
 }
 
+void MLIRSubgraphExtractionPass::encapsulate_nodes(NodeVector& ck_nodes)
```
nmostafa: Again, this should be part of CK construction logic and not part of the pass. In case we need to construct CKs elsewhere.

src/contrib/mlir/compiler/compiler.cpp
```diff
@@ -629,9 +629,22 @@ mlir::Operation* MLIRCompiler::create_generic_op(const ngraph::Node* ng_node)
 {
     std::vector<mlir::Value*> arg_values;
     std::vector<mlir::Type> res_types;
+    auto input_map = m_compiled_kernel->get_input_map();
+    std::shared_ptr<descriptor::Tensor> arg_tensor;
     for (auto& arg : ng_node->get_arguments())
     {
-        auto arg_tensor = arg->get_output_tensor_ptr();
+        if (std::dynamic_pointer_cast<ngraph::op::Parameter>(arg))
```
diyessi: `is_type<ngraph::op::Parameter>(arg)`

src/ngraph/op/experimental/compiled_kernel.cpp
```diff
@@ -45,7 +45,15 @@ shared_ptr<Node> ngraph::op::CompiledKernel::copy_with_new_args(const NodeVector
         OutputVector cur_args;
         for (auto a : n->input_values())
         {
-            cur_args.push_back(a.for_node(nm.at(a.get_node())));
+            if (dynamic_cast<ngraph::op::Parameter*>(a.get_node()))
```
diyessi: `as_type<op::Parameter>(a.get_node())`

src/contrib/mlir/compiler/compiler.cpp
```diff
@@ -629,9 +629,22 @@ mlir::Operation* MLIRCompiler::create_generic_op(const ngraph::Node* ng_node)
 {
     std::vector<mlir::Value*> arg_values;
     std::vector<mlir::Type> res_types;
+    auto input_map = m_compiled_kernel->get_input_map();
+    std::shared_ptr<descriptor::Tensor> arg_tensor;
     for (auto& arg : ng_node->get_arguments())
```
diyessi: If you switched this to use `input_values` it would work with values from multi-output nodes.

src/contrib/mlir/compiler/compiler.cpp
```diff
@@ -637,25 +637,39 @@ namespace ngraph
 template <typename Op>
 mlir::Operation* MLIRCompiler::createGenericOp(const ngraph::Node* ngNode)
 {
-    std::vector<mlir::Value*> argValues;
-    std::vector<mlir::Type> resTypes;
-    for (auto& arg : ngNode->get_arguments())
+    std::vector<mlir::Value*> arg_values;
```
dcaballe: nit: we are following MIR naming convention in this file so we should use `lowerCamelCase` case for all the variable names, params, function names, etc. You can check the naming convention used in MLIR files by looking at the NOTE at the beginning of the file.

PR #3586, opened 2019-09-10
[MLIR] Add padding support to Convolution

nmostafa:
Do a range check on the Img spatial indices and perform computation if within non-padded region. 

Depends on #3584


src/contrib/mlir/compiler/lowerer.cpp
```diff
@@ -804,11 +794,13 @@ namespace
         //           Output[n, k, r_1, .. r_f] +=
         //             Images[n, c, i_1 + j_1, .. i_f + j_f] * Filters[k, c, j_1, .. j_f]
 
-        // TODO: With padding, we need to check (using IntegerSets) whether each spatial dim in
-        // Images lie within paddings
-        // If yes, we init value to zero, else load from MemRef.
-        // Q: Can this be done using a map from padded tensor to  unpadded one ? Will we load zero
-        // if OOB ?
+        // With padding, we check (using IntegerSets) whether each spatial dim in Images lie within
+        // paddings. If true, we perform the computation:
```
dcaballe: "lie within paddings" sounds as if we performed computation for elements in the padding. Should we say outside of padding instead?

src/contrib/mlir/compiler/lowerer.cpp
```diff
@@ -842,6 +837,22 @@ namespace
...
+                withPadding = true;
+            }
+            padBelowIntValues.push_back(padValue);
+
+            iAttr = padAbove[i].cast<IntegerAttr>();
+            padValue = iAttr.getInt();
+            if (padValue)
+            {
+                withPadding = true;
+            }
```
dcaballe: For my understanding... we don't need `padAboveValues`?

src/contrib/mlir/compiler/lowerer.cpp
```diff
@@ -860,6 +871,41 @@ namespace
...
+        IntegerSet nonPaddedRange;
+        if (withPadding)
+        {
+            // Create affine expressions and IntegerSet
+            // IntegerSet (d0, d1, .. d_N-1)[LB_0, LB_1, .. LB_N-1, UB_0, UB_1, .. UB_N-1], where
+            // for each dim:
+            //   (d_dim - padBelow[dim] - LB_dim >= 0),
+            //   (padBelow[dim] + UB_dim - d_dim - 1 >= 0)
+            SmallVector<AffineExpr, 4> affineExprs;
+            SmallVector<bool, 4> isEq;
```
dcaballe: Please, doc for `isEq`

src/contrib/mlir/compiler/lowerer.cpp
```diff
@@ -928,8 +976,47 @@ namespace
                                                   filtersSpatialIndices.begin(),
                                                   filtersSpatialIndices.end());
 
-                            iRes(resIndices) =
-                                iRes(resIndices) + (iImages(imgIndices) * iFilters(filtersIndices));
+                            if (withPadding)
+                            {
+                                // if args : img dims, img lbs, img ubs
+                                SmallVector<IndexHandle, 4>::iterator it = imgIndices.begin();
```
dcaballe: std::advance(imgIndices.begin(), 2)

test/backend/convolution.in.cpp
```diff
@@ -99,3 +99,40 @@ NGRAPH_TEST(${BACKEND_NAME}, convolution_simple)
...
+    vector<float> expected_result{0.0f, 0.0f, 0.0f, 0.0f, 0.0f,
+                                  0.0f, 2.0f, 4.0f, 0.0f, 0.0f,
+                                  0.0f, 6.0f, 8.0f, 0.0f, 0.0f,
+                                  0.0f, 0.0f, 0.0f, 0.0f, 0.0f,
+                                  0.0f, 0.0f, 0.0f, 0.0f, 0.0f};
+    // clang-format on
+    auto handle = backend->compile(f);
+    handle->call_with_validate({result}, {a, b});
+    EXPECT_TRUE(test::all_close_f(vector<float>{expected_result}, read_vector<float>(result)));
+}
```
dcaballe: Could you please add a lit test to cover the lowering of a convolution with padding? For example to `mlir/lowering/convolution.mlir`

nmostafa: Wouldn't we still need to fix ngraph-opt to run the lowering pass ?

src/contrib/mlir/compiler/lowerer.cpp
nmostafa: Not to the check if we are inside the non-padded range. PadAbove is actually included in the inferred output shape early on in nGraph, so the loop bounds are adjusted here to include padAbove though.

test/backend/convolution.in.cpp
dcaballe: Yes, working on that. Sorry

PR #3587, opened 2019-09-10
Changed main cmake script

ilyachur:
I changed CMake script in order to support:

- Ninja generator
- ARM platforms
- Custom build path


CMakeLists.txt
```diff
@@ -292,6 +288,14 @@ message(STATUS "NGRAPH_INTERPRETER_STATIC_LIB_ENABLE: ${NGRAPH_INTERPRETER_STATI
 message(STATUS "NGRAPH_CPU_STATIC_LIB_ENABLE:         ${NGRAPH_CPU_STATIC_LIB_ENABLE}")
 message(STATUS "NGRAPH_DYNAMIC_COMPONENTS_ENABLE:     ${NGRAPH_DYNAMIC_COMPONENTS_ENABLE}")
 
+if("${CMAKE_GENERATOR}" STREQUAL "Ninja" AND (NGRAPH_TOOLS_ENABLE OR NGRAPH_CPU_ENABLE OR NGRAPH_MLIR_ENABLE OR
+        NGRAPH_INTELGPU_ENABLE OR NGRAPH_GPU_ENABLE OR NGRAPH_INTERPRETER_ENABLE OR NGRAPH_NOP_ENABLE OR
+        NGRAPH_GPUH_ENABLE OR NGRAPH_GENERIC_CPU_ENABLE OR NGRAPH_ONNX_IMPORT_ENABLE OR NGRAPH_DEX_ONLY OR
+        NGRAPH_ENABLE_CPU_CONV_AUTO OR NGRAPH_CODE_COVERAGE_ENABLE OR NGRAPH_PLAIDML_ENABLE OR NGRAPH_JSON_ENABLE))
+    message(FATAL_ERROR "Ninja generator not supported.")
```
jbobba: Not sure if you need "NGRAPH_ENABLE_CPU_CONV_AUTO", "NGRAPH_CODE_COVERAGE_ENABLE" , "NGRAPH_DEX_ONLY"

ilyachur: Ok, I will remove it. Just added it because I wasn't sure that Ninja is supported for these options.

ilyachur: Fixed

silee2:
@ilyachur Let me see if I correctly understand what you are trying to do with this PR. Looks like you are setting up nGraph as a git sub repository in your project. Any reason why nGraph cannot be used as a cmake externalproject? Then you don't need all the NGRAPH_BUILD_DIR related changes. I think your change disables all nGraph backends when using Ninja. Then all you are trying to do build libngraph.so. I see new variables `ARM` and `AARCH64` which are supposed to be defined somewhere in your project. What do those variables mean? If it is for your project only, they shouldn't be used in nGraph.

PR #3590, opened 2019-09-10
[Spec] Implement Gather:v1

mbencer:
Change to add a `v1` version of Gather op, which replaces static attributes `m_axis` with the dynamic input.
As agreed `axis` input can have only one element.


src/ngraph/op/gather.cpp
```diff
@@ -15,23 +15,36 @@
...
+#include <limits>
+
 using namespace std;
 using namespace ngraph;
 
 static int PARAMS = 0;
 static int INDICES = 1;
+static int AXIS = 2;
+
+static size_t AXIS_NOT_SET_VALUE = std::numeric_limits<int64_t>::max();
```
tomdol: This and all of those 3 ints above should be declared `const`.

src/ngraph/op/gather.cpp
```diff
@@ -82,3 +95,119 @@ void op::Gather::validate_and_infer_types()
...
+{
+    const auto& input_rank = get_input_partial_shape(PARAMS).rank();
+    const auto& axis_shape = get_input_partial_shape(AXIS);
+    const auto& axis_rank = axis_shape.rank();
+
+    if (axis_rank.is_static())
+    {
+        NODE_VALIDATION_CHECK(this,
+                              static_cast<size_t>(axis_rank) == 1 &&
+                                  static_cast<size_t>(axis_shape[0]) == 1,
```
tomdol: I think this line will not work if `axis_shape.is_static() == false`. You should add this check to the `if` above.

src/ngraph/op/gather.cpp
```diff
@@ -82,3 +95,119 @@ void op::Gather::validate_and_infer_types()
...
+{
+    int64_t axis = AXIS_NOT_SET_VALUE;
+    auto axes_input_node = input_value(AXIS).get_node_shared_ptr();
+    if (auto const_op = dynamic_pointer_cast<op::Constant>(axes_input_node))
+    {
+        axis = const_op->get_vector<int64_t>()[0];
+    }
+    else
+    {
+        axis = AXIS_NOT_SET_VALUE;
```
tomdol: I think this `else` branch is redundant since you've already initialized `axis` to `AXIS_NOT_SET_VALUE`.

src/ngraph/op/gather.cpp
```diff
@@ -82,3 +95,119 @@ void op::Gather::validate_and_infer_types()
...
+    if (axis < 0)
+    {
+        const auto& input_rank = get_input_partial_shape(PARAMS).rank();
+        if (input_rank.is_static())
+        {
+            axis += static_cast<size_t>(input_rank);
+        }
+        else
+        {
+            axis = AXIS_NOT_SET_VALUE;
```
tomdol: This `else` branch also seems reduntant

src/ngraph/pass/opset1_upgrade.cpp
```diff
@@ -91,6 +93,18 @@ bool pass::Opset1Upgrade::run_on_node(shared_ptr<Node> node)
         modified = true;
         break;
     }
+    case OP_TYPEID::Gather:
+    {
+        auto tmp = dynamic_cast<const op::v0::Gather*>(node.get());
+        int64_t axis = tmp->get_axis();
+
+        auto axis_node = make_shared<op::Constant>(element::i64, Shape{1}, vector<int64_t>{axis});
```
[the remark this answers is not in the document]
mbencer: Done

src/ngraph/op/gather.hpp
```diff
@@ -22,38 +22,67 @@ namespace ngraph
...
+            /// \brief Gather slices from axis of params according to indices
+            class Gather : public Op
             {
-                constructor_validate_and_infer_types();
-            }
+            public:
+                NGRAPH_API
+                static const std::string type_name;
+                const std::string& description() const override { return type_name; }
+                Gather() = default;
```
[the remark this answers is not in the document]
mbencer: Done

PR #3592, opened 2019-09-10
[PRIVATE] Build-time configuration for datatypes that require optimized kernels

jbobba:
E.g., 
cmake -DNGRAPH_CPU_OPTIMIZED_DATATYPES=common $SRC_DIR (default, Optimize for 'common' datatypes defined in CPU cmake file)
cmake -DNGRAPH_CPU_OPTIMIZED_DATATYPES=f32;u8  $SRC_DIR  (Optimize for FP32 and UINT8)
cmake -DNGRAPH_CPU_OPTIMIZED_DATATYPES=all $SRC_DIR (default, Optimize for 'all' datatypes defined in CPU cmake file)

Moved all selector macros to kernel_selectors.hpp and changed macros to permit nesting and reuse to quickly spin up many different combinations.


src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -156,9 +156,42 @@ if (NGRAPH_MLIR_ENABLE)
...
+set(NGRAPH_CPU_COMMON_DATATYPES
+    f32
+    i64
+    )
+
+
 if (NGRAPH_CPU_ENABLE)
     set(NGRAPH_CPU_DEBUGINFO_ENABLE 0 CACHE STRING "Enable debuginfo in the CPU backend")
 
+    set(NGRAPH_CPU_OPTIMIZED_DATATYPES "common"
```
nmostafa: Going to hard-code it here instead of a cmake option ?

jbobba: These are cached variables. Will be set here only if they were not previously specified on the cmdline.

PR #3593, opened 2019-09-11
FakeQuantize python binding fix

tomdol:
This fixes the CI build (py27 + mac + clang)


tomdol:
Closing since more bindings need to be fixed

PR #3595, opened 2019-09-11
[FUSED] Add lstm sequence operator

tsocha:
```
Inputs:
0: X input data
1: initial_hidden_state
2: initial_cell_state
3: sequence_lengths
4: W forward weights
5: R recurse weights
6: B bias
7: P pipholes(optional)[Required by ONNX]

Attributes:
0. hidden_size
1. direction
2. activations(optional)
3. activations_alpha(optional)
4. activations_beta(optional)
5. clip(optional)
6. input_forget(optional)[Required by ONNX]
```


src/ngraph/op/fused/lstm_sequence.hpp
```diff
@@ -0,0 +1,154 @@
...
+            ///
+            std::shared_ptr<Node> get_masked_node(const std::shared_ptr<Node>& data,
+                                                  std::int32_t time_step,
+                                                  std::size_t batch_axis = 0,
+                                                  const std::shared_ptr<Node>& default_value = {
+                                                      nullptr}) const;
+
+            NodeVector lstm_pass(bool is_reverse = false) const;
+
+            std::shared_ptr<Node> prepare_input(Output<Node> node, bool is_reverse) const;
```
arogowie-intel: Some documentation would be helpful :)

jbobba:
@tsocha I dont see any shape inference here? Is that planned for a separate PR?

src/ngraph/op/fused/lstm_sequence.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include "ngraph/op/util/broadcasting.hpp"
+
+using namespace ngraph;
+using namespace std;
+
+constexpr NodeTypeInfo op::LSTMSequence::type_info;
+NodeVector op::LSTMSequence::decompose_op() const
+{
+    NodeVector results;
+    if (m_direction == "forward" || m_direction == "reverse")
```
diyessi: If we really must be forced to make these strings, let's at least define constants for them and use the constants rather than the strings.

src/ngraph/op/fused/lstm_sequence.cpp
```diff
@@ -0,0 +1,230 @@
...
+    {
+        tmp = builder::split(node, 2).at(is_reverse ? 1 : 0);
+    }
+    // Since we have forward LSTM we can squeeze `num_directions` axis from inputs.
+    return builder::squeeze(tmp);
+}
+
+void op::LSTMSequence::validate_and_infer_types()
+{
+    NGRAPH_CHECK(m_direction == "bidirectional" || m_direction == "forward" ||
```
diyessi: Replace with constants if these can't be enums.

src/ngraph/op/fused/lstm_sequence.hpp
```diff
@@ -0,0 +1,154 @@
...
+            std::int64_t get_hidden_size() const { return m_hidden_size; }
+            bool get_input_forget() const { return m_input_forget; }
+        private:
+            ///
+            /// \brief      Gets the masked node according to sequence lenght in a batch.
+            ///
+            /// \note       Zeros out values or sets them to default value for inputs with
+            ///             sequence lenght shorter than currently procssed time step.
+            ///
+            /// \param[in]  data           The input node.
```
diyessi: Use `value` instead of `node`. Nodes can have more than one output, so these are not nodes.
If you are calling it the input in the doc, use `input` for the variable name.

src/ngraph/op/fused/lstm_sequence.hpp
tsocha: Done

src/ngraph/op/fused/lstm_sequence.cpp
tsocha: m_direction is a enum now.

src/ngraph/op/fused/lstm_sequence.cpp
tsocha: m_direction is a enum now.

src/ngraph/op/fused/lstm_sequence.hpp
tsocha: Done

tsocha:
> @jbobba 
> @tsocha I dont see any shape inference here? Is that planned for a separate PR?

I used `constructor_validate_and_infer_types` method, is it enough or do I need to make more changes?

jbobba:
> > @jbobba
> > @tsocha I dont see any shape inference here? Is that planned for a separate PR?
> 
> I used `constructor_validate_and_infer_types` method, is it enough or do I need to make more changes?

Should be good. Can you add some type_prop tests as well.

tsocha:
> > > @jbobba
> > > @tsocha I dont see any shape inference here? Is that planned for a separate PR?
> > 
> > 
> > I used `constructor_validate_and_infer_types` method, is it enough or do I need to make more changes?
> 
> Should be good. Can you add some type_prop tests as well.

Done.
@arogowie-intel Could you take a look at this type_prop test? Does it have sense?

test/type_prop/lstm_sequence.cpp
```diff
@@ -0,0 +1,46 @@
...
+using namespace ngraph;
+
+TEST(type_prop, lstm_sequence)
+{
+    const auto X = make_shared<op::Parameter>(element::f32, Shape{1, 2, 4});
+    const auto W = make_shared<op::Parameter>(element::f32, Shape{1, 12, 4});
+    const auto R = make_shared<op::Parameter>(element::f32, Shape{1, 12, 3});
+    const auto initial_hidden_state = make_shared<op::Parameter>(element::f32, Shape{1, 2, 3});
+    const auto initial_cell_state = make_shared<op::Parameter>(element::f32, Shape{1, 2, 3});
+    const auto B = make_shared<op::Parameter>(element::f32, Shape{1, 24});
```
arogowie-intel: I'm wondering why B shape is so big? Shouldn't it be only 4 * hidden_size, that is 12? Bias input should be a sum of gates biases and recurrence biases.

arogowie-intel:
@tsocha  Will take a closer look on Monday.

test/type_prop/lstm_sequence.cpp
tsocha: This change isn't merged yet.

PR #3607, opened 2019-09-12
[MLIR] Add flag to dump jitted MLIR to object file

dcaballe:
This PR adds a flag to dump the jitted MLIR code to an object
file and another flag to be able to specify the object file name.


src/contrib/mlir/compiler/compiler.cpp
```diff
@@ -713,6 +724,11 @@ void MLIRCompiler::execute()
     // Please, note that 'invoke' method is overloaded with a parameter pack version.
     // Make sure the MutableArrayRef version is invoked.
     auto invocationResult = m_engine->invoke("main", llvm::MutableArrayRef<void*>(m_invoke_args));
+
+    if (clDumpObjectFile)
```
nmostafa: nit: braces around single stmt body.

PR #3614, opened 2019-09-13
[MLIR] Fix non-determ. order of nodes in MLIR sub-graphs

dcaballe:
Replace set with vector container in MLIR sub-graphs for
nodes, inputs and outputs. Using a set caused different order of
operations in MLIR code across different executions and this led to
differences in optimizations.

### DIFF
--- a/src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.hpp
+++ b/src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.hpp
@@ -53,17 +53,15 @@ namespace ngraph
                 /// Get sub-graph id
                 int get_id() const { return m_graph_id; }
                 /// Get all nodes in the sub-graph.
-                NodeSet& get_nodes() { return m_nodes; }
+                NodeVector& get_nodes() { return m_nodes; }
                 /// Get input nodes. Predecessors to head nodes.
-                NodeSet& get_inputs() { return m_input_nodes; }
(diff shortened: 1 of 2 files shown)


src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.hpp
```diff
@@ -73,10 +71,13 @@ namespace ngraph
...
                 int m_graph_id;
                 // Actual nodes of the sub-graph
-                NodeSet m_nodes;
-                // Predecessor to head nodes in the sub-graph.
-                NodeSet m_input_nodes;
-                NodeSet m_output_nodes;
+                NodeVector m_nodes;
+                // Predecessor to head nodes in the sub-graph. Both containers have the same
+                // elements. Set is only used for efficient look-up operations.
+                NodeVector m_input_node_vector;
```
nmostafa: Why the order of input/output nodes causes non-determinism ? Rather avoid the duplication of data if not needed.

dcaballe: IIRC, I saw function arguments in different order but I can double check if you think this is a problem. For outputs, I did it for consistency. I'm not sure if there is order is required for them. Having two container for performance reasons is usually acceptable (see VectorSet in LLVM). Do you have any other suggestion?

PR #3621, opened 2019-09-14
Support for RandomUniformOp in CPU

gaurides:


src/ngraph/runtime/cpu/builder/random_uniform.cpp
```diff
@@ -0,0 +1,160 @@
...
+            void Builder::BUILDER_DECL(ngraph::op::RandomUniform)
+            {
+                auto& functors = external_function->get_functors();
+                CPUKernelFunctor functor;
+                if (args[2].get_element_type() != element::i64)
+                {
+                    throw ngraph_error("Unsupported index 2 element type");
+                }
+                auto element_type = args[0].get_element_type();
+                if (element_type == element::f32)
```
rkimballn1: Use a switch statement and set the #pragma flags so that all element types are required to be here or it won't compile. That way you don't get a runtime error for not supported element type

PR #3622, opened 2019-09-14
Remove some dynamic_ptr_casts

diyessi:


src/ngraph/pass/implicit_broadcast_elimination.cpp
```diff
@@ -27,14 +27,14 @@ using namespace ngraph;
 template <typename optype>
 static bool broadcast_and_replace(std::shared_ptr<ngraph::Node>& node)
```
aprocter: Looks like this probably does not need to be templated anymore.

src/ngraph/node.hpp
```diff
@@ -504,6 +489,36 @@ namespace ngraph
         size_t m_placement_index = placement_invalid;
     };
 
+    /// Casts a Node* to a NodePtr* if it is of type NodePtr, nullptr otherwise
```
aprocter:
```suggestion
    /// Casts a Node* to a NodeType* if it is of type NodeType, nullptr otherwise
```

PR #3624, opened 2019-09-16
Mmahbubu/provenance fix grpah replace

tachyon77:
Fixes provenance algorithm when subgraphs are replaced.
Unit test change explains what changed.

### DIFF
--- a/src/ngraph/graph_util.cpp
+++ b/src/ngraph/graph_util.cpp
@@ -153,14 +153,20 @@ void ngraph::replace_node(std::shared_ptr<Node> target,
     {
         auto common_args = ngraph::find_common_args(target, replacement);
 
-        auto set_replacement_prov = [replacement](std::shared_ptr<Node> node) {
-            replacement->merge_provenance_tags_from(node);
+        std::unordered_set<string> removed_subgraph_tags;
+
+        auto set_replacement_prov = [&removed_subgraph_tags](std::shared_ptr<Node> node) {
+            for (auto tag : node->get_provenance_tags())
+            {
+                removed_subgraph_tags.insert(tag);
+            }
         };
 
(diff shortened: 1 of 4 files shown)


src/ngraph/graph_util.cpp
```diff
@@ -153,14 +153,20 @@ void ngraph::replace_node(std::shared_ptr<Node> target,
     {
         auto common_args = ngraph::find_common_args(target, replacement);
 
-        auto set_replacement_prov = [replacement](std::shared_ptr<Node> node) {
-            replacement->merge_provenance_tags_from(node);
+        std::unordered_set<string> removed_subgraph_tags;
```
diyessi: Use an ordered set so that results are consistent on different runs.

src/ngraph/node.cpp
```diff
@@ -328,6 +328,14 @@ void Node::add_provenance_tag(const std::string& tag)
     m_provenance_tags.insert(tag);
 }
 
+void Node::add_provenance_tags(const std::unordered_set<std::string>& tag_set)
```
diyessi: ordered set

src/ngraph/node.hpp
```diff
@@ -368,6 +368,7 @@ namespace ngraph
 
         const std::unordered_set<std::string>& get_provenance_tags() const;
         void add_provenance_tag(const std::string& tag);
+        void add_provenance_tags(const std::unordered_set<std::string>& tag_set);
```
diyessi: ordered set

src/ngraph/graph_util.cpp
tachyon77: @diyessi Provenance tags (m_provenance_tags) is unordered. I am not sure making this one sorted would be enough.

PR #3625, opened 2019-09-16
[MLIR] Add `ngraph` prefix to MLIR flags

dcaballe:
Some flags collision with some MLIR flags.


src/contrib/mlir/compiler/compiler.cpp
```diff
@@ -93,27 +93,27 @@ static llvm::cl::opt<bool> clPrintIRAfterAll(
 // *** Optimization flags ***
```
nmostafa: I would change `print-ngraph-ir-after-all` to be `ngraph-print-after-all` to be consistent. There is also the other flags you recently added to dump obj file.

dcaballe: Good point, thanks. Changing it to `ngraph-print-ir-after-all` for consistency with MLIR counterpart

PR #3629, opened 2019-09-17
Re-enable sign-compare and double-promotion warnings

diyessi:


src/ngraph/frontend/onnx_import/op/conv.cpp
```diff
@@ -68,7 +68,7 @@ namespace ngraph
                                 filters->get_shape().size());
                             std::vector<std::size_t> filters_upper_bounds{filters->get_shape()};
 
-                            for (std::size_t group{0}; group < groups; ++group)
+                            for (int64_t group{0}; group < groups; ++group)
```
aprocter: (Not going to reject on this nor even look for the same pattern elsewhere, but...)

`groups` was declared `int`. Shouldn't we use the same type here?

PR #3630, opened 2019-09-18
LayerNorm

silee2:
Implement LayerNorm as a fused op
TODO:
- [x] Add unit test
- [x] Add type prop test


silee2:
PlaidML IntelGPU Mac fails tests nut PlaidML IntelGPU ubuntu passes. Very strange.

test/backend/layer_norm.in.cpp
```diff
@@ -0,0 +1,196 @@
...
+    copy_data(scale, s_input);
+    vector<float> b_input{-4.0f, -3.0f, -2.0f, -1.0f};
+    copy_data(bias, b_input);
+    // Create tensors for output
+    auto norm = backend->create_tensor(element::f32, Shape{2, 4});
+    auto mean = backend->create_tensor(element::f32, Shape{2});
+    auto var = backend->create_tensor(element::f32, Shape{2});
+
+    // Expected results
+    vector<float> exp_norm{-2.658364534378051758f,
```
rkimballn1: It would be nice to know where these magic numbers came from. Just a comment would be OK or the python script or whatever would be even better, but I won't hold up this PR for that

silee2: @rkimballn1 Added a comment that the result are manually computed.

PR #3631, opened 2019-09-18
[ONNX] Handling of unknown domains in ONNX models

tomdol:


test/models/onnx/unknown_domain.prototxt
```diff
@@ -0,0 +1,67 @@
+ir_version: 4
+producer_name: "nGraph ONNX Importer"
+graph {
+  node {
+    input: "A"
+    input: "B"
+    output: "C"
+    op_type: "Add"
```
postrational: I would like to see another test, which contains an `unknown.domain.Add` and does throw the `UnknownDomain` exception with a friendly message.

tomdol: It doesn't throw an `UnknownDomain` in this case but I've added a test to make sure an exception is thrown.

PR #3633, opened 2019-09-18
[Spec] Implement Reshape:v1

mbencer:
1. Change `DynReshape`(`v0`) to `Reshape`(`v1`)
~~2. According to specification `Reshape:v1` has no attributes - `zero_flag` was removed. However, `Reshape:v1` should process data as  `DynReshape:v0` with `zero_flag=true` (
`0` in `pattern` means "copy the respective dimension of the input blob")~~


aprocter:
Quick comment before I dive in for a full review later: we will still need the `zero_flag` in `v1`, since this is needed for some frameworks.

src/ngraph/op/experimental/reshape.hpp
```diff
@@ -0,0 +1,65 @@
+//*****************************************************************************
```
postrational: I think this op in `v1` should no longer be in `experimental` directory.

src/ngraph/op/experimental/reshape.hpp
```diff
@@ -0,0 +1,65 @@
...
+            ///
+            /// "Converts" an input tensor into a new shape with the same number of elements.
+            /// This op does not touch the actual data. If needed, use Transpose for that purpose.
+            ///
+            class Reshape : public Op
+            {
+            public:
+            public:
+                NGRAPH_API
+                static constexpr NodeTypeInfo type_info{"DynReshape", 1};
```
postrational: `DynReshape` => `Reshape`

mbencer: It was discussed earlier (for `ReduceSum` and `ReduceProd`). In current implementation `get_typeid` mechanism is based on op description (it is a key in map where the value is `OP_TYPEID` - enum represents op). When we change the description to `Reshape` using `switch` will be not working correctly. The conclusion of preview discussion was that it should be fixed in separate PR.

src/ngraph/op/experimental/reshape.hpp
```diff
@@ -0,0 +1,65 @@
...
+                ///        transpose.
+                ///
+                /// \param arg The tensor to be reshaped.
+                /// \param pattern The node that defines output shape pattern.
+                ///        If the input shape is \f$(a_0,\dots,a_{k-1})\f$ then the output shape
+                ///        must
+                ///        be of the form \f$(b_0,\dots,b_{j-1})\f$ where \f$\Pi(a_i) = \Pi(b_i)\f$.
+                ///        A value of -1 is allowed for at most one dimension, in which case the
+                ///        dimension size is inferred based on element count of input tensor.
+                Reshape(const Output<Node>& arg, const Output<Node>& pattern);
```
aprocter: We'll need to continue to support the `zero_flag`. In some frontends (e.g., TensorFlow), `0` actually means `0`. In others (ONNX, PaddlePaddle), `0` means "copy the dimension". At a glance it looks like `v1` as you've defined it here is always treating `0` as "copy".

src/ngraph/op/experimental/reshape.hpp
aprocter: I agree here---there's a higher-order issue with this stuff we'll need to figure out.

src/ngraph/serializer.cpp
```diff
@@ -1151,7 +1152,14 @@ shared_ptr<Node> JSONDeserializer::deserialize_node(json node_js)
         }
         case OP_TYPEID::DynReshape:
         {
-            node = make_shared<op::DynReshape>(args[0], args[1]);
+            if (op_version == 0)
+            {
+                node = make_shared<op::v0::DynReshape>(args[0], args[1]);
```
aprocter: Looks like deserializer and was missing the zero_flag. :(

Please update the serializer as well.

src/ngraph/op/experimental/reshape.hpp
mbencer: Unfortunately, in `ngraph/op` folder there is already `reshape.hpp(.cpp)` file. How to deal with it? @aprocter Any ideas?

src/ngraph/op/experimental/reshape.hpp
mbencer: Done, `zero_flag` was restored. Unfortunately, `Reshape:v1` is now almost complete copy of `DynReshape:v0` (the only difference are `NodeTypeInfo type_info{"DynReshape", 1}` and `get_version()`). @aprocter The inheritance from `DynReshape:v0` and overriding only indicated differences can be solution to this case?

src/ngraph/serializer.cpp
mbencer: Serializer was updated.

src/ngraph/op/experimental/reshape.hpp
aprocter: @mbencer Let's just add the new `v1` stuff into the existing `reshape.*pp` files.

src/ngraph/op/experimental/reshape.hpp
aprocter: @mbencer Let's just live with the duplication for now. `DynReshape` will go away, so the duplication will be short lived.

src/ngraph/op/experimental/reshape.hpp
mbencer: Done

PR #3634, opened 2019-09-18
[ONNX] ReverseSequence attributes validation

tomdol:


test/onnx/onnx_import.in.cpp
```diff
@@ -1609,3 +1609,29 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_reverse_sequence_1_batch_0)
...
+        << "ReverseSequence batch_axis attribute can only equal 0 or 1. Value of '2' has been "
+           "accepted.";
+}
+
+NGRAPH_TEST(onnx_${BACKEND_NAME}, model_reverse_sequence_incorrect_time_axis)
+{
+    EXPECT_THROW(onnx_import::import_onnx_model(file_util::path_join(
+                     SERIALIZED_ZOO, "onnx/reverse_sequence_incorrect_time_axis.prototxt")),
+                 ngraph_error)
+        << "ReverseSequence time_axis attribute can only equal 0 or 1. Value of '2' has been "
```
mbencer: The same as before, shouldn't it be `Value of '2' is not accepted`?

test/onnx/onnx_import.in.cpp
```diff
@@ -1609,3 +1609,29 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_reverse_sequence_1_batch_0)
 
     test_case.run();
 }
+
+NGRAPH_TEST(onnx_${BACKEND_NAME}, model_reverse_sequence_incorrect_batch_axis)
+{
+    EXPECT_THROW(onnx_import::import_onnx_model(file_util::path_join(
+                     SERIALIZED_ZOO, "onnx/reverse_sequence_incorrect_batch_axis.prototxt")),
+                 ngraph_error)
+        << "ReverseSequence batch_axis attribute can only equal 0 or 1. Value of '2' has been "
```
mbencer: Shouldn't it be Value of '2' is not accepted?

tomdol: This error message will only be shown if the test assertion fails. It will fail only if the `import_onnx_model` doesn't throw and that will only happen if someone removes/modifies the `NGRAPH_CHECK`s from onnx_importer. 

I wanted to give the message to the developer here that an incorrect behavior occurred (when it occurs) - a value that is hardcoded in the model passed the validation, while it shouldn't.

Does that make sense or should I change it anyway @diyessi @mbencer ?

PR #3636, opened 2019-09-18
[SPEC] Add v1:Convolution operator

tsocha:
```
Strides m_window_movement_strides -> m_strides
Strides m_window_dilation_strides -> m_dilations
CoordinateDiff m_padding_below    -> m_pads_begin
CoordinateDiff m_padding_above    -> m_pads_end
PadType m_pad_type                -> m_auto_pad
Strides m_data_dilation_strides   -> REMOVED
```


src/ngraph/op/convolution.cpp
```diff
@@ -25,16 +25,476 @@
...
+    if (m_pads_end.size() == 0)
+    {
+        m_pads_end = conv_default_padding(this, data_batch_shape, filters_shape);
+    }
+
+    if (m_auto_pad == PadType::SAME_UPPER || m_auto_pad == PadType::SAME_LOWER)
+    {
+        if (data_batch_shape.is_static() && filters_shape.is_static())
+        {
+            // TODO: data dilation
```
aprocter: I guess we can delete this comment now, happily :)

src/ngraph/op/convolution.cpp
```diff
@@ -25,16 +25,476 @@
...
+                           x, f_shape, delta, m_strides, m_dilations, m_pads_begin, m_pads_end));
+}
+
+constexpr NodeTypeInfo op::v1::ConvolutionBackpropData::type_info;
+shared_ptr<Node> op::v1::Convolution::get_default_value() const
+{
+    return ngraph::make_constant_from_string("0", get_element_type(), get_shape());
+}
+
+op::v1::ConvolutionBackpropData::ConvolutionBackpropData(const Shape& data_batch_shape,
```
aprocter: At some point in the near future we are going to need to make this dynamic, but let's stick with this for this PR.

src/ngraph/pass/opset1_upgrade.cpp
```diff
@@ -75,6 +76,107 @@ bool pass::Opset1Upgrade::run_on_node(shared_ptr<Node> node)
...
+        auto data_dilation_strides = tmp->get_data_dilation_strides();
+        auto auto_pad = tmp->get_pad_type();
+
+        bool is_dds_valid = true;
+        for (auto value : data_dilation_strides)
+        {
+            is_dds_valid = is_dds_valid && (value == 1);
+        }
+
+        NGRAPH_CHECK(is_dds_valid,
```
aprocter: (Just thinking out loud...) this _would_ be possible if we had an "interior padding" operation, but we got rid of that. With DynReplaceSlice we could rig something up, but since data dilation was (AFAIK) totally unused anyway, it may not be worth the effort.

src/ngraph/op/convolution.cpp
tsocha: Done :)

src/ngraph/pass/opset1_upgrade.cpp
tsocha: We should avoid adding code that will not be used IMHO.

PR #3642, opened 2019-09-19
Convert remaining dynamic_pointer_casts for ops

diyessi:
Make `is_type` a function instead of a method
Direct implementation of `is_parameter`


src/ngraph/runtime/cpu/pass/cpu_rnn_fusion.cpp
```diff
@@ -219,12 +219,12 @@ void ngraph::runtime::cpu::pass::LSTMFusion::construct_onnx_lstmcell_fprop()
             lstm_ht_ct_output, Coordinate{batch_size, 0}, Coordinate{(2 * batch_size), dic});
 #endif
         // find the user's for {ht} and replace them with lstm_goe_0
```
indie: Find the user's what? Find the users for `{ht}`? Users might be the wrong word here if users means "framework node" (nodes?) or something like that.

src/ngraph/frontend/onnx_import/op/reshape.cpp
```diff
@@ -50,9 +50,8 @@ namespace ngraph
                         ASSERT_IS_SUPPORTED(node, ng_inputs.at(1)->description() == "Constant")
                             << "doesn't support shape input of other type than Constant.";
 
-                        auto output_shape_node =
-                            std::dynamic_pointer_cast<ngraph::op::Constant>(ng_inputs.at(1));
-                        output_shape = output_shape_node->get_vector<std::size_t>();
+                        output_shape = ngraph::as_type_ptr<ngraph::op::Constant>(ng_inputs.at(1))
+                                           ->get_vector<std::size_t>();
                     }
                     // Do nothing if there is no shape argument nor second node input.
```
indie: When there is neither shape argument, nor second node input, do nothing.

src/ngraph/runtime/cpu/pass/cpu_rnn_fusion.cpp
```diff
@@ -427,8 +426,7 @@ void ngraph::runtime::cpu::pass::LSTMFusion::construct_lstm_fprop()
             swap_lstm_inputs();
         }
 #if MKLDNN_VERSION_MAJOR < 1
-        else if (std::dynamic_pointer_cast<ngraph::op::GetOutputElement>(
-                     cell_state->get_argument(0)))
+        else if (is_type<ngraph::op::GetOutputElement>(cell_state->get_argument(0)))
         {
             // swap the inputs if the cell_state and hidden state does not
```
indie: "swap the inputs if cell_state and hidden state do not belong..."

src/ngraph/runtime/cpu/pass/cpu_rnn_fusion.cpp
```diff
@@ -440,7 +438,7 @@ void ngraph::runtime::cpu::pass::LSTMFusion::construct_lstm_fprop()
             }
         }
 #else
-        else if (std::dynamic_pointer_cast<ngraph::op::GetOutputElement>(cell_state))
+        else if (is_type<ngraph::op::GetOutputElement>(cell_state))
         {
             // swap the inputs if the cell_state and hidden state does not
```
indie: same grammar issue

test/builder.cpp
```diff
@@ -112,27 +112,25 @@ TEST(builder, numpy_transpose)
     // 2D Transpose
     Shape shape{2, 4};
     auto param = make_shared<op::Parameter>(element::f32, shape);
-    auto transposed = dynamic_pointer_cast<op::Reshape>(builder::numpy_transpose(param));
+    auto transposed = as_type_ptr<op::Reshape>(builder::numpy_transpose(param));
     EXPECT_EQ(Shape({4, 2}), transposed->get_output_shape());
 
     // Multidimensional Transpose
```
indie: Multi-dimensional transpose

PR #3647, opened 2019-09-21
Master edition of Provenance for builders

diyessi:


src/ngraph/builder/make_constant.hpp
```diff
@@ -103,7 +103,7 @@ namespace ngraph
                 val = std::make_shared<ngraph::op::Broadcast>(val, shape, axes);
             }
 
-            return val;
+            return val->add_provenance_group_members_above({});
```
aprocter: What effect does this have? Not quite sure I understand.

PR #3654, opened 2019-09-23
[MLIR] Add support for parsing nGraph tensor type

dcaballe:
Initial commit that enables nGraph parsing. It's needed for testing.


src/contrib/mlir/compiler/dialect/dialect.cpp
```diff
@@ -37,6 +39,49 @@ NGraphOpsDialect::NGraphOpsDialect(mlir::MLIRContext* ctx)
...
+        SmallVector<int64_t, 4> shape;
+        for (unsigned i = 0, end = subStrings.size() - 1; i < end; ++i)
+        {
+            StringRef dimStr = subStrings[i];
+            int64_t dim = -1;
+            // NOTE: `consumeInteger` returns false if an integer was parsed successfully.
+            if (dimStr.consumeInteger(/*Radix=*/10, dim) || !dimStr.empty())
+            {
+                emitError(loc, "expected a list of '[0-9]+x' dimension specifiers: ") << tyData;
+                return Type();
```
ayzhuang: Why do we need  "return Type();" here?

dcaballe: Because `emitError` doesn't raise any exception so we still have to return the expected type. I could add an `NGRAPH_CHECK` or `NGRAPH_UNREACHABLE` but I'm worried that breaks MLIR's error handling, which is also used in LIT tests.

ayzhuang: emitError on line 50 and line 76 do not have "return Type();"

dcaballe: Oh, good catch! That's wrong because the execution will continue. I'll fix it following approach in line 82. Thanks!

PR #3656, opened 2019-09-23
Materialize reshapes for multi-output operators

csullivan:
Reshape sinking does not re-materialize existing reshapes attached to multi-output ops. This causes a PDPD unit test failure. Add support for materializing reshapes for multi-output ops and a test in ngraph-core to reproduce the behavior on the CPU backend --when ReshapeSinking is manually enabled.


src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -266,16 +266,10 @@ static void materialize_shapes(shared_ptr<Node> n,
                                ReshapeMap& reorders,
                                set<shared_ptr<Node>>& reshapes_to_delete)
 {
-    // skip multiple output nodes and deal with GOEs exclusively
-    if (n->get_output_size() > 1)
-    {
-        return;
-    }
-
     for (size_t i = 0; i < n->get_arguments().size(); i++)
```
diyessi: Anything calling `get_arguments` is likely to be broken when it comes to multiple outputs.

csullivan: The whole ReshapeSinking pass needs to be updated to use Input/Output<Node> and Node* instead of shared_ptr<Node> and get_argument(s).

diyessi: #4040 is a first step in adding support for Output<Node>.

PR #3657, opened 2019-09-23
[MLIR] Enable nGraph dialect in ngraph-opt

dcaballe:
This PR registers nGraph dialect in ngraph-opt and prepares
nGraph lowering pass for LIT testing, fixing all the related issues.
It adds `convert-ngraph-to-affine` for that purpose.


src/contrib/mlir/compiler/pass/memory_optimization.cpp
```diff
@@ -158,6 +158,3 @@ namespace mlir
         return std::make_unique<MemoryOptimizationPass>();
     }
 } // namespace mlir
-
-static PassRegistration<MemoryOptimizationPass> pass("ng-inplace-mem-opt",
-                                                     "Performs in-place memory optimizations");
```
dcaballe: This flag for testing causes an error in `ngraph-opt` that says it is registered twice. I decided to remove it since Nagy couldn't finally add a test for it and it's not being used. We'll investigate what's going on once the flag is used on a test.

PR #3659, opened 2019-09-24
Element wise ops bf16 support by float fallback for CPU backend

nbpatel:


src/ngraph/runtime/cpu/cpu_builder.hpp
```diff
@@ -32,46 +32,101 @@
...
-
-#define BUILD_BINARY_ELEMWISE_FUNCTOR(OP)                                                          \
-    (void)node;                                                                                    \
-    auto& functors = external_function->get_functors();                                            \
-    std::function<void(void*, void*, void*, size_t, int)> kernel;                                  \
+        auto functor = [&, kernel, element_count, arg0_buffer_index, out0_buffer_index](           \
+            CPURuntimeContext* ctx, CPUExecutionContext* ectx) {                                   \
+            void* fp_src = std::malloc(element_count * 4);                                         \
+            void* fp_dst = std::malloc(element_count * 4);                                         \
+            ngraph::bf16_to_float(ctx->buffer_data[arg0_buffer_index], fp_src, element_count);     \
```
jbobba: This seems hacky. Instead of doing it this way, it would be better to write a graph pass to replace BF16 elementwise ops with Convert->FP32 Eltwise->Convert.

nbpatel: Agree. We need #3689 for it to be a graph pass thing ...so once thats in I'll modify this.

PR #3661, opened 2019-09-24
[WIP][FusedOp] Implement StridedSlice

mbencer:
Assumptions: `StridedSlice` should be implemented as FusedOp (with using `DynSlice`).

**[HELP WANTED]**
Problems:
- `FusedOpDecomposition` in current implementation is performed after `DynElimination` (and `ConstantFolding`) - as a result backend implementation of `StridedSlice` is not found (`StridedSlice` is calculated by pass)
- When for tests `FusedOpDecomposition` was added (redundantly) before `DynElimination`  the machter does NOT use `DynElimination` for `DynSlice` node.
In diagnostic logs we can find:
`matcher.cpp 344	[MATCHER] Aborting at DynSlice_211 for pattern DynSlice_203`
(BTW: I am not sure why `decompose_op` is called so many times - there are constructed many instances of `StridedSlice`)
- Other problems were described by comments

The main differences between `DynSlice` and `StridedSlice`:
- `strides` input can be optional
- Attributes are represented by `0` and `1` values mask vector where the position determines which axes should be used


src/ngraph/op/util/fused_op.cpp
```diff
@@ -57,8 +57,17 @@ void op::util::FusedOp::validate_and_infer_types()
...
-                i, output_node->get_output_element_type(j), output_node->get_output_shape(j));
+            //TODO [REMOVE] How to deal with dynamic output?
+            if (output_node->get_output_partial_shape(j).is_static())
+            {
+                set_output_type(
+                    i, output_node->get_output_element_type(j), output_node->get_output_shape(j));
+            }
+            else
+            {
+                set_output_type(
```
mbencer: Can we add support for `dynamic` output to `FusedOp` base class?

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -170,6 +172,8 @@ bool runtime::dynamic::DynamicExecutable::call(
     }
 
     pass::Manager passes;
+    //TODO [REMOVE] FUSED PASS SHOULD BE PERFORMED BEFORE ConstantFolding and DynElimination?
+    passes.register_pass<pass::FusedOpDecomposition>();
```
mbencer: `FusedOpDecomposition` added before `DynElimination` for tests

src/ngraph/op/fused/strided_slice.cpp
```diff
@@ -0,0 +1,228 @@
...
+                                      convert_mask_to_axis_set(m_shrink_axis_mask),
+                                      convert_mask_to_axis_set(m_ellipsis_mask));
+    }
+    else
+    {
+        const auto added_axes = std::count_if(
+            m_new_axis_mask.begin(), m_new_axis_mask.end(), [](size_t i) { return i == 1; });
+        const auto shrinked_axes = std::count_if(
+            m_shrink_axis_mask.begin(), m_shrink_axis_mask.end(), [](size_t i) { return i == 1; });
+        const auto stride_size = m_begin_mask.size() + added_axes - shrinked_axes;
```
mbencer: I am sure if we can use mask size to calculate `stride` shape. The better option is probably `get_input_partial_shape(0).rank()`, but it is not available for dynamic case

PR #3665, opened 2019-09-25
[MLIR] Add support for parsing nGraph element types

dcaballe:
It introduces initial support for parsing nGraph signed/unsigned
integer and floating point data types.


dcaballe:
> Can you add ngraph tensor type to the two tests?

Let me do that in a separate PR, since this one it's only addressing only element types and I'm not sure if more work is needed for tensor types

PR #3670, opened 2019-09-25
[SPEC] Implement Pad:v1 downgrade pass

mbencer:
1. Downgrade pass implementation for `Pad:v1`
2. Using `Pad:v1` in `onnx_importer`


src/ngraph/op/pad.cpp
```diff
@@ -305,23 +305,6 @@ void op::v1::Pad::validate_and_infer_types()
     const auto& pads_begin_coord = get_pads_begin();
     const auto& pads_end_coord = get_pads_end();
 
-    for (const auto& pads_begin_dim : pads_begin_coord)
```
mbencer: According to specification all elements should be non-negative. However, `Pad:v0` supports negative elements of `pad_begin` and `pad_end` (it causes removing elements from indicated axes). @ilyachur @GlebKazantaev  Can we remove this constraint from op?

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1202,6 +1203,7 @@ void runtime::cpu::CPU_ExternalFunction::register_common_passes(
         return true;
     };
 
+    REGISTER_KNOBBED_PASS(Opset1Downgrade, true, ngraph::pass)
```
mbencer: @aprocter Is it correct place to perform `Opset1Downgrade` pass?

src/ngraph/op/pad.cpp
aprocter: To be honest I'm having trouble remembering why we had this capability in `v0`.

src/ngraph/runtime/cpu/cpu_external_function.cpp
aprocter: I think so! CPU BE folks should double check.

test/opset_pass/pad_opset_pass.cpp
```diff
@@ -3,13 +3,14 @@
...
 #include "ngraph/pass/manager.hpp"
+#include "ngraph/pass/opset1_downgrade.hpp"
 #include "ngraph/pass/opset1_upgrade.hpp"
 #include "util/type_prop.hpp"
 
 using namespace std;
 using namespace ngraph;
 
-TEST(serialize, opset1_pad_pass)
+TEST(pad_opset_pass, opset1_pad_upgrade_pass)
```
tsocha:
```suggestion
TEST(opset_upgrade, opset1_pad_upgrade_pass)
```

test/opset_pass/pad_opset_pass.cpp
```diff
@@ -37,3 +38,32 @@ TEST(serialize, opset1_pad_pass)
     EXPECT_EQ(pad_v1_node->get_pads_begin(), padding_below);
     EXPECT_EQ(pad_v1_node->get_pads_end(), padding_above);
 }
+
+TEST(pad_opset_pass, opset1_pad_downgrade_pass)
```
tsocha:
```suggestion
TEST(opset_downgrade, opset1_pad_downgrade_pass)
```

src/ngraph/op/pad.cpp
diyessi: It might have been so an op with padding (which can be negative) could be decomposed into an unpadded op preceded by a pad op.

src/ngraph/pass/opset1_downgrade.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#pragma once
+
+#include "ngraph/pass/pass.hpp"
+
+namespace ngraph
+{
+    namespace pass
+    {
+        class Opset1Downgrade : public NodePass
```
postrational: This class should be named `Opset0Downgrade`, please rename.

src/ngraph/pass/opset1_downgrade.cpp
```diff
@@ -0,0 +1,98 @@
...
+//*****************************************************************************
+#include "ngraph/pass/opset1_downgrade.hpp"
+#include "ngraph/graph_util.hpp"
+#include "ngraph/op/get_output_element.hpp"
+#include "ngraph/op/pad.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+#define NGRAPH_OP(a, b) a,
```
postrational: Is there any way we could avoid duplicating this `OP_TYPEID` code yet again?

src/ngraph/pass/opset1_downgrade.cpp
```diff
@@ -0,0 +1,98 @@
...
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wswitch-enum"
+#endif
+    switch (get_typeid(node))
+    {
+    case OP_TYPEID::Pad:
+    {
+        auto tmp = dynamic_cast<const op::v1::Pad*>(node.get());
+        auto replacement_node = make_shared<op::v0::Pad>(node->input(0).get_source_output(),
```
postrational: Let's assign `node->input(0).get_source_output()` and `node->input(3).get_source_output()` to variables to make code self-documenting.

test/type_prop/pad.cpp
```diff
@@ -602,58 +602,6 @@ TEST(type_prop, pad_v1_arg_pads_end_incompatible_type)
     }
 }
 
-TEST(type_prop, pad_v1_pads_begin_elem_negative_value)
```
postrational: Why are these tests removed?

src/ngraph/pass/opset1_downgrade.hpp
mbencer: Done

src/ngraph/pass/opset1_downgrade.cpp
mbencer: Good point, but I'll create separate PR for this request ;)

test/type_prop/pad.cpp
mbencer: `ONNX` standard requires support for negative indexes (it means removing elements at the beginning and end of each axis).

PR #3675, opened 2019-09-26
[MLIR] Modify mlir subgraph extraction pass.

ayzhuang:
Process each node in topological order:
1)start a new sub-graph when an unsupported node is encountered and there is no supported node ready.
2)supported node is added to current sub-graph and its
   predecessors outside current sub-graphs are marked as input to the sub-graph.
Sub-graph may contain multiple disjoint clusters.


src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.cpp
```diff
@@ -139,85 +133,50 @@ bool MLIRSubgraphExtractionPass::run_on_function(std::shared_ptr<Function> func)
 void MLIRSubgraphExtractionPass::build_subgraphs(std::shared_ptr<Function> func)
 {
     NGRAPH_DEBUG << "[CK Extract] Construct sub-graphs";
+    bool create_subgraph = true;
+    int current_subgraph_id = 0;
     for (auto op : func->get_ordered_ops())
```
nmostafa: We need to have our own topological sort here to minimize sub-graphs. When multiple nodes are ready (all predecessors assigned a sub-graph), we need to pick the one that matches the last selected node. For example, if the last node assigned is an MLIR node, then we should pick another MLIR node to add to the sub-graph, if any is available. Similarily, if last node is unsupported, then we should pick and skip the unsupported node.

ayzhuang: Will change it.

src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.cpp
```diff
@@ -136,89 +138,159 @@ bool MLIRSubgraphExtractionPass::run_on_function(std::shared_ptr<Function> func)
...
+                    }
                 }
-
-                first_subgraph.add_node(op);
-                first_subgraph.add_inputs(inputs);
             }
             else
             {
-                // we have a cycle, start a new sub-graph
+                // create a new sub-graph
```
nmostafa: This code is duplicated above. Outline into a function ?

src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.cpp
```diff
@@ -136,89 +138,159 @@ bool MLIRSubgraphExtractionPass::run_on_function(std::shared_ptr<Function> func)
...
+                    }
+                    // add inputs and op to current sub-graph
+                    MLIRSubgraph& current_subgraph = get_subgraph(current_subgraph_id);
+                    current_subgraph.add_node(node);
+                    current_subgraph.add_inputs(inputs);
+                    NGRAPH_DEBUG << "[CK Extract] Node Processed " << *node;
+                    process_successors(node, node_to_size_map, nodes_ready);
+                    nodes_ready.erase(it);
+                    change_mode = false;
+                    break;
```
nmostafa: Do we need to break here ? Can't we just keep looking for the next supported node in the list ?
The idea is that if last node is supported (mode 0), we keep scanning the worklist and adding supported nodes to the sub-graph until we cannot find any (then it is a change of mode to mode 1). If last node is unsupported (mode 1), we keep scanning until and visiting unsupported nodes until none is found (then switch back to mode 0).

ayzhuang: I got segfault when tried to do that because we're removing node from the list while doing the iteration. I could try again to see if that would work.

src/contrib/mlir/compiler/pass/mlir_subgraph_extraction.cpp
```diff
@@ -45,6 +45,10 @@ using namespace ngraph::op;
 using namespace ngraph::pass;
 
 #define TI(x) std::type_index(typeid(x))
+#define ERASE_NODE                                                                                 \
```
nmostafa: Function would be better.

PR #3687, opened 2019-09-30
add get_preferred_pipeline_depth to executable

ashokei:


src/ngraph/runtime/executable.hpp
```diff
@@ -112,4 +116,6 @@ class ngraph::runtime::Executable
 private:
     ngraph::ParameterVector m_parameters;
     ngraph::ResultVector m_results;
+    // default pipeline_depth of 2, specific executable can override.
+    size_t m_default_pipeline_depth = 2;
```
rkimballn1: Remove this. Put the value inside get_default_pipeline_depth() implementation. Because you made m_default_pipeline_depth private it can't be changed by a derived class anyway.

src/ngraph/runtime/executable.cpp
```diff
@@ -108,6 +108,11 @@ const ngraph::ResultVector& runtime::Executable::get_results() const
     return m_results;
 }
 
+size_t runtime::Executable::get_default_pipeline_depth() const
+{
+    return m_default_pipeline_depth;
```
rkimballn1: Just return the number here. It is a virtual function so a user can override it and change the value without the need for m_default_pipeline_depth

src/ngraph/runtime/executable.hpp
```diff
@@ -69,6 +69,10 @@ class ngraph::runtime::Executable
     /// \returns an ngraph::ResultVector of all input parameters
     const ngraph::ResultVector& get_results() const;
 
+    /// \brief Get the default pipeline_depth for this executable
+    /// \returns default pipeline_depth
+    virtual size_t get_default_pipeline_depth() const;
```
rkimballn1: What does default depth mean? Is it the preferred depth?

ashokei: @rkimballn1  "default" is misleading, changed to "preferred"; also addressed other feedback. thanks!

PR #3688, opened 2019-09-30
Support non-mkldnn fallback for Batchnorm training bprop in CPU backend

jbobba:


src/ngraph/runtime/cpu/builder/batch_norm.cpp
```diff
@@ -395,120 +395,195 @@ namespace ngraph
...
+                    std::function<decltype(runtime::cpu::kernel::batch_norm_backprop<float>)>
+                        kernel;
+
+                    SELECT_KERNEL(kernel,
+                                  args[0].get_element_type(),
+                                  runtime::cpu::kernel::batch_norm_backprop)
+
+                    auto arg2_shape = args[2].get_shape();
+                    auto arg0_buffer_index =
+                        external_function->get_buffer_index(args[0].get_name()); /* gamma */
```
diyessi: `gamma_buffer_index` etc. would be easier to follow.

PR #3689, opened 2019-09-30
Modify convert op to convert from bf16 to float and vice-versa

nbpatel:
Current op only does regular casting. Modified it to do bf16 --> float and float --> bf16 conversion as well


src/ngraph/runtime/reference/convert.hpp
```diff
@@ -41,6 +41,30 @@ namespace ngraph
                     out[i] = static_cast<char>(static_cast<bool>(arg[i]));
                 }
             }
+
+            template <typename TI>
+            void convert_float_to_bf16(void* arg, void* out, size_t count)
```
aprocter: Does `convert<TI, ngraph::bfloat16>` not work? If not, could we find a way to make it work?

FWIW, we do have casts between bfloat16 and float32.

diyessi: I think `static_cast` should be made to work in `bfloat16.hpp`

nbpatel: ok, that seems to work...how about the other way round? do we support that as well?

aprocter: It should already work, though it _might_ only work if you're going directly to/from `float`. (I dunno when/if/how the whole promotion machinery kicks in for "custom" numeric types in C++...)

aprocter: I think there is an `operator float()` in `bfloat16`...

PR #3701, opened 2019-10-01
Enable Gather with negative indices.

ayzhuang:


src/ngraph/runtime/cpu/kernel/gather.hpp
```diff
@@ -69,6 +69,8 @@ namespace ngraph
                     Eigen::array<Eigen::Index, Rank1> in_dims;
                     Eigen::array<Eigen::Index, Rank2> out_dims;
 
+                    auto bound_value = inputs_shape[axis];
```
diyessi: `axis_length` would be a better name for this than `bound_value`.

test/backend/gather.in.cpp
```diff
@@ -162,6 +162,32 @@ NGRAPH_TEST(${BACKEND_NAME}, gather_2d_indices_no_axis_2d_input)
...
 
+NGRAPH_TEST(${BACKEND_NAME}, gather_2d_negative_and_positive_indices_no_axis_2d_input)
+{
+    Shape params_shape{3, 2};
+    Shape indices_shape{2, 2};
+    Shape out_shape{2, 2, 2};
+    auto P = make_shared<op::Parameter>(element::f32, params_shape);
+    auto I = make_shared<op::Parameter>(element::i32, indices_shape);
+    auto G = make_shared<op::Gather>(P, I);
+    auto f = make_shared<Function>(make_shared<op::GetOutputElement>(G, 0), ParameterVector{P, I});
```
jbobba: Do we need GOE here?

ayzhuang: I reused existing unit test. Will change gather unit tests.

PR #3705, opened 2019-10-02
Hotfix for negative axes in unsqueeze op

tsocha:


src/ngraph/frontend/onnx_import/op/unsqueeze.cpp
```diff
@@ -30,8 +31,10 @@ namespace ngraph
                 {
                     auto data = node.get_ng_inputs().at(0);
                     auto axes = node.get_attribute_value<std::vector<std::int64_t>>("axes", {});
+                    std::vector<std::size_t> valid_axes =
+                        common::validate_axes(node, axes, data->get_shape().size() + axes.size());
```
postrational: Please store this in an extra variable named `expanded_rank`, so the code is self-documenting.

PR #3707, opened 2019-10-02
[SPEC] Implement Reverse:v1 downgrade pass

mbencer:
1. Downgrade pass implementation for `Reverse:v1` was implemented (`INDEX` and `MASK` mode support).
2. `Reverse` is not used in `onnx_importer` so it was not changed

**The branch is depended on: https://github.com/NervanaSystems/ngraph/pull/3670 which should be merged as first.**


src/ngraph/pass/opset1_downgrade.cpp
```diff
@@ -88,6 +91,38 @@ bool pass::Opset1Downgrade::run_on_node(shared_ptr<Node> node)
...
     }
+    case OP_TYPEID::Reverse:
+    {
+        auto tmp = as_type_ptr<op::v1::Reverse>(node);
+        auto axes_node = tmp->input_value(1).get_node_shared_ptr();
+        NGRAPH_CHECK(axes_node->is_constant(),
+                     "Unable to convert Reverse:v1 to Reverse:v0 "
+                     "if reduction axes are not constant. Node: ",
+                     *node);
+        const auto axes_node_const = dynamic_pointer_cast<op::Constant>(axes_node);
```
tsocha: If we know that axes_node is constant could we use static_pointer_cast here?
```suggestion
        const auto axes_node_const = static_pointer_cast<op::Constant>(axes_node);
```

jbobba: as_type_ptr?

mbencer: I changed it to `as_type_ptr`

PR #3708, opened 2019-10-02
Concat operator, negative indexing support

mozga-intel:
Negative indexing support for a concat operator:
1) Added new `concat_negative_axis` test
2) Replaced `size_t -> int64_t`
3) Support for a negative indexing, calculate `axis = axis + int64(input_rank)`
  ` m_axis = m_axis < 0 ? m_axis + int64_t(this_input_rank) : m_axis;`


src/ngraph/op/concat.cpp
```diff
@@ -50,8 +50,10 @@ void op::Concat::validate_and_infer_types()
         Dimension this_input_rank = this_input_shape.rank();
         if (this_input_rank.is_static())
         {
+            m_axis = m_axis < 0 ? m_axis + int64_t(this_input_rank) : m_axis;
```
baojun-nervana: It seems this requires  a static rank and won't work if rank is dynamic

src/ngraph/op/concat.cpp
```diff
@@ -46,12 +46,16 @@ void op::Concat::validate_and_infer_types()
...
     for (uint64_t i = 0; i < get_input_size(); i++)
     {
+        NODE_VALIDATION_CHECK(this,
+                              element::Type::merge(inputs_et, inputs_et, get_input_element_type(i)),
+                              "Argument element types are inconsistent.");
         PartialShape this_input_shape = get_input_partial_shape(i);
         Dimension this_input_rank = this_input_shape.rank();
         if (this_input_rank.is_static())
         {
+            m_axis = m_axis < 0 ? m_axis + int64_t(this_input_rank) : m_axis;
```
silee2: Don't update m_axis. This will not work for dynamic shape.
Example) m_axis originally is -1 and Rank is 3. Then your code will update m_axis to 2. Then let's say on the next call Rank becomes 4, since m_axis was -1 originally the correct concat axis will be 3. But your code has updated m_axis to 2 and you get an incorrect concat axis.
For node validation, assign the computed concat axis to a new variable like concat_axis and use it in the NODE_VALIDATION_CHECK below.

src/ngraph/runtime/reference/concat.hpp
```diff
@@ -32,12 +32,11 @@ namespace ngraph
                         T* out,
                         const std::vector<Shape>& in_shapes,
                         const Shape& out_shape,
-                        size_t concatenation_axis)
+                        int64_t concatenation_axis)
             {
```
silee2: You need to handle negative concatenation_axis.
concatenation_axis = concatenation_axis < 0 ? static_cast<int64_t>(in_shapes.size()) + concatenation_axis : concatenation_axis;

src/ngraph/op/concat.cpp
jbobba: Agree. You can provide a size_t get_axis() and int64_t get_axis(). size_t get_axis() will return the adjusted axes and throws an error if input shape is not static. int64_t get_axis() will return the value specified through the constructors and setters.

src/ngraph/runtime/reference/concat.hpp
```diff
@@ -32,12 +32,15 @@ namespace ngraph
...
                 // We will copy the inputs to the output one at a time. As we go, we will move out
                 // along the concatenation axis, starting at 0.
-                size_t concatenation_pos = 0;
 
+                concatenation_axis = concatenation_axis < 0
+                                         ? concatenation_axis + int64_t(in_shapes.size())
```
silee2: I mad a mistake in my last comment this is the correct way.
change in_shapes.size() to out_shape.size()

src/ngraph/runtime/cpu/kernel/concat.hpp
```diff
@@ -38,8 +38,9 @@ namespace ngraph
                             std::vector<Shape> input_shapes,
                             void* output,
                             Shape output_shape,
-                            size_t axis)
+                            int64_t axis)
                 {
+                    // axis = axis < 0 ? axis + int64_t(Rank) : axis;
```
silee2: Don't comment out this line. This line is need to compute the real axis.

src/ngraph/op/concat.cpp
mozga-intel: If axis will not be updated in this way, then the value of axis is not correct...  Axis and rank are defined as follows: {axis = -1, Rank = 2} thus the axis is equal 1.  For a reference version we have { axis = -1 Rank = 3} axis = 2. But we need to have 1

src/ngraph/runtime/reference/concat.hpp
mozga-intel: I will do fallback to the previous version. Thanks.

src/ngraph/op/concat.cpp
mozga-intel: The following code is also run for a dynamic_shape if the rank will not be updated then the axis will be equal -1 for each iteration

src/ngraph/op/concat.hpp
```diff
@@ -37,29 +37,33 @@ namespace ngraph
...
-            size_t get_axis() const { return m_axis; }
-            void set_axis(size_t axis) { m_axis = axis; }
+            int64_t get_axis() const { return m_axis; }
+            void set_axis(int64_t axis) { m_axis = axis; }
         protected:
             virtual void generate_adjoints(autodiff::Adjoints& adjoints,
                                            const NodeVector& deltas) override;
-            size_t m_axis;
+            int64_t m_axis;
+            int64_t m_concat_axis = -1;
```
jbobba: Should this be a size_t? Please add a comment to explain the difference between m_axis and m_concat_axis. It would also be good to keep the setter/getter name consistent with this variable name. 'concat' vs 'concatenation'

src/ngraph/op/concat.hpp
```diff
@@ -37,29 +37,35 @@ namespace ngraph
...
 
             void validate_and_infer_types() override;
 
             virtual std::shared_ptr<Node>
                 copy_with_new_args(const NodeVector& new_args) const override;
 
             /// \return The concatenation axis.
-            size_t get_concatenation_axis() const { return get_axis(); }
-            void set_concatenation_axis(size_t concatenation_axis) { set_axis(concatenation_axis); }
+            int64_t get_concatenation_axis() const { return m_concat_axis; }
```
jbobba: @baojun-nervana @mozga-intel please change this to get_concat_axis() and I wouldn't expose a setter since this is an internal state variable, not to be manipulated directly by the op user.

PR #3711, opened 2019-10-03
Shape as a node in ConvBpropData/Filters (v1)

nbpatel:


src/ngraph/op/convolution.cpp
```diff
@@ -443,16 +473,16 @@ void op::v1::ConvolutionBackpropFilters::validate_and_infer_types()
                           delta_shape,
                           ").");
 
-    set_output_type(0, forward_result_et, m_filters_shape);
+    set_output_type(0, forward_result_et, get_filters_shape());
```
jbobba: if get_filters_shape() returns empty shape then set_output_type to PartialShape::Dynamic(). Also add set_input_relevant_to_shape

src/ngraph/op/convolution.cpp
```diff
@@ -231,12 +245,19 @@ void op::v1::ConvolutionBackpropData::validate_and_infer_types()
                           delta_shape,
                           ").");
 
-    set_output_type(0, forward_result_et, m_data_batch_shape);
+    set_output_type(0, forward_result_et, get_data_batch_shape());
```
jbobba: set shape to PartialShape::dynamic() if data_shape is not available. Add shape relevancy hints.

src/ngraph/op/convolution.cpp
```diff
@@ -231,12 +251,20 @@ void op::v1::ConvolutionBackpropData::validate_and_infer_types()
...
 
-    set_output_type(0, forward_result_et, m_data_batch_shape);
+    set_input_is_relevant_to_shape(2);
+    set_output_type(0, forward_result_et, data_batch_shape);
 }
 
 void op::v1::ConvolutionBackpropData::generate_adjoints(autodiff::Adjoints& adjoints,
                                                         const NodeVector& deltas)
 {
+    if (auto const_op = as_type<op::Constant>(input_value(2).get_node()))
```
jbobba: use is_constant() helper https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/node.hpp#L258

nbpatel:
@jbobba can you take a look at this again?

PR #3715, opened 2019-10-03
Add support for bf16 when creating constant.

XWangIntel:


src/ngraph/frontend/onnx_import/op/flatten.cpp
```diff
@@ -21,6 +21,7 @@
 #include "ngraph/builder/reshape.hpp"
 #include "utils/common.hpp"
 
+
```
diyessi: Extra file.

XWangIntel:
> Backends currently assume that op inputs have the same type; this change is a major API change that would break every current backend if mixed types showed up. The current approach is that framewor bridges should add explicit conversions according to framework semantics when a conversion is required.

Understandable. It is kind of hacky, and the PR is more like a showcase of it.

rkimballn1:
@diyessi Since Node is the basis of everything in ngraph perhaps we should not put validation there. Node should not care what the types of the inputs are but particular ops should care. If we put constructor validation of inputs in Node then we can't ever create a node with dissimilar inputs.

diyessi:
It's a method that a class may call. If your op doesn't want to call it, don't call the methods. Most ops don't call it.

rkimballn1:
The test in question was already in elementwise_args so it is fine. These changes belong in nnptr. We already have the ability to override ngraph core validation, we just need to use it

XWangIntel:
Revert the removal of validation. Only fix for constant builder remains

XWangIntel:
already merged in #3741

PR #3719, opened 2019-10-04
Restrict check to specific input for get_argument.

pchan:
The get_argument(index) checks for all inputs. I am facing an instance where the validation of Sum node[1] is being done for both inputs (output and reduction_axes) while my guess is it should be restircted only to reduction_axes.  I have put more context of the problem in ngraph slack. 

[1]
 Sum(const Output<Node>& arg, const AxisSet& reduction_axes)

### DIFF
--- a/src/ngraph/node.cpp
+++ b/src/ngraph/node.cpp
@@ -345,13 +345,10 @@ void Node::merge_provenance_tags_from(const std::shared_ptr<const Node>& source)
 
 std::shared_ptr<Node> Node::get_argument(size_t index) const
 {
-    for (auto& i : m_inputs)
-    {
-        NGRAPH_CHECK(i.get_output().get_node()->get_output_size() == 1,
-                     "child ",
-                     i.get_output().get_node()->get_name(),
-                     " has multiple outputs");
-    }
+    NGRAPH_CHECK(m_inputs.at(index).get_output().get_node()->get_output_size() == 1,
+                 "child ",
+                 m_inputs.at(index).get_output().get_node()->get_name(),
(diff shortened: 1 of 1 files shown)


rkimballn1:
This function does in fact work correctly and does exactly what it is defined to do. The user wants to find out which `Node` is producing the requested input, it is not defined to return the output port of that node, just return the Node. This is a very useful function and should not be deprecated. It is a helper function. I use it when I want to print out the name of the node supplying an input.
The current error check is wrong. It is a bug. The only error check that belongs in there is if the requested index is within the range of inputs and currently that is handled with `at(index)`
We should not deprecate this function.
If you are going to deprecate it then at least fix the bug until you remove it.

src/ngraph/node.cpp
```diff
@@ -345,13 +345,10 @@ void Node::merge_provenance_tags_from(const std::shared_ptr<const Node>& source)
...
 {
-    for (auto& i : m_inputs)
-    {
-        NGRAPH_CHECK(i.get_output().get_node()->get_output_size() == 1,
-                     "child ",
-                     i.get_output().get_node()->get_name(),
-                     " has multiple outputs");
-    }
+    NGRAPH_CHECK(m_inputs.at(index).get_output().get_node()->get_output_size() == 1,
+                 "child ",
```
diyessi: `shared_ptr<Node> result = input_value(index)->get_node_shared_ptr();
NGRAPH_CHECK(result->get_output_size() == 1 ...)
return result`

rkimballn1:
Replace by #3721

PR #3720, opened 2019-10-04
Fix Node::get_argment() check

rkimballn1:
The check inside this function was incorrect and not needed. With Input<Node> there is no need to check anything in this function for multiple outputs.
This is the r0.26 version of PR #3721


rkimballn1:
I am not addressing get_argument() in this I am addressing get_argument(index). It returns the correct result regardless of the number of outputs of the input node but the check prevents that from working. The check is left over from pre Input/Output<Node> days. The function works correctly but the check in the function is incorrect.

PR #3722, opened 2019-10-04
[SPEC] Implement StridedSlice:v1

mbencer:
Assumptions of StridedSlice:v1
- `v0` is `Slice`
- `stride` argument should be optional
-  attributes should be represented by `0` and `1` values mask vector where the position determines which axes should be used

Implemented elements:
- upgrade pass
- downgrade pass
- support for serialization


src/ngraph/op/experimental/dyn_slice.hpp
```diff
@@ -77,5 +127,6 @@ namespace ngraph
             AxisSet m_shrink_axis;
             AxisSet m_ellipsis_mask;
         };
+        using StridedSlice = DynSlice;
```
mbencer: I am not sure if the alias name is needed? @GlebKazantaev @ilyachur

src/ngraph/op/experimental/dyn_slice.cpp
```diff
@@ -45,6 +45,61 @@ op::DynSlice::DynSlice(const Output<Node>& arg,
...
+                       const std::vector<int64_t>& begin_mask,
+                       const std::vector<int64_t>& end_mask,
+                       const std::vector<int64_t>& new_axis_mask,
+                       const std::vector<int64_t>& shrink_axis_mask,
+                       const std::vector<int64_t>& ellipsis_mask)
+    : DynSlice(data,
+               begin,
+               end,
+               op::Constant::create(
+                   element::i64, Shape{begin_mask.size()}, vector<int64_t>(begin_mask.size(), 1)),
```
mbencer: Can we use `mask` size to determine shape of default stride?

src/ngraph/op/experimental/dyn_slice.hpp
```diff
@@ -54,6 +54,58 @@ namespace ngraph
...
+            /// \param shrink_axis_mask If shrink_axis_mask[i] is 1, the dimension
+            ///                         on the i-th position is deleted.
+            /// \param ellipsis_mask    It inserts missing dimensions
+            ///                         on a position of a non-zero bit.
+            DynSlice(const Output<Node>& data,
+                     const Output<Node>& begin,
+                     const Output<Node>& end,
+                     const Output<Node>& stride,
+                     const std::vector<int64_t>& begin_mask,
+                     const std::vector<int64_t>& end_mask,
```
silee2: Default argument?

silee2: TF default is 0 which would be an empty vector for nGraph.

mbencer: Current implementation introduces `StridedSlice:v1`. Default values of masks arguments are empty vectors (according to specification). @silee2 Is it something wrong with it?

mbencer:
> @jbobba Can we just rename `DynSlice` to `StridedSlice`?

@postrational, @jbobba  In accordance with arrangements I've introduced `StridedSlice:v1`

src/ngraph/op/experimental/strided_slice.hpp
```diff
@@ -0,0 +1,120 @@
+//*****************************************************************************
```
jbobba: I think of StridedSlice as v1 of Slice (https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/slice.hpp)
We could leave out DynSlice unversioned and eventually drop it from experimental. Upgrade/Downgrade should convert ops between v0::Slice and v1::StridedSlice

mbencer: I see the the following disadvantages of introduction `StridedSlice` as v1 of `Slice`:
- during downgrade pass: `begin`, `end`, `strides` must be constant
- during downgrade pass: conversion `begin_mask` and `end_mask` seems to be easy, but handling `new_axis_mask`, `shrink_axis_mask`, `ellipsis_mask` is rather problematic
The only limit of using  `StridedSlice` as v1 of `DynSlice` which I see is requirement of static rank of input during upgrade pass.
Execution downgrade pass for `dynamic backend` was added anyway for `Reshape:v1`.
In case of op location I think you are right that `experimental` is probably not good place for `StridedSlice:v1` (for `DynSlice` I don't know?)
@jbobba  What do you think about these arguments? Can we leave `StrideSlice` as v1 of `DynSlice`? :)

jbobba: During downgrade passes, we expect certain inputs to be constants anyway. For dealing with new_axis_mask, shrink_mask etc. isn't it the same as https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/pass/dyn_elimination.cpp#L180

mbencer: @jbobba Using `Slice` as `v0` as `StridedSlice` was implemented

src/ngraph/op/strided_slice.cpp
```diff
@@ -0,0 +1,191 @@
...
+    auto are_mask_elem_in_range = [](size_t e) { return e == 0 || e == 1; };
+    NODE_VALIDATION_CHECK(
+        this,
+        std::all_of(m_begin_mask.begin(), m_begin_mask.end(), are_mask_elem_in_range) &&
+            std::all_of(m_end_mask.begin(), m_end_mask.end(), are_mask_elem_in_range) &&
+            std::all_of(m_new_axis_mask.begin(), m_new_axis_mask.end(), are_mask_elem_in_range) &&
+            std::all_of(
+                m_shrink_axis_mask.begin(), m_shrink_axis_mask.end(), are_mask_elem_in_range) &&
+            std::all_of(m_ellipsis_mask.begin(), m_ellipsis_mask.end(), are_mask_elem_in_range),
+        "All maks elements of StridedSlice must have be 0 or 1");
```
itikhono: minor typo: maks -> mask 
same in other places

src/ngraph/op/strided_slice.hpp
```diff
@@ -0,0 +1,120 @@
...
+    {
+        namespace v1
+        {
+            /// \brief Takes a slice of an input tensor, i.e., the sub-tensor that resides within a
+            ///        bounding box, optionally with stride.
+            class StridedSlice : public Op
+            {
+            public:
+                NGRAPH_API
+                static constexpr NodeTypeInfo type_info{"Slice", 1};
```
itikhono: "Slice" or "StridedSlice"?

src/ngraph/op/strided_slice.hpp
```diff
@@ -0,0 +1,120 @@
...
+                             const Output<Node>& begin,
+                             const Output<Node>& end,
+                             const Output<Node>& strides,
+                             const std::vector<int64_t>& begin_mask,
+                             const std::vector<int64_t>& end_mask,
+                             const std::vector<int64_t>& new_axis_mask = std::vector<int64_t>{},
+                             const std::vector<int64_t>& shrink_axis_mask = std::vector<int64_t>{},
+                             const std::vector<int64_t>& ellipsis_mask = std::vector<int64_t>{});
+
+                /// \brief Constructs a dynamic tensor strided slice operation.
```
itikhono: As I can see, this constructor for the case when strides are optional and have a default value, could you please shortly describe it here?

src/ngraph/op/strided_slice.cpp
mbencer: Fixed

src/ngraph/op/strided_slice.hpp
mbencer: `StridedSlice` is `v1` (newest version of `op`) of `Slice`. According to previous conclusion name of `v0` op should be kept

src/ngraph/op/strided_slice.hpp
mbencer: Default value of `strides` is set here: https://github.com/NervanaSystems/ngraph/blob/7f2c9b275e53bf2de230b484db29eb442120e096/src/ngraph/op/strided_slice.cpp#L58 as `Constant` with vector of `1` (length of this vector is input rank -> the same as mask vector size)

PR #3733, opened 2019-10-07
[SPEC] LSTMCell, RNNCell updates.

arogowie-intel:
Changes:

**LSTMCelll**
* Changed input order from [X, W, R, Ht, Ct, B, P] -> [X, Ht, Ct, W, R, B, P]
* Added parameter `weights_format` which controls gates' order in weights and bias tensors.
* Remove weights and bias reordering from `LSTMFusion`

**RNNCell**
* Changed input order from [X, W, R, Ht, B]  -> [X, Ht, W, R, B]
* renamed attribute `activation_alpha` -> `activations_alpha`
* renamed attribute `activation_beta` -> `activations_beta`


src/ngraph/op/fused/lstm_cell.cpp
```diff
@@ -225,12 +210,19 @@ NodeVector op::LSTMCell::decompose_op() const
...
-    Output<Node> bias = get_bias();
-    NodeVector p_iof = get_peephole_weights();
+    Output<Node> H_t = input_value(1);
+    Output<Node> C_t = input_value(2);
+    Output<Node> W = input_value(3);
+    Output<Node> R = input_value(4);
+    Output<Node> bias = input_value(5);
+    NodeVector p_iof = builder::split(input_value(6), s_peepholes_count);
+
+    if (m_weights_format != op::LSTMWeightsFormat::IFCO)
```
tsocha: It is worth to add a comment why default format of weights is `op::LSTMWeightsFormat::IFCO` IMHO.

src/ngraph/op/fused/lstm_cell.cpp
```diff
@@ -319,11 +312,7 @@ shared_ptr<Node> op::LSTMCell::copy_with_new_args(const NodeVector& new_args) co
                                      new_args.at(3),
                                      new_args.at(4),
                                      get_hidden_size(),
-                                     get_activations(),
-                                     get_activation_alpha(),
-                                     get_activation_beta(),
-                                     get_clip(),
-                                     m_input_forget);
```
tsocha: Are you removed all of these parameters intentional?

src/ngraph/op/fused/lstm_cell.hpp
```diff
@@ -31,13 +31,40 @@ namespace ngraph
 {
     namespace op
     {
+        enum class LSTMWeightsFormat
+        {
+            FICO, // IE
+            ICOF, // PyTorch
+            IFCO, // DNNL, TF, MxNet
+            IFOC, // Caffe
+            IOFC, // ONNX
```
tsocha: Nice idea to document it like this!

src/ngraph/runtime/cpu/unit_test.manifest
```diff
@@ -28,3 +28,6 @@ random_uniform_all_static_seed_used
 random_uniform_seed_use_dynamic
 random_uniform_all_static_range_dynamic
 random_uniform_dynamic_shapes
+
+# Failing on CPU
+lstm_cell_no_bias_no_peepholes
```
tsocha: Thats weird because default bias and peepholes should be equal to zero.

src/ngraph/op/fused/lstm_cell.cpp
arogowie-intel: Forgot to undo :)

src/ngraph/op/fused/lstm_cell.cpp
arogowie-intel: Hmm this is already explained in the doc of `convert_node_format` function. But will add it here also.

src/ngraph/runtime/cpu/unit_test.manifest
tsocha: Could you add an unit test with bias and peepholes filled with default values?
If this kind of test passed, that would mean something strange.

src/ngraph/frontend/onnx_import/op/lstm.cpp
```diff
@@ -93,17 +94,20 @@ namespace ngraph
                             m_map[LSTMInput::LSTM_INPUT_W]->get_shape().front();
 
                         // ------ Optional inputs ------
-                        // The bias tensor for input gate. Shape [num_directions, 8*hidden_size]
+                        // The bias tensor for input gate. Shape [num_directions, 4*hidden_size]
                         if (ng_inputs.size() > 3 && !ng_inputs.at(3)->is_null())
                         {
-                            m_map[LSTMInput::LSTM_INPUT_B] = ng_inputs.at(3);
+                            auto bias = ng_inputs.at(3);
+                            auto splitted_bias = builder::split(bias, 2, 1);
```
postrational: `splitted` is an nonstandard or archaic form of the past participle. Should be `split`
```suggestion
                            auto split_bias = builder::split(bias, 2, 1);
```

src/ngraph/frontend/onnx_import/op/lstm.cpp
```diff
@@ -93,17 +94,20 @@ namespace ngraph
...
+                            m_map[LSTMInput::LSTM_INPUT_B] =
+                                splitted_bias.at(0) + splitted_bias.at(1);
```
postrational:
```suggestion
                                split_bias.at(0) + split_bias.at(1);
```

src/ngraph/op/fused/lstm_cell.cpp
```diff
@@ -274,38 +266,38 @@ NodeVector op::LSTMCell::decompose_op() const
...
-    Output<Node> P =
-        op::Constant::create(input(0).get_element_type(),
-                             Shape{s_peepholes_count * get_hidden_size()},
-                             vector<float>(s_peepholes_count * get_hidden_size(), 0.f));
-    set_argument(6, P);
+    static const std::map<op::LSTMWeightsFormat, std::vector<size_t>> gate_order_conversion_map{
+        {op::LSTMWeightsFormat::FICO, {1, 0, 2, 3}},
+        {op::LSTMWeightsFormat::ICOF, {0, 3, 1, 2}},
+        {op::LSTMWeightsFormat::IFOC, {0, 1, 3, 2}},
+        {op::LSTMWeightsFormat::IOFC, {0, 2, 3, 1}},
```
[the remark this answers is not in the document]
arogowie-intel: We're not missing it, since we're converting other supported formats to `IFCO` as it is in [function doc](https://github.com/NervanaSystems/ngraph/blob/tsocha/fused-lstm/src/ngraph/op/fused/lstm_cell.hpp#L169-L178).

src/ngraph/runtime/cpu/pass/cpu_rnn_fusion.cpp
```diff
@@ -160,13 +154,17 @@ void ngraph::runtime::cpu::pass::LSTMFusion::construct_onnx_lstmcell_fprop()
...
-        auto R_iofc = pattern_map[R];
-        auto W_ifco = get_weights_ifco_gate_order(W_iofc);
-        auto R_ifco = get_weights_ifco_gate_order(R_iofc);
-        // here onnx bias will be of shape (2 * gates_count * hidden_size) bias of Wb and Rb are
-        // concatenated, we will split the bias, add and rearrange in order IFCO
-        auto bias_ifco = get_bias_ifco_gate_order(bias_iofc);
+        // we need to reorder W, R and bias from IOFC to IFCO gate order
+        // Note: ONNX runtime provides W, R and bias in the gate order [IOFC] but
+        // MKLDNN computes LSTM kernel in the [IFCO] order.
+        if (lstmcell_op->get_weights_format() != op::LSTMWeightsFormat::IFCO)
```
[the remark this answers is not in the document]
arogowie-intel: @pruthviIntel Only the following test cases should run with MKLDNN on CPU backend:
* [lstm_cell_no_bias_no_peepholes](https://github.com/NervanaSystems/ngraph/blob/tsocha/fused-lstm/test/backend/fused_op.in.cpp#L1286-L1346)
* [lstm_cell_zero_bias_peepholes_constant](https://github.com/NervanaSystems/ngraph/blob/tsocha/fused-lstm/test/backend/fused_op.in.cpp#L1419-L1484),
since they're passing last two inputs as `Constant` which satisfies [this condition](https://github.com/NervanaSystems/ngraph/blob/tsocha/fused-lstm/src/ngraph/runtime/cpu/cpu_external_function.cpp#L1199). 
When CPU backend has support for `LSTMCell` then the op decomposition is **not** run. I added the check you cited to make sure we're not converting weights which are already in correct order.

src/ngraph/op/fused/gru_cell.hpp
```diff
@@ -65,55 +65,59 @@ namespace ngraph
...
-            ///                               in order respective to activation list.
-            /// \param[in]  clip              The value defining clipping range [-clip, clip] on
-            ///                               input of activation functions.
+            /// \param[in]  X                  The input tensor with shape: [batch_size,
+            ///                                input_size].
+            /// \param[in]  W                  The weight tensor with shape:
+            ///                                [gates_count * hidden_size, input_size].
+            /// \param[in]  R                  The recurrence weight tensor with shape:
+            ///                                [gates_count * hidden_size, hidden_size].
+            /// \param[in]  H_t                The hidden state tensor at current time step with
```
itikhono: I think we should have the same names for the similar inputs of the LSTMCell, LSTMSequence, GRUCell, and RNN Cell operations.  I suggest rename H_t, C_t to initial_hidden_state, initial_cell_state as in LSTMSequence operation. I will fix the specification on our side.

tsocha: Done

PR #3736, opened 2019-10-07
Use Eigen kernel for REFLECT mode Pad.

ayzhuang:


src/ngraph/runtime/cpu/builder/pad.cpp
```diff
@@ -50,7 +50,8 @@ namespace ngraph
                 auto padding_above = pad->get_padding_above();
                 auto pad_mode = pad->get_pad_mode();
 
-                if (pad_mode == ngraph::op::PadMode::CONSTANT &&
+                if ((pad_mode == ngraph::op::PadMode::CONSTANT ||
+                     pad_mode == ngraph::op::PadMode::REFLECT) &&
                     is_optimized_et(args[0].get_element_type()))
```
gaurides: I see you removed this element_type check in PR #3745 . Do we need it here?

ayzhuang: I removed it from the PR for r0.26 because the PR introduced this check is not in r0.26. We need it here.

PR #3737, opened 2019-10-08
[SPEC] Added support for v1 Broadcast op specification

jbobba:
- [x] Add more unit tests for v1 ops including PDPD broadcast spec


src/ngraph/op/broadcast.cpp
```diff
@@ -15,29 +15,278 @@
...
+                             const AutoBroadcastSpec& broadcast_spec)
+    : Op({arg, target_shape, axes_mapping})
+    , m_broadcast_spec(broadcast_spec)
+{
+    constructor_validate_and_infer_types();
+}
+
+op::v1::Broadcast::Broadcast(const Output<Node>& arg,
+                             const Output<Node>& target_shape,
+                             const AutoBroadcastSpec& broadcast_spec)
```
GlebKazantaev: Set it to NUMPY broadcasting by default.

src/ngraph/op/broadcast.cpp
```diff
@@ -15,29 +15,278 @@
...
+    : Op({arg, target_shape, op::Constant::create(element::u8, Shape{}, {0})->output(0)})
```
GlebKazantaev: Also I realized that in case with 2 inputs broadcast_spec must be NUMPY. So let's move it from constructor to set m_broadcast_spec(NUMPY) directly.

src/ngraph/op/broadcast.cpp
```diff
@@ -15,29 +15,278 @@
...
+            for (auto i = axes_mapping_val.rbegin(); i != axes_mapping_val.rend(); ++i)
+            {
+                axes.erase(axes.begin() + *i);
+            }
+            broadcast_axes.insert(axes.begin(), axes.end());
+            axes_known = true;
+        }
+    }
+    else if (m_broadcast_spec.m_type == AutoBroadcastType::NUMPY ||
+             m_broadcast_spec.m_type == AutoBroadcastType::PDPD)
```
GlebKazantaev: I'm not sure about AutoBroadcastType::PDPD, what is the specific here?

src/ngraph/op/broadcast.cpp
```diff
@@ -15,29 +15,278 @@
...
+    }
+
+    PartialShape result_shape{PartialShape::dynamic()};
+    if (input_value(1).get_node_shared_ptr()->is_constant())
+    {
+        result_shape = static_pointer_cast<op::Constant>(input_value(1).get_node_shared_ptr())
+                           ->get_shape_val();
+    }
+
+    if (m_broadcast_spec.m_type == AutoBroadcastType::NONE)
```
GlebKazantaev: According to spec there are only two types of broadcasting NUMPY and EXPLICIT, so what is NONE and PDPD?

jbobba: NONE is the same as EXPLICIT. AutoBroadcastSpec is also being used in elementwise with implicit broadcasting. So I left the name as 'NONE'. i can create an alias if that helps. 
PDPD is paddle paddle auto-broadcast. The definition is in util/attr_types.hpp. We cant assume that numpy is the only possibility here.

GlebKazantaev: I agree with NONE for me it's looks more natural than EXPLICIT. I just want spec to be aligned with operations, so in this case I would suggest us to update spec with NONE. Also if PDPD has different broadcast logic it should be also described in spec.

test/opset_pass/broadcast_opset_pass.cpp
```diff
@@ -0,0 +1,63 @@
...
+    pass_manager.run_passes(f);
+
+    auto bcast_v0 = static_pointer_cast<op::v0::Broadcast>(
+        f->get_results().at(0)->input_value(0).get_node_shared_ptr());
+
+    EXPECT_EQ(bcast_v0->description(), "Broadcast");
+    EXPECT_EQ(bcast_v0->get_version(), 0);
+    EXPECT_EQ(bcast_v0->get_broadcast_shape(), (Shape{3, 1, 4, 2, 3}));
+    EXPECT_EQ(bcast_v0->get_broadcast_axes(), (AxisSet{0, 2}));
+}
```
postrational:
```suggestion
}

```

src/ngraph/op/broadcast.hpp
```diff
@@ -18,88 +18,169 @@
...
+                /// \param arg            Node that produces the input tensor to be broadcast.
+                /// \param target_shape   The shape of the output tensor.
+                /// \param broadcast_spec Broadcast specification to use for determining broadcast
+                ///                       axes
+                Broadcast(const Output<Node>& arg,
+                          const Output<Node>& target_shape,
+                          const AutoBroadcastSpec& broadcast_spec =
+                              AutoBroadcastSpec(AutoBroadcastType::NUMPY));
+
+                size_t get_version() const override { return 1; }
```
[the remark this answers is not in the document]
jbobba: get_version() in base class returns 0

src/ngraph/op/broadcast.hpp
```diff
@@ -18,88 +18,169 @@
...
+                virtual std::shared_ptr<Node>
+                    copy_with_new_args(const NodeVector& new_args) const override;
+
+                /// \return Broadcast Specification.
+                const AutoBroadcastSpec& get_broadcast_spec() const { return m_broadcast_spec; }
+                void set_broadcast_spec(const AutoBroadcastSpec& broadcast_spec)
+                {
+                    m_broadcast_spec = broadcast_spec;
+                }
+
```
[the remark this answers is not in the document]
jbobba: will add in a separate PR.

src/ngraph/op/broadcast.cpp
jane-intel: @jbobba, should we update the spec with this new broadcasting type? PDPD? Can you please talk with @slyalin, or please create a merge request to the spec.

src/ngraph/op/broadcast.cpp
GlebKazantaev: @jbobba let's set `broadcast_spec` to NUMPY by default.

src/ngraph/op/broadcast.hpp
```diff
@@ -18,88 +18,169 @@
...
+            {
+            public:
+                NGRAPH_API
+                static constexpr NodeTypeInfo type_info{"Broadcast", 1};
+                const NodeTypeInfo& get_type_info() const override { return type_info; }
+                /// \brief Constructs a broadcast operation.
+                Broadcast() = default;
+                /// \brief Constructs a broadcast operation.
+                ///
+                /// \param arg            Node that produces the input tensor to be broadcast.
```
diyessi: "Value" instead of "Node" since nodes can gave more than one output. Same for other node params.

PR #3742, opened 2019-10-08
Implements proveance tag propagation for reshape sinking pass

tachyon77:
Adds some special case handling for provenance tag propagation for reshape sinking.

There are two cases where nodes are inserted or graph is modified that do not use replace_node() function where provenance is normally handled. These cases are node insertion and calling replace_source_output() directly (which is what replace_node call does normally).

Also the delete_node function does not need to do anything for provenance.

### DIFF
--- a/src/ngraph/pass/reshape_sinking.cpp
+++ b/src/ngraph/pass/reshape_sinking.cpp
@@ -105,6 +105,7 @@ static void insert_reshape(shared_ptr<Node> target, shared_ptr<Node> reshape, si
     auto arg = target->input(input_index).get_source_output();
     NGRAPH_DEBUG << "Arg shape: " << arg.get_shape();
     auto new_reshape = reshape->copy_with_new_inputs({arg});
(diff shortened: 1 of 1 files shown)


src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -115,7 +116,7 @@ static void delete_reshape(shared_ptr<Node> reshape)
     NGRAPH_DEBUG << "Removing reshape " << reshape->get_name();
     if (!reshape->get_users().empty())
     {
-        ngraph::replace_node(reshape, reshape->get_argument(0));
+        ngraph::replace_node(reshape, reshape->get_argument(0), true);
```
diyessi: Use `reshape->input_value(0).get_node_shared_ptr()` or this will fail if you are reshaping something not from an output 0.

tachyon77: @diyessi I think input_values() isn't there in r0.251

PR #3745, opened 2019-10-08
Migrate #3736 from master.

ayzhuang:


src/ngraph/runtime/cpu/kernel/pad.hpp
```diff
@@ -88,6 +92,8 @@ namespace ngraph
                         {
                             NGRAPH_CHECK(padding_below[i] > INT_MIN);
                             indices[i] = -padding_below[i];
+                            temp_dims[i] -= padding_below[i];
```
gaurides: Since padding_below[i] is < 0, shouldn't this be `temp_dims[i] += padding_below[i]` ?

ayzhuang: temp_dims[i] is larger than the out_dims[i] when padding_below[i] < 0. In that case, we slice temp to get the correct output. Suppose the input is [1, 2, 3] and the padding is {-1, 2}, temp would be [1, 2, 3, 2, 1] and output would be [2, 3, 2, 1].

PR #3748, opened 2019-10-08
[MLIR] Change order of dot operation loop nest

nmostafa:
Change loop ordering for better cache locality on RHS matrix.


test/mlir/affine_conversion/core_ops.mlir
```diff
@@ -15,3 +15,21 @@ func @simple_gather(%arg0: !ng.tensor<16x!ng.i64>, %arg1: !ng.tensor<512x32xf32>
    %0 = "ng.gather"(%arg1, %arg0) {axis = 0 : i64} : (!ng.tensor<512x32xf32>, !ng.tensor<16x!ng.i64>) -> !ng.tensor<16x32xf32>
   "ng.return"(%0) : (!ng.tensor<16x32xf32>) -> ()
 }
+
```
dcaballe: Please, add the `// -----` for split-input-file (line 5)

test/mlir/affine_conversion/core_ops.mlir
```diff
@@ -15,3 +15,21 @@ func @simple_gather(%arg0: !ng.tensor<16x!ng.i64>, %arg1: !ng.tensor<512x32xf32>
...
+// Dot Op
+// CHECK:       affine.for [[I:%.*]] = 0 to 16
+// CHECK-NEXT:  affine.for [[J:%.*]] = 0 to 32
+// CHECK-NEXT:  affine.store %{{.*}}, [[RESULT:%.*]]{{\[}}[[I]], [[J]]{{\]}}
```
dcaballe: We can leave `%` out of the capture and then we wouldn't need to escape `[`:
```
// CHECK:       affine.for %[[I:.*]] = 0 to 16
// CHECK-NEXT:  affine.for %[[J:.*]] = 0 to 32
// CHECK-NEXT:  affine.store %{{.*}}, %[[RESULT:.*]][%[[I]], %[[J]]]
```
Feedback provided here: https://github.com/tensorflow/mlir/pull/162/files#diff-ea810dda41c98feaab3d122f33648fb5R2336

test/mlir/affine_conversion/core_ops.mlir
```diff
@@ -15,3 +15,21 @@ func @simple_gather(%arg0: !ng.tensor<16x!ng.i64>, %arg1: !ng.tensor<512x32xf32>
...
+//
```
dcaballe: Could we check for `}` so that we are sure we have two separate loop nests?

PR #3749, opened 2019-10-08
Add Softmax::validate_and_infer_types

csullivan:
Currently `element::Types` and `Shapes` are not propagated for Softmax. This adds `Softmax::validate_and_infer_types` to fix that. Without this, bf16 support via provenance metatype tags in NNP does not function correctly as the assumption is that types will always be inferred when validation is run.


jbobba:
Thank you!

PR #3751, opened 2019-10-09
[SPEC] Implement Reshape downgrade pass

mbencer:
- downgrade pass for `Reshape:v1` was implemented (`v0` version of `Reshape:v1` is `DynReshape:v0` - I know it is a bit confusing)
- `Opset0Downgrade` was added to dynamic wrapper of backend


src/ngraph/op/reshape.hpp
```diff
@@ -117,7 +117,7 @@ namespace ngraph
             {
             public:
                 NGRAPH_API
-                static constexpr NodeTypeInfo type_info{"Reshape", 1};
+                static constexpr NodeTypeInfo type_info{"DynReshape", 1};
```
mbencer: `Reshape:v1` description is `DynReshape` in order to deal with current `get_typeid` implementation (`Reshape:v1` is higher version of `DynReshape:v0`)

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -174,6 +175,7 @@ bool runtime::dynamic::DynamicExecutable::call(
     }
 
     pass::Manager passes;
+    passes.register_pass<pass::Opset0Downgrade>();
```
mbencer: In current version for dynamic ops `Opset0Downgrade` is performed twice (firstly for `dynamic shape` backend and secondly for inner `static`). Any idea how to deal with it?

test/opset_pass/dyn_reshape_opset_pass.cpp
```diff
@@ -19,13 +19,14 @@
...
 #include "ngraph/pass/manager.hpp"
+#include "ngraph/pass/opset0_downgrade.hpp"
 #include "ngraph/pass/opset1_upgrade.hpp"
 #include "util/type_prop.hpp"
 
 using namespace std;
 using namespace ngraph;
 
-TEST(serialize, opset1_dyn_reshape_upgrade)
+TEST(opset_upgrade, opset1_dyn_reshape_upgrade)
```
postrational:
```suggestion
TEST(opset_transform, opset1_dyn_reshape_upgrade_pass)
```

test/opset_pass/dyn_reshape_opset_pass.cpp
```diff
@@ -40,8 +41,30 @@ TEST(serialize, opset1_dyn_reshape_upgrade)
...
     const auto pass_replacement_node =
         f->get_result()->input(0).get_source_output().get_node_shared_ptr();
-    const auto reshape_v1 = static_pointer_cast<op::v1::Reshape>(pass_replacement_node);
+    const auto reshape_v1 = as_type_ptr<op::v1::Reshape>(pass_replacement_node);
 
     EXPECT_EQ(reshape_v1->description(), "DynReshape");
     EXPECT_EQ(reshape_v1->get_version(), 1);
 }
+
+TEST(opset_downgrade, opset1_reshape_downgrade)
```
postrational:
```suggestion
TEST(opset_transform, opset1_reshape_downgrade_pass)
```

src/ngraph/runtime/dynamic/dynamic_backend.cpp
tsocha: What if some other backend would like to get and opset1 instead of opset0?
example:
1. Produce opset1 operators.
2. Downgrade operators to opset0
3. Pass these operators to backend which require opset1.
4. Upgrade operators to opset1
5. Run computation

This flow looks invalid IMHO.

mbencer: @jbobba @diyessi Have you idea how to handle such scenario?

mbencer: In current state passes flow in `dynamic_backend`:
- `ConstantFolding`
- `Opset0Downgrade`
- `DynElimination`
works both for `Broadcast:v1`, `Reshape:v1` and  `StridedSlice:v1`
@jbobba Is it good solution?

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -178,8 +178,8 @@ bool runtime::dynamic::DynamicExecutable::call(
 
     pass::Manager passes;
     passes.register_pass<pass::ConstantFolding>();
-    passes.register_pass<pass::DynElimination>();
     passes.register_pass<pass::Opset0Downgrade>(); // Converts dynamic v1 variants to v0 ops
+    passes.register_pass<pass::DynElimination>();
```
jbobba: Is this required?

mbencer: You are right, thank you, I confirm that order (from your `Broadcast` PR):
- ConstantFolding
- DynElimination
- Opset0Downgrade
works here too ;)

src/ngraph/runtime/dynamic/dynamic_backend.cpp
jbobba: @mbencer We should make these upgrade/downgrade passes more configurable so backends can decide which ops to downgrade/upgrade. Something similar to FusedOpDecomposition.

src/ngraph/op/reshape.hpp
jbobba: This is going to get confusing. Why cant we stick to "Reshape" here?

mbencer: @jbobba As I mentioned in first comment to this line `Reshape:v1` is `v1` of `DynReshape`. Current implementation of `get_typeid` (which provides using `switch` over op types) doesn't support it. However, I agree is confusing. Should we create separate task to work on it?

jbobba: Yes, please do.

PR #3755, opened 2019-10-10
Unify test names for opset transformation passes

postrational:


test/opset_pass/convolution_opset_pass.cpp
```diff
@@ -10,7 +10,7 @@
 using namespace std;
 using namespace ngraph;
 
-TEST(upgrade_pass, opset1_convolution_pass)
+TEST(opset_transform, opset1_convolution_upgrade_pass)
```
tsocha: I'm not a fan of putting all of upgrade/downgrade passes in the same test group.
In my opinion that will be better if we would have `opset_upgrade` and `opset_downgrade` test groups.

PR #3759, opened 2019-10-10
Add partial slice fused op for pdpd use

baojun-nervana:


src/ngraph/op/fused/partial_slice.hpp
```diff
@@ -0,0 +1,109 @@
...
+                                           const NodeVector& deltas) override;
+
+        private:
+            AxisVector m_axes;
+            std::vector<int64_t> m_lower_bounds;
+            std::vector<int64_t> m_upper_bounds;
+            AxisVector m_decrease_axes;
+        };
+
+        /// \brief Layer Normalization Backprop
```
jbobba: Modify

src/ngraph/runtime/executable.cpp
```diff
@@ -81,6 +81,7 @@ void runtime::Executable::validate(const vector<std::shared_ptr<runtime::Tensor>
     for (size_t i = 0; i < results.size(); i++)
     {
         if (outputs[i]->get_element_type().is_static() &&
+            results[i]->get_element_type() != element::dynamic &&
```
jbobba: @diyessi would we want this to be results[i]->get_element_type().is_static() as well? Looks like we are treating 'undefined' type as a static type as well?

PR #3764, opened 2019-10-11
[MLIR] New Core Ops (V0) and Ops Versioning in NG dialect

nmostafa:
This change adds two operation interfaces, one for each version of ngraph opcodes. The interfaces don't hold any methods at the moment (but could possible do, for example, OpVersion0::upgradeToV1()) . To check if an op holds a specific version we do

```
auto operation =...
if (dyn_cast<OpVersion0>(operation)) { ... }
```
Opcode names and class-names changed to reflect version number. Version0 ops don't have a version suffix. 

The PR also adds declarations of most of core v0 ops (-experimental - fused).


src/contrib/mlir/compiler/dialect/ops_interfaces.td
```diff
@@ -0,0 +1,64 @@
...
+#define NG_OP_INTERFACES
+
+#ifdef OP_BASE
+#else
+include "mlir/IR/OpBase.td"
+#endif // OP_BASE
+
+// Op Interfaces for Op Versions
+// They are empty for now. To check the version of an op, we do:
+// 	Operation *op = …;
```
diyessi: Only up to 9. Then you need a regexp. If `Foo` the same as `Foo0` and `Foo000`? I think you just need to define version syntax and say "Don't to that" for variant ways to write the number since they'd be different classes.

src/contrib/mlir/compiler/dialect/fused_ops.td
```diff
@@ -0,0 +1,45 @@
...
+//
+// This is the nGraph Dialect Fused Ops definition file
+// All Operations in this file implement FusedOp interface.
+//===----------------------------------------------------------------------===//
+
+#ifdef NG_FUSED_OPS
+#else
+#define NG_FUSED_OPS
+// Squeeze Op
+def NGSqueezeOp :
```
nmostafa: More ops to be added in a separate PR.

PR #3771, opened 2019-10-14
[ONNX] Add support for ONNX 1.6 TopK

postrational:
Add support for ONNX `TopK` attributes: `largest` and `sorted`


src/ngraph/runtime/cpu/unit_test.manifest
```diff
@@ -24,3 +24,6 @@ lrn_6D_across_2_axes
 
 # ONNX TopK with dynamic K
 top_k_opset_10
+
+# ONNX TopK with dynamic K
+top_k_opset_10
```
arogowie-intel: There is duplicate `top_k_opset_10`. Didn't you intended to write: `top_k_opset_11_const_k_smallest` ?

tsocha: You should remove it.

src/ngraph/frontend/onnx_import/op/topk.cpp
```diff
@@ -79,6 +92,31 @@ namespace ngraph
                 }
             }
 
+            namespace set_11
+            {
+                NodeVector topk(const Node& node)
+                {
+                    // Process inputs
+                    auto data = node.get_ng_inputs().at(0);
```
tsocha:
```suggestion
                    auto data = get_data(node);
```
Is it worth to add a new function for consistency?

src/ngraph/frontend/onnx_import/op/topk.cpp
```diff
@@ -79,6 +92,31 @@ namespace ngraph
...
+                    auto k = get_k(node);
+
+                    // Process attributes
+                    const auto axis = get_axis(node);
+                    const auto largest = node.get_attribute_value<std::int64_t>("largest", 1);
+                    const auto sorted = node.get_attribute_value<std::int64_t>("sorted", 1);
```
tsocha: You can read these attributes to `bool` variable.

example from `gemm` operator:
```
                    bool trans_a = node.get_attribute_value<int64_t>("transA", 0);
                    bool trans_b = node.get_attribute_value<int64_t>("transB", 0);
```

src/ngraph/frontend/onnx_import/op/topk.cpp
```diff
@@ -20,25 +20,38 @@
...
-static ngraph::NodeVector get_outputs(const std::shared_ptr<ngraph::Node>& top_k)
+/// \return Return the second input to the TopK node reshaped to a scalar.
+static std::shared_ptr<ngraph::Node> get_k(const ngraph::onnx_import::Node& node)
 {
-    std::shared_ptr<ngraph::Node> indices =
-        std::make_shared<ngraph::op::GetOutputElement>(top_k, 0);
-    std::shared_ptr<ngraph::Node> values = std::make_shared<ngraph::op::GetOutputElement>(top_k, 1);
+    auto k_node = node.get_ng_inputs().at(1);
+    NGRAPH_CHECK(shape_size(k_node->get_shape()) == 1,
+                 "ONNX TopK operator: 'K' parameter must contain a single positive value.",
```
tomdol: The error message mentions the sign of a value but there's no check for a sign here.

src/ngraph/frontend/onnx_import/utils/reshape.cpp
```diff
@@ -100,6 +99,15 @@ namespace ngraph
                              "Scalar value can't be derived from a node with ",
                              node_shape);
 
+                // If node is a Constant, recreate as Constant with Shape{}
+                if (node->is_constant())
+                {
+                    const auto value =
+                        ngraph::as_type_ptr<ngraph::op::Constant>(node)->get_data_ptr();
+                    return std::make_shared<ngraph::op::Constant>(
```
tomdol: I think this should be unit tested to verify that values in the reshaped constant propagate/broadcast correctly.

src/ngraph/frontend/onnx_import/op/topk.cpp
postrational: The value is unknown at the time of graph import. We're only checking the shape, but I think the informative description of the expected value (from ONNX spec) may be helpful. I can remove the word `single` if you insist.

src/ngraph/frontend/onnx_import/op/topk.cpp
postrational: I don't think so. If the function did any preparation of the input node, then yes, but in this situation I think this is clearer.

src/ngraph/frontend/onnx_import/op/topk.cpp
postrational: I could, but we're changing this line in the `v1` upgrade this sprint.

src/ngraph/frontend/onnx_import/utils/reshape.cpp
postrational: Done.

test/CMakeLists.txt
```diff
@@ -342,9 +342,10 @@ if (NGRAPH_ONNX_IMPORT_ENABLE)
     list(APPEND MULTI_TEST_SRC
             onnx/onnx_import.in.cpp
             onnx/onnx_import_convpool.in.cpp
+            onnx/onnx_import_quant.in.cpp
             onnx/onnx_import_reshape.in.cpp
             onnx/onnx_import_rnn.in.cpp
-            onnx/onnx_import_quant.in.cpp)
+            onnx/onnx_import_utils_reshape.in.cpp)
```
tsocha: Could you sort this list?

PR #3776, opened 2019-10-15
[Spec] Donwgrade for Convolution op

arogowie-intel:


src/ngraph/pass/opset0_downgrade.cpp
```diff
@@ -83,6 +84,62 @@ bool pass::Opset0Downgrade::run_on_node(shared_ptr<Node> node)
 #endif
     switch (get_typeid(node))
     {
+    case OP_TYPEID::Convolution:
+    {
+        auto tmp = as_type_ptr<op::v1::Convolution>(node);
+        const auto data_arg = node->input(0).get_source_output();
+        const auto filters_arg = node->input(1).get_source_output();
+        const size_t num_spatial_dims = data_arg.get_shape().size() - 2;
```
tsocha: What if shape will be dynamic?
You should read `data_arg` rank and check if it is dynamic or not IMO.

src/ngraph/pass/opset0_downgrade.cpp
```diff
@@ -83,6 +84,62 @@ bool pass::Opset0Downgrade::run_on_node(shared_ptr<Node> node)
...
+        replace_node(node, replacement_node);
+        modified = true;
+        break;
+    }
+    case OP_TYPEID::ConvolutionBackpropData:
+    {
+        auto tmp = as_type_ptr<op::v1::ConvolutionBackpropData>(node);
+        const auto filters_arg = node->input(0).get_source_output();
+        const auto delta_arg = node->input(1).get_source_output();
+        const size_t num_spatial_dims = delta_arg.get_shape().size() - 2;
```
tsocha: the same as above

src/ngraph/pass/opset0_downgrade.cpp
```diff
@@ -83,6 +84,62 @@ bool pass::Opset0Downgrade::run_on_node(shared_ptr<Node> node)
...
+        replace_node(node, replacement_node);
+        modified = true;
+        break;
+    }
+    case OP_TYPEID::ConvolutionBackpropFilters:
+    {
+        auto tmp = as_type_ptr<op::v1::ConvolutionBackpropFilters>(node);
+        const auto data_arg = node->input(0).get_source_output();
+        const auto delta_arg = node->input(1).get_source_output();
+        const size_t num_spatial_dims = data_arg.get_shape().size() - 2;
```
tsocha: the same as above

PR #3780, opened 2019-10-16
Added aliases for v0 ops that are compatible v1

GlebKazantaev:
In this PR I've made an aliases for v0 ops that are compatible with v1. That will help us to divide operation between v0 and v1 more transparently.
This PR affects ops that are currently supported by IE. So for other ops I want you guys to divide them to v0 and v1.

### DIFF
--- a/src/ngraph/op/experimental/layers/detection_output.hpp
+++ b/src/ngraph/op/experimental/layers/detection_output.hpp
@@ -74,5 +74,11 @@ namespace ngraph
         private:
             DetectionOutputAttrs m_attrs;
         };
+
+        namespace v1
+        {
+            using op::DetectionOutput;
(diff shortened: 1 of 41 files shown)


src/ngraph/op/experimental/layers/detection_output.hpp
```diff
@@ -74,5 +74,11 @@ namespace ngraph
         private:
             DetectionOutputAttrs m_attrs;
         };
+
+        namespace v1
+        {
+            using op::DetectionOutput;
+            using op::DetectionOutputAttrs;
+        }
```
jbobba: Might be better to move all the layer ops into namespace v1 directly instead of aliasing.

PR #3782, opened 2019-10-17
Check for element-wise equality at construction and simplify zero/one…

jbobba:
… checks later

### DIFF
--- a/src/ngraph/op/constant.hpp
+++ b/src/ngraph/op/constant.hpp
@@ -70,6 +70,7 @@ namespace ngraph
                     write_values(values);
                 }
                 constructor_validate_and_infer_types();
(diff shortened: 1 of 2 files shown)


src/ngraph/op/constant.hpp
```diff
@@ -246,6 +249,10 @@ namespace ngraph
 
             bool is_constant() const override { return true; }
             bool are_all_data_elements_bitwise_identical() const;
```
csullivan: Should we change the impl. of this method to return the precomputed member after it's been initialized once?

jbobba:
AlgebraicSimplification on BERT runs 20X faster (60s->3s)

PR #3786, opened 2019-10-17
[MLIR] MLIR Compiler refactoring

nmostafa:
This changes breaks down MLIR Compiler into re-usable pieces:
1. MLIRCompiler: core compiler to translate from nGraph IR to nGraph dialect. Applies nGraph dialect level optimizations. 
2. MLIRBackend/MLIRCPUBackend: A backend with common interface to lower nGraph dialect. For CPU, it lowers all the way to LLVM dialect and applies optimizations along the way. 
3. MLIRRuntime/MLIRCPURuntime: A runtime with a common interface to run an MLIR module. For CPU, it invokes LLVM's ORC JIT which lazily compiles from LLVM dialect down to assembly.


src/contrib/mlir/runtime/cpu/cpu_runtime.cpp
```diff
@@ -0,0 +1,168 @@
+//*****************************************************************************
```
dcaballe: Same history comment for these files.

src/contrib/mlir/CMakeLists.txt
```diff
@@ -14,13 +14,97 @@
...
 add_subdirectory(tools/ngraph-opt)
+
+set(SRC
+    backend/cpu_backend.cpp
+    backend/pass/affine_lowerer.cpp
+    backend/pass/memory_optimization.cpp
+    core/compiler.cpp
+    core/dialect/dialect.cpp
+    core/dialect/type.cpp
+    core/dialect/ops.cpp
```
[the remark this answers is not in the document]
nmostafa: Renamed folder to ngraph_dialect.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -17,10 +17,10 @@
 // NOTE: This file follows nGraph format style and MLIR naming convention since it does
 // not expose public API to the rest of nGraph codebase and heavily depends on MLIR API.
 
-#include "lowerer.hpp"
+#include "affine_lowerer.hpp"
```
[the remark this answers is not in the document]
nmostafa: Rather change that later since we use lowerer throughout the file itself.

src/contrib/mlir/backend/cpu/cpu_backend.cpp
```diff
@@ -0,0 +1,274 @@
...
+        optCacheLevelSize = targetInfo.getCacheSize(llvm::TargetTransformInfo::CacheLevel::L2D);
+        break;
+    default:
+        NGRAPH_UNREACHABLE("Unsupported cache level: ", cacheLevel, ". Only 1 and 2 are supported");
+    }
+
+    NGRAPH_CHECK(optCacheLevelSize.hasValue() && "Cache level size is not available in TTI");
+    return optCacheLevelSize.getValue();
+}
+
```
diyessi: See `runtime/cpu/cpu_backend.cpp` for what to do so things work with static linking as well as dynamically loaded libraries. It's mainly having the correctly-named extern "C" function available, which could just call your init here.

nmostafa: I don't think this applies here since MLIRCPUBackend is not a replacement for `runtime/cpu/cpu_backend`. It is sub-component of the MLIR lib that gets linked into ngraph cpu backend. There is still more refactoring before we worry about static linking. Currently, Core, Backend and Runtime are all in one lib that we need to split up into 3. The Core will get linked to ngraph lib, and the rest into the ngraph backends.

PR #3788, opened 2019-10-18
[SPEC] Implement MaxPool:v1 and AvgPool:v1 downgrade pass

mbencer:
- implemented downgrade pass for `MaxPool:v1` and `AvgPool:v1`
- using `MaxPool:v1` and `AvgPool:v1` in `onnx_importer` (both for `average_pool`, `max_pool` and for `global_average_pool`, `global_max_pool`)


test/backend/pool.in.cpp
```diff
@@ -1374,7 +1374,7 @@ NGRAPH_TEST(${BACKEND_NAME}, avg_pool_bprop_2d_2channel_2image_dyn_shape)
                                                                padding_below,
                                                                padding_above,
                                                                window_shape,
-                                                               false);
+                                                               true);
```
mbencer: Changed because `Include padding in average not yet implemented in PlaidML` error for `plaidml` backend. In `master` version is bug.
`master`: `exclude_pad = tmp->get_exclude_pad();`
`current:` `include_padding_in_avg_computation = !tmp->get_exclude_pad();`
This param is passed to `AvgPoolBackprop:v0` during downgrade pass

PR #3789, opened 2019-10-18
[SPEC] Implement TopK:v1 downgrade

mbencer:
- implemented downgrade pass for `TopK:v1`
- using `v1` in `onnx_importer`


src/ngraph/pass/opset0_downgrade.cpp
```diff
@@ -212,6 +213,37 @@ bool pass::Opset0Downgrade::run_on_node(shared_ptr<Node> node)
...
+        switch (tmp->get_mode())
+        {
+        case op::v1::TopK::Mode::MAX: comnpute_max = true; break;
+        case op::v1::TopK::Mode::MIN: comnpute_max = false; break;
+        default:
+            NGRAPH_CHECK(true,
+                         "Mode ",
+                         static_cast<int>(tmp->get_mode()),
+                         " is not supported during TopK:1 to TopK:0 conversion. Node: ",
+                         *node);
```
tsocha: This check will always pass.
Since TopK::Mode is an enum it should be impossible to have an invalid value here.

mbencer: Because `clang` compiler check if all enum are used in swtich as default I removed this assert.

src/ngraph/frontend/onnx_import/op/topk.cpp
```diff
@@ -107,11 +119,17 @@ namespace ngraph
...
-                    std::shared_ptr<ngraph::Node> top_k = std::make_shared<ngraph::op::TopK>(
-                        data, k, axis, element::i64, compute_max, sort_type);
+                    const auto sort_type = sorted ? ngraph::op::v1::TopK::SortType::SORT_VALUES
+                                                  : ngraph::op::v1::TopK::SortType::NONE;
+
+                    std::shared_ptr<ngraph::Node> top_k = std::make_shared<ngraph::op::v1::TopK>(
+                        data,
+                        k,
+                        axis,
+                        compute_max ? ngraph::op::v1::TopK::Mode::MAX
```
postrational: Let's move this ternary operator up to keep consistent with `sort_type`

mbencer: Done

mbencer:
> `Topk:v1` returns: `values`, `indices`
> `Topk:v0` returns: `indices`, `values`
> During `downgrade` pass replacement output of nodes is needed. `ngraph::replace_node` should support determining order of outputs - it blocks this PR.
> PS.: Other solution can be introduction new `FusedOp` responsible for determining outputs order

I added extended implementation of `replace_node`. @jbobba, @diyessi Please for review ;)

src/ngraph/graph_util.cpp
```diff
@@ -131,13 +132,21 @@ NodeVector ngraph::find_common_args(std::shared_ptr<Node> node1, std::shared_ptr
     return common_args;
 }
 
-void ngraph::replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> replacement)
+void ngraph::replace_node(std::shared_ptr<Node> target,
```
diyessi: If this was `replace_value(Output<Node> target, Output<Node> replacement)` it wouldn't need these output size checks and `output_order` arguments.

mbencer: In current implementation: `replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> replacement)` (line: 198) call extended `replace_node(std::shared_ptr<Node> target, std::shared_ptr<Node> replacement, const std::vector<int64_t>& output_order)` with default `output_order` value (**to avoid copying code**), but you are right that for version without `output_order` it cases additional checking and vector allocation. 
@diyessi  Do you expect fully separate version of `replace_node` for both cases?

diyessi: Let's keep this one for now and think about how we can simplify replace node in general. You've had a fair amount of experience with it now.

PR #3791, opened 2019-10-18
Add constant folding for v1 reduce ops.

ayzhuang:


src/ngraph/runtime/reference/mean.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+                for (const Coordinate& output_coord : output_transform)
+                {
+                    out[output_transform.index(output_coord)] = 0;
+                    cs[output_transform.index(output_coord)] = 0;
+                }
+
+                CoordinateTransform input_transform(in_shape);
+                std::map<size_t, int> index_to_count_map;
+                ;
```
jbobba: extra semicolon

test/constant_folding.cpp
```diff
@@ -507,6 +507,62 @@ TEST(constant_folding, const_product)
...
+    ASSERT_EQ(count_ops_of_type<op::Constant>(f), 1);
+
+    auto new_const = as_type_ptr<op::Constant>(f->get_results().at(0)->get_argument(0));
+    ASSERT_TRUE(new_const);
+    auto values_out = new_const->get_vector<int32_t>();
+
+    vector<int32_t> values_expected{6, 120, 504};
+
+    ASSERT_EQ(values_expected, values_out);
+}
```
gaurides: i think we should also check the keep_dims part in these `keepdims` test (and others), check the output_shape

ayzhuang: Added output_shape checks.

PR #3794, opened 2019-10-21
Add constant folding for Squeeze and Unsqueeze.

ayzhuang:


test/constant_folding.cpp
```diff
@@ -24,6 +24,60 @@
...
+    pass_manager.run_passes(f);
+
+    ASSERT_EQ(count_ops_of_type<op::Squeeze>(f), 0);
+    ASSERT_EQ(count_ops_of_type<op::Constant>(f), 1);
+
+    auto new_const = as_type_ptr<op::Constant>(f->get_results().at(0)->get_argument(0));
+    ASSERT_TRUE(new_const);
+    auto values_out = new_const->get_vector<float>();
+
+    ASSERT_TRUE(test::all_close_f(values_in, values_out, MIN_FLOAT_TOLERANCE_BITS));
```
gaurides: Should we also check the output shape to ensure the squeeze/unsqueeze worked as expected?

ayzhuang: Done.

PR #3795, opened 2019-10-21
add serialize_types and deserialize_types API

ashokei:


src/ngraph/serializer.cpp
```diff
@@ -458,6 +458,37 @@ std::string ngraph::serialize(std::shared_ptr<ngraph::Function> func, size_t ind
     return ::serialize(func, indent, false);
 }
 
+template <typename T>
+std::string ngraph::serialize_attrs(const std::vector<std::shared_ptr<T>>& nodes)
```
diyessi: It's best if the serialize and deserialize have the same kinds of arguments.
Call it `serialize_types` and have it take a `const std::vector<std::pair<PartialShape, element::Type>>&` argument.

src/ngraph/serializer.cpp
```diff
@@ -458,6 +458,37 @@ std::string ngraph::serialize(std::shared_ptr<ngraph::Function> func, size_t ind
...
+        attrs.push_back(j);
+    }
+    return attrs.dump();
+}
+template std::string ngraph::serialize_attrs<ngraph::op::Result>(
+    const std::vector<std::shared_ptr<ngraph::op::Result>>& nodes);
+template std::string ngraph::serialize_attrs<ngraph::op::Parameter>(
+    const std::vector<std::shared_ptr<ngraph::op::Parameter>>& nodes);
+
+std::vector<std::pair<PartialShape, element::Type>>
```
diyessi: This one would be `deserialize_types`

src/ngraph/serializer.cpp
ashokei: agree..will make the change.

PR #3796, opened 2019-10-21
Added set_parameters_and_results with params and results

jiminha:
This is for AOT(Ahead of time compile) use case. 
During compile-time, we generate backend blob, so no need to keep the serialized function information other than inputs and outputs (to save space, and also to save time for serialize/deserialize). 

During run-time, we don't need function but we still need to set parameters and results, so the front end can create input and output tensors. Ngraph-bridge is using executable->get_parameters(), and executable->get_results(). To support this, I'm adding new api to just set the parameters and results without constructing new function.

### DIFF
--- a/src/ngraph/runtime/executable.cpp
+++ b/src/ngraph/runtime/executable.cpp
@@ -119,6 +119,13 @@ void runtime::Executable::set_parameters_and_results(const Function& func)
     m_results = func.get_results();
 }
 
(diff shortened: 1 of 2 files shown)


jiminha:
Ok will do, @diyessi should I submit both master and r0.25.1?

jiminha:
I had to open a new PR after baseline from r0.25.1
Please use https://github.com/NervanaSystems/ngraph/pull/3801 instead.

PR #3797, opened 2019-10-22
Fixed paths for python build

ilyachur:


silee2:
@ilyachur What issue are you trying to fix? All this PR is doing is moving the work directory deeper down the build tree.

silee2:
@ilyachur I see that you would need to change CMAKE_SOURCE_DIR CMAKE_CURRENT_SOURCE_DIR if you embed nGraph source code to find the correct location of sources. But you don't have to do the same for CMAKE_BINARY_DIR. Anyway LGTM.

silee2:
One unwanted side effect of PR would be different location for generated Wheel. If that is undesirable, CMAKE_CURRENT_BINARY_DIR should be reverted back to CURRENT_BINARY_DIR.

ilyachur:
@silee2 This PR doesn't change binary and source paths if you are build only ngraph, because before you had a code `${CMAKE_BINARY_DIR}/python/...` and I change it to `${CMAKE_CURRENT_BINARY_DIR}/...`. In other words these paths are equal if you build nGraph only.

These changes affect only path to pybind11 project. After these changes it will clone to `<ngraph_build_dir>/python/` folder.

ilyachur:
@aslepko Please, look at this PR.

PR #3798, opened 2019-10-22
[PY] Add check if pointer in capsule contains a nullptr

tsocha:


python/pyngraph/function.cpp
```diff
@@ -58,7 +58,7 @@ void regclass_pyngraph_Function(py::module m)
         auto* capsule_ptr = PyCapsule_GetPointer(pybind_capsule_ptr, CAPSULE_NAME);
 
         auto* ngraph_function = static_cast<std::shared_ptr<ngraph::Function>*>(capsule_ptr);
-        if (ngraph_function)
+        if (ngraph_function && *ngraph_function)
```
tomdol: You should be checking the contents of the `shared_ptr`, right now, you're dereferencing a `std::shared_ptr<...>*` which means in the second part of the if you're getting `std::shared_ptr<...>&` and that's not what should be checked.

Try this instead:
` if (ngraph_function && ngraph_function->get())`

PR #3802, opened 2019-10-23
Constant folding with shape inference; Updated specialize_function

GlebKazantaev:
In this PR I've combined constant folding with shape inference. This is necessary when part of nGraph Functions has dynamic shapes.
Also `specialize_function` method was updated to preserve names and support specialization to different shape.


src/ngraph/specialize_function.hpp
```diff
@@ -107,5 +107,6 @@ namespace ngraph
         specialize_function(std::shared_ptr<Function> f,
                             const std::vector<element::Type>& parameter_element_types,
                             const std::vector<PartialShape>& parameter_shapes,
-                            const std::vector<void*>& parameter_values);
+                            const std::vector<void*>& parameter_values,
+                            bool constant_folding = true);
```
jbobba: Can we make the default 'false'. ConstantFolding can be quite slow since we use reference kernels and it might come as a surprise to users if we enable by default.

ilyachur:
CI was fixed

PR #3808, opened 2019-10-25
Added filter for v1 operations

GlebKazantaev:


AnastasiaKazantaeva:
@diyessi please, review

I checked all tests if it covered v1 op or not and added to name mark "opv1". So, it helps to filter backend tests via filter "opv1"

AnastasiaKazantaeva:
@rkimballn1 Ok, I see. But in our validation we need to get percentage of passed and failed test for v1 operation to track it and improve. How should we do it?

PR #3809, opened 2019-10-25
[MLIR] Use normalized axis as a common name for negative axis support

baojun-nervana:
address comment https://github.com/NervanaSystems/ngraph/pull/3708#pullrequestreview-307343338


baojun-nervana:
> Could you please add `MLIR` somewhere in the title and rerun CI so that MLIR tests are executed? Thanks!

added to title and re-running

ilyachur:
I agree with @diyessi. Let's call it `axis`.

From my point of view it correlates with OpenVINO.

baojun-nervana:
We will need a different solution than just renaming. I will close this PR.

PR #3813, opened 2019-10-29
Opset1 Definition

jbobba:
Created Opset1 definition (src/ngraph/opsets/opset1.hpp)

Added versioning to most of the ops (version numbers start from zero but this is an internal detail).


test/build_graph.cpp
```diff
@@ -219,3 +221,9 @@ TEST(build_graph, validate_function_for_dynamic_shape)
     EXPECT_EQ(true, make_function(true)->is_dynamic());
     EXPECT_EQ(false, make_function(false)->is_dynamic());
 }
+
+TEST(build_graph, opset1)
+{
```
diyessi: Could do `ASSERT_TRUE(is_type<op::v0::Add>(opset1::Add())`
I don't think we actually ever need to know the op version number, just the versioned op.

src/ngraph/CMakeLists.txt
```diff
@@ -464,10 +464,10 @@ set (SRC
     pass/nop_elimination.hpp
     pass/pass.cpp
     pass/pass.hpp
-    pass/opset0_downgrade.cpp
-    pass/opset0_downgrade.hpp
-    pass/opset1_upgrade.cpp
-    pass/opset1_upgrade.hpp
+    pass/op_downgrade_v1tov0.cpp
```
diyessi: Good renaming.

src/ngraph/op/util/opset1.hpp
```diff
@@ -0,0 +1,29 @@
+//*****************************************************************************
```
postrational: Is `util` the best subdirectory to place this file? I would place it one or 2 levels up.

test/opset1.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "gtest/gtest.h"
+
+#include "ngraph/ngraph.hpp"
+#include "ngraph/opsets/opset1.hpp"
+
+#include <memory>
+using namespace std;
+using namespace ngraph;
+
+#define CHECK_OPSET(op1, op2) EXPECT_TRUE(is_type<op1>(make_shared<op2>()));
```
tomdol: Just wondering if instead of testing it in runtime, we could switch to `std::is_same` and have it checked in compile time.

https://ideone.com/WYkqla something like this

jbobba: We could possibly do both. We are checking the 'is_type' implementation as well in this test.

PR #3815, opened 2019-10-29
Fixed python build for some debug configurations

ilyachur:
I found an issue with python build on debug configuration for some platforms.
Fixed issue with different build tools.


python/setup.py
```diff
@@ -73,6 +74,9 @@ def find_pybind_headers_dir():
           'correctly'.format(NGRAPH_CPP_DIST_DIR))
     sys.exit(1)
 
+NGRAPH_CPP_LIBRARY_NAME = 'ngraph'
+if len([fn for fn in os.listdir(NGRAPH_CPP_LIBRARY_DIR) if re.search('ngraphd', fn)]):
```
postrational: Please add a comment explaining why we need to do this and where the name `ngraphd` comes from.

ilyachur: It is happened because for OpenVINO package on some platforms we provided Debug and Release binary files and in this case debug files have postfix `d`

ilyachur:
@postrational  @silee2 @aslepko Please look at this ASAP. It fixed some problems with Python API on current master.

python/setup.py
postrational: Yes, please add this as a comment to the code for future reference.

ilyachur: Done

ilyachur:
@postrational All comments were resolved. Please look at this PR

PR #3819, opened 2019-10-30
[SPEC] [MLIR] Adjust names of logical operators to the spec.

tsocha:
- And -> LogicalAnd
- Not -> LogicalNot
- Or -> LogicalOr
- LessEq -> LessEqual
- Xor -> LogicalXor (this branch have base on top of #3811 PR)


nmostafa:
I think this should pass, but might be safer to run MLIR CI for this type of changes as well. Just rename title to include `[MLIR]`.

PR #3820, opened 2019-10-30
[Spec][FusedOp]Adjust DepthToSpace fused op to specification

mbencer:
According to new specification for `DepthToSpace `:
- add support `mode` attribute as required string. Values: `blocks_first` or `depth_first`
- `block_size` is optional

According ONNX standard (https://github.com/onnx/onnx/blob/master/docs/Operators.md#DepthToSpace):
- `mode` is optional string. Values: `DCR` or `CRD`
- `blocksize` is required


src/ngraph/op/fused/depth_to_space.cpp
```diff
@@ -83,5 +107,18 @@ shared_ptr<Node> op::DepthToSpace::copy_with_new_args(const NodeVector& new_args
     {
         throw ngraph_error("Incorrect number of new arguments");
     }
-    return make_shared<DepthToSpace>(new_args.at(0), m_blocksize);
+    return make_shared<DepthToSpace>(new_args.at(0), m_mode, m_blocksize);
+}
+
+op::DepthToSpace::DepthToSpaceMode
+    op::DepthToSpace::depth_to_space_mode_from_string(const std::string& sort) const
```
tsocha: Non blocking:
Name of the operator is already in a class name.
```suggestion
    op::DepthToSpace::mode_from_string(const std::string& sort) const
```

src/ngraph/op/fused/depth_to_space.hpp
```diff
@@ -35,24 +35,42 @@ namespace ngraph
...
+            DepthToSpaceMode get_mode() const { return m_mode; }
             virtual NodeVector decompose_op() const override;
 
             virtual std::shared_ptr<Node>
                 copy_with_new_args(const NodeVector& new_args) const override;
 
         protected:
             std::size_t m_blocksize;
+            DepthToSpaceMode m_mode;
+            DepthToSpaceMode depth_to_space_mode_from_string(const std::string& sort) const;
```
tsocha:
```suggestion
            DepthToSpaceMode mode_from_string(const std::string& sort) const;
```

python/ngraph/ops.py
```diff
@@ -1078,12 +1078,13 @@ def depth_to_space(node, block_size, name=None):  # type: (Node, int, str) -> No
     [N, C * :code:`block_size` * :code:`block_size`, H / :code:`block_size`, W / :code:`block_size`]
 
     :param node: The node with input tensor data.
+    :param mode: Specifies how the input depth dimension is split to block coordinates
```
postrational: Please provide acceptable values in doc string.

diyessi:
Fails python tests.

PR #3823, opened 2019-10-30
[SPEC] Rename element_type->destination_type attribute in Convert op

tsocha:


tsocha:
> @diyessi 
> This either needs to be a V1 or you need to add the change to `changes.md` since it will affect all the transformers.

I just added a new setter and getter(the old ones are still avaliable) so transformers should not be affected.

PR #3827, opened 2019-10-31
LRU cache for dynamic shapes

nbpatel:


src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -117,130 +117,158 @@ bool runtime::dynamic::DynamicExecutable::call(
...
 
-    NGRAPH_CHECK(m_wrapped_function->get_parameters().size() == inputs.size());
+    auto lru = std::make_shared<runtime::LRUCache>(4);
+    std::vector<size_t> merged_input_shapes;
+    std::ostringstream key;
 
-    std::vector<std::shared_ptr<runtime::Tensor>> wrapped_inputs;
-    std::vector<element::Type> arg_element_types;
-    std::vector<PartialShape> arg_shapes;
+    for (auto& input : inputs)
```
diyessi: With variable rank, this will do the wrong thing: [2], [3, 4] and [2, 3], [4] will get the same key.
You can directly use a vector of shapes as a key and skip the conversion to a string.

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+void runtime::LRUCache::add_entry(Shape shape, shared_ptr<runtime::Executable> exec)
+{
+    ostringstream key;
+    // check if the list is empty
+    if (m_list.size() == m_size)
+    {
+        ostringstream key = convert_shape_to_string(m_list.back());
+        m_list.pop_back();
+        m_map.erase(key.str());
```
sayantan-nervana: There could be use cases where 2 threads are running `call` together. In which case do we want to make this cache thread safe and protect some data members?

nbpatel: yes, I have plan to add locks and make it thread safe

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -117,130 +117,160 @@ bool runtime::dynamic::DynamicExecutable::call(
     // (1) all shapes;
     // (2) all values of shape-relevant input tensors.
 
-    NGRAPH_CHECK(m_wrapped_function->get_parameters().size() == inputs.size());
+    auto lru = std::make_shared<runtime::LRUCache>(4);
```
sayantan-nervana: So the `LRUCache` local object `lru` is created everytime we do `dyn_exec->call`

But we just create the `lru` object to get access to the static (hence global like) objects `m_list` and `m_map`

also the depth of the cache is passed in everytime we create the local `lru` object.
which means this situation is possible:

call(shape=[2]) {
}

nbpatel:
Ping for review

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include "ngraph/runtime/cache.hpp"
+
+using namespace ngraph;
+using namespace std;
+
+// Constructor
+runtime::LRUCache::LRUCache()
+{
+    char* cache_size = getenv("NGRAPH_CACHE_SIZE");
```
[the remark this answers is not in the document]
sayantan-nervana: We may want tests for the cache

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return false;
+}
+
+shared_ptr<runtime::Executable> runtime::LRUCache::get_cached_entry(const vector<int>& shape)
+{
+    std::lock_guard<std::mutex> guard(m_mutex);
+    // find the entry and return the function
+    ostringstream key;
+    key = convert_shape_to_string(shape);
+    auto it = m_map.find(key.str());
```
sayantan-nervana: Should we handle cases where the entry is not found?

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -117,130 +117,210 @@ bool runtime::dynamic::DynamicExecutable::call(
...
-                arg_value_base_pointers[i] = nullptr;
+                merged_input_shapes.emplace_back(input->get_shape()[i]);
             }
+        }
+        // -1 is the separator.
+        // So if shape of Input 1 = {2, 2, 3, 3} & Input 2 = {4, 5}
+        // the key would be 2, 2, 3, 3, -1, 4, 5, -1
+        merged_input_shapes.emplace_back(-1);
+        loop_count++;
+    }
```
sayantan-nervana: Minor:
This computation of the key could be moved inside the cache class. then the calling code does not have to do this. The class could just accept a `vector<vector<int>>` or `vector<ng::Shape>` or something else and compute its key as it sees fit. The user does not have to know about how the class implements the key calculation

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -117,130 +117,210 @@ bool runtime::dynamic::DynamicExecutable::call(
...
+        pass::Manager passes;
+        passes.register_pass<pass::ConstantFolding>();
+        passes.register_pass<pass::DynElimination>();
+        passes.register_pass<pass::Opset0Downgrade>(); // Converts dynamic v1 variants to v0 ops
+        passes.set_per_pass_validation(false);
+
+        // FIXME(amprocte): Vile, temporary hack: we need to do repeated rounds of
+        // ConstantFolding/DynElimination until everything that DynElimination is supposed to
+        // eliminate has actually been eliminated. We could do this by monitoring the return values
+        // of
```
sayantan-nervana: extra `of`

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+        {
+        public:
+            using GraphCache = unordered_map<string, shared_ptr<Executable>>;
+            using ClonedFunctionMap = unordered_map<string, shared_ptr<Function>>;
+
+            LRUCache();
+
+            virtual ~LRUCache();
+
+            void add_entry(const vector<int>& shape,
```
rkimballn1: Why are shapes vector<int> instead of Shape?

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <list>
+#include <mutex>
+#include <sstream>
+#include <string>
+#include <unordered_map>
+#include "ngraph/function.hpp"
+#include "ngraph/runtime/executable.hpp"
+#include "ngraph/shape.hpp"
+
+using namespace std;
```
rkimballn1: No using namespace in header.

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+            LRUCache();
+
+            virtual ~LRUCache();
+
+            void add_entry(const vector<int>& shape,
+                           shared_ptr<Executable> exec,
+                           shared_ptr<Function> func);
+            bool is_cached(const vector<int>& shape);
+            shared_ptr<Executable> get_cached_entry(const vector<int>& shape);
+            void convert_shape_to_string(const vector<int>& shape, ostringstream& key);
```
rkimballn1: Could be better names `to_string` to match the existing string functions
```suggestion
            std::string to_string(const vector<int>& shape);
```

rkimballn1: If you want to keep the function name use this instead
```suggestion
            void convert_shape_to_string(const vector<int>& shape, std::ostream& key);
```

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+            shared_ptr<Function> get_cloned_function(const vector<int>& shape);
+
+        private:
+            int m_size; // cache size
+            GraphCache m_map;
```
rkimballn1: These names should be more descriptive.

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -120,130 +120,207 @@ bool runtime::dynamic::DynamicExecutable::call(
...
-    // seem to work so here we are. Probably a better fix is to somehow combine the matchers in CF
-    // and DE into one pass.
-    size_t num_dyn_nodes_last_pass = std::numeric_limits<size_t>::max();
-
-    while (num_dyn_nodes_last_pass != 0)
+    else
     {
-        passes.run_passes(clone);
-        auto num_dyn_nodes_this_pass = count_dyn_nodes(clone);
+        NGRAPH_INFO << "Cache Miss";
```
rkimballn1: remove

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -120,130 +120,207 @@ bool runtime::dynamic::DynamicExecutable::call(
...
+                if (m_wrapped_function->get_parameters()[i]->is_relevant_to_shapes())
+                {
+                    // TODO(amprocte): Move has_storage() to runtime::Tensor?
+                    if (auto dynamic_tensor =
+                            std::dynamic_pointer_cast<runtime::dynamic::DynamicTensor>(input))
+                    {
+                        NGRAPH_CHECK(dynamic_tensor->has_storage());
+                    }
+
+                    arg_buffers.emplace_back(input->get_size_in_bytes(), /*alignment=*/64);
```
rkimballn1: Make a veriable called alignment and set it to 64, then pass alignment into this function. It will read much better that way. Or better yet, remove the 64. Just use the default alignment.

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,134 @@
...
+using namespace ngraph;
+using namespace std;
+
+// Constructor
+runtime::LRUCache::LRUCache()
+{
+    char* cache_size = getenv("NGRAPH_CACHE_SIZE");
+    if (cache_size == nullptr)
+    {
+        m_size = 1024; // TODO(nbpatel): Figure out a default size for the cache
```
baojun-nervana: We use a much smaller cache size of 5 in brige. https://github.com/NervanaSystems/ngraph-paddle/blob/develop/paddle/fluid/operators/ngraph/ngraph_engine.cc#L514

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include "ngraph/runtime/cache.hpp"
+
+using namespace ngraph;
+using namespace std;
+
+// Constructor
+runtime::LRUCache::LRUCache()
+{
+    char* cache_size = getenv("NGRAPH_CACHE_SIZE");
```
baojun-nervana: Can we have an API to enable bridge to set the cache size through bridge? The FW can be restrictive in env var use.

src/ngraph/runtime/cache.hpp
nbpatel: because shapes are of type size_t and my shapes have -1 in them as a separator between different args.
so if the input shapes are {2, 4} & {5,6,7} it would be stored as a key like {2, 4, -1, 5, 6, 7, -1}

src/ngraph/runtime/dynamic/dynamic_backend.cpp
nbpatel: haven't touched this part of the code..not sure why its showing up in the diff

src/ngraph/runtime/cache.cpp
diyessi: @gaurides is working on this.

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,130 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
rkimballn1:
```suggestion
// Copyright 2017-2020 Intel Corporation
```

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,61 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
rkimballn1:
```suggestion
// Copyright 2017-2020 Intel Corporation
```

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,130 @@
...
+    ostringstream key;
+    convert_shape_to_string(shape, key);
+    // find the entry and return the function
+    auto it = m_clone_function_map.find(key.str());
+    if (it == m_clone_function_map.end())
+    {
+        throw ngraph_error("Cloned function not found");
+    }
+    return it->second;
+}
```
rkimballn1:
```suggestion
}

```

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -120,130 +120,207 @@ bool runtime::dynamic::DynamicExecutable::call(
...
-        std::vector<void*> arg_value_base_pointers(inputs.size());
-
-        size_t i = 0;
-
-        for (auto& input : inputs)
+        if (m_wrapped_function->get_parameters()[loop_count]->is_relevant_to_shapes())
         {
-            if (m_wrapped_function->get_parameters()[i]->is_relevant_to_shapes())
+            // Caching on values of Shape relevant inputs
+            void* data_ptr = malloc(input->get_size_in_bytes());
```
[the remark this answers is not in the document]
rkimballn1: Really, just use a vector<int64_t> here instead of malloc. malloc is old-school and dangerous. You could also use an AlignedBuffer which is more efficient that a buffer for this use and is RAII so you don't need to worry about deleting it.

PR #3832, opened 2019-10-31
[MLIR] In-place memory optimization for elt-wise and concat ops.

nmostafa:
This PR implements a memory re-use optimization for elt-wise ops and Concat. The output of the pass is annotated ops with buffer id and offset. At this stage, the optimization is OFF by default. 
The optimization scans the code backwards and tracks liveness and "cannot-alias" relationship. If a tensor source is re-used for write (e.g. an OpAdd writes result to one of its dead sources), then that source is marked as cannot-alias all live tensors at that point. 

Next steps:

- [x] Implement affine lowering support for Concat 
- [x] Add LIT tests with FileCheck on affine lowered code
Separate PR
-  Use Op Interfaces to indicate inplace and destructive traits. 
- Support Slice and Reshape
- Improve concat re-use if only one of the srcs is pre-assigned a buffer (see comment in code)


src/contrib/mlir/backend/pass/memory_optimization.cpp
```diff
@@ -43,114 +56,567 @@ namespace
...
+#endif
+    private:
+        using Row = SmallVector<int8_t, 10>;
+        std::unordered_map<Value*, unsigned> m_valueToIdx;
+        SmallVector<Row, 10> m_reachability;
+        unsigned m_maxIdx;
+        bool m_needsTransitiveClosure;
+    };
+
+    class Liveness
```
dcaballe: Is this an analysis? -> `LivenessAnalysis`?

src/contrib/mlir/core/ngraph_dialect/ops.cpp
```diff
@@ -328,6 +328,28 @@ mlir::IntegerAttr getBufferId(mlir::Operation* op)
...
+{
+    op->setAttr(getBufferOffsetAttrName(), attr);
+}
+
+mlir::IntegerAttr setBufferOffset(mlir::Operation* op, unsigned val)
+{
+    auto attr = mlir::IntegerAttr::get(IntegerType::get(32, op->getContext()), val);
+    setBufferOffset(op, attr);
+    return attr;
+}
```
dcaballe: Have you considered storing this information in a separate analysis class? I think that adding analysis information to the main IR doesn't scale and it's more difficult to update/invalidate/retrieve the information since we have to traverse the IR for that. An analysis that contains this information in a map<Operation *, BufferInfo> could work pretty well and it would make update/invalidate/retrieve very simple without modifying the main IR. Some examples of this in LLVM, like SCEV.

test/mlir/ngraph_dialect/memory_opt.mlir
```diff
@@ -0,0 +1,205 @@
+// RUN: ngraph-opt %s --split-input-file --ng-memory-opt --ngraph-memory-opt-concat --ngraph-memory-opt-eltwise | FileCheck %s
```
dcaballe: Would it make sense to have concat and eltwise optimizations in separate passes. They seem two independent passes that we could be applied, customized (flags) and test independently.

nmostafa: Not really. Why make separate passes over the same IR ?

src/contrib/mlir/core/ngraph_dialect/ops.cpp
nmostafa: I see your point. But it would be difficult to invalidate the map due to IR modifications.  Generally, attaching info to the IR is better IMO, because if a value is modified/deleted we can detect the invalid state, and bail out.  
For example, consider
```
X = ...
V = Concat X, Y, Z

The side map would say:
Concat --> BufferInfo{ Operand0-> buffer0, offset: 16, Operand1--> ...}
```
If the value of X is modified or replaced, for example the operation is decomposed, the information in the map will be invalid, and we wouldn't be able to tell. Yielding to incorrect in-place concat. 

OTH if the info is attached to X's def directly, the attribute will just be removed, and when we try to do the concat in-place, we will detect the missing data, and just bail out.

src/contrib/mlir/backend/pass/memory_optimization.cpp
nmostafa: Will rename.

src/contrib/mlir/core/ngraph_dialect/ops.cpp
dcaballe: The example that you describe is how analysis works in LLVM in general. When a pass modifies the IR, it invalidates any prior analysis unless the pass knows how to preserve/update it. You can see how vectorizer updates loop info, dominator tree and scalar evolution (all storing analysis information separately) after IR modification here: https://llvm.org/doxygen/LoopVectorize_8cpp_source.html#l04338. Scalar evolution, for example, uses a map to store SCEV for each Instruction (http://llvm.org/doxygen/ScalarEvolution_8h_source.html#l01181), which is pretty similar to what we would need here. Among other things, storing analysis information to the IR is problematic because we would have to traverse the whole IR for invalidation, cloning the IR would clone all the analysis information for that IR (now invalid for the clone?), and analysis data structures may become corrupted after a change in the IR if they point to info stored in the IR. It's kind of separation of concerns principle.

test/mlir/ngraph_dialect/memory_opt.mlir
dcaballe: Basically they are different transformations. It's better to keep things simple and decoupled since transformations usually get complicated over time. Similar examples in MLIR: tiling, fusion, vectorization... all of them separate passes.

nmostafa: Nothing in this pass is a transformation. The analysis required for both ops is the same. The only difference is that src operands for concat get a non-zero offset.

src/contrib/mlir/backend/analysis/memory_analysis.cpp
```diff
@@ -0,0 +1,628 @@
...
+        for (auto opnd : op->getOperands())
+        {
+            if (!m_liveness.isLive(opnd) && !isInputOrOutputValue(opnd))
+            {
+                int uses = 0;
+                for (auto& i : opnd->getUses())
+                {
+                    uses++;
+                }
+                if (useCount == -1 || uses < useCount)
```
ayzhuang: useCount is never updated.

PR #3841, opened 2019-11-01
Fixed compilation warnings to build nGraph unit tests in OV CI

GlebKazantaev:


AnastasiaKazantaeva:
Could you please look at this PR? We want to add backend tests to our CI, so this commit fix some compilation warnings. 

Ans could you give me permissions for push and to Jenkins? @diyessi

GlebKazantaev:
@diyessi @silee2 please review

test/type_prop_layers.cpp
```diff
@@ -130,7 +130,7 @@ TEST(type_prop_layers, proposal)
     attrs.base_size = 1;
     attrs.pre_nms_topn = 20;
     attrs.post_nms_topn = 200;
-    const uint64_t batch_size = 7;
+    const size_t batch_size = 7;
```
diyessi: Can you keep this as `uint64_t` and change other things to `uint64_t` to make it work? Because `size_t` is defined in different ways in different environments, we'd prefer to make changes that help reduce the places where we are using `size_t` when it's not too inconvenient.

AnastasiaKazantaeva: Ok, to make this test work I have to change  `post_nms_topn` Proposal attribute from size_t to uint64_t .I think this is not ok to change operation attribute type due to test implementation specific.  And in this test case size_t for batch size looks fine. So let's fix test instead of changing attribute type.

AnastasiaKazantaeva:
Could you give me permissions for push and to Jenkins? @diyessi

PR #3845, opened 2019-11-04
specialize_function with shared constants

GlebKazantaev:


src/ngraph/specialize_function.hpp
```diff
@@ -198,5 +200,6 @@ namespace ngraph
                             const std::vector<element::Type>& parameter_element_types,
                             const std::vector<PartialShape>& parameter_shapes,
                             const std::vector<void*>& parameter_values,
-                            bool constant_folding);
+                            bool constant_folding,
+                            bool share_constants);
```
postrational: Signature of this function is getting unwieldy. Soon we'll need to speak in a binary code:
`specialize_function(..., 0, 1, 0, 1, 0);`

Should we consider some solution to this?

GlebKazantaev: I hope this is the last key for specialize_function :) 
I found a comment "TODO(amprocte): convert this to a pass." and maybe this would be a future solution.

PR #3848, opened 2019-11-05
Fix float16 denorm conversion, add rounding

diyessi:


src/ngraph/type/float16.cpp
```diff
@@ -49,12 +49,17 @@ float16::float16(float value)
...
     };
     fv = value;
+    uint32_t hidden_one = 0x00800000;
     uint32_t sign = iv & 0x80000000;
     uint32_t biased_exp = (iv & 0x7F800000) >> 23;
     uint32_t raw_frac = (iv & 0x007FFFFF);
     int32_t exp = biased_exp - 127;
     int32_t min_exp = -14 - frac_size;
-    if (biased_exp == 0 || exp < min_exp)
+    if (biased_exp == 0 && raw_frac == 0)
```
ilyachur: Why do we need this empty case?
The last `else` could be changed to `!(biased_exp == 0 && raw_frac == 0)` and it won't change the current logic, but you will avoid empty code block.

ilyachur:
I left a comment. But in general LGTM. 

I set approve status, in order to don't wait for one more day to merge.

src/ngraph/type/float16.cpp
diyessi: Will change to something similar.

PR #3852, opened 2019-11-05
[SPEC] Add new v1::FloorMod operator

tsocha:


src/ngraph/op/floor_mod.hpp
```diff
@@ -0,0 +1,60 @@
...
+                /// `[d0, ...]`
+                /// \param arg1 Output that produces the second input tensor.<br>
+                /// `[d0, ...]`
+                /// \param auto_broadcast Auto broadcast specification
+                ///
+                /// Output `[d0, ...]`
+                ///
+                FloorMod(const Output<Node>& arg0,
+                         const Output<Node>& arg1,
+                         const AutoBroadcastSpec& auto_broadcast = AutoBroadcastSpec());
```
GlebKazantaev: Should be set to NUMPY by default.

tsocha: Done

PR #3855, opened 2019-11-06
[SPEC] Softplus as a fused op

tomdol:


src/ngraph/op/fused/softplus.hpp
```diff
@@ -0,0 +1,46 @@
...
+namespace ngraph
+{
+    namespace op
+    {
+        /// \brief A Softplus (SmoothReLU) operation ln(1 + e^x)
+        class Softplus : public ngraph::op::util::FusedOp
+        {
+        public:
+            NGRAPH_API
+            static constexpr NodeTypeInfo type_info{"Softplus", 0};
```
tsocha: I just wondering if the new fused ops should be `v1` or `v0`.

src/ngraph/frontend/onnx_import/op/softplus.cpp
```diff
@@ -37,34 +29,11 @@ namespace ngraph
...
+                    const auto fused_softplus = std::make_shared<ngraph::op::Softplus>(data);
 
-                    //
-                    // This equation represents:
-                    //     x + log(exp(-x) + 1) - for x > 0; to manage exponent overflow,
-                    //     log(exp(x) + 1)      - elsewhere.
-                    //
-                    return {std::make_shared<ngraph::op::Select>(
-                        condition_node, positive_val_node, negative_val_node)};
+                    return fused_softplus->decompose_op();
```
postrational: Decomposing a fused op in the ONNX importer would defeat the purpose of creating the fused op.
ONNX importer must output fused op without decomposing them.

src/ngraph/op/fused/softplus.cpp
```diff
@@ -0,0 +1,67 @@
...
+NodeVector op::Softplus::decompose_op() const
+{
+    const auto data = input_value(0);
+
+    const auto zero_node = builder::make_constant(data.get_element_type(), data.get_shape(), 0.f);
+    const auto one_node = builder::make_constant(data.get_element_type(), data.get_shape(), 1.f);
+
+    const auto positive_val_node =
+        data + std::make_shared<op::Log>(
+                   std::make_shared<op::Exp>(std::make_shared<op::Negative>(data)) + one_node);
```
GlebKazantaev: As I know `+` is overloaded for V0 add op. Let's use V1 Add here.

tomdol: We shouldn't use `v1` here because the decomposition of those ops happens after the upgrade/downgrade passes. If I used `v1` here, it would not be possible to execute the resulting `Function` by our current backends.

postrational:
This operation was removed from `v1`.

PR #3857, opened 2019-11-06
[SPEC] HardSigmoid adjustments

tomdol:


tomdol:
I still thing we should change the spec so that `alpha` and `beta` inputs are defined as scalars. I will add more commits if/when the decision is made.

src/ngraph/op/fused/hard_sigmoid.cpp
```diff
@@ -29,41 +29,70 @@ using namespace ngraph;
...
 
-    std::shared_ptr<ngraph::Node> one_node = ngraph::op::Constant::create<float>(
-        data.get_element_type(), data_shape, std::vector<float>(elem_count, 1.0));
+    const auto alpha_node = input_value(1).get_node_shared_ptr();
+    const auto beta_node = input_value(2).get_node_shared_ptr();
 
-    std::shared_ptr<ngraph::Node> zero_node = ngraph::op::Constant::create<float>(
-        data.get_element_type(), data_shape, std::vector<float>(elem_count, 0.0));
+    std::shared_ptr<Node> alpha_x_plus_beta =
+        std::make_shared<op::Multiply>(alpha_node, data, AutoBroadcastType::NUMPY);
```
GlebKazantaev: Let's use V1 Add and V1 Multiply.

src/ngraph/op/fused/hard_sigmoid.cpp
```diff
@@ -29,41 +29,70 @@ using namespace ngraph;
...
+
+    alpha_x_plus_beta =
+        std::make_shared<op::Add>(alpha_x_plus_beta, beta_node, AutoBroadcastType::NUMPY);
 
     return {std::make_shared<op::Minimum>(
-        std::make_shared<op::Maximum>(alpha_node * data + beta_node, zero_node), one_node)};
+        std::make_shared<op::Maximum>(alpha_x_plus_beta, zero_node), one_node)};
```
GlebKazantaev: Let's use V1 Maximum

src/ngraph/op/fused/hard_sigmoid.cpp
```diff
@@ -29,41 +29,70 @@ using namespace ngraph;
...
 {
     constructor_validate_and_infer_types();
 }
 
+void op::HardSigmoid::pre_validate_and_infer_types()
+{
+    const auto& alpha_pshape = get_input_partial_shape(1);
+    const auto& beta_pshape = get_input_partial_shape(2);
+    NODE_VALIDATION_CHECK(this,
+                          alpha_pshape.is_static() && beta_pshape.is_static(),
```
GlebKazantaev: Since FusedOp supports dynamic inputs it's not correct to check that shapes are static. So please handle dynamic shapes also.
BTW if you take a look into validate_and_infer_types method for FusedOp it calls is_dynamic() function and in case of not static shapes it will skip decomposition.

tomdol: I thought about it for a moment and since the specs says that `alpha` and `beta` should be scalars - do we really need to handle dynamic shapes for both of those inputs? I've added this restriction intentionally to reflect the requirements I was given.

GlebKazantaev: This scalars can be calculated after constant folding pass for example. So this inputs can be dynamic at some time.

tomdol: Ok, got it. I've relaxed the requirements for alpha and beta.

PR #3858, opened 2019-11-06
enable env pass filtering in core

ashokei:


src/ngraph/pass/pass_config.cpp
```diff
@@ -87,6 +87,11 @@ void pass::PassConfig::set_pass_enable(const string& name, bool enable)
     m_pass_enables[name] = enable;
 }
 
+bool pass::PassConfig::get_pass(const string& name) const
```
nmostafa: IIUC, any pass is enabled by default: if it is not in `m_pass_enables` then it is enabled. Then you can fix `get_pass_enable` to return true if the pass is not in the map, and remove `get_pass`.

ashokei: m_pass_enables has passes that maybe disabled by env var.
`m_pass_enables.emplace(split_str[0], parse_string<bool>(split_str[1]))`

ashokei: i guess each backend chooses which passes they enable hardcoded, users cant really pick to enable non-enabled backend passes, will make the change.

ashokei: @nmostafa pass_config is expected to be independent of which passes backend enabled by default. In other words, pass_config role is to report what users configured on the command line env (irrespective if the pass is enabled by default on backend). thoughts?

nmostafa: I see. Perhaps changing it to `is_pass_disabled()` would be better, and the condition above should be moved to that function.

ashokei:
@rkimballn1 currently each backend/transformer decide which passes they want to use/enable hardcoded. CPU backend has a utility using env var to filter-out/disable user given passes at runtime. This PR just generalizes it.

src/ngraph/pass/manager.hpp
```diff
@@ -44,6 +44,11 @@ class ngraph::pass::Manager
     template <typename T, class... Args>
     std::shared_ptr<T> register_pass(Args&&... args)
     {
+        auto pass = get_pass_name(typeid(T).name());
```
rkimballn1: since type_info.name() produces an undefined output it can't be used to get the pass name. You need some other way to do this.

ashokei: `Why put all of this code in the pass manager when all you need to do is at the point where you populate the pass::Manager just don't register passes`

@rkimballn1 this check will have to be done by each backend. Currently CPU backend does this. This PR just moves/generalizes that logic into core, so all backends can take advantage of this env var filter for passes. `typeid(T).name()` is previously used in code in this file, just reusing here.

rkimballn1: No, typeid was never used for `functional` code where the value returned is used in a conditional. You can't do that. It is perfectly valid for typeid(T).name() to return a null string.
I am very against this PR. The code should be done in the backend. Don't generalize when it makes things too complex.

ashokei: @rkimballn1 not a crucial feature, just a utility to make it easier if user wants to bypass some passes (even though backend enabled them).

Backends can explicitly enable this as needed (as CPU does), closing this PR

PR #3859, opened 2019-11-06
Atan2 op

diyessi:


src/ngraph/runtime/reference/atan2.hpp
```diff
@@ -0,0 +1,38 @@
...
+    namespace runtime
+    {
+        namespace reference
+        {
+            template <typename X, typename Y, typename Z>
+            void atan2(const X* px, const Y* py, Z* pout, size_t count)
+            {
+                for (size_t i = 0; i < count; i++)
+                {
+                    *pout++ = static_cast<Z>(std::atan2(*px++, *py++));
```
sayantan-nervana: The constructor has `(y,x)`
std's atan2 is also `atan2( float y, float x )`
but this function has `const X* px, const Y* py`

Maybe we can reverse order?

diyessi: Missed one.

PR #3860, opened 2019-11-07
[MLIR] Fused Ops dialect declaration

nmostafa:
This change adds all ops under ngraph/ops/fused to nGraph dialect. All fused ops implement a fused op interface with a decompose method that is TBD. 

Also fixes naming convention for Op v0 ops.


src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,895 @@ def NGSqueezeOp :
...
+  }];
+
+  let parser = [{ NGRAPH_CHECK(false, "No parser support"); return mlir::failure(); }];
+  let verifier = [{ return mlir::success(); /* TBD */ }];
+}
+
+// Squared Difference Op
+def NGSquaredDiffOp : 
+    NG_Binary_Op<"sqrdDiff", [DeclareOpInterfaceMethods<FusedOp>]>
+{
```
gaurides: 1. So, we also need to define the arguments here, right? comparing with https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/fused/squared_difference.hpp#L42
2. Why is this NG_Binary_Op vs NG_OneResult_Op? `y[i] = (x1[i] - x2[i])^2` that is the equation, right?

nmostafa: `NG_Binary_Op` will have the results and arguments defined (lhs, rhs). Doesn't matter what the equation is, just a binary operation.

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,895 @@ def NGSqueezeOp :
...
+
+  let extraClassDeclaration = [{
+    void setBlockSize(const Attribute& attr) { this->setAttr("blockSize", attr); }
+  }];
+}
+    
+// ShuffleChannels Op
+def NGShuffleChannelsOp :
+    NG_OneResult_Op<"shuffleChannels", [NoSideEffect, DeclareOpInterfaceMethods<FusedOp>]>,
+    Arguments<(ins NG_TensorType:$data, I64Attr:$axis, I64Attr:$groups)>
```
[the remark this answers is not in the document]
nmostafa: No unsigned attributes unfortunately, not worth adding our own tblgen class.

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,895 @@ def NGSqueezeOp :
...
+    void setAxis(const Attribute& attr)                 { this->setAttr("axis", attr);            }
+    void setNumSplits(const ArrayAttr& arrayAttr)  { this->setAttr("numSplits", arrayAttr);  }
+  }];
+ 
+}
+
+// SpaceToDepth Op
+def NGSpaceToDepthOp :
+    NG_OneResult_Op<"spaceToDepth", [NoSideEffect, DeclareOpInterfaceMethods<FusedOp>]>,
+    Arguments<(ins NG_TensorType:$data, I64Attr:$blockSize)>
```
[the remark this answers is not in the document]
nmostafa: Good catch. Fixed. Default values are done using `DefaultValuedAttribute`, where when the attribute is missing, we return a default expression (in this case "1")

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,899 @@ def NGSqueezeOp :
...
+def NGLayerNormOp : 
+    NG_Op<"layernorm", [NoSideEffect, DeclareOpInterfaceMethods<FusedOp>]>,
+    Results<(outs NG_TensorType:$res, NG_TensorType:$mean, NG_TensorType:$var)>,
+    Arguments<(ins NG_TensorType:$data, NG_TensorType:$scale, NG_TensorType:$bias,
+               DefaultValuedAttr<BoolAttr, "true">:$keepStats,
+               DefaultValuedAttr<I64Attr, "1">    :$beginNormAxis,
+               DefaultValuedAttr<F32Attr, "1e-5"> :$epsilon)>
+{
+  let summary = "LayerNorm Op";
+  let description = "Constructs an LayerNorm operation.";
```
gaurides: typo `an` -> `a`

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,899 @@ def NGSqueezeOp :
...
+       tblgen_state.addOperands(data);
+       tblgen_state.addAttribute("keepStats", keepStats);
+       tblgen_state.addAttribute("beginNormAxis", beginNormAxis);
+       tblgen_state.addAttribute("epsilon", epsilon);
+       tblgen_state.addTypes(res);
+    }]>
+  ];
+
+  let extraClassDeclaration = [{
+    void setKeepstats(const Attribute& attr)      { this->setAttr("keepStats", attr);    }
```
gaurides: just checking: would you need the getter functions too? ngraph Fusedop defines getters.

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,899 @@ def NGSqueezeOp :
...
+    Arguments<(ins NG_TensorType:$data, NG_TensorType:$delta, 
+               NG_TensorType:$mean, NG_TensorType:$variance, NG_TensorType:$scale,
+               DefaultValuedAttr<I64Attr, "1">:$beginNormAxis,
+               DefaultValuedAttr<F32Attr, "1e-5">:$epsilon)>
+{
+  let summary = "LayerNorm Op";
+  let description = "Constructs an LayerNorm operation.";
+    
+  let builders = [
+    OpBuilder<
```
gaurides: did you miss the OpBuilder with 5 args including scale?

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,899 @@ def NGSqueezeOp :
...
+    "Builder *builder, OperationState &tblgen_state, ArrayRef<Type> res,"
+    "Value *data, Attribute keepStats, Attribute beginNormAxis, Attribute epsilon", [{
+       tblgen_state.addOperands(data);
+       tblgen_state.addAttribute("keepStats", keepStats);
+       tblgen_state.addAttribute("beginNormAxis", beginNormAxis);
+       tblgen_state.addAttribute("epsilon", epsilon);
+       tblgen_state.addTypes(res);
+    }]>
+  ];
```
gaurides: in LayerNormBackprop, you defined multiple OpBuilders for each definition. So, we need another definition here too for the definition with only data as input?

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,899 @@ def NGSqueezeOp :
...
+                  
+// GroupConv Op
+def NGGroupConvOp : 
+    NG_OneResult_Op<"groupConv", [NoSideEffect, DeclareOpInterfaceMethods<FusedOp>]>,
+    Arguments<(ins NG_TensorType:$images, NG_TensorType:$filters, 
+               I64ArrayAttr:$strides,
+               I64ArrayAttr:$padBelow,
+               I64ArrayAttr:$padAbove,
+               I64Attr:$groups,
+               DefaultValuedAttr<PadTypeEnumAttr,
```
gaurides: window_dilation_Strides and data_dilation_strides are missing. 
Does mlir not support dilation?

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,899 @@ def NGSqueezeOp :
...
+               DefaultValuedAttr<PadTypeEnumAttr, 
+                                "static_cast<int64_t>(MLIRPadType::EXPLICIT)">:$padType)>
+{
+  let summary = "Group Convolution Op";
+  let description = [{
+    Group Convolution
+  }];
+
+  let builders = [
+    OpBuilder<
```
gaurides: this builder has all args and attrs, so it will be auto-generated, right? not really needed here

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,899 @@ def NGSqueezeOp :
...
+    void setLevels(const Attribute& attr) { this->setAttr("levels", attr); }
+  }];
+}
+
+// DepthToSpace Op
+def NGDepthToSpaceOp : 
+    NG_OneResult_Op<"depthToSpace", [NoSideEffect, DeclareOpInterfaceMethods<FusedOp>]>,
+    Arguments<(ins NG_TensorType:$data, 
+               DefaultValuedAttr<I64Attr, "1">:$blockSize, 
+               DepthSpaceModeEnumAttr:$mode)>
```
gaurides: There's a definition where the `mode` is a string. will this work in that case too?

src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -32,14 +33,899 @@ def NGSqueezeOp :
...
+    set of floating point values.
+    
+    Implementation This class creates a node which performs the following
+    operation:
+      round((data - input_low) / (input_high - input_low) * (levels-1)) /
+      (levels-1) * (output_high - output_low) + output_low
+  }];
+
+  let extraClassDeclaration = [{
+    void setLevels(const Attribute& attr) { this->setAttr("levels", attr); }
```
gaurides: does autobroadcast need a setter?

PR #3864, opened 2019-11-07
Fix an error in ONNXRT build

tomdol:
ONNXRT uses different settings (warnings as errors) and this is what it's complaining about:

https://aipg-rancher.intel.com/jenkins/algo/job/CLX-onnxrt-performance-master/288/console

```
In file included from /onnxruntime/build/Release/external/ngraph/include/ngraph/ngraph.hpp:100:0,
                 from /onnxruntime/onnxruntime/core/providers/ngraph/ngraph_custom_op.h:12,
                 from /onnxruntime/onnxruntime/core/providers/ngraph/ngraph_execution_provider.cc:12:
/onnxruntime/build/Release/external/ngraph/include/ngraph/op/binary_convolution.hpp:102:45: error: type qualifiers ignored on function return type [-Werror=ignored-qualifiers]
                 const float get_pad_value() const { return m_pad_value; }
```


aslepko:
TESTNOW

PR #3868, opened 2019-11-08
[SPEC] Add new v1::VariadicSplit operator

tsocha:


src/ngraph/op/variadic_split.hpp
```diff
@@ -0,0 +1,52 @@
...
+        {
+        public:
+            NGRAPH_API
+            static constexpr NodeTypeInfo type_info{"VariadicSplit", 0};
+            const NodeTypeInfo& get_type_info() const override { return type_info; }
+            /// \brief Constructs a variadic split operation.
+            VariadicSplit() = default;
+            /// \brief Constructs a variadic split operation.
+            ///
+            /// \param data           The tensor to be splited.
```
diyessi: `the tensor to be split`

silee2:
I don't see any reference to `v1` although the PR title states `v1`

src/ngraph/op/variadic_split.hpp
tsocha: Done

src/ngraph/op/variadic_split.hpp
```diff
@@ -0,0 +1,57 @@
...
+    {
+        namespace v1
+        {
+            /// \brief VariadicSplit operation splits an input tensor into pieces along some axis.
+            /// The pieces may have variadic lengths depending on "split_lengths" attribute.
+            class VariadicSplit : public Op
+            {
+            public:
+                NGRAPH_API
+                static constexpr NodeTypeInfo type_info{"VariadicSplit", 0};
```
tomdol: The version should be set to `1` here.

src/ngraph/op/variadic_split.hpp
```diff
@@ -0,0 +1,57 @@
...
+                const NodeTypeInfo& get_type_info() const override { return type_info; }
+                /// \brief Constructs a variadic split operation.
+                VariadicSplit() = default;
+                /// \brief Constructs a variadic split operation.
+                ///
+                /// \param data           The tensor to be split.
+                /// \param axis           An axis along "data" to split.
```
postrational:
```suggestion
                /// \param axis           The index of an axis in "data" along which to perform the split.
```

src/ngraph/op/variadic_split.hpp
```diff
@@ -0,0 +1,57 @@
...
+                /// \param split_lengths  A list containing the sizes of each output tensor along
+                /// the split "axis". Size of "split_lengths" should be equal to the number of
+                /// inputs. The sum of sizes must match data.shape[axis]
```
postrational:
```suggestion
                /// outputs. The sum of split_lengths must match data.shape[axis]
```

src/ngraph/op/variadic_split.hpp
tsocha:
```suggestion
                static constexpr NodeTypeInfo type_info{"VariadicSplit", 1};
```

src/ngraph/op/variadic_split.cpp
```diff
@@ -0,0 +1,36 @@
...
+using namespace ngraph;
+
+constexpr NodeTypeInfo op::v1::VariadicSplit::type_info;
+
+op::v1::VariadicSplit::VariadicSplit(const Output<Node>& data,
+                                     const Output<Node>& axis,
+                                     const Output<Node>& split_lengths)
+    : Op({data, axis, split_lengths})
+{
+    constructor_validate_and_infer_types();
```
GlebKazantaev: I see that validate_and_infer_types function is not implemented here, so how it will propagate output shapes and data type?

tsocha: I set all of the inputs to be relevant in output shape computation and set an output type if rank of `split_lengths` is known.
I can't calculate output shapes if `axis` and s`plit_lengths` are not constant.

GlebKazantaev: Ok, but if they are constants you can do it.

PR #3872, opened 2019-11-11
[MLIR] Affine lowering support for Comparison Ops

pruthviIntel:
Affine lowering support for
    1) GreaterEq
    2) LessEq
    3) Equal
    4) NotEqual


src/contrib/mlir/core/ngraph_dialect/ops.cpp
```diff
@@ -171,7 +171,7 @@ static mlir::LogicalResult verifyCmpOp(T* op)
     NGTensorType resType = r0.cast<NGTensorType>();
 
     // result of same shape as input and has bool type
-    if (!resType.isCompatibleShape(opType0) || !resType.getElementType().isa<NGBoolType>())
+    if (!resType.isCompatibleShape(opType0))
```
nmostafa: We agreed to preserve BoolType. We need the check back here.

src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
```diff
@@ -477,13 +481,13 @@ bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node
     // Otherwise, we would generate code like this:
     //   %0 = icmp %a, %b : i1
     //   store %0, %c[%arg1] : i8  // Type error: trying to store an i1 into an i8.
-    for (auto& output : node->get_outputs())
+    /*for (auto& output : node->get_outputs())
```
nmostafa: Delete?

nmostafa:
Looks good. What we still need:
1. nGraph bool should map to ng.bool 
2. introduce an ng.tensor_cast operation. 

When lowering to affine, ng.bool should be converted to i8. The cast operation will have both RHS/LHS as i8 and can be erased during lowering.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1410,6 +1476,30 @@ namespace
         }
         NGRAPH_UNREACHABLE("Unsupported type");
     }
+
+    ValueHandle createOneConstant(mlir::Type type)
```
nmostafa: Can you unify this and the func above to `template<typename T> createConstant(mlir::Type type, T value)`

src/contrib/mlir/core/ngraph_dialect/type.hpp
```diff
@@ -162,8 +162,8 @@ namespace mlir
             return Base::get(context, kind);
         }
 
-        static bool kindof(unsigned kind) { return kind == NGTypeKind::NG_BOOL_TYPE_ID; }
-        static NGBoolType get(mlir::MLIRContext* ctx) { return get(NG_BOOL_TYPE_ID, ctx); }
+        static bool kindof(unsigned kind) { return kind == NGTypeKind::NG_U8_TYPE_ID; }
```
nmostafa: What do we need this ?

test/core_fusion.cpp
```diff
@@ -830,6 +830,12 @@ TEST(core_fusion, softmax_crossentropy_bprop_with_ignore_mask)
 }
 #endif
 
+// TODO(pthoreho): MLIR currently doesnot support all the op's needed for CrossEntropy+Softmax
+// this results in multiple CompiledKernels and we cannot able to safely check for certain op's
+// from the function created by user.
+// Note: remove this gaurds once we have full support for CE and Softmax through MLIR
+
+#ifndef NGRAPH_MLIR_ENABLE
```
dcaballe: Instead, you can add MLIR_DISABLE_TEST to the test using this function. For example:

```
TEST(core_fusion, crossentropy)
```
-->
```
TEST(core_fusion, MLIR_DISABLE_TEST(crossentropy))
```

test/core_fusion.cpp
```diff
@@ -830,6 +830,12 @@ TEST(core_fusion, softmax_crossentropy_bprop_with_ignore_mask)
 }
 #endif
 
+// TODO(pthoreho): MLIR currently doesnot support all the op's needed for CrossEntropy+Softmax
```
dcaballe: doesnot -> does not

test/mlir/ngraph_dialect/elementwise_binary_ops.mlir
```diff
@@ -11,3 +11,44 @@ func @add_float(%arg0: !ng.tensor<2x2xf32>, %arg1: !ng.tensor<2x2xf32>) -> !ng.t
   "ng.return"(%0) : (!ng.tensor<2x2xf32>) -> ()
 }
 
+// CHECK-LABEL: func @equal_float
+func @equal_float(%arg0: !ng.tensor<2x2xf32>, %arg1: !ng.tensor<2x2xf32>) -> !ng.tensor<2x2x!ng.u8> {
+  // CHECK: %{{[0-9]+}} = "ng.equal"(%{{.*}}, %{{.*}}) : (!ng.tensor<2x2xf32>, !ng.tensor<2x2xf32>) -> !ng.tensor<2x2x!ng.u8>
```
dcaballe: Please use `{{.*}}` instead of `{{[[0-9]+]}}`. MLIR printing is constantly changing. I recently fixed some tests that were failing because of this.

pruthviIntel:
@nagy @dcaballe addressed the feedback, w.r.to templating `createConstant` function, looks we need to specialize for `int` and ` float ` types based on the Operand types. I am creating a patch for MLIR to include `zero_extendi` after that we don't need this helpers. So for now, I am just leaving as it is, does it sound okay?

pruthviIntel:
@nmostafa @dcaballe can you guys take another peek on this? :)

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1245,6 +1270,8 @@ namespace
         auto pivs = makeIndexHandlePointers(ivs);
         // Steps
         auto steps = vLHS.getSteps();
+        // element type of the operand
+        Type elemTy = result->getType().dyn_cast<MemRefType>().getElementType();
```
dcaballe: `dyn_cast` -> `cast` since we want a proper error if type is not a MemRef here.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1410,6 +1476,30 @@ namespace
...
         NGRAPH_UNREACHABLE("Unsupported type");
     }
+
+    ValueHandle createOneConstant(mlir::Type type)
+    {
+        if (auto floatTy = type.dyn_cast<FloatType>())
+        {
+            if (floatTy.isF32())
+            {
+                return intrinsics::constant_float(llvm::APFloat(1.0f), floatTy);
```
dcaballe: Could you please check if there is something to create one constant given a type. There has to be something in the builder or constant op.

src/contrib/mlir/core/ngraph_dialect/type.hpp
dcaballe: yeah, if you are lowering to boolean type, this should be no longer necessary, right?

test/core_fusion.cpp
```diff
@@ -830,6 +830,11 @@ TEST(core_fusion, softmax_crossentropy_bprop_with_ignore_mask)
 }
 #endif
 
+// TODO(pthoreho): MLIR currently does not support all the op's needed for CrossEntropy+Softmax
+// this results in multiple CompiledKernels and we cannot able to safely check for certain op's
+// from the function created by user.
+// Note: remove this gaurds once we have full support for CE and Softmax through MLIR
```
dcaballe: typo

test/mlir/affine_conversion/core_ops.mlir
```diff
@@ -18,6 +18,108 @@ func @simple_gather(%arg0: !ng.tensor<16x!ng.i64>, %arg1: !ng.tensor<512x32xf32>
...
+// Equal Op
+// CHECK:       affine.for %[[I:.*]] = 0 to 2
+// CHECK-NEXT:  affine.for %[[J:.*]] = 0 to 2
+// CHECK-NEXT:  %[[C1:.*]] = constant
+// CHECK-NEXT: 	%[[C2:.*]] = constant
+// CHECK:	%[[O1:.*]] = affine.load  %{{.*}}[%[[I]], %[[J]]]
+// CHECK:	%[[O2:.*]] = affine.load  %{{.*}}[%[[I]], %[[J]]]
+// CHECK-NEXT:	%[[R1:.*]] = cmpf "oeq"
+// CHECK:       %[[R2:.*]] = select
+// CHECK-NEXT:  affine.store %[[R2]], %{{.*}}[%[[I]], %[[J]]]
```
dcaballe: They look great! Thanks! Some suggestions:
- I think we should verify that the `cmpf` operand are actually `O1` and `O2` and that they are in the right order.
- Same for select (and therefore, constant ops): we should check operands to make sure the order is correct.
- We can add a `CHECK-LABEL: func @simple_equal` to make sure we prevent other checks from matching against other functions.
- nit: it would be great if you could use the same indentation for all the instructions within the same loop. Otherwise, they look like they are at different nesting levels.

Same comments for tests below.

pruthviIntel:
@dcaballe changed it

pruthviIntel:
TESTNOW

PR #3873, opened 2019-11-12
 Cumulative Sum 

pruthviIntel:
### CumSum

Performs cumulative sum of the input elements along the given axis. By default, it will do the sum inclusively meaning the first element is copied as is. Through an exclusive attribute, this behavior can change to exclude the first element. It can also perform summation in the opposite direction of the axis. For that, set reverse attribute to 1.
### 
Example:

```
input_x = [1, 2, 3]
axis=0
output = [1, 3, 6]
exclusive=1
output = [0, 1, 3]
exclusive=0
reverse=1
output = [6, 5, 3]
exclusive=1
reverse=1
output = [5, 3, 0]
```
_Version

This version of the operator has been available since version 11 of the default ONNX operator set.

Attributes

exclusive : int (default is 0)
If set to 1 will return exclusive sum in which the top element is not included. In other terms, if set to 1, the j-th output element would be the sum of the first (j-1) elements. Otherwise, it would be the sum of the first j elements.
reverse : int (default is 0)
If set to 1 will perform the sums in reverse direction.
Inputs

x : T
An input tensor that is to be processed.
axis : T2
(Optional) A 0-D tensor. Must be in the range [-rank(x), rank(x)-1]. Negative value means counting dimensions from the back._


src/ngraph/op/cum_sum.cpp
```diff
@@ -0,0 +1,74 @@
...
+op::CumSum::CumSum(const Output<Node>& arg,
+                   const Output<Node>& axis,
+                   const int exclusive,
+                   const int reverse)
+    : Op({arg, axis})
+    , m_exclusive(exclusive)
+    , m_reverse(reverse)
+{
+    auto const_op = as_type_ptr<op::Constant>(axis.get_node_shared_ptr());
+    m_axis = *(static_cast<const int64_t*>(const_op->get_data_ptr()));
```
tsocha: What will happen if `axis` isn't `Constant`?
Will we get a segfault here?

src/ngraph/op/cum_sum.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    if (m_axis < 0)
+    {
+        m_axis = m_axis + arg.get_shape().size();
```
tsocha: We should validate if `m_axis` is valid after this operation.
The valid range for `m_axis` is `[-rank, rank)`

src/ngraph/op/cum_sum.cpp
```diff
@@ -0,0 +1,74 @@
...
+    auto x = input_value(0);
+    auto& x_shape = x.get_shape();
+
+    adjoints.add_delta(
+        x, make_shared<op::Broadcast>(delta, x_shape, AxisSet{static_cast<size_t>(get_axis())}));
+}
+
+shared_ptr<Node> op::CumSum::get_default_value() const
+{
+    return ngraph::make_constant_from_string("0", get_element_type(), get_shape());
```
tsocha: Please use a `Constant::create` method.

src/ngraph/op/cum_sum.hpp
```diff
@@ -0,0 +1,99 @@
...
+           int64_t get_axis() const { return m_axis; }
+           int is_exclusive() const { return m_exclusive; }
+           int is_reverse() const { return m_reverse; }
+       protected:
+           virtual void generate_adjoints(autodiff::Adjoints& adjoints,
+               const NodeVector& deltas) override;
+    
+       private:
+	   int m_exclusive;
+	   int m_reverse;
```
tsocha: There is a proposal to the spec which define these attributes as boolean values.
Please use `bool` instead of `int` here.

test/backend/cum_sum.in.cpp
```diff
@@ -0,0 +1,177 @@
...
+    auto backend = runtime::Backend::create("${BACKEND_NAME}");
+
+    // Create some tensors for input/output
+    auto a = backend->create_tensor(element::f32, shape);
+    copy_data(a, vector<float>{1, 2, 3, 4});
+    auto result = backend->create_tensor(element::f32, shape);
+
+    auto handle = backend->compile(f);
+    handle->call_with_validate({result}, {a});
+    EXPECT_TRUE(test::all_close_f((vector<float>{1, 3, 6, 10}), read_vector<float>(result)));
```
tsocha: Could you use `test::NgraphTestCase` for all of these tests?

src/ngraph/op/cum_sum.hpp
```diff
@@ -0,0 +1,99 @@
...
+       /// | ----- | --------------------------------------------------------------- |
+       /// | `axis`| An axis position along which cumulative sum must be performed.  |
+       ///
+       /// ## Output
+       ///
+       /// | Description                                                                         |
+       /// | ------------------------------------------------------------------------------------|
+       /// | Output tensor of the same type as `arg` with cumulative sums of the arg's elements  |
+       //
+       // clang-format off
```
mitruska: I suggest to `// clang-format on` here and apply style for the rest of code in this file. 
There is some indentation issues.

src/ngraph/op/cum_sum.hpp
sayantan-nervana: In TF, exclusive and reverse are bools too: https://www.tensorflow.org/versions/r1.14/api_docs/cc/struct/tensorflow/ops/cumsum/attrs.html#structtensorflow_1_1ops_1_1_cumsum_1_1_attrs

pruthviIntel: I will change them too to `bool` I initially had `bool` in mind, but onnx doc misguided me :)

src/ngraph/op/cum_sum.hpp
```diff
@@ -0,0 +1,99 @@
...
+           bool is_exclusive() const { return m_exclusive; }
+           bool is_reverse() const { return m_reverse; }
+       protected:
+           virtual void generate_adjoints(autodiff::Adjoints& adjoints,
+               const NodeVector& deltas) override;
+    
+       private:
+	   bool m_exclusive;
+	   bool m_reverse;
+	   int64_t m_axis;
```
postrational: In order to support the ONNX version of `CumSum`, we will need `axis` to be a dynamic input, not a static attribute of the op.

src/ngraph/op/cum_sum.cpp
mitruska: Output from debugging ngraph-onnx tests:
```
tests/test_backend.py::OnnxBackendNodeModelTest::test_cumsum_1d_cpu
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
```
``` 
#4  0x00007fff4f7c3509 in ngraph::op::CumSum::CumSum (this=0x157e910, arg=..., axis=..., exclusive=0, reverse=0)
    at /root/projects/ngraph/src/ngraph/op/cum_sum.cpp:36
36          m_axis = *(static_cast<const int64_t*>(const_op->get_data_ptr()));
```

pruthviIntel:
@tsocha @mitruska can you re-review and approve if it looks good? :)

PR #3874, opened 2019-11-12
[SPEC] ReduceLogicalAnd & ReduceLogicalOr

tomdol:


src/ngraph/op/reduce_logical_and.hpp
```diff
@@ -0,0 +1,53 @@
...
+{
+    namespace op
+    {
+        namespace v1
+        {
+            /// \brief Performs a reduction using "logical and"
+            ///
+            /// The reduction is performed over slices of the first input. The slices shape depends
+            /// on the values passed to the second input - the axes.
+            class ReduceLogicalAnd : public util::ArithmeticReductionKeepDims
```
GlebKazantaev: I don't see changes related to constant folding passes for this new ops, so please add it. Also according to "nGraph/DLDT common operation set" table you have implementation for v0 ops.

tomdol: Done, please have a look.

PR #3879, opened 2019-11-13
[MLIR] Bump MLIR commit to c61db4bb

nmostafa:
Fixes:
* Rename Affine loop builders. 
* Parser/Printer call-backs take an AsmParser/Printer. 
* StaticFloatMemRef is removed. Added our own struct. The struct just holds the first data pointer and skips offset, shape and strides. This is OK since we only support static shapes/strides now. 
* Added more unit-tests to cover NGTensorType parsing.


src/contrib/mlir/core/ngraph_dialect/dialect.cpp
```diff
@@ -39,63 +39,60 @@ NGraphOpsDialect::NGraphOpsDialect(mlir::MLIRContext* ctx)
...
-        {
-            return (emitError(loc, "Unexpected element type in tensor type: " + tyData), Type());
-        }
+        // Parse the current element type.
+        Type eltType;
+
+        parser.parseType(eltType);
 
-        return NGTensorType::get(context, elem_ty, shape);
+        parser.parseGreater();
```
dcaballe: Nice! These are great improvements! Should we keep error messages in lines 81 and 70?

src/contrib/mlir/core/ngraph_dialect/dialect.cpp
```diff
@@ -39,63 +39,60 @@ NGraphOpsDialect::NGraphOpsDialect(mlir::MLIRContext* ctx)
...
+        return NGTensorType::get(context, eltType, shape);
     }
+    else
+    {
+        // parse nGraph scalar type
+        return parseEltType(parser);
```
dcaballe: 64 goes through `parser` and 72 calls `parseEltType` directly. Shouldn't we use the same for both?

src/contrib/mlir/runtime/cpu/cpu_runtime.cpp
```diff
@@ -129,39 +129,37 @@ void MLIRCPURuntime::cleanup()
     // Free void double pointer arguments without freeing external tensor data.
     for (auto* arg : m_invokeArgs)
     {
-        auto* memRefArg = *(reinterpret_cast<mlir::StaticFloatMemRef**>(arg));
+        auto* memRefArg = *(reinterpret_cast<StaticMemRef**>(arg));
```
dcaballe: How is MLIR replacing `StaticFloatMemRef`?

test/mlir/ngraph_dialect/types.mlir
```diff
@@ -55,23 +55,23 @@ func @i64(%arg0: !ng.i64) {
...
+// CHECK-SAME: (%{{.*}}: !ng.u16)
 func @u16(%arg0: !ng.u16) {
   "ng.return"() : () -> ()
 }
 
 // -----
 
 // CHECK-LABEL: func @u32
-// CHECK-SAME: (%{{.*}}: !ng.i32)
+// CHECK-SAME: (%{{.*}}: !ng.u32)
```
dcaballe: Thanks for fixing this

src/contrib/mlir/core/ngraph_dialect/dialect.cpp
```diff
@@ -39,63 +39,60 @@ NGraphOpsDialect::NGraphOpsDialect(mlir::MLIRContext* ctx)
...
+    // failure is true
+    if (!parser.parseOptionalKeyword("tensor"))
     {
-        if (!tyData.consume_front("<") || !tyData.consume_back(">"))
+        llvm::SMLoc typeLoc = parser.getCurrentLocation();
+        if (parser.parseLess())
         {
-            return (emitError(loc, "expected '<' and '>' enclosing the tensor shape: " + tyData),
-                    Type());
+            parser.emitError(typeLoc, "expected '<' and '>' enclosing the tensor shape");
```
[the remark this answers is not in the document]
nmostafa: No need to. The parser emitError will print the offending line along with a location.

src/contrib/mlir/core/ngraph_dialect/dialect.cpp
nmostafa: The difference is subtle: @line 64 the parser is at `ng.` if we have an nGraph type (e.g ng.i32) or at the type token directly if we have std type (e.g f32). The call to `parser.parseType` will decide whether to call us back or not based on the existence of that prefix. IOW, it is capable of handling both ng and std types. 
@line 72, the parser is already past the `ng.` prefix (if any), when we go the callback. If I call `parser.parseType` here on `ng.i32`, the parser will assume intType token (i32) and return std type. I gave that a try.

src/contrib/mlir/runtime/cpu/cpu_runtime.cpp
nmostafa: I haven't seen any replacement. JITRunner has its own struct as well. If you know of any, please point me to it.

src/contrib/mlir/core/ngraph_dialect/dialect.cpp
nmostafa: Right. Fixed.

PR #3884, opened 2019-11-13
[SPEC] Add OneHot:v1

mbencer:
Assumptions `OneHot:v1` (all required):
Inputs:
- `indices` (any integer)
- `depth` (0D scalar)
- `on_value` (0D scalar)
- `off_value` (0D scalar)
Attributes:
- `axis` (int)

Steps:
- implementation `OneHot:v1` ngraph op
- `upgrade pass`
- `downgrade pass`
- using `OneHot:v1` in `onnx_importer`
- implementation `serialization`


src/ngraph/validation_util.hpp
```diff
@@ -102,4 +102,12 @@ namespace ngraph
...
                                    const AxisSet& shrink_axis_mask,
                                    const AxisSet& ellipsis_mask);
+
+    int64_t normalize_axis(const Node* node, std::int64_t axis, std::int64_t tensor_rank);
+
+    int64_t normalize_axis(const Node* node,
+                           std::int64_t axis,
+                           std::int64_t tensor_rank,
+                           std::int64_t axis_range_min,
+                           std::int64_t axis_range_max);
```
tsocha: Since you have added these functions please remove the old ones(`validate_axis`) from `onnx_importer`.

mbencer: Done. `onnx_impoter` now uses ngraph version of `normalize_axis` .

src/ngraph/validation_util.hpp
```diff
@@ -103,5 +103,17 @@ namespace ngraph
                                    const AxisSet& shrink_axis_mask,
                                    const AxisSet& ellipsis_mask);
 
-    std::size_t normalize_axis(const Node* node, std::int64_t axis, std::int64_t tensor_rank);
+    int64_t normalize_axis(const Node* node, std::int64_t axis, std::int64_t tensor_rank);
```
postrational: Please add documentation for all of these functions.

src/ngraph/frontend/onnx_import/utils/common.cpp
```diff
@@ -60,23 +61,8 @@ namespace ngraph
...
-                             "].");
-
-                if (axis < 0)
-                {
-                    axis = axis + tensor_rank;
-                }
-
-                return static_cast<size_t>(axis);
+                return ngraph::normalize_axis(
+                    node.get_description(), axis, tensor_rank, axis_range_min, axis_range_max);
```
postrational: We can do this in a separate story, but we should remove the `validate_axis` function altoghether, as it now just adds an unneeded extra layer.

src/ngraph/validation_util.hpp
mbencer: Done

src/ngraph/frontend/onnx_import/utils/common.cpp
mbencer: I agree with you, but for transparency let's do it in separate PR.

PR #3885, opened 2019-11-13
Round the right bit with denorms

diyessi:


GlebKazantaev:
I've checked your changes with our FP32 to FP16 transformation and conversion for this cases works as expected. But we still have some accuracy difference between path via FP16 IR and FP32 IR passed through transformation. I can't troubleshoot it right now, but probably I'll send you another cases when conversion works different.
Thank you for fast response!

PR #3890, opened 2019-11-14
Changes to get nnp-transformer building with ngraph

rkimballn1:
# Note
This is PR against branch r0.27


diyessi:
Unfortunately, this breaks some other things. I'll play with it on Friday and see if I can come up with something that works for everything.

PR #3893, opened 2019-11-15
CropAndResize op

diyessi:
Just the op, no implementation.


src/ngraph/op/crop_and_resize.hpp
```diff
@@ -0,0 +1,76 @@
...
+            {
+                unspecified,
+                bilinear,
+                nearest
+            };
+
+            NGRAPH_API
+            static const std::string type_name;
+            const std::string& description() const override { return type_name; }
+            /// \brief Constructs an absolute value operation.
```
sayantan-nervana: "absolute" -- > "cropandresize"

PR #3906, opened 2019-11-18
Fix for the bug with as_type_ptr for TensorIterator::Input/Ouput desc

itikhono:
this issue reproduces on Windows platforms (MSVC 17/19)


diyessi:
I added exports

PR #3908, opened 2019-11-18
[SPEC][FusedOp] Add Mod operator

mbencer:
Requirements:
The result consistent with `truncated(x / y) * y + truncated_mod(x, y) = x`. It means that `x/y` value is truncated like in C language (casting to `int`) instead of using `floor`. It is equivalent to `numpy.fmod`.

New `mod` operation was also added to `onnx_importer` with support only `fmod=1`.
To support `fmod=0` `FloorMod` can be used, but it was implemented as `v1` op without `downgrade` pass.


postrational:
Seems you should exclude `onnx_PlaidML.model_mod` in the PlaidML manifest file.

mbencer:
> You'll need to add the op to `op/op_v[0,1]_tbl.hpp` whichever is appropriate and add your header a few new places.

@diyessi Should `Mod` as new `FusedOp` belong to `op/op_v0_tbl.hpp` or to `op/op_v1_tbl.hpp`?

src/ngraph/op/fused/mod.cpp
```diff
@@ -0,0 +1,59 @@
...
+    auto division = make_shared<op::Convert>(
+        make_shared<op::Divide>(dividend, divisor, m_auto_broadcast), ngraph::element::i64);
+    division = make_shared<op::Convert>(division, dividend_et);
+    // truncated(a / b) * b
+    const auto multiplication = make_shared<op::Multiply>(division, divisor, m_auto_broadcast);
+    // a mod b = a - truncated(a / b) * b
+    const auto mod = make_shared<op::Subtract>(dividend, multiplication, m_auto_broadcast);
+
+    // apply sign of dividend
+    return {make_shared<op::Multiply>(dividend_sign, mod, m_auto_broadcast)};
```
GlebKazantaev: Can we use V1 ops here?

mbencer: Yes, I changed used ops to `v1` (if `v1` were provided)

PR #3909, opened 2019-11-18
[SPEC] Add support string as AutoBroadcastSpec

mbencer:
According to specification `AutoBroadcastSpec` should be passed as string. It's my proposition to support this case.


src/ngraph/op/util/attr_types.cpp
```diff
@@ -126,4 +128,17 @@ namespace ngraph
     {
         return s << as_string(type);
     }
+
+    op::AutoBroadcastType op::AutoBroadcastSpec::type_from_string(const std::string& type) const
+    {
+        static const std::map<std::string, AutoBroadcastType> allowed_values = {
+            {"none", AutoBroadcastType::NONE},
```
diyessi: Let's keep the strings and enum values consistent.

mbencer: Done

src/ngraph/op/util/attr_types.hpp
```diff
@@ -257,6 +257,10 @@ namespace ngraph
                 , m_axis(0)
             {
             }
+            AutoBroadcastSpec(const char* type)
```
tsocha: Could we use `std::string` here?

mbencer: I consider it but `std::string` can not be used to implicit conversion here. It doesn't compile.

PR #3914, opened 2019-11-19
Added u1 precision for binary weights

GlebKazantaev:
To support BinaryConvolution we have to add `u1` precision due to we store BinaryConvolution weights packed. And it's not possible to use any of existing ngraph types to express such packed tensor.
Current support for u1 is limited.


src/ngraph/op/binary_convolution.cpp
```diff
@@ -114,18 +114,7 @@ void op::v1::BinaryConvolution::validate_and_infer_types()
         }
     }
 
-    element::Type result_et;
     PartialShape result_shape;
-
```
nmostafa: Shouldn't the filters type be of compatible types ? Do need checks on `filters_et` here ?

src/contrib/mlir/core/pass/ng_dialect_builder.cpp
```diff
@@ -264,6 +264,7 @@ mlir::Type NgDialectConversionPass::getMlirType(const element::Type& type)
     {
     case ngraph::element::Type_t::undefined:
     case ngraph::element::Type_t::dynamic:
+    case ngraph::element::Type_t::u1:
```
nmostafa: You could map it to u8 here (check out boolean below).

src/ngraph/op/binary_convolution.cpp
GlebKazantaev: This check was wrong because data type and filters type can not be the same.

src/contrib/mlir/core/pass/ng_dialect_builder.cpp
GlebKazantaev: size of `u1` is 1 that is differ from size of u8. Also it's not related to my initial purpose (just to add u1 precision), so let's handle it in future.

GlebKazantaev: size of u1 is 1 that is differ from size of u8. Also it's not related to my initial purpose (just to add u1 precision), so let's handle it in future.

src/ngraph/op/constant.cpp
```diff
@@ -78,6 +78,7 @@ string op::Constant::convert_value_to_string(size_t index) const
     case element::Type_t::u16: rc = to_string(get_vector<uint16_t>()[index]); break;
     case element::Type_t::u32: rc = to_string(get_vector<uint32_t>()[index]); break;
     case element::Type_t::u64: rc = to_string(get_vector<uint64_t>()[index]); break;
+    case element::Type_t::u1: throw runtime_error("unsupported type");
```
rkimballn1: u1 should be supported in constant

src/ngraph/type/element_type.cpp
```diff
@@ -80,6 +81,7 @@ static const map<element::Type_t, const TypeInfo>& get_type_info_map()
         {element::Type_t::i16, TypeInfo(16, false, true, false, "int16_t", "i16")},
         {element::Type_t::i32, TypeInfo(32, false, true, true, "int32_t", "i32")},
         {element::Type_t::i64, TypeInfo(64, false, true, false, "int64_t", "i64")},
+        {element::Type_t::u1, TypeInfo(1, false, false, false, "u1", "u1")},
```
rkimballn1: parameter 5 "u1" is the C type used to represent this number and u1 is not a valid C type.

GlebKazantaev: Yes `u1` is not a valid C type, this is normal. What's your proposal? This time we are going to add limited support for u1 only. There is no need to manipulate with u1 values except just storing it in a constant node and consuming in BinaryConvolution. It shouldn't affect any parts of nGraph.

src/ngraph/op/constant.cpp
GlebKazantaev: Added convert_value_to_string support for u1.

PR #3916, opened 2019-11-19
Enable pipelining in CPU Backend

magrawal128:
Added code similar to INT_Execfutable in CPU_Executable to enable pipelining.


src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -210,6 +210,76 @@ vector<runtime::PerformanceCounter> runtime::cpu::CPU_Executable::get_performanc
...
+                                                          parameter->get_shape());
+        tensor = static_pointer_cast<runtime::cpu::CPUTensorView>(t);
+        tensors.push_back(tensor);
+    }
+    vector<shared_ptr<runtime::Tensor>> result_tensors;
+    for (const shared_ptr<runtime::cpu::CPUTensorView>& tensor : tensors)
+    {
+        result_tensors.push_back(tensor);
+    }
+    return result_tensors;
```
pruthviIntel: do we need to have two temp vectors here ? `tensors` and `result_tensors` . we can just update result tensor in the first loop?

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -210,6 +210,76 @@ vector<runtime::PerformanceCounter> runtime::cpu::CPU_Executable::get_performanc
...
+                                                          result->get_shape());
+        tensor = static_pointer_cast<runtime::cpu::CPUTensorView>(t);
+        tensors.push_back(tensor);
+    }
+    vector<shared_ptr<runtime::Tensor>> result_tensors;
+    for (const shared_ptr<runtime::cpu::CPUTensorView>& tensor : tensors)
+    {
+        result_tensors.push_back(tensor);
+    }
+    return result_tensors;
```
pruthviIntel: same comment as above ^^

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -210,6 +210,76 @@ vector<runtime::PerformanceCounter> runtime::cpu::CPU_Executable::get_performanc
...
+
+shared_ptr<runtime::Tensor> runtime::cpu::CPU_Executable::create_output_tensor(size_t output_index)
+{
+    shared_ptr<op::Result> result = get_result(output_index);
+    return make_shared<runtime::cpu::CPUTensorView>(result->get_element_type(),
+                                                    result->get_shape());
+}
+
+vector<shared_ptr<runtime::Tensor>>
+    runtime::cpu::CPU_Executable::create_input_tensor(size_t input_index, size_t pipeline_depth)
```
tomdol: You're doing a lot of unnecessary copying in this method(and in `create_output_tensor` too). This method is supposed to return a vector of tensor views with `pipeline_depth` number of elements. All of them have the same type and shape, so you can just create just one and use constructor number 3 https://en.cppreference.com/w/cpp/container/vector/vector which will let you remove both `for` loops from both methods.

jbobba:
thanks @tomdol and @pruthviIntel for the feedback. We will get the improvements done in another PR. I will unblock this PR to let it go into 0.27.

PR #3926, opened 2019-11-22
[SPEC] Add negative axes support for ReverseSequence

mbencer:


src/ngraph/op/reverse_sequence.cpp
```diff
@@ -41,21 +42,23 @@ void op::ReverseSequence::validate_and_infer_types()
...
+                              "batch_axis=",
+                              m_batch_axis,
+                              ", seq_axis=",
+                              m_seq_axis,
+                              ")");
+    }
+
+    if (input_rank.is_static())
+    {
+        m_batch_axis = ngraph::normalize_axis(this, m_batch_axis, static_cast<int64_t>(input_rank));
```
diyessi: Have separate members `m_batch_normalized_axis` and `m_seq_normalized_axis` and set them. That way if the inputs change rank and validation gets re-run, the right thing happens. It also means the serialized graph will have the provided attribute values as opposed to the normalized attributes.

mbencer: Done, but I am sure to all aspects of you review. I added additional methods to serialize origin axes values. `get_batch_axis` and `get_sequence_axis` methods return normalized axes values in order to leave `size_t` type expected by current backends.

PR #3933, opened 2019-11-24
Make Placement enum a Node member and simplify enum

rkimballn1:


src/ngraph/node.hpp
```diff
@@ -407,18 +407,22 @@ namespace ngraph
         /// True if this and node have one output with same element type and shape
         bool has_same_type(std::shared_ptr<const Node> node) const;
 
+        enum class Placement
```
diyessi: If this is a class with an enum member, and the class has `is_on_host()` etc. methods, it will be easier to support more than one device in the future, as will be needed for model parallelism.

rkimballn1: Tis not a class, tis an enum class and enum classes cannot contain methods.

rkimballn1: putting the enum inside another class makes it that much harder to get at, something like `Placement::PlacementEnum::UNDEFINED` instead of just `Placement::UNDEFINED`.
I like solving future problems in the future because I find they often never materialize or change drastically.

PR #3935, opened 2019-11-25
[SPEC] Adjust ConvolutionBackpropData op.

tsocha:
```
inputs:
  1. filters-------+
  2. output_delta  |  -> 1. data
                   +---> 2. filters
  3. data_batch_shape -> 3. output_shape(+optional)

attributes:
  1. strides          -> 1. strides
  2. dilations-----+
  3. pads_begin    |  -> 2. pads_begin
  4. pads_end      |  -> 3. pads_end
                   +---> 4. dilations
		      -> 5. +auto_pad(optional)[PadType::EXPLICIT]
		      -> 6. +output_padding(optional)[zeros]
```


src/ngraph/pass/opset0_downgrade.cpp
```diff
@@ -181,22 +181,37 @@ bool pass::Opset0Downgrade::run_on_node(shared_ptr<Node> node)
...
+                     "if output_shape is not constant. Node: ",
+                     *node);
         const size_t num_spatial_dims = static_cast<size_t>(delta_arg_pshape.rank()) - 2;
+
+        auto output_padding = tmp->get_output_padding();
+
+        bool is_op_valid = true;
+        for (auto value : output_padding)
+        {
+            is_op_valid = is_op_valid && (value == 0);
```
tomdol: TBH I don't like this check very much. Can you use `std::all_of` instead with a lambda that checks the equality of elements with zero instead?

tsocha: Done.

PR #3936, opened 2019-11-25
Enable 4d bgemm support

jeromean:
Signed-off-by: Jerome <jerome.anand@intel.com>


rkimballn1:
not an op in an official opset so it's going away

PR #3938, opened 2019-11-25
Adding default ctor for Constant

gaurides:
Missed in previous PR


diyessi:
Already fixed in #3813

PR #3940, opened 2019-11-25
Add groupconvolution bprop

baojun-nervana:
TODO:
- [x] add ut


src/ngraph/op/fused/group_conv.hpp
```diff
@@ -85,5 +85,83 @@ namespace ngraph
         private:
             bool has_groups_in_filters_shape() const;
         };
+
+        /// \brief Group Convolution data batch backprop
+        class GroupConvolutionBackpropData : public ngraph::op::util::FusedOp
+        {
+        public:
+            NGRAPH_API
```
diyessi: As of this morning, it's `class NGRAPH_API GroupConvolutionBackpropData ...` instead of the `NGRAPH_API` before the `type_info`.

src/ngraph/op/fused/group_conv.hpp
```diff
@@ -85,5 +85,83 @@ namespace ngraph
...
+            CoordinateDiff m_padding_below;
+            CoordinateDiff m_padding_above;
+            Dimension m_groups;
+        };
+
+        /// \brief Group Convolution filters backprop
+        class GroupConvolutionBackpropFilters : public ngraph::op::util::FusedOp
+        {
+        public:
+            NGRAPH_API
```
diyessi: The `NGRAPH_API` also needs to be moved.

src/ngraph/op/fused/group_conv.hpp
baojun-nervana: updated

src/ngraph/op/fused/group_conv.hpp
baojun-nervana: updated

src/ngraph/op/fused/group_conv.cpp
```diff
@@ -229,3 +229,241 @@ bool ngraph::op::GroupConvolution::has_groups_in_filters_shape() const
...
+    {
+        set_output_type(0, data_element_type, PartialShape::dynamic());
+    }
+}
+
+size_t ngraph::op::GroupConvolutionBackpropData::get_groups() const
+{
+    NODE_VALIDATION_CHECK(this,
+                          m_groups.is_static(),
+                          "get_groups() can only be called if the number of groups is static.");
```
tsocha: There is no option to initialize groups by other type than `size_t` so this check is unnecessary.
Groups are a required `size_t` attribute for constructor and public setter is unavailable so `m_groups` will be always static.

src/ngraph/op/fused/group_conv.cpp
```diff
@@ -229,3 +229,241 @@ bool ngraph::op::GroupConvolution::has_groups_in_filters_shape() const
...
+    {
+        set_output_type(0, filters_element_type, PartialShape::dynamic());
+    }
+}
+
+size_t ngraph::op::GroupConvolutionBackpropFilters::get_groups() const
+{
+    NODE_VALIDATION_CHECK(this,
+                          m_groups.is_static(),
+                          "get_groups() can only be called if the number of groups is static.");
```
tsocha: As above

src/ngraph/op/fused/group_conv.cpp
baojun-nervana: updated

PR #3941, opened 2019-11-26
Switch interpreter to opset0 + a few extras

diyessi:


src/ngraph/runtime/interpreter/int_extra_op_tbl.hpp
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+//*****************************************************************************
+
+// This table contains ops not in opset0 handled by the interpreter
+
+NGRAPH_OP(LessEqual, op::v1)
+NGRAPH_OP(LogicalAnd, op::v1)
+NGRAPH_OP(LogicalOr, op::v1)
+NGRAPH_OP(LogicalXor, op::v1)
+NGRAPH_OP(LogicalNot, op::v1)
```
jbobba: This can get confusing if the interpreter supports op::v2::LessEqual. We could do NGRAPH_OP(LessEqual, v1) and adjust the macro in int_executable.cpp

PR #3943, opened 2019-11-26
[SPEC] Adjust Split

mbencer:
Requirements:
`axis` of `Split` should be `Node` (because of `FusedOp` it must be `Constant`)

Steps:
- updated `serializer`
- updated (and refactored) `onnx_importer`


src/ngraph/frontend/onnx_import/utils/common.cpp
```diff
@@ -79,6 +80,17 @@ namespace ngraph
                 return new_axes;
             }
 
+            ngraph::NodeVector get_outputs(const std::shared_ptr<ngraph::Node>& node,
+                                           int64_t outputs_number)
+            {
+                ngraph::NodeVector outputs(outputs_number);
+                for (int i = 0; i < outputs_number; ++i)
+                {
+                    outputs[i] = std::make_shared<ngraph::op::GetOutputElement>(node, i);
```
diyessi: `outputs[i] = Output(node, i).as_single_output_node()` re-uses existing `GetOutputElement` nodes and only creates `GetOutputElement` if the output needs one.
There is also `as_node_vector(node.outputs())` which would do everything this method does without the need for `outputs_number`.

mbencer: In `as_node_vector` is called `as_single_output_node` which requires `node->get_output_size() == 1'` which is not true for some tests (such as `model_split_equal_parts_default`).
Similar problem occurs for `as_single_output_node`. It also requires that output node must produce exactly one value.

mbencer: @diyessi In `as_node_vector` is called `as_single_output_node` which requires `node->get_output_size() == 1'` which is not true for some tests (such as `model_split_equal_parts_default`). Assert in `check_single_output_arg` fails.
Similar problem occurs for version `outputs[i] = node->get_output_as_single_output_node(i)` - assert `check_single_output_args` fails.
I am not sure how to deal with it.

src/ngraph/op/fused/split.hpp
```diff
@@ -37,24 +37,29 @@ namespace ngraph
...
+                ///                should be split. Negative values mean counting from
+                ///                the back of the input tensor's shape.
+                /// \param splits  a list of lengths that the input tensor should be
+                ///                split to. Use this constructor to split the input
+                ///                tensor to variable length chunks.
+                Split(const Output<Node>& data,
+                      const Output<Node>& axis,
+                      const std::vector<size_t>& splits);
+
+                // TODO REMOVE IT. INTRODUCED TO PROVIDE CI COMPATIBILITY
```
postrational:
```suggestion
                // TODO REMOVE THIS CONSTRUCTOR. INTRODUCED TO PROVIDE CI COMPATIBILITY
```

src/ngraph/frontend/onnx_import/utils/common.hpp
```diff
@@ -115,6 +115,15 @@ namespace ngraph
                                                    std::vector<std::int64_t> axes,
                                                    std::int64_t tensor_rank);
 
+            /// \brief Return the outputs of the node as vector.
+            ///
+            /// \param[in] node            Node with multiple outputs.
+            /// \param[in] outputs_number  Number of outputs in returned vector.
+            ///
+            /// \return                    Vector of outputs of input node.
+            ngraph::NodeVector get_outputs(const std::shared_ptr<ngraph::Node>& node,
```
postrational: It would be great if `outputs_number` was optional and the function returned all outputs by default.

src/ngraph/op/fused/split.hpp
mbencer: Done

src/ngraph/frontend/onnx_import/utils/common.hpp
mbencer: Thanks. `outputs_number` was not needed. I used `get_output_size`

src/ngraph/op/fused/split.hpp
```diff
@@ -37,24 +37,32 @@ namespace ngraph
...
+                ///                should be split. Negative values mean counting from
+                ///                the back of the input tensor's shape.
+                /// \param splits  a list of lengths that the input tensor should be
+                ///                split to. Use this constructor to split the input
+                ///                tensor to variable length chunks.
+                Split(const Output<Node>& data,
+                      const Output<Node>& axis,
+                      const std::vector<size_t>& splits);
+
+                // TODO REMOVE THIS CONSTRUCTOR. INTRODUCED TO PROVIDE CI COMPATIBILITY
```
mbencer: Ticket for removing unnecessary `TODO` code after CI integration was created

src/ngraph/frontend/onnx_import/utils/common.cpp
diyessi: It should be returning the real node itself for single-output nodes, and GetOutputElement(real node, i) for the other nodes.

mbencer: @diyessi Thanks. I introduced this check.

PR #3945, opened 2019-11-26
v1::Reshape zero_flag renamed. Default value unset

jane-intel:


src/ngraph/op/reshape.hpp
```diff
@@ -130,26 +130,25 @@ namespace ngraph
                 ///        be of the form \f$(b_0,\dots,b_{j-1})\f$ where \f$\Pi(a_i) = \Pi(b_i)\f$.
                 ///        A value of -1 is allowed for at most one dimension, in which case the
                 ///        dimension size is inferred based on element count of input tensor.
-                /// \param zero_flag Treats zeros in `pattern` as wildcard flags indicating a copy
+                /// \param special_zero Treats zeros in `pattern` as wildcard flags indicating a
+                /// copy
                 /// from input shape at the same index.
-                Reshape(const Output<Node>& arg,
-                        const Output<Node>& pattern,
-                        bool zero_flag = false);
```
tsocha: Why default value for this flag has been removed?
I can't see any benefit for this change for nGraph code.

GlebKazantaev: We have a merge request for the spec for a new name for this attribute. We had an old request to propose a new name for the attribute.

tsocha: I'm asking about the default value not the attribute name.

GlebKazantaev: Sorry. Some time ago we discussed with @jbobba and @diyessi that we should over-use default values. This case is a good example when there is no default value, because TF, which is used for many-many models has this flag set to false, but other FMs have true value. So no obvious default value here.

tsocha: Ok, I see it now.
I will accept this change but please consider to bring the default value back to nGraph(not to the spec).
It will help us to keep internal code tidy and will avoid to write `, false` in all usages of `Reshape` op.
Default value for this flag don't break nGraph compatibility with the spec IMHO.

diyessi: Not really an issue since bridges/importers are the primary builder of graphs other than for unit tests.

tsocha:
Please, do not force push so often it is difficult to track what has changed.
Merge commit is good enough.

PR #3947, opened 2019-11-26
Fix compose/decompose of group_conv for CPU backend. 

nmostafa:


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1227,6 +1227,14 @@ void runtime::cpu::CPU_ExternalFunction::register_common_passes(
             return false;
 #endif
         }
+        else if (typeid(ngraph::op::GroupConvolution) == typeid(node))
+        {
+            Strides stride_1{1, 1};
+            Node* p_node = const_cast<Node*>(&node);
+            auto conv = as_type<ngraph::op::GroupConvolution>(p_node);
+            return conv->get_window_dilation_strides() == stride_1 &&
```
ayzhuang: window dilation is supported.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1227,6 +1227,14 @@ void runtime::cpu::CPU_ExternalFunction::register_common_passes(
             return false;
 #endif
         }
+        else if (typeid(ngraph::op::GroupConvolution) == typeid(node))
```
diyessi: `else if (auto conv = as_type<ngraph::op::GroupConvolution>(&node))`

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1230,6 +1230,11 @@ void runtime::cpu::CPU_ExternalFunction::register_common_passes(
             return false;
 #endif
         }
+        else if (auto conv = as_type<ngraph::op::GroupConvolution>(const_cast<Node*>(&node)))
+        {
+            Strides stride_1{1, 1};
+            return conv->get_data_dilation_strides() == stride_1;
+        }
```
jbobba: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/cpu/pass/cpu_assignment.cpp#L145 would be a better check.

PR #3948, opened 2019-11-26
Add redrawn illustrations

evelynfay:


indie:
These are great, thank you @evelynfay  :shipit:

PR #3950, opened 2019-11-26
[MLIR] Update MLIR/LLVM repos

dcaballe:


src/contrib/mlir/core/ngraph_dialect/fused_ops.td
```diff
@@ -33,7 +33,7 @@ def NGSqueezeOp :
   }];
 
   let parser = [{ NGRAPH_CHECK(false, "No parser support"); return mlir::failure(); }];
-  let verifier = [{ return mlir::success(); /* TBD */ }];
+  // TODO: verifier
```
nmostafa: Not an issue, but what went wrong here ?

dcaballe: Lit tests failed with "unsupported verifier" or something like that for all these ops.

PR #3954, opened 2019-11-27
[SPEC] Add DeformablePSROIPooling v1

mbencer:


src/ngraph/op/deformable_psroi_pooling.hpp
```diff
@@ -0,0 +1,109 @@
...
+namespace ngraph
+{
+    namespace op
+    {
+        namespace v1
+        {
+            class DeformablePSROIPooling : public Op
+            {
+            public:
+                NGRAPH_API
```
diyessi: `class NGRAPH_API DeformablePSROIPooling ...` now instead of before `type_info`

mbencer: Done

src/ngraph/op/deformable_psroi_pooling.cpp
```diff
@@ -0,0 +1,157 @@
...
+                              ")");
+    }
+    int64_t output_rank = 4;
+    std::vector<Dimension> output_dim_vec(output_rank, Dimension::dynamic());
+    if (box_coords_pshape[0].is_static())
+    {
+        output_dim_vec[0] = box_coords_pshape.to_shape()[0];
+    }
+    output_dim_vec[1] = m_output_dim;
+    for (int i = 2; i < 4; ++i)
```
tomdol: Shouldn't `output_rank` be used in this loop?

mbencer: Right, thanks

src/ngraph/op/deformable_psroi_pooling.cpp
```diff
@@ -0,0 +1,157 @@
...
+    , m_spatial_bins_y(spatial_bins_y)
+    , m_no_trans(no_trans)
+    , m_trans_std(trans_std)
+    , m_part_size(part_size)
+{
+    constructor_validate_and_infer_types();
+}
+
+op::v1::DeformablePSROIPooling::DeformablePSROIPooling(const Output<Node>& input,
+                                                       const Output<Node>& coords,
```
slyalin: The specification has been recently updated to have all 3 inputs required. Sorry for the late response, but the MR with update has been committed just a minute ago.

mbencer: Op updated to match new specification

PR #3956, opened 2019-11-27
[SPEC] Add new v1::GroupConvolution and v1::GroupConvolutionBackpropData operators.

tsocha:


src/ngraph/op/group_convolution.hpp
```diff
@@ -0,0 +1,196 @@
...
+                ///
+                GroupConvolution(const Output<Node>& data_batch,
+                                 const Output<Node>& filters,
+                                 const Strides& strides,
+                                 const CoordinateDiff& pads_begin,
+                                 const CoordinateDiff& pads_end,
+                                 const Strides& dilations,
+                                 const PadType& auto_pad = PadType::EXPLICIT);
+
+                size_t get_version() const override { return 1; }
```
tomdol: Shouldn't this method be removed or return the value from `type_info` instead?

src/ngraph/op/group_convolution.hpp
```diff
@@ -0,0 +1,196 @@
...
+                GroupConvolutionBackpropData(const Output<Node>& data,
+                                             const Output<Node>& filter,
+                                             const Strides& strides,
+                                             const CoordinateDiff& pads_begin,
+                                             const CoordinateDiff& pads_end,
+                                             const Strides& dilations,
+                                             const PadType& auto_pad = PadType::EXPLICIT,
+                                             const CoordinateDiff& output_padding = {});
+
+                size_t get_version() const override { return 1; }
```
tomdol: Same question as above

src/ngraph/op/group_convolution.hpp
diyessi: Remove

src/ngraph/op/group_convolution.hpp
diyessi: Remove

src/ngraph/op/group_convolution.cpp
```diff
@@ -0,0 +1,331 @@
+//*****************************************************************************
```
jbobba: Why is this not a fused op?

src/ngraph/op/group_convolution.hpp
```diff
@@ -0,0 +1,196 @@
...
+{
+    namespace op
+    {
+        namespace v1
+        {
+            /// \brief Batched convolution operation, with optional window dilation and stride.
+            class GroupConvolution : public Op
+            {
+            public:
+                NGRAPH_API
```
jbobba: Move NGRAPH_API up to the class declaration.

src/ngraph/op/group_convolution.cpp
tsocha: The original fused op was already used in passes, kernels and template functions.
Since name of the attributes(and his getters) would change it will make conflicts.
That's why I made a decision to make a new op instead of adjusting the existing fused operator.
The work on these operators started when fused ops didn't use namespaces for opsets.
We can consider to make these operators a v1 fused ops again but I would prefer to do it in another PR.

src/ngraph/op/group_convolution.hpp
tsocha: Done.

src/ngraph/op/group_convolution.hpp
tsocha: Done.

src/ngraph/op/group_convolution.hpp
tsocha: Done.

src/ngraph/op/fused/group_conv.cpp
```diff
@@ -28,17 +28,323 @@
...
+    return make_shared<v1::GroupConvolution>(new_args.at(0),
+                                             new_args.at(1),
+                                             m_strides,
+                                             m_pads_begin,
+                                             m_pads_end,
+                                             m_dilations,
+                                             m_auto_pad);
+}
+
+void op::v1::GroupConvolution::generate_adjoints(autodiff::Adjoints& adjoints,
```
diyessi: Don't need this.

PR #3957, opened 2019-11-27
Add v1 version of Subtract with Numpy broadcasting as default

mitruska:
In order to match with other `v1` arithmetic ops, added a `v1` version of Subtract which use `Numpy` broadcasting by default.


src/ngraph/pass/opset0_downgrade.cpp
```diff
@@ -611,6 +611,12 @@ bool pass::Opset0Downgrade::run_on_node(shared_ptr<Node> node)
         modified = true;
         break;
     }
+    case OP_TYPEID::Subtract_v1:
+    {
+        upgrade_binary_elementwise_node<op::v0::Subtract, op::v1::Subtract>(node);
```
mitruska: Now I see, I sould use `downgrade` here.

src/ngraph/op/subtract.hpp
```diff
@@ -50,9 +50,43 @@ namespace ngraph
                 virtual void generate_adjoints(autodiff::Adjoints& adjoints,
                                                const NodeVector& deltas) override;
             };
-        }
+        } // namespace v0
+
+        namespace v1
+        {
+            /// \brief Elementwise subtraction operation.
+            class Subtract : public util::BinaryElementwiseArithmetic
```
mbencer: After last nGraph changes (on master) it should be `class NGRAPH_API Subtract`.

src/ngraph/op/subtract.hpp
```diff
@@ -50,9 +50,43 @@ namespace ngraph
...
+            {
+            public:
+                NGRAPH_API
```
mbencer: `NGRAPH_API` macro for `type_info` should be removed

src/ngraph/op/subtract.hpp
```diff
@@ -50,9 +50,43 @@ namespace ngraph
...
+                         const Output<Node>& arg1,
+                         const AutoBroadcastSpec& auto_broadcast =
+                             AutoBroadcastSpec(AutoBroadcastType::NUMPY));
+
+                virtual std::shared_ptr<Node>
+                    copy_with_new_args(const NodeVector& new_args) const override;
+
+                virtual void generate_adjoints(autodiff::Adjoints& adjoints,
+                                               const NodeVector& deltas) override;
+                size_t get_version() const override { return 1; }
```
mbencer: In current version `get_version` is redundant. Version is set based on `type_info`

src/ngraph/op/subtract.hpp
mitruska: Ah yes, I've just resolved conflicts and added `class NGRAPH_API Subtract` only in `v0` namespace.
I missed `v1`, thanks:)

src/ngraph/op/subtract.hpp
mitruska: Okay, so can i just remove it?

src/ngraph/op/subtract.hpp
mitruska: Done

src/ngraph/op/subtract.hpp
```diff
@@ -50,9 +50,43 @@ namespace ngraph
...
+        {
+            /// \brief Elementwise subtraction operation.
+            class Subtract : public util::BinaryElementwiseArithmetic
+            {
+            public:
+                NGRAPH_API
+                static constexpr NodeTypeInfo type_info{"Subtract", 1};
+                const NodeTypeInfo& get_type_info() const override { return type_info; }
+                Subtract()
+                    : util::BinaryElementwiseArithmetic(AutoBroadcastSpec::NONE)
```
tsocha: Are you sure `AutoBroadcastType::NONE` this is correct value since default value for Broadcasting in the second constructor we are using `AutoBroadcastType::NUMPY`?
```suggestion
                    : util::BinaryElementwiseArithmetic(AutoBroadcastSpec::NUMPY)
```

src/ngraph/op/subtract.hpp
tsocha: Yes

src/ngraph/serializer.cpp
```diff
@@ -2592,9 +2592,20 @@ shared_ptr<Node> JSONDeserializer::deserialize_node(json node_js)
...
+                break;
+            }
+            if (op_version == 1)
+            {
+                node = make_shared<op::v1::Subtract>(
+                    args[0],
+                    args[1],
+                    read_auto_broadcast(node_js, "auto_broadcast", op::AutoBroadcastType::NUMPY));
+                break;
+            }
```
tsocha: Can you differ this serialization by `OP_TYPEID` instead of `op_version`?

mitruska: Sounds good.

src/ngraph/op/subtract.hpp
mitruska: Changed to `NUMPY`.

PR #3958, opened 2019-11-27
[nBench] Add support to dump result tensors

dcaballe:
A new flag '--dump_results' is added to dump result tensors to std::out.


src/tools/nbench/benchmark.cpp
```diff
@@ -26,12 +26,56 @@
...
+    cout << "============================================================================\n";
+    cout << "---- Dumping result tensors \n";
+    cout << "============================================================================\n";
+
+    unsigned i = 0;
+    for (auto& result : results)
+    {
+        cout << "Result tensor #" << i << ": " << result->get_name() << endl;
+        auto element_type = result->get_element_type();
+        switch (element_type)
```
rkimballn1: You should put the pragma that requires all element types are handled in this enum.
boolean is missing
I would also look at supporting bfloat16 and f16. I think you should be able to dump those too.

src/tools/nbench/benchmark.cpp
```diff
@@ -113,6 +157,10 @@ vector<runtime::PerformanceCounter> run_benchmark(shared_ptr<Function> f,
     float time = t1.get_milliseconds();
     cout << time / iterations << "ms per iteration" << endl;
 
+    if (dump_results) {
```
rkimballn1: this won't pass the style check. Run `make style` on your code.

PR #3963, opened 2019-11-28
[ONNX] Add GatherND op to ONNX importer

Ewa21:


Ewa21:
> Please create a matching branch also named `etusien/gatherND` in nGraph-ONNX to unblock the `GatherND` tests from the ONNX suite.
> https://github.com/NervanaSystems/ngraph-onnx/blob/f03ebac2a59b37f497f7927835072512a46dfd36/tests/test_backend.py#L264-L266
> 
> Please also make sure that your tests are not exactly the same as the ones from ONNX. Provide additional test cases instead of making a copy of the same ones.

`GatherND` tests in nGraph-ONNX are now unblocked. And I changed my test cases after checking that they were copies of tests provided by ONNX.

PR #3964, opened 2019-11-28
[SPEC] DeformableConvolution

tomdol:
Please pay extra attention to the `validate_and_infer_types()` method of this op and let me know if I'm doing something wrong there.


src/ngraph/op/deformable_convolution.cpp
```diff
@@ -0,0 +1,177 @@
...
+                               filter_shape,
+                               m_strides,
+                               m_dilations,
+                               m_auto_pad,
+                               m_pads_end,
+                               m_pads_begin);
+        }
+    }
+
+    if (deformable_values_shape.is_static())
```
mbencer: I think that better (less restrictive) is using `deformable_values_shape.rank()` instead of `deformable_values_shape.to_shape().size()`

src/ngraph/op/deformable_convolution.cpp
```diff
@@ -0,0 +1,177 @@
...
+    {
+        NODE_VALIDATION_CHECK(
+            this,
+            deformable_values_shape.to_shape().size() >= 3,
+            "The deformable values tensor rank is expected to be at least 3, got: ",
+            deformable_values_shape.to_shape().size());
```
tsocha: Please use casted `rank()` instead of `to_shape().size()`.

src/ngraph/op/deformable_convolution.cpp
```diff
@@ -0,0 +1,177 @@
...
+                              static_cast<size_t>(data_batch_shape[1]) % m_group == 0,
+                              "The input data shape must be evenly divisible by the 'group' value "
+                              "along the channels axis. Current input shape: ",
+                              data_batch_shape,
+                              ", 'group' attribute value: ",
+                              m_group);
+
+        NODE_VALIDATION_CHECK(
+            this,
+            static_cast<size_t>(filters_shape[0]) % m_group == 0,
```
tsocha: `NON-BLOCK`
Just wonder if we can check only this one dimension for its dynamic(as below).
for example:
PartialShape = `{2, ?, ?, ?}`

test/type_prop/convolution.cpp
```diff
@@ -2862,3 +2862,83 @@ TEST(type_prop, conv_v1_partial_rank)
...
+                                                   1,
+                                                   7);
+
+        FAIL() << "DeformableConvolution created with incorrect 'deformable group' value";
+    }
+    catch (const NodeValidationFailure& error)
+    {
+        EXPECT_HAS_SUBSTRING(error.what(), "deformable values input must be evenly divisible");
+    }
+}
```
tsocha:
```suggestion
}

```

PR #3965, opened 2019-11-28
Add stack fused op for pdpd use

mozga-intel:


src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+shared_ptr<Node> op::Stack::copy_with_new_args(const NodeVector& new_args) const
+{
+    return make_shared<Stack>(new_args, m_axis);
+}
+
+void op::Stack::pre_validate_and_infer_types()
+{
+    for (uint64_t i = 0; i < get_input_size(); ++i)
```
baojun-nervana: use size_t instead of uint64_t

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,81 @@
...
+                              input_element_type,
+                              ").");
+    }
+}
+
+NodeVector op::Stack::decompose_op() const
+{
+    auto axis = get_axis();
+    std::vector<std::shared_ptr<ngraph::Node>> args;
+    for (uint64_t i = 0; i < get_input_size(); ++i)
```
baojun-nervana: use size_t

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,81 @@
...
+    {
+        element::Type input_element_type = get_input_element_type(i);
+
+        NODE_VALIDATION_CHECK(this,
+                              input_element_type.is_dynamic() || input_element_type.is_real(),
+                              "Argument element type must be f16, bf16, f32, f64 or dynamic (got ",
+                              input_element_type,
+                              ").");
+    }
+}
```
baojun-nervana: can you set the output element type? The type is known.
Example: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/fused/gelu.cpp#L79

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,81 @@
...
+            data_shape.insert(data_shape.begin() + axis, 1);
+            std::vector<size_t> input_order(data_shape.size() - 1);
+            std::iota(std::begin(input_order), std::end(input_order), 0);
+            args.push_back(
+                std::make_shared<op::Reshape>(data, AxisVector(input_order), data_shape));
+        }
+    }
+    auto concat = std::make_shared<op::Concat>(args, axis);
+    return {concat};
+}
```
baojun-nervana: add generated_adjoints and set it as "not implemented"?

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    auto axis = get_axis();
+    std::vector<std::shared_ptr<ngraph::Node>> args;
+    for (uint64_t i = 0; i < get_input_size(); ++i)
+    {
+        const PartialShape data_pshape = get_input_partial_shape(i);
+        {
+            auto data = input_value(i);
+            auto data_shape = data.get_shape();
+            axis = (axis < 0) ? axis + data_shape.size() + 1 : axis;
```
baojun-nervana: All input args must have same dimensions in stack op, we may handle this outside of the loop.

src/ngraph/op/fused/stack.hpp
```diff
@@ -0,0 +1,72 @@
...
+                /// \return The stack axis.
+                int64_t get_stack_axis() const { return m_stack_axis; }
+                /// \set The stack axis
+                void set_stack_axis(const int64_t stack_axis) { m_stack_axis = stack_axis; }
+                /// \return The stack axis
+                int64_t get_axis() const { return m_axis; }
+                void set_axis(int64_t axis) { m_axis = axis; }
+            private:
+                int64_t m_axis;
+                int64_t m_stack_axis = -1;
```
baojun-nervana: Not sure if we need this? Negative index is handled in the decompose_op.

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,81 @@
...
+            data_shape.insert(data_shape.begin() + axis, 1);
```
baojun-nervana: we may define another variable here such as stacked_data_shape to avoid this type of reusing a name while it has changed to a different meaning from its name.

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,81 @@
...
+            std::vector<size_t> input_order(data_shape.size() - 1);
+            std::iota(std::begin(input_order), std::end(input_order), 0);
```
baojun-nervana: If data_shape hasn't been reused, you may use get_default_order to get AxisVector.

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,102 @@
...
+    ngraph_error("Not Yet Implemented");
+}
+
+void op::Stack::pre_validate_and_infer_types()
+{
+    for (size_t i = 0; i < get_input_size(); ++i)
+    {
+        element::Type input_element_type = get_input_element_type(i);
+
+        NODE_VALIDATION_CHECK(this,
```
silee2: I don't think we need to restrict element type for this op

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,102 @@
...
+{
+    return make_shared<Stack>(new_args, m_axis);
+}
+
+void op::Stack::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
+{
+    ngraph_error("Not Yet Implemented");
+}
+
+void op::Stack::pre_validate_and_infer_types()
```
silee2: This is optional but it may be better to override validate_and_infer_types() instead of pre_validate_and_infer_types() like here https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/frontend/fluid/operators/reduce_sum.cpp#L74-L85

mozga-intel:
@diyessi done

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,102 @@
...
+        NODE_VALIDATION_CHECK(this,
+                              input_element_type.is_dynamic() || input_element_type.is_real(),
+                              "Argument element type must be f16, bf16, f32, f64 or dynamic (got ",
+                              input_element_type,
+                              ").");
+    }
+    element::Type input_element_type = get_input_element_type(0);
+    PartialShape input_pshape = get_input_partial_shape(0);
+
+    if (input_element_type.is_dynamic())
```
baojun-nervana: Here should be 
if (input_pshape.is_dynamic())
{
      set_output_type(0, input_element_type, PartialShape::dynamic());
}

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,102 @@
...
+    {
+        set_output_type(0, input_element_type, input_pshape);
+    }
+}
+
+NodeVector op::Stack::decompose_op() const
+{
+    auto axis = get_axis();
+    std::vector<std::shared_ptr<ngraph::Node>> args;
+    PartialShape inputs_shape_scheme{PartialShape::dynamic()};
```
baojun-nervana: In decompose_op, the shape becomes static. No need to set the dynamic shape and validate here.
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/fused/layer_norm.cpp#L68

src/ngraph/op/fused/stack.cpp
```diff
@@ -0,0 +1,102 @@
...
+    : Stack(as_output_vector(args), axis)
+{
+}
+
+shared_ptr<Node> op::Stack::copy_with_new_args(const NodeVector& new_args) const
+{
+    return make_shared<Stack>(new_args, m_axis);
+}
+
+void op::Stack::generate_adjoints(autodiff::Adjoints& adjoints, const NodeVector& deltas)
```
baojun-nervana: use outputvector instead https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/op/fused/partial_slice.cpp#L149

PR #3966, opened 2019-11-28
[Spec] Add 3-input constructor to DetectionOutput

postrational:


src/ngraph/op/experimental/layers/detection_output.cpp
```diff
@@ -50,6 +60,21 @@ void op::DetectionOutput::validate_and_infer_types()
...
+    }
+    else
+    {
+        return make_shared<DetectionOutput>(new_args.at(0),
+                                            new_args.at(1),
+                                            new_args.at(2),
+                                            new_args.at(3),
+                                            new_args.at(4),
+                                            m_attrs);
+    }
```
mbencer: You can probably throw exception when `new_args.size()` is not `3` and `5`

PR #3967, opened 2019-11-28
[SPEC] Add GatherTree:v1

mbencer:


src/ngraph/op/gather_tree.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace ngraph
+{
+    namespace op
+    {
+        namespace v1
+        {
+            /// \brief Generates the complete beams from the ids per each step and the parent beam
+            /// ids.
+            class NGRAPH_API GatherTree : public Op
```
GlebKazantaev: AFAIK we don't use `NGRAPH_API` for other ops. So why do we keep it here?

GlebKazantaev: And in other ops it's usually used before type_infor

mbencer: @GlebKazantaev You are right but it was changed after https://github.com/NervanaSystems/ngraph/pull/3767 introduction

PR #3971, opened 2019-11-29
Enable multitple int types for OneHot 'depth' param

postrational:


src/ngraph/op/one_hot.cpp
```diff
@@ -162,7 +162,32 @@ void op::v1::OneHot::validate_and_infer_types()
...
+        NODE_VALIDATION_CHECK(this,
+                              depth_element_type == element::i8 ||
+                                  depth_element_type == element::i32 ||
+                                  depth_element_type == element::i64,
+                              "'depth' input element type must be i8, i32 or i64 (got ",
+                              depth_element_type,
+                              ").");
+
+        NODE_VALIDATION_CHECK(this,
+                              is_scalar(depth->get_shape()),
```
tsocha: The old solution allows us to threat one element vector as a scalar.
It can be worth to keep this functionality.
```suggestion
                              shape_size(depth->get_shape()) == 1),
```

diyessi: They aren't scalars.

PR #3972, opened 2019-11-30
Change generate_adjoints to take an OutputVector

rkimballn1:
Change generate_adjoints to take an OutputVector instead of a NodeVector for deltas. This is the start of a cleanup of adjoints which currently don't work without GetOutputElement.

The only files that changed beyond simply changing the input type are:
src/ngraph/autodiff/adjoints.cpp
src/ngraph/op/dot.cpp
src/ngraph/op/relu.?pp
The rest of the files simply replace NodeVector with OutputVector.


jbobba:
This would break DLDT and other backends as well. Might be good to give them some visibility into the change as well.

diyessi:
Included in #3973

PR #3974, opened 2019-12-01
[SPEC] Select op with implicit broadcast

jbobba:
- [x] Constant Folding
- [x] More type prop tests for error cases


src/ngraph/op/select.cpp
```diff
@@ -25,15 +25,105 @@
...
 }
 
-void op::Select::validate_and_infer_types()
+void op::v1::Select::validate_and_infer_types()
+{
+    // Condition element type check
+    NODE_VALIDATION_CHECK(this,
+                          get_input_element_type(0).is_dynamic() ||
+                              get_input_element_type(0) == element::boolean,
+                          "Argument 0 does not have boolean element type (element type: ",
```
tomdol: Not sure here but `does` or `does not`?

diyessi: `must have`

PR #3976, opened 2019-12-02
Fixed Selu op decomposition

GlebKazantaev:
Due to `*` and `+` operators converts to v0 eltwise ops, they don't support NUMPY broadcast, but Selu op consumes 1D tensors for 2nd and 3rd inputs.


src/ngraph/op/fused/selu.cpp
```diff
@@ -39,12 +39,18 @@ NodeVector op::v0::Selu::decompose_op() const
...
+    return {std::make_shared<op::v1::Multiply>(
+        lambda,
+        std::make_shared<op::v1::Add>(
+            std::make_shared<op::v1::Maximum>(data, zero_node),
+            std::make_shared<op::Subtract>(
+                std::make_shared<op::v1::Multiply>(
+                    alpha,
+                    std::make_shared<op::Exp>(std::make_shared<op::v1::Minimum>(data, zero_node))),
+                alpha,
+                AutoBroadcastSpec::NUMPY)))};
```
postrational: Do you need to specify the `AutoBroadcastSpec::NUMPY` here?

src/ngraph/op/fused/selu.cpp
```diff
@@ -39,12 +39,18 @@ NodeVector op::v0::Selu::decompose_op() const
...
-        data.get_element_type(), data.get_shape(), std::vector<double>{0});
-    return {lambda *
-            (std::make_shared<op::Maximum>(data, zero_node) +
-             alpha * std::make_shared<op::Exp>(std::make_shared<op::Minimum>(data, zero_node)) -
-             alpha)};
+    const auto zero_node = op::Constant::create(data.get_element_type(), Shape{1}, {0});
+
+    return {std::make_shared<op::v1::Multiply>(
+        lambda,
+        std::make_shared<op::v1::Add>(
```
postrational: Please add a comment with the formula above this block, to make it easier to understand what we're going for here.

src/ngraph/op/fused/selu.cpp
```diff
@@ -39,12 +39,18 @@ NodeVector op::v0::Selu::decompose_op() const
     const auto data = input_value(0);
     const auto alpha = input_value(1);
     const auto lambda = input_value(2);
-    const auto zero_node = std::make_shared<ngraph::op::Constant>(
-        data.get_element_type(), data.get_shape(), std::vector<double>{0});
-    return {lambda *
-            (std::make_shared<op::Maximum>(data, zero_node) +
-             alpha * std::make_shared<op::Exp>(std::make_shared<op::Minimum>(data, zero_node)) -
-             alpha)};
+    const auto zero_node = op::Constant::create(data.get_element_type(), Shape{1}, {0});
```
mbencer: not blocking: In this case better is using scalar instead of 1-element 1D vector - `Shape{}`

src/ngraph/op/fused/selu.cpp
GlebKazantaev: V1 Subtract was merge 15 hours ago. I'll switch to V1.

GlebKazantaev: Done.

src/ngraph/op/fused/selu.cpp
GlebKazantaev: Done.

PR #3981, opened 2019-12-03
Make sure examples compile

diyessi:


doc/examples/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@
 if (NGRAPH_CPU_ENABLE)
```
silee2: Any reason why we only build examples if cpu backend is enabled? Interpreter too slow for mnist_mlp?

indie:
Review / fixes on PR #3995

doc/examples/CMakeLists.txt
diyessi: Maybe it was to make sure they built once.

silee2:
Please run `style-apply`

PR #3989, opened 2019-12-03
Don't imbue locale to cout, imbue a stringstream instead so that it d…

rkimballn1:
…oes not contaminate ngraph output

I was seeing outputs like this float{4,096,768} which is clearly wrong. The correct output is float{4096,768} but the comma was added in 4096 because nbench had imbued locale into cout.


diyessi:
It would work just as well to add the missing space after the comma, so that it would be `float{4,096, 768}`

PR #3994, opened 2019-12-04
Update ONNX importer to use nGraph ops from new opset header

mitruska:
Include opset headers instead of operators explicitly.


src/ngraph/opsets/opset1.hpp
```diff
@@ -16,7 +16,7 @@
 
 #pragma once
 
-#include "ngraph/ngraph.hpp"
+#include "ngraph/ops.hpp"
```
tsocha: Is this line really needed here?
Can we move this line into `ngraph/opsets/opset1_tbl.hpp`?

src/ngraph/opsets/opset1_tbl.hpp
```diff
@@ -53,6 +53,7 @@
 NGRAPH_OP(Abs, ngraph::op::v0)
 NGRAPH_OP(Acos, ngraph::op::v0)
 NGRAPH_OP(Add, ngraph::op::v1)
+NGRAPH_OP(ArgMax, ngraph::op::v0) // ArgMax is not in opset1 spec
```
tsocha: If `ArgMax` is not in the opset1 spec it should be added to it.
We can't add random ops to opsets.

src/ngraph/frontend/onnx_import/op/argmax.cpp
```diff
@@ -14,9 +14,8 @@
 // limitations under the License.
 //*****************************************************************************
 
-#include "ngraph/op/argmax.hpp"
-#include "core/node.hpp"
-#include "ngraph/node.hpp"
+#include "argmax.hpp"
+#include "default_opset.hpp"
```
tsocha: Since `ArgMax` is not a part of the opset1 we should use opset0 here.

src/ngraph/opsets/opset1.hpp
mitruska: I think this PR [#3985](https://github.com/NervanaSystems/ngraph/pull/3985) fixed it.

src/ngraph/frontend/onnx_import/op/add.cpp
```diff
@@ -0,0 +1,57 @@
...
+                NodeVector add(const Node& node)
+                {
+                    auto left_rank = node.get_ng_inputs().at(0)->get_shape().size();
+                    auto right_rank = node.get_ng_inputs().at(1)->get_shape().size();
+                    auto axis =
+                        node.get_attribute_value<std::int64_t>("axis", left_rank - right_rank);
+                    NodeVector ng_inputs{ngraph::op::legacy_style_broadcast_for_binary_operation(
+                        node.get_ng_inputs().at(0), node.get_ng_inputs().at(1), axis)};
+
+                    return {std::make_shared<Add>(ng_inputs.at(0), ng_inputs.at(1))};
```
postrational: You should be using `v0` `Add` in this case.

tsocha: Since input data is already broadcasted there is no difference between V0 and V1.
We should leave v1 here IMO.

postrational: OK, you may be right in this case. Let's see if tests pass.
But in general, this story is not about making too many changes in the opset we use, but rather how we include ops.

src/ngraph/opsets/opset0.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+namespace ngraph
+{
+    namespace opset0
+    {
+#define NGRAPH_OP(a, b) using b::a;
+#include "ngraph/opsets/opset0_tbl.hpp"
+#undef NGRAPH_OP
+    }
+}
```
mitruska: This is one of my idea, I created `opset0.hpp` file like for opset1 to enable include operators from opset0.

src/ngraph/frontend/onnx_import/op/sum.hpp
```diff
@@ -40,7 +41,7 @@ namespace ngraph
             {
                 inline NodeVector sum(const Node& node)
                 {
-                    return variadic::make_ng_variadic_op<ngraph::op::v1::Add>(node);
+                    return variadic::make_ng_variadic_op<ngraph::opset1::Add>(node);
```
mitruska: We can use operators from different opsets within different namespaces.

src/ngraph/frontend/onnx_import/op/add.cpp
mitruska: I understand, I made some new changes with opset includes. Take a look.

src/ngraph/opsets/opset1_tbl.hpp
mitruska: I just wanted to show that, this is why I left this comment here. Removed.

src/ngraph/frontend/onnx_import/op/argmax.cpp
mitruska: I've already created `opset0.hpp` file to enable including different opsets same way.
Let's talk about it and other possible ways.

src/ngraph/frontend/onnx_import/op/abs.hpp
```diff
@@ -19,6 +19,7 @@
 #include <memory>
 
 #include "core/node.hpp"
+#include "default_opset.hpp"
```
mitruska: While *.hpp file should't contain `using namespace` we can create *.cpp file for each operator or include `ngraph/opsets/opset1.hpp` and use `ngraph::opset1` explicit instead of `default_opset.hpp`.

src/ngraph/frontend/onnx_import/op/abs.hpp
```diff
@@ -32,7 +33,7 @@ namespace ngraph
             {
                 inline NodeVector abs(const Node& node)
                 {
-                    return {std::make_shared<ngraph::op::Abs>(node.get_ng_inputs().at(0))};
+                    return {std::make_shared<Abs>(node.get_ng_inputs().at(0))};
```
mitruska: Opset1 includes `v0::Abs` so we can use `default_opset.hpp` here.
But we have to decide if `default_opset.hpp` is a good idea or rather explicit include `ngraph/opsets/opset0.hpp`/`ngraph/opsets/opset1.hpp` and use their namespaces explicit before ops.

src/ngraph/frontend/onnx_import/op/acos.hpp
```diff
@@ -32,7 +33,7 @@ namespace ngraph
             {
                 inline NodeVector acos(const Node& node)
                 {
-                    return {std::make_shared<ngraph::op::Acos>(node.get_ng_inputs().at(0))};
+                    return {std::make_shared<ngraph::opset1::Acos>(node.get_ng_inputs().at(0))};
```
mitruska: This is another approach (without `default_opset.hpp` file), include `ngraph/opsets/opset1.hpp` and use `ngraph::opset1` namespace.

src/ngraph/frontend/onnx_import/op/acosh.cpp
```diff
@@ -21,6 +21,7 @@
 #include "ngraph/op/multiply.hpp"
 #include "ngraph/op/sqrt.hpp"
 #include "ngraph/op/subtract.hpp"
+#include "ngraph/opsets/opset0.hpp"
```
mitruska: I created this file to enable include opset0 like opset1.

src/ngraph/frontend/onnx_import/op/add.cpp
```diff
@@ -0,0 +1,59 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include "add.hpp"
+#include "default_opset.hpp"
```
mitruska: This `add.cpp` file was created to avoid include `default_opset.hpp` in `add.hpp` file,
(default_opset file contain `using namespace` statement).
It should be done for all ops with only `*.hpp` files, if we want to use `default_opset.hpp` idea.

src/ngraph/frontend/onnx_import/op/sub.hpp
```diff
@@ -48,8 +49,8 @@ namespace ngraph
             {
                 inline NodeVector sub(const Node& node)
                 {
-                    return {std::make_shared<ngraph::op::v1::Subtract>(node.get_ng_inputs().at(0),
-                                                                       node.get_ng_inputs().at(1))};
+                    return {std::make_shared<Subtract>(node.get_ng_inputs().at(0),
```
mitruska: This is example how we can use different opsets within one file.  Opset1 is included as `default_opset.hpp`.

src/ngraph/frontend/onnx_import/op/sum.hpp
```diff
@@ -18,7 +18,8 @@
 
 #include "core/node.hpp"
 #include "ngraph/node.hpp"
-#include "ngraph/op/add.hpp"
+#include "ngraph/opsets/opset0.hpp"
+#include "ngraph/opsets/opset1.hpp"
```
mitruska: This is example how we can use different opsets within one file. 
Rather by explicit opsets includes than `default_opset.hpp`.

postrational:
I think in the end we should be able to use both approaches, combined.

Perhaps the best way to write code it is to explicitly state where each operation is coming from.
Either: `ngraph::opset0::Add`, `ngraph::opset1::Add` or `onnx_importer::default_opset::Add`.

In a file which uses all defaults we could include `using namespace ngraph::default_opset`. 

@jbobba @diyessi What do you think?

tsocha:
> @postrational 
> 
> I think in the end we should be able to use both approaches, combined.
> 
> Perhaps the best way to write code it is to explicitly state where each operation is coming from.
> Either: `ngraph::opset0::Add`, `ngraph::opset1::Add` or `onnx_importer::default_opset::Add`.
> 
> In a file which uses all defaults we could include `using namespace onnx_importer::default_opset`.
> 
> @jbobba @diyessi What do you think?

In my opinion `default_opset` can cause some shady problems when it will change from opset1 to opset2.
Some of the operations would have the same signature but different default parameters etc.

mitruska:
> I think in the end we should be able to use both approaches, combined.
> 
> Perhaps the best way to write code it is to explicitly state where each operation is coming from.
> Either: `ngraph::opset0::Add`, `ngraph::opset1::Add` or `onnx_importer::default_opset::Add`.
> 
> In a file which uses all defaults we could include `using namespace ngraph::default_opset`.

So what about creating `onnx_importer::default_opset` alias for `ngraph::opset1` namespace. We will be able to change it with opset upgrade on `ngraph::opset2`,  `ngraph::opset3` etc. in the `default_opset.hpp` and then write explicitly older opset namespace only there, where ops version in the new opset will be incompatible.

In default_opset.hpp file, something like this:
```
#include "ngraph/opsets/opset1.hpp"
namespace ngraph::default_opset = ngraph::opset1;
```
Instead of this:
```
#include "ngraph/opsets/opset1.hpp"
using namespace ngraph::opset1;
```

src/ngraph/frontend/onnx_import/default_opset.hpp
```diff
@@ -0,0 +1,6 @@
+#include "ngraph/opsets/opset1.hpp"
+
+namespace ngraph
```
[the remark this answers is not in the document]
mitruska: Ok, I can change it. But do we still want to use in code full namespace `ngraph::onnx_import::default_opset::Add` or rather `default_opset::Add` because we actually are in  `ngraph::onnx_import` namespace ?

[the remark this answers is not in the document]
postrational: I think we should use the short form `default_opset::Add` wherever possible.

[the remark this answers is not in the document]
mitruska: Done

PR #3995, opened 2019-12-04
Leona/doc examples

indie:
Tested a build of the doc and saw opset versioning broke a few more ops, so those are updated as well.


indie:
This is not the main PR; should be merged after https://github.com/NervanaSystems/ngraph/pull/3981/

PR #3998, opened 2019-12-04
Bug fix for reshape swimming [NGCORE-735]

csullivan:
Reshapes currently are propagated past ops which have multiple users without materializing inverse reshapes along the other output edges. Avoid this for now by restricting swimming to only ops that have a single user.

I didn't try compiling it @ashokei, but this is my guess as to whats going wrong for NNP-I in NGCORE-735. Give it a shot.


src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -175,6 +175,20 @@ void swim(Input<Node> input, shared_ptr<op::Reshape> reshape)
...
+
+        auto materialize = [csw, n]() {
+            auto new_reshape = csw.reshape->copy_with_new_args({n});
+            new_reshape->merge_provenance_tags_from(n);
+            NGRAPH_DEBUG << "Materializing new reshape " << describe_reshape(new_reshape);
+            csw.input.replace_source_output(new_reshape->output(0));
+        };
+
+        // Only swim past nodes which have a single user
+        if (n->get_users(true) > 1)
```
diyessi: `== 1`?

csullivan: I had it this way with a big conditional around the swimming if/elsif block but it made the github diff huge and hid the change. It can be productized however if @ashokei confirms that it helps.

ashokei: build issue... `n->get_users(true).size()` ? 

seems to fix the issue though reshape cant be avoided, backend can optimize further.

csullivan:
```suggestion
        if (n->get_users(true).size() > 1)
```

src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -175,6 +175,20 @@ void swim(Input<Node> input, shared_ptr<op::Reshape> reshape)
...
+
+        auto materialize = [csw, n]() {
+            auto new_reshape = csw.reshape->copy_with_new_args({n});
+            new_reshape->merge_provenance_tags_from(n);
+            NGRAPH_DEBUG << "Materializing new reshape " << describe_reshape(new_reshape);
+            csw.input.replace_source_output(new_reshape->output(0));
+        };
+
+        // Only swim past nodes which have a single user
+        if (n->get_users(true).size() > 1)
```
jiminha: would this get_users(true) will increase the compile time? 
I remember we had an issue sometime back on slice reshape sinking, this caused significant time increase on reshape sinking, so we had to change to just get_users().

csullivan:
@ashokei, great glad its working. And right, this fixes a bug in swimming. In the graph you showed me you ended up with two back to back reshapes of the same input order which is a correctness issue. 

This change will prevent the Reshape from swimming past the Relu but it won't remove the swam reshape. If you run CommonSubExpressionElimination after ReshapeSinking it should merge the two Reshapes on the output edges of the Relu in the graph you showed me. Then there will only be a single reshape.

src/ngraph/pass/reshape_sinking.cpp
csullivan: It could if there are many ops which have long dangling output chains that never make it to a result -- since get_users(true) is making sure the users it counts are used by Results (calling ngraph::is_used).

I suspect that is unlikely but if you have evidence in the past to the contrary then it's worth checking the compile time.

jiminha: We will run all CI tests first with all our models to make sure that no regression. Will update.

PR #4000, opened 2019-12-05
Helper in Constant to allow casting values to a different type

postrational:
Simplify logic needed to extract values from a Constant node, when
the expected data type is specified only as integral or floating point.


src/ngraph/op/constant.hpp
```diff
@@ -240,6 +240,87 @@ namespace ngraph
...
+                {
+                    auto vector = get_vector<uint32_t>();
+                    return std::vector<T>(vector.begin(), vector.end());
+                }
+                case element::Type_t::u64:
+                {
+                    auto vector = get_vector<uint64_t>();
+                    return std::vector<T>(vector.begin(), vector.end());
+                }
+                case element::Type_t::u1: throw std::runtime_error("unsupported type");
```
tomdol: A single throw below `case element::Type_t::dynamic:` should be enough here for all 3 cases.

src/ngraph/op/one_hot.cpp
```diff
@@ -165,9 +165,7 @@ void op::v1::OneHot::validate_and_infer_types()
 
         auto depth_element_type = depth->get_output_element_type(0);
         NODE_VALIDATION_CHECK(this,
-                              depth_element_type == element::i8 ||
-                                  depth_element_type == element::i32 ||
-                                  depth_element_type == element::i64,
+                              depth_element_type.is_integral(),
                               "'depth' input element type must be i8, i32 or i64 (got ",
```
tsocha:
```suggestion
                              "'depth' input element type must be integral type (got ",
```

PR #4001, opened 2019-12-05
[SPEC] Calculate output shape for NonMaxSuppression

mbencer:


src/ngraph/op/non_max_suppression.cpp
```diff
@@ -126,9 +126,60 @@ void op::v1::NonMaxSuppression::validate_and_infer_types()
 
     // NonMaxSuppression produces triplets
     // that have the following format: [batch_index, class_index, box_index]
-    // The number of returned triplets depends entirely on the computation, thus one dynamic dim
-    const PartialShape out_shape = {Dimension::dynamic(), 3};
-
+    PartialShape out_shape = {Dimension::dynamic(), 3};
+
+    const auto max_output_boxes_per_class = input_value(2).get_node_shared_ptr();
+    if (num_boxes_boxes.is_static() && scores_ps[1].is_static() &&
```
tsocha: What will happen if `scores_ps` would have dynamic rank?
I'm afraid that `scores_ps[1]` will throw an exception, but I'm not 100% sure.

PR #4013, opened 2019-12-06
Change Input<Node> to NodeInput and Output<Node> to NodeOutput

rkimballn1:
Windows can't export template classes so this is my answer to that problem, make Input and Output non-template classes. All of the changes are simple search and replace except for
src/ngraph/node.hpp
src/ngraph/node.cpp
I left function definitions in the node.hpp header file even though they really should be moved to the cpp file now that they are not templated but we can do that in a later PR.


src/ngraph/node.hpp
```diff
@@ -531,32 +526,21 @@ namespace ngraph
...
-        Input(Node* node, size_t index)
+        NodeInput(const Node* node, size_t index)
             : m_node(node)
             , m_index(index)
         {
         }
 
         /// \return A pointer to the node referenced by this input handle.
-        Node* get_node() const { return m_node; }
+        Node* get_node() const { return const_cast<Node*>(m_node); }
```
diyessi: `NodeInput(const_node).get_node()` loses constness.

src/ngraph/node.hpp
```diff
@@ -596,139 +580,65 @@ namespace ngraph
...
+    class NGRAPH_API NodeOutput
     {
     public:
-        /// \brief Constructs a Output.
+        /// \brief Constructs a NodeOutput.
         /// \param node A pointer to the node for the output handle.
         /// \param index The index of the output.
-        Output(Node* node, size_t index)
-            : m_node(node->shared_from_this())
+        NodeOutput(const Node* node, size_t index)
```
diyessi: `const_node.output(0).get_node()` loses constness.

src/ngraph/node.hpp
rkimballn1: Yes, but in order to keep the original API this is what I have to do. If you want to change the API more I would recommend we do a separate PR. Changing to const breaks other things.

src/ngraph/node.hpp
rkimballn1: Yes, but it does exactly what it needs to do. That is the intent of the const_cast. This is what we need to make this work without the ugly templates

PR #4020, opened 2019-12-07
Node class should not have type info since it is not a type

rkimballn1:
This PR prevents the use of the incorrect `n.type_info.name` if n is a Node. By making Node have type info it allowed a bug in the serialize to be introduced. With this change it is not possible to mistakenly get node.type_info.


diyessi:
Check #4043

diyessi:
Incorporate in type info refactoring.

PR #4035, opened 2019-12-10
Add upgrade and downgrade pass for GroupConvolutionBackpropData ops

tsocha:
- Add up/downgrade passes for GroupConvolutionBackpropData operators
- Improve decompose operatorion of v0::GroupConvolutionBackpropData to support N-dimensional data
- Add UT for up/downgrade passes.


postrational:
Since this change only deals with `v0` ops, @tsocha removed some reviewers.

PR #4038, opened 2019-12-10
TensorIterator: reshape support

itikhono:
The "copy_with_new_args" function in the TensorIterator layer is fixed: added copying of BodyLambda via "specialize_function". It's necessary to support the Reshape feature.
Added temporary workaround to call the "validate_and_infer_types" function of the TensorIterator layer. This function was not called in the constructor due to the implementation of the TensorIterator layer (body and input/outputs are not set in the moment of call constructor). So this leads to a bug: in some cases, the shapes are not propagated (e.g. in case of the "specialize_function" call for networks with TensorIterator).
Added temporary workaround for deleting GetOutputElements nodes after call "specialize_function" in TensorIterator (known issue: https://jira.devtools.intel.com/browse/CVS-24517)


postrational:
TESTNOW

PR #4039, opened 2019-12-11
Add constant folding for v1 Split and VariadicSplit.

ayzhuang:


test/constant_folding.cpp
```diff
@@ -1866,6 +1866,163 @@ TEST(constant_folding, constant_v1_select)
...
+    ASSERT_EQ(count_ops_of_type<op::v1::Split>(f), 0);
+    ASSERT_EQ(count_ops_of_type<op::Constant>(f), num_splits);
+
+    auto new_const = as_type_ptr<op::Constant>(f->get_results().at(0)->get_argument(0));
+    ASSERT_TRUE(new_const);
+    auto values_out = new_const->get_vector<int>();
+
+    vector<int> sliced_values{0, 1, 2, 3, 16, 17, 18, 19, 32, 33, 34, 35, 48, 49, 50, 51};
+    ASSERT_EQ(sliced_values, values_out);
+}
```
gaurides: can we check values of all new constants? to make sure the split worked correctly? 
Also, in following tests.

mbencer:
I closed it because target solution is implemented in https://github.com/NervanaSystems/ngraph/pull/4041

PR #4049, opened 2019-12-12
[MLIR]Add clang flag and skip bf16 tests for MLIR.

ayzhuang:


cmake/external_llvm.cmake
```diff
@@ -91,6 +91,7 @@ ExternalProject_Add(
                 -DLIBOMP_OMPT_SUPPORT=OFF
                 -DCLANG_BUILD_TOOLS=OFF
                 -DCLANG_ENABLE_ARCMT=OFF
+                -DCLANG_ANALYZER_ENABLE_Z3_SOLVER=OFF
```
silee2: Why is this change related to MLIR? This LLVM is the one used for CPU codegen.

ayzhuang: For CI we build LLVM only when MLIR is enabled. This flag is needed when building LLVM8 with clang9 or clang10.

test/cpu_test.cpp
```diff
@@ -2157,10 +2157,10 @@ TEST(cpu_test, tensor_copy_from_different_layout)
 
 TEST(cpu_test, max_pool_bf16)
 {
-    if (!runtime::cpu::mkldnn_utils::is_bf16_supported())
+    if (!runtime::cpu::mkldnn_utils::is_bf16_supported() || std::getenv("NGRAPH_MLIR") != nullptr)
```
diyessi: Why not just compile in the disabling of the test?

ayzhuang: You mean to put #ifndef outside to not compile this test when MLIR is enabled?

PR #4051, opened 2019-12-13
Update MKLDNN to v1.1.1.

ayzhuang:


src/ngraph/runtime/cpu/mkldnn_utils.cpp
```diff
@@ -1755,8 +1755,10 @@ bool runtime::cpu::mkldnn_utils::is_bf16_supported()
     }
     catch (const mkldnn::error& e)
     {
+        std::cout << "bf16 is not supported: " << e.message << std::endl;
```
jbobba: use NGRAPH_ERR

src/ngraph/runtime/cpu/mkldnn_utils.cpp
```diff
@@ -1755,8 +1755,10 @@ bool runtime::cpu::mkldnn_utils::is_bf16_supported()
...
         return false;
     }
+    std::cout << "bf is supported\n";
```
jbobba: remove.

cmake/external_mkldnn_v1.cmake
```diff
@@ -256,6 +256,7 @@ else()
         BINARY_DIR "${EXTERNAL_PROJECTS_ROOT}/mkldnn/build"
         INSTALL_DIR "${EXTERNAL_PROJECTS_ROOT}/mkldnn"
         EXCLUDE_FROM_ALL TRUE
+        BUILD_ALWAYS 1
```
jbobba: why are we adding this?

src/ngraph/runtime/cpu/mkldnn_utils.cpp
ayzhuang: I will revert this commit. This is for debugging mac.

PR #4058, opened 2019-12-13
    Support Different Trace Files

HenryTangIntel:
Support different trace files based on process id of each process.  So that for multiple chip case, each chip/process gets its own distinct trace file.


omrialmog:
What is the reason for using pid vs rank? We lose rank naming with this update which would make it harder to understand what rank this trace came from no? How do you easily map pid to ranks?

HenryTangIntel:
> What is the reason for using pid vs rank? We lose rank naming with this update which would make it harder to understand what rank this trace came from no? How do you easily map pid to ranks?

The rank happens to be a MPI related term, and we may not going to use MPI in the further, that's why we use pid number instead of rank number to mark the trace files.

omrialmog:
> > What is the reason for using pid vs rank? We lose rank naming with this update which would make it harder to understand what rank this trace came from no? How do you easily map pid to ranks?
> 
> The rank happens to be a MPI related term, and we may not going to use MPI in the further, that's why we use pid number instead of rank number to mark the trace files.

We can still keep the concept of rank or nodeID, as it is easier to understand than pid, even when not using MPI. nnp-transformer/ngraph still has rank, even though the communication interface might not be MPI.

My opinion is you can add PID, but should still keep rank or NodeID identifier to make logs easier to understand with respect to which card is used. Having PID, then trying to map it to nodeID is not easy when looking at large numbers of cards.

HenryTangIntel:
> > > What is the reason for using pid vs rank? We lose rank naming with this update which would make it harder to understand what rank this trace came from no? How do you easily map pid to ranks?
> > 
> > 
> > The rank happens to be a MPI related term, and we may not going to use MPI in the further, that's why we use pid number instead of rank number to mark the trace files.
> 
> We can still keep the concept of rank or nodeID, as it is easier to understand than pid, even when not using MPI. nnp-transformer/ngraph still has rank, even though the communication interface might not be MPI.
> 
> My opinion is you can add PID, but should still keep rank or NodeID identifier to make logs easier to understand with respect to which card is used. Having PID, then trying to map it to nodeID is not easy when looking at large numbers of cards.

Let's assume that we don't want to use MPI, the issue here is we don't know exactly what the nodeID is if we don't parse the nnp_network.json file from nnp_transformer.

PR #4062, opened 2019-12-14
Enable self-concat to broadcast replacement for a single concat

csullivan:
Ayan's original PR avoided doing the replacement for self-concats when they existed in isolation -- without a chain of concats. This PR enables the Concat to Broadcast replacement. It avoids a currently unsupported concat on NNP in TF/BERT.


csullivan:
Closed as subsumed into r0.28.

PR #4074, opened 2019-12-17
 - emit mlkddn for LRN only for normalizing acorss/within channel

pruthviIntel:


jbobba:
Please change the LRN builder as well.

pruthviIntel:
@jbobba changed it.

PR #4076, opened 2019-12-17
[SPEC] Add using Reshape:v1

mbencer:
- `onnx_importer` should produce Reshape:v1
- Some of fused ops such as `DepthToSpace` and `SpaceToDepth` should produce `v0` ops

`Reshape:v1` doesn't support axes reordering. To this purpose `Transpose` can be used, but it is not fully supported by ngaph backends.
As solution `Transpose` was changed to `v1` and it is downgraded to `Reshape:v0`


mbencer:
> Changing existing builders is likely to break some things.
> It would be better to use fused ops that have the functionality you want, and use those fused ops in ONNX as it adjusts to V1. Since these fused ops would be versioned and associated with v1 but not part of v1, maybe they go in something like v1.fe (front end).

@diyessi For me introduction `Reshape` FusedOp can be confused (In current state we have: `Reshape:v0`, `Reshape:v1` and `DynReshape`). What's more, for for FusedOp order of passes can be problematic.
I've introduced `builder::v1` for these cases. Is it acceptable?

mbencer:
@diyessi @jbobba @postrational 
After internal discussion we are not sure if using namespace `v1` (`builder::v1`) is good idea (it can be confused with ops versioning).
Possible solutions:
1. Using names with `_v1` prefix for builders which produces v1 (such as  `builder::reshape_v1`).
2. Moving builders which produces `v0` ops to `v0` namespaces (to consistency)
3. Keeping previous version: `builder::reshape` (and others when needed) produce `v1` without others actions
4. Keeping current version with `v1` namespaces for builders which produce `v1` ops

What do you think about it?

postrational:
@diyessi 
> Changing existing builders is likely to break some things.

Agreed, but I think these specific builders were added by us when we moved code from ONNX importer to fused ops implementations. I would be very surprised if they are already used in another context.

test/opset_pass/transpose_opset_pass.cpp
```diff
@@ -0,0 +1,108 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
rkimballn1:
```suggestion
// Copyright 2017-2020 Intel Corporation
```

jbobba:
> @diyessi
> 
> > Changing existing builders is likely to break some things.
> 
> Agreed, but I think these specific builders were added by us when we moved code from ONNX importer to fused ops implementations. I would be very surprised if they are already used in another context.

I would leave the existing ones and probably add a deprecated comment for those. I would not use the v0/v1 namespace for builders. 'opset' namespace makes more sense. so opset1::builder::build_blah() will only use ops from opset1. Alternatively, we could go with build_blah_opset1() to keep things simple.

mbencer:
> > @diyessi
> > > Changing existing builders is likely to break some things.
> > 
> > 
> > Agreed, but I think these specific builders were added by us when we moved code from ONNX importer to fused ops implementations. I would be very surprised if they are already used in another context.
> 
> I would leave the existing ones and probably add a deprecated comment for those. I would not use the v0/v1 namespace for builders. 'opset' namespace makes more sense. so opset1::builder::build_blah() will only use ops from opset1. Alternatively, we could go with build_blah_opset1() to keep things simple.

For me `opset1` version is consistent and flexible. @jbobba thanks for suggestion. I implemented it.

src/ngraph/builder/reshape.cpp
```diff
@@ -173,3 +174,56 @@ shared_ptr<Node> builder::expand_dims(const Output<Node>& value, size_t axis)
                value, get_default_order(value.get_shape().size()), output_shape)
         ->add_provenance_group_members_above({value});
 }
+
+shared_ptr<Node> builder::opset1::reshape(const Output<Node>& value, const Shape& shape)
+{
+    const auto out_pattern = op::Constant::create(
+        element::i64, Shape{shape.size()}, vector<int64_t>(shape.begin(), shape.end()));
+    const bool special_zero = false;
+    return make_shared<op::v1::Reshape>(value, out_pattern, special_zero)
```
jbobba: can we use make_shared<opset1::Reshape>(...) instead of directly calling out the version.

src/ngraph/builder/reshape.cpp
```diff
@@ -173,3 +174,56 @@ shared_ptr<Node> builder::expand_dims(const Output<Node>& value, size_t axis)
...
+        ->add_provenance_group_members_above({value});
+}
+
+shared_ptr<Node> builder::opset1::reorder_axes(const Output<Node>& value, vector<size_t> axes_order)
+{
+    const auto axes_order_const =
+        op::Constant::create(element::i64,
+                             Shape{axes_order.size()},
+                             vector<int64_t>(axes_order.begin(), axes_order.end()));
+    return make_shared<op::v1::Transpose>(value, axes_order_const)
```
jbobba: opset1::Transpose.

src/ngraph/opsets/opset1_tbl.hpp
```diff
@@ -156,7 +156,7 @@ NGRAPH_OP(Tanh, ngraph::op::v0)
 NGRAPH_OP(TensorIterator, ngraph::op::v0)
 NGRAPH_OP(Tile, ngraph::op::v0)
 NGRAPH_OP(TopK, ngraph::op::v1)
-NGRAPH_OP(Transpose, ngraph::op::v0)
+NGRAPH_OP(Transpose, ngraph::op::v1)
```
jbobba: why are we switching to ngraph::op::v1? Opset1_tbl.hpp should be pretty much locked down now.

src/ngraph/builder/reshape.cpp
mbencer: done

src/ngraph/builder/reshape.cpp
mbencer: done

src/ngraph/opsets/opset1_tbl.hpp
mbencer: `Transpose` before this PR has no implementation in nGraph (experimental op), but it exist in opset1 specification.
`Reshape:v1` doesn't support axes reorder which is required in onnx importer.
In order to provide axes reordering I set `Transpose` as `v1`. During downgrade pass it is converted to `Reshape:v0`

PR #4084, opened 2019-12-18
[ONNX] Update Reduction ops to use v1 operators.

arogowie-intel:
Additionally add upgrade/downgrade pass for `ReductionMax/Min` operators.


src/ngraph/frontend/onnx_import/op/reduce.cpp
```diff
@@ -45,17 +45,19 @@ namespace ngraph
...
 
                     auto const_node = default_opset::Constant::create(
                         sum_node->get_element_type(),
                         sum_node->get_shape(),
                         std::vector<std::size_t>(shape_size(sum_node->get_shape()),
                                                  elem_count_product));
 
-                    return {std::make_shared<ngraph::opset0::Divide>(sum_node, const_node)};
+                    return {std::make_shared<default_opset::Divide>(
+                        sum_node, const_node, ngraph::op::AutoBroadcastSpec())};
```
tomdol: Just a thought: `sum_node` and `const_node` already match in terms of a shape, so I think we could ommit the last param here.

src/ngraph/frontend/onnx_import/op/reduce.hpp
```diff
@@ -257,13 +258,14 @@ namespace ngraph
                 inline NodeVector reduce_sum_square(const Node& node)
                 {
                     auto input = std::shared_ptr<ngraph::Node>{node.get_ng_inputs().at(0)};
-                    auto square_node = input * input;
+                    auto square_node = std::make_shared<ngraph::op::v1::Multiply>(input, input);
```
mitruska: To be consistent you should use `default_opset` namespace for `Multiply` here.

PR #4088, opened 2019-12-19
Pass _GLIBCXX_USE_CXX11_ABI to protobuf

dkalinowski:
This change allows building nGraph with newer GLIBC: v5.1+


dkalinowski:
TESTNOW

PR #4093, opened 2019-12-20
[WIP] Introduce optional MemRefType lowering

dcaballe:
This PR introduces an alternative Std->LLVM lowering for MemRefType. This
lowering converts MemRefType to a plain LLVM pointer instead of to the
struct descriptor that is used by default. This alternative lowering can
be enabled with an experimental flag.

### DIFF
--- a/src/contrib/mlir/runtime/cpu/cpu_runtime.cpp
+++ b/src/contrib/mlir/runtime/cpu/cpu_runtime.cpp
@@ -53,6 +53,11 @@ static llvm::cl::opt<std::string>
     clObjectFilename("ngraph-mlir-object-filename",
                      llvm::cl::desc("Dump MLIR JITted-compiled object to file jitted_mlir.o"));
 
+// *** Experimental flags ***
+
+// Enable a custom memref lowering to LLVM
+extern llvm::cl::opt<bool> clEnableCustomMemRefLowering;
+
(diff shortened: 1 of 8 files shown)


src/contrib/mlir/runtime/cpu/cpu_runtime.cpp
```diff
@@ -102,9 +107,19 @@ void MLIRCPURuntime::bindArguments(std::vector<void*>& externalTensors)
...
+        {
+            // Default memref lowering lowers memrefs to StaticMemRef descriptors.
+            auto* memRefArg = *(reinterpret_cast<StaticMemRef**>(m_invokeArgs[i]));
+            memRefArg->allocatedPtr = (*m_externalTensors)[i];
+            memRefArg->alignedPtr = (*m_externalTensors)[i];
+        }
+        else
+        {
+            // Custom memref lowering lowers memrefs to plain pointers to tensors.
+            auto** memRefArg = reinterpret_cast<void**>(m_invokeArgs[i]);
```
ayzhuang: With custom memref lowering we still need to use struct to pass the layout information. It's needed for callbacks. https://github.com/NervanaSystems/ngraph/blob/171f5d4dcf5a57043f6ef7c21b62d3ad359385fc/src/contrib/mlir/runtime/cpu/cpu_runtime.cpp#L122

PR #4103, opened 2019-12-21
[MLIR] Add some sane utilities for environment variable handling

rkimballn1:
env_var.?pp files are from the nnp-transformer


environment_variables.md
```diff
@@ -0,0 +1,47 @@
...
+| NGRAPH_ENABLE_SERIALIZE_TRACING | |
+| NGRAPH_ENABLE_TRACING | |
+| NGRAPH_ENABLE_VISUALIZE_TRACING | |
+| NGRAPH_FAIL_MATCH_AT | |
+| NGRAPH_GRAPH_REWRITE_RERUN_DYNAMIC_CHECK | |
+| NGRAPH_GTEST_INFO | |
+| NGRAPH_INTER_OP_PARALLELISM | |
+| NGRAPH_INTRA_OP_PARALLELISM | |
+| NGRAPH_MLIR | |
+| NGRAPH_MLIR_MAX_CYCLE_DEPTH | |
```
nmostafa: `NGRAPH_MLIR_MAX_CYCLE_DEPTH ` is not used anymore. I will clean up sub-graph extraction code. Also, you missed `NGRAPH_MLIR_OPTIONS`

rkimballn1: :( There are so many env vars

rkimballn1:
@diyessi Yes it could maintain a cache but that is beyond the scope of the PR. Here I am only trying to normalize access to environment variables which were previously read and interpreted in a very haphazard way. As an example some environment variables only checked for the variables existence to be true, even if the user set that variable to false. This leads to user confusion.
In the NNP transformer we have another file which creates a static cache of each environment variable used so the first time it is accessed it is read and all subsequent reads use the cached value, but as I said above, that is beyond the scope of this PR. I want to keep things small and focused. This PR is no worse in caching efficiency than the existing code it replaces.

nmostafa:
Added [MLIR] to title to enable MLIR CI.

PR #4108, opened 2019-12-23
Add provenance tags to ONNX importer

mitruska:
Provenance tags can be enabled by set environment varviable `NGRAPH_PROVENANCE_ENABLE=true`.
Then list of provenance tags is serialized to json.

Example of serialize_onnx tool usage:
`./build/src/tools/serialize_onnx/serialize_onnx -i ./test/models/onnx/add_abc.prototxt -o add.json`

> This PR introduces:
> 
> * [x]   ONNX importer's output node has provenance tags
> * [x]   ONNX importer's intermediate nodes have provenance tags
> 
> These functionalities will be imlemented in the next PR:
> 
> * [ ]   Input Parameter nodes have provenance tags
> * [ ]   Input Constant nodes have provenance tags
> * [ ]   Instead of using ONNX Node's name (which is rarely used), use output name


src/ngraph/frontend/onnx_import/core/graph.hpp
```diff
@@ -46,9 +46,19 @@ namespace ngraph
             }
 
             const std::string& get_name() const { return m_graph_proto->name(); }
-            NodeVector make_ng_nodes(const Node& node) const
+            NodeVector make_ng_nodes(const Node& onnx_node) const
             {
-                return m_model->get_operator(node.op_type(), node.domain())(node);
+                auto ng_node_vector =
+                    m_model->get_operator(onnx_node.op_type(), onnx_node.domain())(onnx_node);
+                for (auto& ng_node : ng_node_vector)
```
postrational: Please move this new code to a helper function with a meaningful name such as `add_provenance_tags`.

postrational:
Please add some unit tests for this feature.

src/ngraph/frontend/onnx_import/core/graph.cpp
```diff
@@ -154,6 +154,19 @@ namespace ngraph
             }
             return results;
         }
+        const NodeVector& Graph::add_provenance_tags(const Node& onnx_node,
```
arogowie-intel: I think this function shouldn't be a part of `Graph` class, since it isn't related in any way with it. This function is independent from `Graph` class. I would suggest to make it static or to move it to some file with helper functions.

mitruska: Ok, I moved it to the `onnx_import/utils/common`. Is it a better place for this function?

test/onnx/onnx_import.in.cpp
```diff
@@ -355,6 +355,21 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_initializer_wo_input)
...
+{
+    auto function = onnx_import::import_onnx_model(
+        file_util::path_join(SERIALIZED_ZOO, "onnx/provenance_tag_add.prototxt"));
+
+    auto ng_nodes = function->get_ordered_ops();
+    for (auto ng_node : ng_nodes)
+    {
+        for (auto tag : ng_node->get_provenance_tags())
+        {
+            EXPECT_EQ("<ONNX Add (add_node1)>", tag);
```
arogowie-intel: Maybe we could be here less restrictive? If provenance tag formatting would change, then this condition fail. I'd suggest using macro `EXPECT_HAS_SUBSTRING` from `test/util/type_prop.hpp` file.  And set condition to *ONNX Add*.
https://github.com/NervanaSystems/ngraph/blob/731959f749ce059b5bb2bec8e028d45a7223a9fb/test/util/type_prop.hpp#L21

mitruska: Good idea, done.

mitruska:
@postrational 
This PR introduces:

> * [x]  ONNX importer's output node has provenance tags
> * [x]  ONNX importer's intermediate nodes have provenance tags

These functionalities will be imlemented by @tomdol in the next PR:

> * [ ]  Input Parameter nodes have provenance tags [2]
> * [ ]  Input Constant nodes have provenance tags [3]
> * [ ]  Instead of using ONNX Node's name (which is rarely used), use output name

PR #4116, opened 2019-12-26
Keep GCPU upto date with INTERPRETER.

silee2:
Improvements:
unit_test.manifest works now.
set of unit_test now on par with Interpreter.
passes all UT.


rkimballn1:
See #4131

silee2:
Closing since #4131 is a longer term solution.

PR #4118, opened 2019-12-27
Fixed output types

ilyachur:


src/ngraph/op/experimental/layers/prior_box_clustered.cpp
```diff
@@ -76,11 +76,11 @@ void op::PriorBoxClustered::validate_and_infer_types()
         // {Prior boxes, variances-adjusted prior boxes}
         const auto num_priors = m_attrs.widths.size();
         set_output_type(
-            0, element::f32, Shape{2, 4 * layer_shape[0] * layer_shape[1] * num_priors});
+            0, m_attrs.outputType, Shape{2, 4 * layer_shape[0] * layer_shape[1] * num_priors});
     }
     else
     {
-        set_output_type(0, element::f32, PartialShape::dynamic());
+        set_output_type(0, m_attrs.outputType, PartialShape::dynamic());
```
slyalin: Why is PriorBoxClustered op only affected? Why not to apply to PriorBox as well?

ilyachur: Fixed

PR #4119, opened 2019-12-27
Fixed creation NonMaxSuppression op with dynamic shapes

ilyachur:


tomdol:
Why is it now fully dynamic with dynamic rank? According to the specs (both internal and ONNX definition) of this op, the output's rank should be `2`. @ilyachur are you sure it's correct?

ilyachur:
@tomdol I am sure that it is correct. Because in another case we cannot create NonMaxSuppression op if previous operation provides dynamic shape. I faced with this issue in next situation when Reshape which exists before NonMaxSuppression op returns dynamic shape because the second input for Reshape is not Constant operation and we should do constant propagation before than Reshape will return static shape.
And in this situation (because Reshape returns dynamic shape) I cannot create NonMaxSuppression operation because rank is also dynamic.
But after the constant propagation I get correct shapes and can easily validate NonMaxSuppression op.

PR #4123, opened 2019-12-30
[ONNX] Use v1 ops in Convolution operator.

arogowie-intel:


src/ngraph/frontend/onnx_import/op/conv.cpp
```diff
@@ -131,11 +131,13 @@ namespace ngraph
...
+                    auto broadcasted_bias = std::make_shared<default_opset::Broadcast>(
                         bias,
-                        new_shape,
-                        ngraph::op::calculate_broadcast_axes(new_shape, bias->get_shape(), 1));
-                    return {std::make_shared<ngraph::opset0::Add>(conv_node, broadcasted_bias)};
+                        default_opset::Constant::create(
+                            element::i64, Shape{new_shape.size()}, new_shape),
+                        default_opset::Constant::create(element::i64, Shape{1}, {1}));
+                    return {std::make_shared<default_opset::Add>(
+                        conv_node, broadcasted_bias, ngraph::op::AutoBroadcastSpec())};
```
postrational: I think we don't need this `ngraph::op::AutoBroadcastSpec()`

diyessi: You should probably defined a helpwe `onnx_autobroadcast_spec` that is `AutoBroadcastSpec` but can be changed in the future if needed. The current default value is mainly to keep pre-dynamic things from breaking. Bridges should be passing the value appropriate for their framework.

PR #4124, opened 2019-12-30
Add Round op

mbencer:
Op implemented based on ONNX specification: https://github.com/onnx/onnx/blob/master/docs/Operators.md#Round


src/ngraph/runtime/reference/round.hpp
```diff
@@ -0,0 +1,52 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
arogowie-intel: Please update year.

src/ngraph/runtime/cpu/kernel/round.hpp
```diff
@@ -0,0 +1,39 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
arogowie-intel: Please update year.

test/backend/round.in.cpp
```diff
@@ -0,0 +1,83 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
arogowie-intel: Please update year.

src/ngraph/op/round.hpp
```diff
@@ -0,0 +1,49 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
arogowie-intel: Please update year.

PR #4129, opened 2020-01-01
Fix ScalarConstantLikeBase

rkimballn1:
ScalarConstantLikeBase did not have a type_info defined and thus a Constant op could be as_type_ptr cast to a ScalarConstantLikeBase since ScalarConstantLikeBase is derived from Constant and thus used Constant's type_info. Confusing enough?

I noticed this because the LikeReplacement pass was taking 52 second to run on bert_squad_training.json and there are no Like ops in that file. The issue was every Constant was treated as a ScalarConstantLikeBase and was therefore replaced with a Constant. Very expensive.

LikeReplacement time went from 52 second to 29 ms or a 1793x speed increase.
NopElimination has the exact same ScalarConstantLikeBase code as LikeReplacement so save another 50 seconds with that pass.

### DIFF
--- a/src/ngraph/op/constant.hpp
+++ b/src/ngraph/op/constant.hpp
@@ -442,6 +442,8 @@ namespace ngraph
         class NGRAPH_API ScalarConstantLikeBase : public Constant
         {
         public:
+            static constexpr NodeTypeInfo type_info{"ScalarConstantLikeBase", 0};
+            const NodeTypeInfo& get_type_info() const override { return type_info; }
             std::shared_ptr<op::Constant> as_constant() const;
             ScalarConstantLikeBase() = default;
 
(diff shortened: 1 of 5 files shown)


src/ngraph/op/constant.hpp
```diff
@@ -442,6 +442,8 @@ namespace ngraph
         class NGRAPH_API ScalarConstantLikeBase : public Constant
         {
         public:
+            static constexpr NodeTypeInfo type_info{"ScalarConstantLikeBase", 0};
```
diyessi: I think there was at one time a reason for `ScalarConstantLikeBase` independent of `ScalarConstantLike` but I don't see one now. Can we just merge them into `ScalarConstantLike` without breaking transformers?

PR #4130, opened 2020-01-02
Call generic CPU kernels for Broadcast and Reshape

rkimballn1:
This change gets NNP ConstantFolding pass for BERT Large down from 114 to 2.5 seconds.


src/ngraph/pass/constant_folding_broadcast.cpp
```diff
@@ -45,11 +45,11 @@ shared_ptr<op::Constant> fold_constant_broadcast(shared_ptr<op::Constant> consta
         auto static_bcast_axes = broadcast_v1->get_broadcast_axes();
         if (static_bcast_axes.first)
         {
-            runtime::reference::broadcast<T>(constant->get_data_ptr<T>(),
-                                             data_ptr,
-                                             constant->get_shape(),
-                                             out_shape,
-                                             static_bcast_axes.second);
+            runtime::gcpu::kernel::broadcast<T>(constant->get_data_ptr<T>(),
```
diyessi: These should be moved out of gcpu and into core if they are to be used in core.

rkimballn1: Yes, this is a good idea, moved to opt_kernel

src/ngraph/runtime/opt_kernel/broadcast.hpp
```diff
@@ -0,0 +1,235 @@
...
+            void broadcast_2d(const T* in,
+                              T* out,
+                              const Shape& in_shape,
+                              const Shape& out_shape,
+                              const AxisSet& broadcast_axes)
+            {
+                size_t index[2];
+                size_t* out_index =
+                    (broadcast_axes.find(0) == broadcast_axes.end() ? &index[0] : &index[1]);
+                for (index[0] = 0; index[0] < out_shape[0]; ++index[0])
```
diyessi: For the index offset calculations here and below, use `row_major_strides` to get the strides once outside the loop and then do a simple dot product of the index and strides.

PR #4139, opened 2020-01-08
Change LogSoftmax to produce v1 and remove FusedOp

mbencer:
- `LogSoftmax` is not present in current spec so it was removed from ngraph (implementation moved to onnx importer)
- `LogSoftmax` in onnx importer produces `v1` ops


src/ngraph/frontend/onnx_import/op/log_softmax.cpp
```diff
@@ -16,8 +16,9 @@
 
 #include <memory>
 
-#include "log_softmax.hpp"
```
arogowie-intel: I think this shouldn't be removed.

src/ngraph/frontend/onnx_import/op/log_softmax.cpp
```diff
@@ -16,8 +16,9 @@
...
+#include "default_opset.hpp"
 #include "ngraph/opsets/opset0.hpp"
```
arogowie-intel: This header is now unnecessary.

src/ngraph/frontend/onnx_import/op/log_softmax.cpp
```diff
@@ -30,11 +31,16 @@ namespace ngraph
...
                     NodeVector inputs{node.get_ng_inputs()};
-                    auto data = inputs.at(0);
-                    auto data_shape = data->get_shape();
-                    int axis = node.get_attribute_value<int64_t>("axis", 1);
+                    const auto data = inputs.at(0);
+                    const auto data_shape = data->get_shape();
 
-                    return {std::make_shared<ngraph::opset0::LogSoftmax>(data, axis)};
+                    const auto axis = node.get_attribute_value<int64_t>("axis", 1);
+                    const auto normalize_axis =
```
arogowie-intel: `normalize`**d**`_axis` would by more suitable name.

PR #4140, opened 2020-01-08
[ONNX] Change Slice and Where to produce v1 ops

mbencer:


src/ngraph/frontend/onnx_import/op/slice.cpp
```diff
@@ -18,9 +18,8 @@
 #include <memory>
 #include <vector>
 
+#include "default_opset.hpp"
 #include "ngraph/node.hpp"
-#include "ngraph/opsets/opset0.hpp"
-#include "slice.hpp"
```
arogowie-intel: This header shouldn't be removed.

mbencer: Done

mbencer: Done

PR #4143, opened 2020-01-08
Finish off some old deprecations, update some others

diyessi:


src/ngraph/node.hpp
```diff
@@ -329,9 +319,6 @@ namespace ngraph
                                                                bool for_get_output_element = true);
 
         /// Checks that there is exactly one output and returns its shape
-        // TODO: deprecate in favor of node->output(0).get_shape() with a suitable check in the
-        // calling code, or updates to the calling code if it is making an invalid assumption of
-        // only one output.
         const Shape& get_shape() const;
```
rkimballn1: This really should be deprecated since it is not clear what shape you are referring to. I prefer get_input_shape(index)/get_output_shape(index) methods as nice, clear helper functions.

src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
```diff
@@ -257,7 +255,7 @@ void MLIRSubgraphExtractionPass::build_subgraphs(std::shared_ptr<Function> func)
                 for (auto it = nodes_ready.begin(); it != nodes_ready.end();)
                 {
                     auto node = *it;
-                    if (TI(Result) == TI(*node))
+                    if (is_type<Result>(node))
```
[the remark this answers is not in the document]
diyessi: There are issues on some platforms when types cross shared library boundaries.

src/ngraph/op/op.hpp
```diff
@@ -34,9 +34,7 @@ namespace ngraph
                 : Node()
             {
             }
-            Op(const NodeVector& arguments);
```
dkurt, line 37: Sorry but why this constructor has been deleted without any deprecation warnings before?

dkurt: /сс @ilyachur @ilya-lavrenov

PR #4146, opened 2020-01-08
Migrate PR#4145 from master.

ayzhuang:


src/ngraph/runtime/cpu/pass/cpu_layout.cpp
```diff
@@ -1796,14 +1796,8 @@ namespace ngraph
                     }
                     catch (const mkldnn::error& e)
                     {
-#if MKLDNN_VERSION_MAJOR < 1
                         throw ngraph_error("MKLDNN Unsupported pooling layout" +
-                                           to_string(fprop_input_md.data.format) +
                                            MKLDNN_ERROR_MESSAGE);
-#else
```
shresthamalik: If the MKLDNN_VERSION _MAJOR is less than 1 then this wont compile right?
```
to_string(fprop_input_md.data.format)
```

ayzhuang: yes, because fprop_input_md declaration was deleted.

PR #4147, opened 2020-01-08
[MLIR] Use more threads when building mlir+llvm

silee2:
Instead of using fixed number of threads use ProcessorCount to find out number of logical cores and pass it to make.


silee2:
Yes. `ProcessorCount(VAR)` Is just a function that tries to get the (max) number of logical cores at config time and store the value in `VAR` .  I have another PR draft #4101 that enables `Ninja` build. This PR is mostly for improving `Unix Makefile` generator.

PR #4154, opened 2020-01-10
[ONNX] Extended support for provenance tags

tomdol:
This PR contains the changes from https://github.com/NervanaSystems/ngraph/pull/4142 and extends onnx importer provenance tagging to constants and parameters, which are produced for graph inputs and initializers.


test/onnx/onnx_import_provenance.in.cpp
```diff
@@ -0,0 +1,93 @@
...
+    for (const auto ng_node : ng_nodes)
+    {
+        for (const auto tag : ng_node->get_provenance_tags())
+        {
+            EXPECT_HAS_SUBSTRING(tag, "ONNX");
+        }
+    }
+}
+
+template <typename NodeToCheck>
```
postrational: Do you need this to be a templated function? `get_provenance_tags` is a function of the base `Node` class.

src/ngraph/frontend/onnx_import/utils/common.hpp
```diff
@@ -38,8 +38,14 @@ namespace ngraph
...
         namespace common
         {
-            const NodeVector& add_provenance_tags(const Node& onnx_node,
-                                                  const NodeVector& ng_node_vector);
+            /// \brief Adds a provenance tag to all nodes in a subgraph representing an ONNX node
+            void add_provenance_tags(const Node& onnx_node, const NodeVector& ng_node_vector);
+
+            /// \brief Adds a provenance tag to a node that represents an input of a graph
+            /// \note  This overload only works with Parameter and Constant nodes
+            void add_provenance_tags(const std::string& input_name,
```
postrational: This should be named `add_provenance_tags_to_input` or just moved to the file where only inputs are tagged.

src/ngraph/frontend/onnx_import/utils/common.cpp
```diff
@@ -28,22 +29,65 @@ namespace ngraph
...
+                /// \brief Builds a provenance tag for an ONNX operator
+                static std::string build_provenance_tag(const Node& onnx_node)
+                {
+                    const auto output_names = concat_strings(onnx_node.get_output_names());
+                    return std::string{"<ONNX " + onnx_node.op_type() + " (" +
+                                       onnx_node.get_name() + " -> " + output_names + ")>"};
+                }
+
+                /// \brief Builds a provenance tag for an input of an ONNX graph
+                static std::string build_provenance_tag(const std::string& input_name)
```
postrational: This should be named `add_provenance_tags_for_input`, since it's only meant to be used for input nodes and always returns `<ONNX Input...`

src/ngraph/frontend/onnx_import/utils/common.cpp
```diff
@@ -28,22 +29,65 @@ namespace ngraph
...
+
+                /// \brief Builds a provenance tag for an ONNX operator
+                static std::string build_provenance_tag(const Node& onnx_node)
+                {
+                    const auto output_names = concat_strings(onnx_node.get_output_names());
+                    return std::string{"<ONNX " + onnx_node.op_type() + " (" +
+                                       onnx_node.get_name() + " -> " + output_names + ")>"};
+                }
+
+                /// \brief Builds a provenance tag for an input of an ONNX graph
```
postrational: It will make more sense to keep all of these helper functions in the same `/core` file, where they are used. They are unlikely to be useful in another context.

src/ngraph/frontend/onnx_import/utils/common.cpp
postrational: It mak make sense add a function which takes a `Tensor` object as input. It could also display the shape of the input which created this `Node`. This would be nice to have, not necessary.

src/ngraph/frontend/onnx_import/utils/common.cpp
```diff
@@ -28,22 +29,65 @@ namespace ngraph
...
+                                           outputs.begin()->get(),
+                                           concat_with_comma);
+                }
+
+                /// \brief Builds a provenance tag for an ONNX operator
+                static std::string build_provenance_tag(const Node& onnx_node)
+                {
+                    const auto output_names = concat_strings(onnx_node.get_output_names());
+                    return std::string{"<ONNX " + onnx_node.op_type() + " (" +
+                                       onnx_node.get_name() + " -> " + output_names + ")>"};
```
postrational: I would get rid of this space, it look strange in the ouput `( ->`
```suggestion
                                       onnx_node.get_name() + "-> " + output_names + ")>"};
```

test/onnx/onnx_import_provenance.in.cpp
tomdol: Yes, this has to be a template because its parameter is used to find a node in the subgraph returned by `function->get_ordered_ops()`. Depending on the test I'm looking for either an op, constant or parameter and they are expected to have different provenance tags. 
This function is just a helper that reduces code duplication in unit tests.

postrational:
It would be a good idea to tag graph outputs as well here:
https://github.com/NervanaSystems/ngraph/blob/fbd99f340d162705a47014339134aea79f9ca3ed/src/ngraph/frontend/onnx_import/core/graph.cpp#L95-L98

src/ngraph/frontend/onnx_import/core/graph.cpp
```diff
@@ -158,12 +195,45 @@ namespace ngraph
 
         NodeVector Graph::make_ng_nodes(const Node& onnx_node) const
         {
-            const auto ng_node_vector =
-                m_model->get_operator(onnx_node.op_type(), onnx_node.domain())(onnx_node);
-            common::add_provenance_tags(onnx_node, ng_node_vector);
+            const auto ng_node_factory =
+                m_model->get_operator(onnx_node.op_type(), onnx_node.domain());
+
+            const auto ng_node_vector = ng_node_factory(onnx_node);
```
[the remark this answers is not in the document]
postrational: I split this line to document that `m_model->get_operator(onnx_node.op_type(), onnx_node.domain())` returns a factory function. Compilers should be able to optimize this `const` variable out, but the code will be more readable.

PR #4163, opened 2020-01-11
extend executable create tensor APIs

ashokei:


ashokei:
@jbobba yes updated now to pass memory_pointers to CPUTensorView. Will see if there is a unittest we can extend/reuse, or add.

shresthamalik:
Just curious:
What are the possible uses/benefits of passing the memory_pointer for tensor creation for backends other than (CPU/Interpreter)? The backend will read/write to/from the memory pointer passed in the read write api calls.

sayantan-nervana:
Adding @MrGeva to comment on this API. Is it ok if we pass in the TF raw pointers in like executable->create_tensor() rather than backend->create_tensor(). This will require some code changes from your end in the nnpi-transformer?

ashokei:
some backends want to directly DMA the data from framework buffer into device memory. Without this memory_pointer (from framework), ngraph tensor will have to allocate/maintain intermediate buffer.

2 new APIs, wait_to_read and wait_to_write, will be added to ngraph::tensor to support this use-case. Instead of read/write calls, users would call these wait_to_xx APIs (no read/write memory pointer needed, these APIs will be implemented by each backend to indicate when the tensor is computed)

MrGeva:
> Adding @MrGeva to comment on this API. Is it ok if we pass in the TF raw pointers in like executable->create_tensor() rather than backend->create_tensor(). This will require some code changes from your end in the nnpi-transformer?

This is an excellent change, and very important for SPH to support scenario where the input/output TF raw pointers keep changing, and also for support pipelining infer requests.

src/ngraph/runtime/executable.cpp
```diff
@@ -130,25 +130,26 @@ void runtime::Executable::save(std::ostream& /* output_stream */)
     throw runtime_error("save operation unimplemented.");
 }
 
-shared_ptr<runtime::Tensor> runtime::Executable::create_input_tensor(size_t /* input_index */)
+shared_ptr<runtime::Tensor> runtime::Executable::create_input_tensor(size_t /* input_index */,
```
diyessi: Why aren't these defined as `=0` so they are detected as not implemented at compile-time?

ashokei: this PR just reuses existing API, if we make it pure virtual this would be a breaking change for other backends.

src/ngraph/runtime/executable.hpp
```diff
@@ -80,32 +80,46 @@ class NGRAPH_API ngraph::runtime::Executable
     /// \brief Create an input Tensor
     /// \param input_index The index position in the input Parameter vector. This would be the same
     /// order of Parameters passed into the inputs in the call() method.
+    /// \param memory_pointer A pointer to a buffer used for this tensor. The size of the buffer
+    ///     must be sufficient to contain the tensor. The lifetime of the buffer is the
+    ///     responsibility of the caller.
```
rkimballn1: responsibility of the caller and must outlive the created Tensor.

src/ngraph/runtime/executable.hpp
```diff
@@ -80,32 +80,46 @@ class NGRAPH_API ngraph::runtime::Executable
     /// \brief Create an input Tensor
     /// \param input_index The index position in the input Parameter vector. This would be the same
     /// order of Parameters passed into the inputs in the call() method.
+    /// \param memory_pointer A pointer to a buffer used for this tensor. The size of the buffer
```
rkimballn1: The default value is nullptr but nowhere is it described what happens when a nullptr is passed in. Is the implementor of this this method supposed to allocate a buffer if none is passed in? Should I throw an error with a nullptr?
These comments apply to all docstrings.

src/ngraph/runtime/executable.cpp
rkimballn1: @diyessi These were never required methods. I think in the future we probably want to switch to a single API for creating tensors but this is just a step along the way.

rkimballn1: @ashokei I'm pretty sure what you have will break the NNP backend where we are already overloading old API which has now been removed. We will get compile errors with this code.

src/ngraph/runtime/executable.hpp
```diff
@@ -80,32 +80,46 @@ class NGRAPH_API ngraph::runtime::Executable
...
+    ///     must be sufficient to contain the tensor. The lifetime of the buffer is the
+    ///     responsibility of the caller.
     /// \returns A Tensor
-    virtual std::shared_ptr<runtime::Tensor> create_input_tensor(size_t input_index);
+    virtual std::shared_ptr<runtime::Tensor> create_input_tensor(size_t input_index,
+                                                                 void* memory_pointer = nullptr);
```
rkimballn1: If you leave the old API and add these methods (without the "= nullptr" default) as a new API then it will be backwards compatible. As it is these API changes will break the NNP backend, but not in a way that is too difficult to fix, but it is not backwards compatible.

src/ngraph/runtime/tensor.hpp
```diff
@@ -103,6 +103,12 @@ namespace ngraph
             /// \param n Number of bytes to read, must be integral number of elements.
             virtual void read(void* p, size_t n) const = 0;
 
+            /// \brief check tensor for new data, call may block.
+            ///    backends may use this to ensure tensor is updated (eg: lazy eval).
+            virtual void wait_to_read() {}
```
rkimballn1: How about `wait_for_read_ready` and `wait_for_write_ready`

ashokei: yes sounds more descriptive, will make the change.

src/ngraph/runtime/executable.cpp
ashokei: will make second API,  with a default implementation.

src/ngraph/runtime/executable.hpp
ashokei: will remove nullptr, and make new API without default nullptr.

src/ngraph/runtime/cpu/cpu_backend.cpp
```diff
@@ -224,30 +224,40 @@ shared_ptr<ngraph::op::Result> runtime::cpu::CPU_Executable::get_result(size_t i
...
+        result->get_element_type(), result->get_shape(), memory_pointer);
 }
 
-vector<shared_ptr<runtime::Tensor>>
-    runtime::cpu::CPU_Executable::create_input_tensor(size_t input_index, size_t pipeline_depth)
+vector<shared_ptr<runtime::Tensor>> runtime::cpu::CPU_Executable::create_input_tensor(
+    size_t input_index, size_t pipeline_depth, std::vector<void*> memory_pointers)
 {
+    bool mem_ptr_size = memory_pointers.size();
+    if (mem_ptr_size > 0)
```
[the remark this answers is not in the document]
ashokei: @rkimballn1 this is reused by API without host raw_ptr

PR #4164, opened 2020-01-11
Restore constant folding for DynReshape until users are converted to …

diyessi:
…v1 Reshape


test/dyn_elimination.cpp
```diff
@@ -266,3 +269,23 @@ TEST(dyn_elimination, range_f64)
     ASSERT_TRUE(test::all_close_f(
         vals, vector<double>{-0.5, -0.25, 0, 0.25, 0.5, 0.75, 1.0, 1.25, 1.5, 1.75}));
 }
+
```
jbobba: Guard with #ifndef NGRAPH_JSON_DISABLE

PR #4165, opened 2020-01-13
[ LSTMCell ] Setting number of outputs for LSTMCell op

jane-intel:


src/ngraph/op/fused/lstm_cell.cpp
```diff
@@ -52,6 +52,7 @@ op::LSTMCell::LSTMCell(const Output<Node>& X,
     , m_input_forget{input_forget}
     , m_weights_format{weights_format}
 {
+    set_output_size(2);
```
diyessi: You should set this in `constructor_validate_and_infer_types` instead of the constructor.

PR #4174, opened 2020-01-14
Porting PR #4130 for faster reshape and broadcast constant folding

jbobba:


ilyachur:
This PR doesn't resolve the issue with performance.

PR #4177, opened 2020-01-14
Remove pragma stuff from headers

rkimballn1:


mbrookhart:
After #4178, it doesn't seem we need this anymore. Thanks for all of the help debugging, @rkimballn1.

PR #4182, opened 2020-01-15
[MLIR] Remove halide.

silee2:
nGraph is actively adding MLIR support. It is time to cut ties with the no longer maintained halide code.


nmostafa:
LGTM. Changing title to make sure MLIR is sane.

nmostafa:
Apparently MLIR tests always run now without the title prefix. Sorry for re-running the ci test.

PR #4186, opened 2020-01-16
Added API to get input node@index for current node

swapna-intel:
We have 16 use cases for this in nnp-transformer:

grep -ro \)\.get_source_output\(\)\.get_node\(\) * | wc -l
16


src/ngraph/node.cpp
```diff
@@ -490,6 +490,20 @@ std::shared_ptr<Node> Node::get_argument(size_t index) const
...
 }
 
+Node* Node::get_input_node_ptr(size_t index) const
+{
+    NGRAPH_CHECK(
+        index < m_inputs.size(), "index '", index, "' out of range in get_argument(size_t index)");
+    return m_inputs[index].get_output().get_node().get();
+}
+
+std::shared_ptr<Node> Node::get_input_node_shr_ptr(size_t index) const
```
jbobba: can we call this get_input_node_shared_ptr()

swapna-intel: done

PR #4188, opened 2020-01-16
[MLIR] Update build system to use LLVM mono-repo for MLIR

dcaballe:
This includes only the changes needed in build system so that CI can be updated accordingly.
Note that `build` directory was change from `mlir_project/llvm-projects/build` to `mlir_project/build` because in the new setup, cmake tries to create the build folder *before* cloning LLVM mono-repo and `llvm-projects` doesn't exist yet.

Source conflicts will be merged with this patch later on.


src/contrib/mlir/core/ngraph_dialect/ops.cpp
```diff
@@ -303,56 +303,42 @@ mlir::LogicalResult verifyOp(NGConvolutionOp* op)
...
-            return op->emitOpError("Invalid result spatial shape");
+            return op.emitOpError("Invalid result spatial shape");
         }
     }
     return mlir::success();
 }
 
 template <>
-mlir::LogicalResult verifyOp(NGMatMulOp* op)
+mlir::LogicalResult verifyOp(NGSoftMaxOp op)
```
nmostafa: Why the renames ?

dcaballe: That's probably git. I removed those that became or were unused. I thing we should remove all of these functions in a separate commit. They are growing, not doing anything and we have to maintain them.

PR #4192, opened 2020-01-16
mlir monorepo conflicts

dcaballe:
Addressing conflicts of moving to MLIR mono-repo. This patch will land into dcaballe/mlir_monorepo first before being merged into master.

Key learning:
* Class Value is now used/passed by value since it's just a wrapper to a storage so no more `Value *`.
* Same applies to any dialect op. All NG...Op should be used/passed by value without a pointer for the same reason.
* We should used LLVM container as much as possible since MLIR implements hashes and other utilities needed to manipulate these wrapper classes in containers.

Tests for views an memory optimizations are failing due to changes in the way maps are printed. I'm temporarily disabling them. @nmostafa, could you please look into them and prepare a follow-up patch?


src/contrib/mlir/backend/analysis/memory_analysis.cpp
```diff
@@ -68,16 +68,16 @@ namespace
...
-        void insertNoAlias(Value* a, Value* b);
+        bool canAlias(Value a, Value b);
+        void insertNoAlias(Value a, Value b);
 
     private:
         using BV = llvm::BitVector;
-        std::unordered_map<Value*, unsigned> m_valueToIdx;
-        std::unordered_map<unsigned, Value*> m_idxToValue;
-        std::unordered_map<Value*, BV*> m_valueToSet;
+        DenseMap<Value, unsigned> m_valueToIdx;
```
nmostafa: `unordered_map --> DenseMap` is needed for MLIR upgarde ?

dcaballe: Yes, basically:
> We should used LLVM container as much as possible since MLIR implements hashes and other utilities needed to manipulate these wrapper classes in containers.

Value is no longer a pointer so we would need to implement a hash for it. MLIR already has all the utilities needed for DenseMap.

src/contrib/mlir/backend/analysis/memory_analysis.cpp
```diff
@@ -392,7 +392,7 @@ namespace
     void MemoryAssignment::processDestructiveInPlace(mlir::Operation* op)
     {
         NGRAPH_CHECK(op->getNumResults() == 1, "Destructive in-place with multi-def ?");
-        Value* use = nullptr;
+        Value use = nullptr;
```
ayzhuang: use is not a pointer anymore.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -337,11 +336,11 @@ namespace
                 //    the linear buffer.
                 // If two memrefs are defined via 2 Views over the same buffer, then they share and
                 // will re-use the same buffer.
-                auto tensorType = origResult->getType().cast<NGTensorType>();
-                Value* newResult = nullptr;
+                auto tensorType = origResult.getType().cast<NGTensorType>();
+                Value newResult = nullptr;
```
ayzhuang: newResult is not a pointer. Also line 343 bufferValue is not a pointer.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -402,10 +401,10 @@ namespace
...
-                                                 Value* buffer,
-                                                 unsigned offset,
-                                                 PatternRewriter& rewriter)
+    Value DialectLoweringPass::createTempMemref(Type type,
+                                                Value buffer,
+                                                unsigned offset,
+                                                PatternRewriter& rewriter)
     {
         MemRefType memRefType = type.cast<MemRefType>();
         if (buffer)
```
ayzhuang: need to do different check now.

PR #4194, opened 2020-01-17
Update broadcasting helpers to use v1 operators.

arogowie-intel:
Additionally enhanced `v1::Broadcast` to support bidirectional NumPy broadcasting mechanism.


arogowie-intel:
@jbobba Could you look especially at `v1::Broadcast` operator: https://github.com/NervanaSystems/ngraph/pull/4194/files?file-filters%5B%5D=.cpp#diff-f13765ceacaf216a44b2b4e7ffa11209. I've added bidirectional broadcasting.

src/ngraph/op/util/broadcasting.hpp
```diff
@@ -53,27 +55,6 @@ namespace ngraph
...
-        ///
-        /// This style of broadcast was used in ONNX Op sets prior to version 7, where it was
-        /// replaced by numpy-style broadcasting.
-        ///
-        /// \param left Node which contain input of binary op.
-        /// \param right Node which contain input of binary op.
-        /// \param start_match_axis position in shape denoting start of the mutually equal shape
-        ///
-        /// \return Left and right node after broadcasting.
-        NodeVector legacy_style_broadcast_for_binary_operation(const std::shared_ptr<Node>& left,
```
tsocha: Could you rename `legacy_style_broadcast_values_for_binary_operation` -> `legacy_style_broadcast_for_binary_operation`?
Since `legacy_style_broadcast_for_binary_operation` has been removed there is a good opportunity to remove `values` from function name.
Especially when opset1 version of `legacy_style_broadcast_values_for_binary_operation` is named `legacy_style_broadcast_for_binary_operation`.

src/ngraph/pass/opset0_downgrade.cpp
```diff
@@ -146,13 +146,48 @@ namespace
     bool op_cast(shared_ptr<op::v1::Broadcast> node)
```
tsocha: Do we need to update upgrade pass as well?

src/ngraph/op/util/broadcasting.cpp
```diff
@@ -515,6 +420,100 @@ namespace ngraph
...
+
+            std::vector<std::size_t> get_axes_mapping(const Shape& output_shape,
+                                                      const AxisSet& broadcast_axes)
+            {
+                NGRAPH_CHECK((broadcast_axes.size() <= output_shape.size()));
+                std::vector<size_t> axes_mapping(output_shape.size());
+                std::iota(axes_mapping.begin(), axes_mapping.end(), 0);
+                for (auto i = broadcast_axes.rbegin(); i != broadcast_axes.rend(); ++i)
+                {
+                    axes_mapping.erase(axes_mapping.begin() + *i);
```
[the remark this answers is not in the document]
tomdol: You cannot erase elements from `axes_mapping` using iterators from `broadcast_axes`, the suggested change leads to undefined behavior. 
The loop erases elements from the back using values pointed to by `i` and is correct.

src/ngraph/builder/reshape.cpp
```diff
@@ -195,6 +195,29 @@ shared_ptr<Node> builder::opset1::reorder_axes(const Output<Node>& value, vector
...
+        return value.get_node_shared_ptr();
+    }
+
+    Shape in_shape{value.get_shape()};
+    for (size_t idx = 0; idx < axes.size(); ++idx)
+    {
+        in_shape.at(axes.at(idx)) = 0;
+    }
+    Shape output_shape;
+    for (auto axis : in_shape)
```
tomdol: This loop can be replaced with `std::copy_if`

src/ngraph/frontend/onnx_import/op/instance_norm.cpp
```diff
@@ -42,47 +42,71 @@ namespace ngraph
...
 
                     const auto sqrt = std::make_shared<default_opset::Sqrt>(variance + eps_node);
 
-                    return {scale * (data - mean) / sqrt + bias};
+                    // scale * (data - mean) / sqrt + bias
+                    std::shared_ptr<ngraph::Node> result{
+                        std::make_shared<default_opset::Subtract>(data, mean)};
+                    result = std::make_shared<default_opset::Multiply>(scale, result);
+                    result = std::make_shared<default_opset::Divide>(result, sqrt);
+                    result = std::make_shared<default_opset::Add>(result, bias);
```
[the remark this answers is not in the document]
arogowie-intel: Yes, but I think this should be done in separate PR.

src/ngraph/op/fused/lstm_sequence.cpp
```diff
@@ -120,8 +120,8 @@ shared_ptr<Node> op::LSTMSequence::get_masked_node(const Output<Node>& data,
     shared_ptr<Node> curr_time_step_node = op::Constant::create(
         element::i32, data.get_shape(), vector<int32_t>(shape_size(data.get_shape()), time_step));
 
-    shared_ptr<Node> batch_seq_length =
-        op::legacy_style_broadcast_for_binary_operation(
+    Output<Node> batch_seq_length =
+        op::legacy_style_broadcast_values_for_binary_operation(
```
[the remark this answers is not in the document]
arogowie-intel: Since this is `FusedOp` we should return only v0 operators.

src/ngraph/op/util/broadcasting.hpp
arogowie-intel: Done

src/ngraph/pass/opset0_downgrade.cpp
arogowie-intel: I've refactored it.

src/ngraph/builder/reshape.cpp
arogowie-intel: I've overlooked that there indeed is one overload since c++11 :)

src/ngraph/op/fused/lstm_sequence.cpp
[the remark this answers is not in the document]
tsocha: Ou I missed that this is a v0.

src/ngraph/op/broadcast.cpp
```diff
@@ -81,20 +81,17 @@ std::pair<bool, AxisSet> op::v1::Broadcast::get_broadcast_axes() const
...
     {
-        if (input(0).get_partial_shape().is_static() &&
-            input_value(1).get_node_shared_ptr()->is_constant())
+        if (input(0).get_partial_shape().is_static())
         {
             auto arg_shape = input(0).get_shape();
-            auto target_shape =
-                static_pointer_cast<op::Constant>(input_value(1).get_node_shared_ptr())
-                    ->get_shape_val();
+            auto result_shape = output(0).get_shape();
```
jbobba: wouldn't this lead to an assert if the result_shape is not static?

arogowie-intel: Good point. Added check.

PR #4197, opened 2020-01-17
[ONNX] Initial support for dynamic shapes

tomdol:
The main change in this PR is that `ValueInfo` class in ONNX importer now uses `PartialShape` instead of `Shape`. With this change it's possible to import (and execute) ONNX models containing dynamic inputs (fully dynamic including the tensor rank or with only some dimentions that are dynamic).

Until today we've only relied on static shapes in our unit tests so this change does not break any of those tests - if an ONNX model contains only static shapes, calling `PartialShape::is_static()` on the partial shape returned from `ValueInfo::get_partial_shape()` will return `true`.

This PR should be considered as pre-work to the ONNX dynamic shapes support in nGraph.


src/ngraph/frontend/onnx_import/core/value_info.hpp
```diff
@@ -68,7 +73,7 @@ namespace ngraph
             ValueInfo& operator=(ValueInfo&&) = delete;
 
             const std::string& get_name() const { return m_value_info_proto->name(); }
-            const Shape& get_shape() const { return m_shape; }
```
tsocha: [NON-BLOCK]
`get_shape()` can be helpful in some places.
It can get a shape from `m_partial_shape`.(or throw an exception if it isn't static)

test/models/onnx/cum_sum_2d_dynamic_axis_input.prototxt
```diff
@@ -39,6 +39,8 @@ graph {
     type {
       tensor_type {
         elem_type: 6
+        shape {
+        }
```
tsocha: Is this change necessary?
We should probably treat missing shape as a dynamic one.

test/models/onnx/provenance_input_tags.prototxt
```diff
@@ -10,7 +10,7 @@ graph {
   }
   name: "test_graph"
   initializer {
-    dims: 0
```
tsocha: Since your code have a support for this case its better to leave it as is and add one extra test & model.

tomdol: I've added a test for Constants produced from initializers which are scalars. I will leave this change in this particular provenance tags testing model though - it reflects better how those tags should be constructed now that dynamic shapes support is added.

src/ngraph/frontend/onnx_import/core/tensor.hpp
```diff
@@ -365,6 +365,10 @@ namespace ngraph
                 : m_tensor_proto{&tensor}
                 , m_shape{std::begin(tensor.dims()), std::end(tensor.dims())}
             {
+                if (m_shape == Shape{0})
```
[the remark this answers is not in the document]
tomdol: This a check to correctly represent ONNX scalar initializers. I've added some comment to the code plus there's a unit tests for this `if`.

src/ngraph/frontend/onnx_import/core/value_info.hpp
```diff
@@ -67,7 +72,7 @@ namespace ngraph
...
+            const PartialShape& get_partial_shape() const { return m_partial_shape; }
```
postrational: I think we can retain the name `get_shape` and just change the function to return `PartialShape`s.
In some cases the shape will be fully static and we can use `input.get_shape().is_static()` to check.
I think this will make for a cleaner API.

test/onnx/onnx_import_provenance.in.cpp
```diff
@@ -85,11 +85,11 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, provenance_multiple_outputs_op)
 NGRAPH_TEST(onnx_${BACKEND_NAME}, provenance_tagging_constants)
 {
     test_provenance_tags<default_opset::Constant>("onnx/provenance_input_tags.prototxt",
-                                                  "<ONNX Input (initializer_of_A) Shape{0}>");
+                                                  "<ONNX Input (initializer_of_A) {1}>");
```
postrational: Let's add `Shape:` before the shape in this case. Will be easier to read if shape is fully dynamic.

PR #4200, opened 2020-01-17
[MLIR] Enable MLIR ninja build

silee2:


silee2:
Need #4204 to build on macOS

PR #4203, opened 2020-01-17
Remove dynreshape use v1:Reshape instead

baojun-nervana:


src/ngraph/pass/opset0_downgrade.cpp
```diff
@@ -277,8 +277,7 @@ namespace
         }
         else
         {
-            replacement_node = make_shared<op::v0::DynReshape>(
-                node->input_value(0), node->input_value(1), node->get_special_zero());
+            return false;
```
tomdol: I think this change is going to introduce some inconsistency and problems with the execution of models containing `v1::Reshape`. The ONNX importer is supposed to produce as many `v1` ops as possible but to still be able to execute those models on our backends, we rely on the opset0 downgrade pass. This pass ensures that all `v1` ops produced bu the importer are downgraded to `v0` because our backends can only operate on those implementations.

After this change we can no longer rely on the downgrade pass because it will leave some `v1` ops in the processed `ngraph::Function` and this will cause an exception in runtime.

baojun-nervana: @tomdol Is it confirmed to be an issue in ONNX? I am thinking you may not ever fall into that logic branch. The v1::Reshape and DynReshape seem to be identical except they belong to different opsets? Is it necessary to have all ops to downgrade to v0?

tomdol: If we can move the `v1::Reshape` to `v0::Reshape` (and that needs to include the backend implementation) we'll be safe. 
To be able to execute a model imported from ONNX on a non-dynamic backend we have to downgrade the `ngraph::Function` from opset1 to opset0. On the other hand we have a requirement to return ops from opset1 when importing ONNX models.
If we move the implementation of dynamic reshape from `v1` to `v0`, we can use it to create a `Reshape` operation in opset1. This way `Reshape` will not have to be downgraded at all.
Do you think that's possible?

PR #4204, opened 2020-01-17
[MLIR] Suppress warning from usage of C99 flexible array members on AppleClang.

silee2:
Fix build error on macOS
```
In file included from ../src/ngraph/runtime/cpu/cpu_executor.cpp:19:
In file included from ../src/ngraph/runtime/cpu/cpu_executor.hpp:24:
In file included from ../src/ngraph/runtime/cpu/cpu_runtime_context.hpp:34:
../src/contrib/mlir/runtime/cpu/cpu_runtime.hpp:41:25: error: flexible array members are a C99 feature [-Werror,-Wc99-extensions]
                int64_t shapeAndStrides[];
                        ^
1 error generated.
```


src/contrib/mlir/runtime/cpu/cpu_runtime.hpp
```diff
@@ -38,7 +38,7 @@ namespace ngraph
                 void* allocatedPtr;
                 void* alignedPtr;
                 int64_t offset;
-                int64_t shapeAndStrides[];
+                std::vector<int64_t> shapeAndStrides;
```
nmostafa: This representation must match an MLIR representation for the call ABIs to work. 
https://mlir.llvm.org/docs/ConversionToLLVMDialect/

dcaballe: This is intended to be used as an input descriptor used in MLIR compiled code so I think data is expected to be statically allocated. It's probably not failing now because we are not generating code that uses this field, I think. Let's ask Amy since she run some tests with this.

nmostafa: Didn't fail because callbacks are currently under an env variable. Maybe we should enable it by default ? 
@ayzhuang for reference

silee2: Reverted back to what is was before. Suppressing warning on macOS instead.

silee2: Changed title accordingly

PR #4205, opened 2020-01-17
[MLIR] Use mkldnn callback for ConvBias.

ayzhuang:


src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
```diff
@@ -418,6 +418,50 @@ void MLIRSubgraphExtractionPass::sanity_check(std::shared_ptr<Function> func, No
     }
 }
 
+template <typename T>
+static bool can_use_mkldnn_conv_callback(ngraph::Node* node)
+{
+    auto convolution = static_cast<const T*>(node);
```
nmostafa: Is this method specifically for convolution nodes ? Why templatized when it seems very type-specific ? Maybe specialize the template instead ?

src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
```diff
@@ -550,7 +604,8 @@ bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node
     if (is_type<ngraph::op::MatMul>(node))
     {
         // MatMul is only supported through callback
-        if (std::getenv("NGRAPH_MLIR_CALLBACK") == nullptr)
+        if (std::getenv("NGRAPH_MLIR_CALLBACK") == nullptr ||
```
nmostafa: Use utilities from https://github.com/NervanaSystems/ngraph/pull/4103

src/contrib/mlir/runtime/cpu/cpu_callbacks.cpp
```diff
@@ -136,13 +136,12 @@ static mkldnn::memory convert_layout_if_diff(const mkldnn::memory::desc& lhs,
     }
 }
 
-static void convert_output_layout(const mkldnn::memory::desc& lhs,
+static void convert_output_layout(mkldnn::memory& reorder_in,
```
nmostafa: We should use MLIR naming convention (camelCase) here except for public APIs. You could fix the ones you introduced in the PR, and the remaining in a separate PR.

test/mlir/affine_conversion/callback_ops.mlir
```diff
@@ -4,9 +4,25 @@
 
 // -----
 
+// Convbias Op
```
nmostafa: nit: we should rename the folder to something like "lowering" instead of affine_conversion

test/mlir/affine_conversion/callback_ops.mlir
```diff
@@ -4,9 +4,25 @@
...
+  %0 = "ng.convBias"(%arg0, %arg1, %arg2) {dilation = [1, 1], padAbove = [0, 0], padBelow = [0, 0], strides = [1, 1], withRelu = false} : (!ng.tensor<1x1x3x3xf32>, !ng.tensor<1x1x3x3xf32>, !ng.tensor<1xf32>) -> !ng.tensor<1x1x1x1xf32>
+  "ng.return"(%0) : (!ng.tensor<1x1x1x1xf32>) -> ()
+}
+
+// -----
+
 // Softmax Op
 // CHECK-LABEL: func @simple_softmax
-//       CHECK: %[[C1:.*]] = constant 0 : i64
+//       CHECK: %[[C1:.*]] = constant {{[0-9]+}} : i64
```
nmostafa: Why this change ?

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -44,6 +44,11 @@
 #define DEBUG_TYPE PASS_NAME
 
 std::vector<ngraph::runtime::ngmlir::opAttrs> opAttrsVec;
+inline size_t static insertAttrs(ngraph::runtime::ngmlir::opAttrs attrs)
+{
+    opAttrsVec.push_back(attrs);
+    return opAttrsVec.size() - 1;
+}
```
jbobba: Why are we moving this out here?

src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
ayzhuang: This is for the family of convolution nodes: Convolution, ConvolutionBias, ConvolutionRelu, GroupConvolution, ...

test/mlir/affine_conversion/callback_ops.mlir
ayzhuang: I had m_attrsVec in DialectLoweringPass and assigned the m_attrsVec to the global vector that holds the attributes for callbacks. As a result,  all lit tests in this file would have 0 as the index into the global vector. Now I insert attributes directly into the global vector. The index would be 0, 1, 2, ... for the lit tests.

src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
nmostafa: I see. Then maybe add a comment to that effect.

src/contrib/mlir/runtime/cpu/cpu_callbacks.cpp
```diff
@@ -97,59 +95,58 @@ static bool compare_mkldnn_md_formats(const mkldnn::memory::desc& lhs,
     auto blk2 = md2.format_desc.blocking;
 
     if (blk1.inner_nblks != blk2.inner_nblks ||
-        !compare_mkldnn_dims(blk1.inner_blks, blk2.inner_blks, blk1.inner_nblks) ||
-        !compare_mkldnn_dims(blk1.inner_idxs, blk2.inner_idxs, blk1.inner_nblks))
+        !compareMkldnnDims(blk1.inner_blks, blk2.inner_blks, blk1.inner_nblks) ||
```
dcaballe: Another suggestion for future PRs: renaming changes not tightly related to the main goal of this PR could go in a separate PR to have them in quickly and separate from other unrelated changes.

src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
```diff
@@ -550,7 +605,8 @@ bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node
     if (is_type<ngraph::op::MatMul>(node))
     {
         // MatMul is only supported through callback
-        if (std::getenv("NGRAPH_MLIR_CALLBACK") == nullptr)
+        if (!getenv_bool("NGRAPH_MLIR_CALLBACK") || node->get_input_shape(0).size() != 2 ||
+            node->get_input_shape(1).size() != 2)
```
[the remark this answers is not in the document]
ayzhuang: With your PR, we can run all the unit tests together with callback enabled :).  These two unit tests onnx_CPU.model_matmul_vec_ten3d and onnx_CPU.matmul_float_type have 3d inputs and should not be supported through callback.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1303,7 +1306,7 @@ void runtime::cpu::CPU_ExternalFunction::register_common_passes(
     REGISTER_KNOBBED_PASS(CPUCollapseDims, true, runtime::cpu::pass)
 
 #ifdef NGRAPH_MLIR_ENABLE
-    if (std::getenv("NGRAPH_MLIR") != nullptr)
+    if (getenv_bool("NGRAPH_MLIR_CALLBACK"))
```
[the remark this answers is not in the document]
ayzhuang: This should be NGRAPH_MLIR instead of NGRAPH_MLIR_CALLBACK, made a mistake when doing copy and paste. These checks were here before. I think for line 1300 we don't need NGRAPH_MLIR_ENABLE because we run CPUFusion pass when 1) NGRAPH_MLIR_ENABLE is false or 2) when NGRAPH_MLIR_ENABLE is true but NGRAPH_MLIR is false.

src/contrib/mlir/runtime/cpu/cpu_callbacks.cpp
ayzhuang: Thanks!

src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
```diff
@@ -531,7 +529,7 @@ bool MLIRSubgraphExtractionPass::is_supported_mlir_op(std::shared_ptr<Node> node
     if (auto softmax = as_type_ptr<ngraph::op::Softmax>(node))
     {
         // Softmax is only supported through callback
-        if (!getenv_bool("NGRAPH_MLIR_CALLBACK"))
+        if (std::getenv("NGRAPH_MLIR_CALLBACK") == nullptr)
```
dcaballe: Why don't we use the `getenv_bool` utility?

ayzhuang: will create another PR to update getenv not related to convBias for CPU backend.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -537,9 +603,548 @@ namespace
...
+                                                 globalType,
+                                                 false,
+                                                 LLVM::Linkage::Internal,
+                                                 builder.getZeroAttr(globalType),
+                                                 builder);
+            auto globalPtr = builder.create<LLVM::AddressOfOp>(builder.getUnknownLoc(), globalVal);
+
+            switch (m_attrsTyVec[i])
+            {
+            case AttrsType::INT:
```
dcaballe: I think adding an example of how the set of attributes looks like for each op would be very useful. It would help verify if the stores are actually doing the right thing.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -537,9 +603,548 @@ namespace
...
+                    createStore(llvmI64Ty,
+                                builder.getI64IntegerAttr(attrs.poolAttrs3d.padAbove[k]),
+                                gepStructOp,
+                                builder);
+                }
+                break;
+            }
+            case AttrsType::CONV1D:
+            {
+                auto conv1dTy = getLLVMType(AttrsType::CONV1D, llvmDialect);
```
dcaballe: Let's try to improve this a bit by refactoring all the switch cases to utility functions and leave this function as a dispatcher. Something like:
```
// Doc showing what is initializing exactly.
static initConv2D(...){
   ...
}

switch(...) {
  case AttrsTrype::CONV1D:
      initConv1D(...);
  case YYY:
      initYYY
}
```

test/mlir/affine_conversion/callback_ops.mlir
```diff
@@ -20,13 +36,13 @@ func @simple_softmax(%arg0: !ng.tensor<2x3xf32>, %arg1: !ng.tensor<1x!ng.i64>) -
...
-//       CHECK: %[[C2:.*]] = constant {{[0-9]+}} : i64
-//       CHECK: %0 = memref_cast %arg0 : memref<3x6xf32> to memref<*xf32>
-//       CHECK: %1 = memref_cast %arg1 : memref<6x4xf32> to memref<*xf32>
-//       CHECK: %2 = memref_cast %arg2 : memref<3x4xf32> to memref<*xf32>
-//       CHECK: %3 = memref_cast %arg3 : memref<3x4xf32> to memref<*xf32>
-//       CHECK: call @__mlir_callback_3_inputs(%0, %1, %2, %3, %[[C1]], %[[C2]]) : (memref<*xf32>, memref<*xf32>, memref<*xf32>, memref<*xf32>, i64, i64) -> ()
+//       CHECK: %0 = llvm.mlir.addressof @globalAttrs{{[0-9]+}} : !llvm<"{ i8, [3 x i64], [3 x i64], [3 x i64], [3 x i64] }*">
+//       CHECK: %[[C1:.*]] = constant {{[0-9]+}} : i64
+//       CHECK: %1 = memref_cast %arg0 : memref<3x6xf32> to memref<*xf32>
+//       CHECK: %2 = memref_cast %arg1 : memref<6x4xf32> to memref<*xf32>
```
dcaballe: Please, use regular expresions to capture variable names. For example, here we can do:
```
%[[MC0:.*]] = memref_cast %arg0 : memref<3x6xf32> to memref<*xf32>
...
call @__mlir_callback_3_inputs(...%[[MC0]]...)
```

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -537,9 +603,548 @@ namespace
...
+                                                    geps[4],
+                                                    ArrayRef<Value>({constants[0], constants[k]}));
+                    createStore(llvmI64Ty,
+                                builder.getI64IntegerAttr(attrs.poolAttrs2d.padAbove[k]),
+                                gepStructOp,
+                                builder);
+                }
+                break;
+            }
+            case AttrsType::POOL3D:
```
dcaballe: If this is just introducing support for ConvBias, why do we need all these code for other ops? Is it possible to leave that for a future patch?

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -537,9 +603,548 @@ namespace
...
+            // Create a global and insert to the module.
+            PatternRewriter::InsertionGuard insertGuard(rewriter);
+            rewriter.setInsertionPointToStart(module.getBody());
+            rewriter.create<LLVM::GlobalOp>(
+                rewriter.getUnknownLoc(), globalType, isConstant, linkageType, name, initVal);
+        }
+        return module.lookupSymbol<LLVM::GlobalOp>(name);
+    }
+
+    void DialectLoweringPass::insertInitFunc()
```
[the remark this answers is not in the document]
ayzhuang: comments in line 310.  need to change vector to variables there.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -537,9 +603,548 @@ namespace
...
+                    llvmI8Ty,
+                    builder.getI8IntegerAttr(static_cast<int8_t>(attrs.gemmAttrs2d.transposeA)),
+                    geps[0],
+                    builder);
+                createStore(
+                    llvmI8Ty,
+                    builder.getI8IntegerAttr(static_cast<int8_t>(attrs.gemmAttrs2d.transposeB)),
+                    geps[1],
+                    builder);
+                createStore(
```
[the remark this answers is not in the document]
ayzhuang: llvm type and Attribute passed to createStore are not the same for all. I can combine the same types into a loop. Otherwise, we need these two lists: one for llvm types like {llvmI8Ty, ...} and one for Attributes like {builder.getI8IntegerAttr(static_cast<int8_t>(attrs.gemmAttrs2d.transposeA)), ...}.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
ayzhuang: Do you mean to copy the definitions of those structs from there https://github.com/NervanaSystems/ngraph/blob/66bf59ef9c9291b85862374dd3e99522f33a84f4/src/contrib/mlir/runtime/cpu/callback_utils.hpp#L97 to comment?

src/contrib/mlir/backend/pass/affine_lowerer.cpp
ayzhuang: We have to include those because we're changing the way we store and pass attributes to callbacks.

src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
```diff
@@ -419,6 +420,48 @@ void MLIRSubgraphExtractionPass::sanity_check(std::shared_ptr<Function> func, No
...
 }
 
+// Check if convolution related nodes such as Convolution, ConvolutionBias,
+// ConvolutionRelu, ... can use callback.
+template <typename T>
+static bool can_use_mkldnn_conv_callback(ngraph::Node* node)
+{
+    auto convolution = static_cast<const T*>(node);
+    auto arg0_rank = node->get_input_shape(0).size();
+
```
nmostafa: minor nit: eliminate extra new lines.

test/cpu_test.cpp
```diff
@@ -1052,7 +1052,7 @@ TEST(cpu_test, thread_safe_calls_convolution_2d_2items)
     unset_environment("NGRAPH_CPU_CONCURRENCY");
 }
 
-TEST(cpu_test, constant_convertlayout)
+TEST(cpu_test, MLIR_DISABLE_TEST(constant_convertlayout))
```
nmostafa: Add a comment as to why it is disabled and what's needed to enable it.

test/mlir/affine_conversion/callback_ops.mlir
```diff
@@ -4,13 +4,29 @@
 
 // -----
 
+// Convbias Op
+// CHECK-LABEL: func @simple_convbias
+//       CHECK: %[[GA0:.*]] = llvm.mlir.addressof @{{[a-zA-Z_][a-zA-Z0-9_]*}} : !llvm<"{ i8, [3 x i64], [3 x i64], [3 x i64], [3 x i64] }*">
+//       CHECK: %[[C0:.*]] = constant {{[0-9]+}} : i64
```
nmostafa: Use CHECK-DAG since there is no ordering constraints on the constant/cast ops ?

src/contrib/mlir/core/pass/mlir_subgraph_extraction.cpp
ayzhuang: done.

test/cpu_test.cpp
ayzhuang: done.

test/mlir/affine_conversion/callback_ops.mlir
ayzhuang: done.

PR #4206, opened 2020-01-18
Allow topological sort in Function to be replaced

rkimballn1:
In NNP we do a slightly different topological sort in order to get certain ops to the edges of the graph. This change allows us to replace the sorting algorithm used by Function->get_ordered_ops() which is used by the pass manager.


src/ngraph/function.cpp
```diff
@@ -38,6 +48,7 @@ Function::Function(const ResultVector& results,
     , m_instance_id(m_next_instance_id.fetch_add(1))
     , m_name(name)
     , m_unique_name("Function_" + to_string(m_instance_id))
+    , m_topological_sorter(topological_sort_wrapper)
```
diyessi: Does `m_topological_sorter(topological_sort<std::vector<std::shared_ptr<Node>>>)` work?

rkimballn1: Yes it does work, no wrapper needed. Changed

PR #4207, opened 2020-01-19
Convert Function::get_ops/get_ordered_ops to return vector instead of list

rkimballn1:
I'm not sure why we chose std::list so I change get_ops and get_ordered_ops to return std::vector instead.
For the pass manager there is only one pass taking a list<Node> so I added a second virtual function that takes a vector<Node> which creates a list and calls list<Node>. The intent is we can upgrade the passes at our leisure and the remove the list<Node> version later.


jbobba:
If I recall correctly, we were going with lists to avoid mem copies due to resizing of vectors. Not sure if we ever quantified the performance impact. Do you feel strongly about moving to vectors right away?

diyessi:
> If I recall correctly, we were going with lists to avoid mem copies due to resizing of vectors. Not sure if we ever quantified the performance impact. Do you feel strongly about moving to vectors right away?

I have seen a number of places where we want a vector. If we get a chance to redesign this, we could track the number of ops in a graph/function and then make a vector that's the right size.

PR #4210, opened 2020-01-21
Faster constant folding for v1::Transpose

jbobba:


jbobba:
Will open PR to master instead.

PR #4213, opened 2020-01-22
Add provenance tags to decompose ops pass

mitruska:
Add provenance tags during fused op decomposition passes.

Example tag for ops created by Gelu decomposition:
`<Decomposed from Gelu>`


src/ngraph/pass/fused_op_decomposition.cpp
```diff
@@ -61,7 +66,6 @@ bool pass::FusedOpDecomposition::run_on_node(shared_ptr<Node> node)
         {
             for (size_t j = 0; j < output_node->get_outputs().size(); j++, i++)
             {
-                // TODO: Provenance
```
mitruska: @diyessi @jbobba Do you know something more about this `TODO`? If we still need to add any provenance here?

diyessi: I remember adding provenance during the defusion, so we don't need more.

PR #4225, opened 2020-01-25
Gauri/Add cache and registry for environment variables

gaurides:


src/ngraph/env_util.hpp
```diff
@@ -17,29 +17,145 @@
...
+        string desc;
+    };
+
     /// \brief Get the names environment variable as a string.
-    /// \param env_var The string name of the environment variable to get.
+    /// \param env_var The enum value of the environment variable to get.
     /// \return Returns string by value or an empty string if the environment
     ///         variable is not set.
-    std::string getenv_string(const char* env_var);
+    std::string getenv_string(const EnvVarEnum env_var);
```
diyessi: These should all be NGRAPH_API.

src/ngraph/env_tbl.hpp
```diff
@@ -0,0 +1,197 @@
...
+//
+// In order to use this list you want to define a macro named exactly NGRAPH_DEFINE_ENVVAR
+// When you are done you should undef the macro
+
+#ifndef NGRAPH_DEFINE_ENVVAR
+#warning "NGRAPH_DEFINE_ENVVAR not defined"
+#define NGRAPH_DEFINE_ENVVAR(ENUMID, NAME, DEFAULT, DESCRIPTION)
+#endif
+
+NGRAPH_DEFINE_ENVVAR(ngraph::EnvVarEnum::NGRAPH_CODEGEN,
```
diyessi: Call these `DEFINE_ENVAR` since `ngraph` comes from the namespace, and defining things should start with `def` not `ngr`. Also, consider `DEFINE_BOOL_ENVAR` etc. so you don't need strings like "FALSE".

src/ngraph/env_util.hpp
```diff
@@ -18,28 +18,117 @@
 
 #include <string>
 
+#include "ngraph/ngraph_visibility.hpp"
```
rkimballn1: I added env_util.hpp to ngraph so that the NNP could use this version, but now you have added code that makes it so that I can no longer use this file. Please move the env cache to its own files so that env_util.?pp files are generic and contains no ngraph nor backend specific variables. It should just be the three original functions to get int, bool, and string env vars.

src/ngraph/file_util.cpp
```diff
@@ -186,15 +186,14 @@ string file_util::get_temp_directory_path()
 {
     const vector<string> potential_tmps = {"NGRAPH_TMP", "TMPDIR", "TMP", "TEMP", "TEMPDIR"};
 
-    string path;
-    for (const string& var : potential_tmps)
+    string path = getenv_string(ngraph::EnvVarEnum::NGRAPH_TMPDIR);
+    /*for (const string& var : potential_tmps)
```
rkimballn1: Don't use multi-line comments
Remove the code or leave it.
Why was this removed? It is useful.

src/ngraph/env_tbl.hpp
```diff
@@ -0,0 +1,201 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
rkimballn1:
```suggestion
// Copyright 2017-2020 Intel Corporation
```

src/ngraph/env_tbl.hpp
```diff
@@ -0,0 +1,201 @@
...
+                     "Enable compiler debug info when codegen is enabled")
+NGRAPH_DEFINE_ENVVAR(ngraph::EnvVarEnum::NGRAPH_COMPILER_DIAG_ENABLE,
+                     "NGRAPH_COMPILER_DIAG_ENABLE",
+                     "FALSE",
+                     "Enable compiler diagnostics when codegen is enabled")
+NGRAPH_DEFINE_ENVVAR(ngraph::EnvVarEnum::NGRAPH_COMPILER_REPORT_ENABLE,
+                     "NGRAPH_COMPILER_REPORT_ENABLE",
+                     "FALSE",
+                     "")
+NGRAPH_DEFINE_ENVVAR(ngraph::EnvVarEnum::NGRAPH_CPU_BIN_TRACER_LOG,
```
rkimballn1: This is an ngraph core file, not a backend file. Backend definitions belong in a backend file. This comment is for all non-ngraph-core enums

src/ngraph/env_util.hpp
```diff
@@ -18,28 +18,117 @@
...
+#include "ngraph/util.hpp"
+
+using namespace std;
+
 namespace ngraph
 {
+    /// \brief List of currently supported environment variables
+    enum class EnvVarEnum : uint32_t
+    {
+        NGRAPH_CODEGEN = 0,
```
rkimballn1: This should be auto-generated from the defines you made earlier in env_tbl.hpp

src/ngraph/env_util.hpp
```diff
@@ -18,28 +18,117 @@
...
 
+#include "ngraph/ngraph_visibility.hpp"
+#include "ngraph/util.hpp"
+
+using namespace std;
+
 namespace ngraph
 {
+    /// \brief List of currently supported environment variables
+    enum class EnvVarEnum : uint32_t
```
rkimballn1: This technique does not allow for other backends to add env vars. If we are to have a solution it should work for backends as well as ngraph

src/ngraph/env_util.hpp
gaurides: ok, will move cache and registry it to it's own file.

src/ngraph/file_util.cpp
gaurides: I added a new environment variable `NGRAPH_TMPDIR` to specify temp directory and removed the ambiguity of potential temp values.

src/ngraph/codegen/compiler.cpp
```diff
@@ -157,9 +157,9 @@ static std::string GetExecutablePath(const char* Argv0)
 }
 
 codegen::CompilerCore::CompilerCore()
-    : m_debuginfo_enabled(getenv_bool("NGRAPH_COMPILER_DEBUGINFO_ENABLE"))
-    , m_enable_diag_output(getenv_bool("NGRAPH_COMPILER_DIAG_ENABLE"))
-    , m_enable_pass_report(getenv_bool("NGRAPH_COMPILER_REPORT_ENABLE"))
+    : m_debuginfo_enabled(getenv_bool(ngraph::EnvVarEnum::NGRAPH_COMPILER_DEBUGINFO_ENABLE))
+    , m_enable_diag_output(getenv_bool(ngraph::EnvVarEnum::NGRAPH_COMPILER_DIAG_ENABLE))
+    , m_enable_pass_report(getenv_bool(ngraph::EnvVarEnum::NGRAPH_COMPILER_REPORT_ENABLE))
```
jbobba: Do we need ngraph:: ?

Issue #4209, opened 2020-01-21
"switch (dt)" does not include case "PLAIDML_DATA_BFLOAT16" in plaidml_translate.cpp


dbonner:
see [https://github.com/tensorflow/ngraph-bridge/issues/447](url)

I added under "switch (dt)" in this file:
case PLAIDML_DATA_BFLOAT16: return "as_bfloat16(" + tensor_name + ", 16)";

It then built successfully.

PR #4226, opened 2020-01-25
Update plaidml_translate.cpp to include bfloat16

dbonner:
ngraph-tensorflow-bridge fails to build with plaidml backend unless case PLAIDML_DATA_BFLOAT16 is included in switch (dt).


dbonner:
This fixes #4209

diyessi:
Moved from fork to branch in #4237

PR #4230, opened 2020-01-27
add set_parameters, set_results and serialize types for r28 

ashokei:


ashokei:
@rkimballn1 the set_parameters and set_results seem generic enough we wanted expose at base class level.

rkimballn1:
No, it is not generic enough. Generic enough is the constructor. Adding to the Executable API is so that the framework can modify these values and for no other purpose. The framework does not modify these values.

PR #4231, opened 2020-01-27
Fetch gtest and use exposed cmake target 

silee2:
Fetch googletest at fetch time and add it as a sub-directory
This simplifies dependency management since nGraph can just link exposed targets `gtest` and `gmock`


silee2:
#4582 replaces this one.

PR #4232, opened 2020-01-27
Max Shape propagation

nbpatel:


src/ngraph/op/experimental/range.cpp
```diff
@@ -240,8 +240,10 @@ void op::Range::validate_and_infer_types()
 #if defined(__GNUC__) && !(__GNUC__ == 4 && __GNUC_MINOR__ == 8)
 #pragma GCC diagnostic pop
 #endif
-
-    set_output_type(0, result_et, result_shape);
+    if (!result_shape.is_dynamic())
+    {
+        set_output_type(0, result_et, result_shape);
```
ashokei: dont we need to set the output shape to dynamic {} if it cannot infer out shape. Also, is this change needed for this feature , as we propagate in pass.

src/ngraph/node.cpp
```diff
@@ -333,6 +333,18 @@ void Node::set_output_type(size_t i,
     descriptor.set_partial_shape(partial_shape);
 }
 
+void Node::set_output_max_partial_shape(size_t i, const PartialShape& partial_shape)
```
ashokei: why not use @diyessi 's PR , where you have set_xx_max_xx APIs on inputs/outputs, instead of node.

test/backend/range.in.cpp
```diff
@@ -81,3 +82,46 @@ NGRAPH_TEST(${BACKEND_NAME}, range)
...
+    passes.register_pass<pass::MinMaxShapePropagation>();
+    passes.run_passes(f);
+
+    auto t_start = backend->create_tensor(element::i32, Shape{});
+    copy_data(t_start, vector<int32_t>{0});
+    auto t_stop = backend->create_tensor(element::i32, Shape{});
+    copy_data(t_stop, vector<int32_t>{10});
+    auto t_step = backend->create_tensor(element::i32, Shape{});
+    copy_data(t_step, vector<int32_t>{1});
+    auto result = backend->create_dynamic_tensor(element::i32, PartialShape::dynamic());
```
ashokei: @sayantan-nervana see this unit-test on how max shape is propagated , and can be used by backend. We will need to explore how to use executable->create_outpu_tensor here instead of create_dynamic_tensor (not sure TF bridge uses this).

src/ngraph/node.cpp
nbpatel: this is to set the output shape to max shape and hide the descriptor mess behind this api in node. We might not need the getter for it probably.

src/ngraph/op/experimental/range.cpp
nbpatel: We defer the shape inference to runtime if we dont run the max propagation pass. 
If we run the pass the output shape is being set to max from the pass

diyessi: Let's assume we always run the max propagation pass. No need to make things complicated for a pass that should be low-overhead if no maxes have been set.

test/backend/range.in.cpp
```diff
@@ -81,3 +82,46 @@ NGRAPH_TEST(${BACKEND_NAME}, range)
...
+NGRAPH_TEST(${BACKEND_NAME}, range_subgraph)
+{
+    // Create a graph for f(start,stop,step) = Range(start,stop,step).
+    auto start = make_shared<op::Parameter>(element::i32, Shape{});
+    auto stop = make_shared<op::Parameter>(element::i32, Shape{});
+    auto step = make_shared<op::Parameter>(element::i32, Shape{});
+    PartialShape out_max_shape{15};
+
+    auto range = make_shared<op::Range>(start, stop, step);
+    range->output(0).set_max_partial_shape(out_max_shape);
```
diyessi: I think we are likely to see the `set_max_partial_shape` happen after the graph is fully constructed (for example, from an annotation file) so let's move it down to the end to avoid any dependency on validate and infer shapes.

src/ngraph/node.cpp
diyessi: These are okay; sometimes output is convenient, sometimes node is convenient.

src/ngraph/op/experimental/range.cpp
nbpatel: The problem is the max output shape gets overwritten by some other pass or elsewhere. we call the validate and infer a bunch of times from various places in the code.

PR #4238, opened 2020-01-28
[MLIR] Use enum class for broadcast hint.

ayzhuang:


src/contrib/mlir/runtime/cpu/cpu_callbacks.cpp
```diff
@@ -713,6 +713,10 @@ static void __mlir_cblas_sgemm_with_bias(StaticMemRef* memRefmatA,
                            matOut,
                            std::max<size_t>(1, ldc));
     }
+    else
+    {
+        NGRAPH_UNREACHABLE("Unsupported broadcast");
+    }
```
dcaballe: IIRC, we've been using 'llvm_unreachable' and 'assert' in mlir-only files. Could you please check if that is the case and change this accordingly?

ayzhuang: Didn't find "llvm_unreachable" in ngraph source code. Only "NGRAPH_UNREACHABLE".

PR #4239, opened 2020-01-29
Added const specifier for argument of specialize_function

ilyachur:


ilyachur:
@diyessi From my understanding `specialize_function` is needed to create a copy of original nGraph function. And in this case I want to be sure that this method doesn't change the original function. Const specifier helps me to be sure in this.
Also I thought that we shouldn't use `specialize_function` often. Am I not right?

PR #4245, opened 2020-01-29
[ONNX] Add new ConstantOfShape operator

tsocha:


src/ngraph/frontend/onnx_import/op/constant_of_shape.cpp
```diff
@@ -0,0 +1,54 @@
...
+                        auto value_tensor = node.get_attribute_value<Tensor>("value");
+                        constant_value = value_tensor.get_ng_constant();
+                        constant_value = builder::opset1::squeeze(constant_value);
+                    }
+                    else
+                    {
+                        constant_value = default_opset::Constant::create(element::f32, {}, {0});
+                    }
+                    return {std::make_shared<default_opset::Broadcast>(constant_value,
+                                                                       node.get_ng_inputs()[0])};
```
tomdol: Just to avoid segfaults, can you change `[0]` to `.at(0)`?

test/onnx/onnx_import.in.cpp
```diff
@@ -1904,3 +1904,59 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_reciprocal)
 
     test_case.run();
 }
+
+NGRAPH_TEST(onnx_${BACKEND_NAME}, model_constant_of_shape_float_zeros)
```
tomdol: I don't see that as a big problem, but we have a separate file for dynamic shapes related unit tests, but you can consider moving these 2 tests to `onnx_import_dyn_shapes.in.cpp`.

test/onnx/onnx_import_dyn_shapes.in.cpp
```diff
@@ -175,3 +175,59 @@ NGRAPH_TEST(onnx_dyn_shapes_${BACKEND_NAME}, dynamic_rank_input_inference)
         test_case.run();
     }
 }
+
+NGRAPH_TEST(onnx_${BACKEND_NAME}, model_constant_of_shape_float_zeros)
+{
+    auto function = onnx_import::import_onnx_model(file_util::path_join(
+        SERIALIZED_ZOO, "onnx/dynamic_shapes/constant_of_shape_float_zeros.prototxt"));
+
+    auto backend = runtime::Backend::create("${BACKEND_NAME}", true);
```
postrational: Let's refactor these tests to use TestCase with dynamic backend.

src/ngraph/pass/opset0_downgrade.cpp
```diff
@@ -148,49 +148,67 @@ namespace
...
+                                       make_constant_from_string("1", element::i64, {})));
         }
-        // Check if arg_shape contains some more empty dimensions marked to broadcast.
-        // If axes_mapping size is less than arg_shape size, then some of arg dimensions may
-        // be equal to one and marked to broadcast.
-        if (axes_mapping.size() < arg_shape.size())
+        else
         {
-            for (size_t a{axes_mapping.size()}; a < arg_shape.size(); ++a)
+            const auto& arg_shape = arg_pshape.to_shape();
```
mbencer: I think that we should add assert to check if `arg_pshape` is static here.

tsocha: Done

postrational:
TESTNOW_DLDT_E2E

PR #4248, opened 2020-01-29
Changed ONNX importer to convert Convolution operation to V1 ops

GlebKazantaev:


src/ngraph/frontend/onnx_import/op/conv.cpp
```diff
@@ -129,13 +129,15 @@ namespace ngraph
                     }
 
                     auto bias = inputs.at(2);
-                    const Shape& new_shape = conv_node->get_shape();
+                    Shape new_shape(conv_node->get_shape().size(), 1);
+                    new_shape[1] = conv_node->get_shape()[1];
 
-                    auto broadcasted_bias = std::make_shared<ngraph::opset0::Broadcast>(
+                    auto reshaped_bias = std::make_shared<ngraph::opset1::Reshape>(
```
postrational:
```suggestion
                    auto reshaped_bias = std::make_shared<default_opset::Reshape>(
```

src/ngraph/frontend/onnx_import/op/conv.cpp
```diff
@@ -129,13 +129,15 @@ namespace ngraph
...
                         bias,
-                        new_shape,
-                        ngraph::op::calculate_broadcast_axes(new_shape, bias->get_shape(), 1));
-                    return {std::make_shared<ngraph::opset0::Add>(conv_node, broadcasted_bias)};
+                        ngraph::op::Constant::create(
```
postrational:
```suggestion
                        default_opset::Constant::create(
```

postrational:
Once this is merged, please make a similar PR to `master`, so this change is not lost.

arogowie-intel:
@jbobba Could you say whether `Reshape` will work with nGraph fusions? I see that there is `construct_conv_bias()` https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/pass/core_fusion.cpp#L1202 fusion but I'm not sure whether it will cover this.

suryasidd:
With these changes, i was able to verify that Convolution biases are getting fused into Convolution

GlebKazantaev:
@arogowie-intel we do not use nGraph transformations just because they are based on opset0 ops except ConstantFolding pass that we asked to update with opset1 support.

PR #4271, opened 2020-02-03
Introduce customized namespace for ONNX library

Ewa21:
This PR's main goal is to link directly to ONNX library. Hence, the namespace `ngraph_onnx` is applied. 
This is the second step (after upgrading ONNX to 1.6 version, PR #4244) to provide support for function in ONNX models.


Ewa21:
> There needs to be more explanation of what problem is being solved here. For example, why couldn't you use a namespace alias instead of a preprocessor macro?

 I've used the macro that ONNX library provides.

postrational:
Let's wait with merging this until other cmake changes are added in first. After the 2020.2 release is done we can come back and merge this.

diyessi:
@mbencer can you add the fully reviewed label if this is ready to merge? Thanks.

PR #4272, opened 2020-02-03
Revert "add set_parameters, set_results and serialize types (#4168)"

rkimballn1:
This reverts commit 431720687e3d53c21940a5426faca19958fdbbda.

This API is no longer needed


src/ngraph/serializer.hpp
```diff
@@ -30,13 +30,6 @@ namespace ngraph
     ///    indent level specified.
     std::string serialize(std::shared_ptr<ngraph::Function> func, size_t indent = 0);
 
-    /// \brief Serialize given vector of shapes/types
-    /// \param types The vector of shape/types to serialize
-    std::string serialize_types(const std::vector<std::pair<PartialShape, element::Type>>& types);
```
ashokei: @rkimballn1 these utilities can be useful, independent of set_params/results.

src/ngraph/runtime/executable.hpp
```diff
@@ -73,14 +73,6 @@ class NGRAPH_API ngraph::runtime::Executable
     /// \returns  preferred pipeline_depth
     virtual size_t get_preferred_pipeline_depth() const;
 
-    /// \brief Set the input Parameters
-    /// \param params ngraph::ParameterVector of all input parameters
-    void set_parameters(const ngraph::ParameterVector& params);
```
ashokei: these APIs were originally added by @jiminha in r0.25 for a specific use-case in nnp-i transformer, can we work with her to close on that use-case before reverting.

PR #4277, opened 2020-02-03
reshape_sinking on reduce op mix max

ashokei:


src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -586,6 +626,14 @@ bool ngraph::pass::ReshapeSinking::run_on_function(shared_ptr<ngraph::Function>
         {
             sink_concat(concat, reorders, reshapes_to_delete);
         }
+        else if (auto reduce = as_type_ptr<op::Max>(n))
```
jbobba: Can we use the same logic and handle all reductions (Arithmetic, Logical, Index)?

ashokei: i think so as they all reduce axis similarly, we can enable it. This PR only enables reducing on one axis, as i havent fullly tested reducing on multiple axis. 
Also, v1 ops that have keep_dims  need to be handled slightly differently.

src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -106,9 +108,9 @@ static void insert_reshape(shared_ptr<Node> target, shared_ptr<Node> reshape, si
 
 static void delete_reshape(shared_ptr<Node> reshape)
 {
-    NGRAPH_DEBUG << "Removing reshape " << reshape->get_name();
     if (!reshape->get_users().empty())
     {
+        NGRAPH_DEBUG << "Removing reshape " << reshape->get_name();
```
diyessi: `NGRAPH_DEBUG << "Removing shape " << reshape;`

src/ngraph/pass/reshape_sinking.cpp
diyessi: Why are you using a template? Using `shared_ptr<LogicalReduction>` should work.
It looks like `IndexReduction` and `LogicalReduction` are suffering from opset1 and dynamic shape change collisions and need to be made consistent.

src/ngraph/pass/visualize_tree.cpp
```diff
@@ -376,6 +377,10 @@ string pass::VisualizeTree::get_attributes(shared_ptr<Node> node)
             label << " " << (node->get_output_size() != 1
                                  ? string("[skipped]")
                                  : pretty_partial_shape(node->get_output_partial_shape(0)));
+            if (auto reshape = dynamic_cast<const op::Reshape*>(&(*node)))
```
diyessi: Use `auto reshape = as_type_ptr<op::Reshape>(node)`. dynamic cast has subtle problems on some platforms.

ashokei: will make the change, since node isnt passed as reference i wasnt sure as_type_ptr would work. Looks like as_type_ptr does support val and ref.

src/ngraph/pass/reshape_sinking.cpp
ashokei: inside this function we need to know the actual op type as we create a new replacement op. Using shared_ptr<LogicalReduction> works to keep things common, but we will have to identify op type for creation. I looked at `copy_with_new_args` but that doesnt work seamlessly either. Template allowed to pass type for op (re)creation.

src/ngraph/pass/reshape_sinking.cpp
ashokei: nice! i recall recent PRs in master has some nice print handling, will convert to this.

src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -508,6 +510,44 @@ static void sink_dequantize(shared_ptr<op::Dequantize> dequantize,
...
+        auto dummy_correct_shape =
+            make_shared<pattern::op::Label>(arg_reshape->get_element_type(), input_shape);
+        AxisSet new_axis_set;
+        for (auto& i : reduce->get_reduction_axes())
+        {
+            new_axis_set.emplace(order[i]);
+        }
+        auto new_reduce = make_shared<T>(dummy_correct_shape, new_axis_set);
+
+        ngraph::replace_node(dummy_correct_shape, reduce->get_argument(0));
```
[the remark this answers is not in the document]
ashokei: agree this is not ideal way to do this, this entire pass uses this approach unfortunately. Will see if we can refactor.

src/ngraph/pass/reshape_sinking.cpp
diyessi: You could do this with `copy_with_new_inputs` or the node factory and then set the attributes , but that is better done with a separate major refactoring PR.

PR #4279, opened 2020-02-04
Add some attribute visitors

diyessi:


aslepko:
TESTNOW_DLDT_E2E

aslepko:
TESTNOW_DLDT_E2E

src/ngraph/attribute_adapter.cpp
```diff
@@ -201,6 +201,26 @@ namespace ngraph
         m_buffer_valid = false;
     }
 
+#ifdef __APPLE__
+    // size_t is not uint_64t on OSX
+    constexpr DiscreteTypeInfo AttributeAdapter<size_t>::type_info;
```
ilyachur: Can we change size_t to uint64_t? It should help to avoid this code. Moreover for other platforms size_t also can have size which is not equal uint64_t.

src/ngraph/op/convolution.cpp
```diff
@@ -45,6 +45,16 @@ op::v1::Convolution::Convolution(const Output<Node>& data_batch,
     constructor_validate_and_infer_types();
 }
 
+bool op::v1::Convolution::visit_attributes(AttributeVisitor& visitor)
+{
+    visitor.on_attribute("strides", m_strides);
+    visitor.on_attribute("dilations", m_dilations);
+    visitor.on_attribute("pads_begin", m_pads_begin);
+    visitor.on_attribute("pads_end", m_pads_end);
+    visitor.on_attribute("auto_pad", m_auto_pad);
```
ilyachur: What is the default value of auto_pad in this case?

We are using ngraph::op::PadType::EXPLICIT as default value. Could we specify it here?

src/ngraph/op/divide.cpp
```diff
@@ -93,6 +100,13 @@ op::v1::Divide::Divide(const Output<Node>& arg0,
     constructor_validate_and_infer_types();
 }
 
+bool op::v1::Divide::visit_attributes(AttributeVisitor& visitor)
+{
+    BinaryElementwiseArithmetic::visit_attributes(visitor);
+    visitor.on_attribute("m_pythondiv", m_pythondiv);
```
ilyachur: What is it `m_pythondiv`?

Also in spec Divide can have only one attribute: `auto_broadcast`.

src/ngraph/op/fused/depth_to_space.hpp
```diff
@@ -56,25 +57,43 @@ namespace ngraph
                 /// \param block_size The size of the block of values to be moved
                 DepthToSpace(const Output<Node>& data,
                              const DepthToSpaceMode& mode,
-                             std::size_t block_size = 1);
+                             size_t block_size = 1);
```
ilyachur: can it be changed to int64_t?

src/ngraph/op/divide.cpp
diyessi: Python and C/C++ do division differently; models expecting the python behavior need the python behavior. For example, https://stackoverflow.com/questions/828092/python-style-integer-division-modulus-in-c

src/ngraph/attribute_adapter.cpp
diyessi: This works as long as size_t ins't bigger than int64_t. Hopefully we can get the size_ts changed before we hit 128-bit adddresses.

src/ngraph/op/convolution.cpp
diyessi: Can't specify it here, and we can't default it because it's framework specific. We could have something to set a global to be used as a default.

src/ngraph/op/fused/depth_to_space.hpp
diyessi: Let's do that separately. But we can undo this particular change which was a partial backing out of a failed attempt to change to int64_t here.

PR #4281, opened 2020-02-04
ngraph 20% faster bert-lrn 

pawelpiotrowicz:
POC . only to see - do not merge.

If you want to review this make sure you set the `Hide whitespace changes` before reviewing. It will make your life much better.

image


pawelpiotrowicz:
diff_result

src/ngraph/coordinate.cpp
```diff
@@ -55,6 +55,12 @@ ngraph::Coordinate::Coordinate(const Coordinate& axes)
 ngraph::Coordinate::Coordinate(size_t n, size_t initial_value)
     : std::vector<size_t>(n, initial_value)
 {
+    static size_t max = 0;
+    if(n > max) 
+    {
+       max = n;
+      std::cout << "Coordinate : New Max " << max  << std::endl;
```
diyessi: Remove or use logging.

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+            LRUCache();
+
+            virtual ~LRUCache();
+
+            void add_entry(const std::vector<int>& shape,
+                           std::shared_ptr<Executable> exec,
+                           std::shared_ptr<Function> func);
+            bool is_cached(const std::vector<int>& shape);
+            std::shared_ptr<Executable> get_cached_entry(const std::vector<int>& shape);
```
diyessi: Change these `int`s to the appropriate `int32_t` or `int64_t`.

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,130 @@
...
+    ostringstream key;
+    // check if the list is empty
+    if (m_list.size() == m_cache_size)
+    {
+        ostringstream key;
+        convert_shape_to_string(m_list.back(), key);
+        m_list.pop_back();
+        m_map.erase(key.str());
+    }
+
```
diyessi: Why convert the shape to a string instead of just using the shape?

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -120,130 +120,206 @@ bool runtime::dynamic::DynamicExecutable::call(
...
-        for (auto& input : inputs)
+        if (m_wrapped_function->get_parameters()[loop_count]->is_relevant_to_shapes())
         {
-            if (m_wrapped_function->get_parameters()[i]->is_relevant_to_shapes())
+            // Caching on values of Shape relevant inputs
+            void* data_ptr = malloc(input->get_size_in_bytes());
+            input->read(data_ptr, input->get_size_in_bytes());
+            // assuming that shape relevant inputs are int64_t
+            int64_t* result = (int64_t*)data_ptr;
+            for (int i = 0; i < input->get_element_count(); i++)
```
diyessi: Just allocate an int64_t* directly; you can pass that to anything that needs a void* and it will be properly aligned. No need for this casting. Use a unique_ptr for it and then we don't need to worry about freeing it in the future if someone rearranges the control flow.

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <unordered_map>
+#include "ngraph/function.hpp"
+#include "ngraph/runtime/executable.hpp"
+#include "ngraph/shape.hpp"
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        class LRUCache : public std::enable_shared_from_this<LRUCache>
```
rkimballn1: This needs a doc string describing what it is caching. Just calling the file cache.?pp is not helpful. What is it caching? Change the name and document better.
Calling the cache LRUCache is not a good idea. You should not expose the implementation method in the name of the cache. What if you change to a new algorithm, no LRU? Do you rename the class breaking users or do you leave it misleadingly named LRUCache. The name of the class should be WHAT it is caching, not HOW it is caching it.

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,61 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
rkimballn1:
```suggestion
// Copyright 2017-2020 Intel Corporation
```

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,130 @@
+//*****************************************************************************
+// Copyright 2017-2019 Intel Corporation
```
rkimballn1:
```suggestion
// Copyright 2017-2020 Intel Corporation
```

src/ngraph/runtime/cache.hpp
rkimballn1: Or you could use Shape instead

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,61 @@
...
+                           std::shared_ptr<Executable> exec,
+                           std::shared_ptr<Function> func);
+            bool is_cached(const std::vector<int>& shape);
+            std::shared_ptr<Executable> get_cached_entry(const std::vector<int>& shape);
+            void convert_shape_to_string(const std::vector<int>& shape, std::ostringstream& key);
+            std::shared_ptr<Function> get_cloned_function(const std::vector<int>& shape);
+
+        private:
+            int m_cache_size;
+            GraphCache m_map;
```
rkimballn1: Since GraphCache is used in one place only you could make the code easier to read if you got rid of the using = statement and just replace GraphCache with the map definition.

src/ngraph/runtime/cache.hpp
```diff
@@ -0,0 +1,61 @@
...
+            ClonedFunctionMap m_clone_function_map;
+            std::list<std::vector<int>> m_list;
```
rkimballn1: Since this list contains Shapes why not name it something descriptive like m_shapes_list?

src/ngraph/runtime/cache.hpp
rkimballn1: m_map is not descriptive.

src/ngraph/runtime/cache.cpp
```diff
@@ -0,0 +1,130 @@
...
+        throw ngraph_error("Entry not found in cache");
+    }
+    else
+    {
+        // update list to push this reference to the front
+        for (auto itr = m_list.begin(); itr != m_list.end(); itr++)
+        {
+            if (*itr == shape)
+            {
+                m_list.remove(shape);
```
rkimballn1: You should be able to .remove(itr) to be more efficient. The call you are making here searches the m_list again to find the shape to remove.

src/ngraph/runtime/cache.hpp
rkimballn1: A doc string stating that the items are stored in order of age with the youngest at the head would be most helpful.

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -120,130 +120,206 @@ bool runtime::dynamic::DynamicExecutable::call(
...
-        std::vector<void*> arg_value_base_pointers(inputs.size());
-
-        size_t i = 0;
-
-        for (auto& input : inputs)
+        if (m_wrapped_function->get_parameters()[loop_count]->is_relevant_to_shapes())
         {
-            if (m_wrapped_function->get_parameters()[i]->is_relevant_to_shapes())
+            // Caching on values of Shape relevant inputs
+            void* data_ptr = malloc(input->get_size_in_bytes());
```
rkimballn1: Using malloc this way is risky. If read below throws then you leak. If you exit this method for any reason before you get to free then you leak. Do as @diyessi says and use a unique_ptr or I like using runtime::AlignedBuffer for temporaries.

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -120,130 +120,206 @@ bool runtime::dynamic::DynamicExecutable::call(
...
+        std::vector<std::shared_ptr<runtime::Tensor>> wrapped_inputs;
+        std::vector<element::Type> arg_element_types;
+        std::vector<PartialShape> arg_shapes;
 
-        num_dyn_nodes_last_pass = num_dyn_nodes_this_pass;
-    }
+        std::shared_ptr<Function> clone;
+        {
+            // We'll use AlignedBuffers to back the base pointers, storing them in this vector for
+            // RAII
```
rkimballn1: I hate the way clang-format wraps comments :(

src/ngraph/runtime/reference/broadcast.hpp
```diff
@@ -55,10 +55,14 @@ namespace ngraph
                 }
                 CoordinateTransform input_transform(adjusted_in_shape);
                 CoordinateTransform output_transform(out_shape);
-
+                auto v = adjusted_axes.to_vector();
+              //  std::sort(adjusted_axes.begin(),adjusted_axes.end());
```
rkimballn1: remove debug code here and below

src/ngraph/shape_util.hpp
```diff
@@ -56,9 +57,50 @@ namespace ngraph
...
+        while(b!=e)
+        {
+             if( *b > v )
+             {
+                return e;
+             }
+             else if ( *b == v)
+             {
+               return b;
+             } else {
```
rkimballn1: You need to run make style on this

src/ngraph/shape_util.hpp
```diff
@@ -56,9 +57,50 @@ namespace ngraph
         return project(axis_values, axes);
     }
 
+    template<class I,class V>
```
rkimballn1: Naming things I and V and not putting any doc strings makes this unfathomable. What is this doing?

PR #4283, opened 2020-02-04
fix reshape_sinking for broadcast & single user swim only

ashokei:
This fixes issues with broadcast swimming on more than one dim change. 
Also has fix for swim logic crashing when graph has nodes with more than one user ( @csullivan 's had PR on this for 0.25, bringing to master)


src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -166,6 +166,18 @@ void swim(Input<Node> input, shared_ptr<op::Reshape> reshape)
         auto csw = work_queue.front();
         work_queue.pop_front();
         auto n = csw.input.get_source_output().get_node_shared_ptr();
+        auto materialize = [csw, n]() {
```
csullivan: Thanks for bringing this bug fix to master @ashokei, I thought we had previously but guess not!

PR #4287, opened 2020-02-05
handle reshapes that squeeze/expand dim size 1

ashokei:


src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -313,14 +343,9 @@ static void sink_reshape(shared_ptr<op::Reshape> reshape,
...
-        auto new_reshape = combine_reshapes(orig_reshape, reshape);
+        new_reshape = combine_reshapes(orig_reshape, reshape);
+        new_reshape_in_map = new_reshape;
         // remove original reshape now it's combined with a new one
-        // should be safe to remove an already detached node
-        mark_reshape_for_deletion(orig_reshape, reshapes_to_delete);
-        // replace reshape with combined one
-        ngraph::replace_node(reshape, new_reshape);
-        mark_reshape_for_deletion(new_reshape, reshapes_to_delete);
-        write_reshapemap(reorders, new_reshape, new_reshape);
```
jbobba: why are we removing this?

src/ngraph/pass/reshape_sinking.cpp
```diff
@@ -294,15 +294,45 @@ static void materialize_shapes(shared_ptr<Node> n,
...
 }
 
+static bool is_shape_diff_one(const Shape& s1, const Shape& s2)
+{
+    Shape ss, ss1(s1), ss2(s2);
+    sort(ss1.begin(), ss1.end());
+    sort(ss2.begin(), ss2.end());
+    set_symmetric_difference(ss1.begin(), ss1.end(), ss2.begin(), ss2.end(), back_inserter(ss));
+    return ss.size() == 1 && ss[0] == 1;
+}
```
jbobba: This could be added as a nice utility to the Shape class. `s1.is_expands(s2)`

src/ngraph/pass/reshape_sinking.cpp
ashokei: missed a call to `replace_reshape();` will add.

PR #4290, opened 2020-02-05
Add OneCCL distributed backend

adk9:
Add a distributed backend to use Intel OneCCL (https://github.com/intel/oneccl)
to support hybrid/sparse/unordered collectives.


adk9:
> Distributed support belongs in the framework not in core.

Fair enough. But I do think that distributed support is important in core for two reasons:

1. There are several compiler optimizations related to communication that can accelerate distributed training, in case of both data and model parallelism.
2. True to nGraph's original motivation, there is incredible value in taking a single-device program and transparently running it on other devices or multiple devices – whether across multiple nodes or across multiple accelerators on the same node.

Having said that, if distributed support doesn't belong in core does it mean that we deprecate the MLSL and MPI backends and remove `BroadcastDistributed` and `Allreduce` ops?

diyessi:
Core provides an interface for the distributed ops to call, but the implementation of the interface belongs outside of core in the component invoking core, since that is the component that knows which distributed system is in use. It doesn't work well if at all to have two different ones in use, or two copies of the same one. 
So a bridge should be supplying an implementation of the distributed interface that is using the framework's distribution choice.

adk9:
Distributed implementations were removed in nGraph (#4380). Closing this as invalid now.

PR #4291, opened 2020-02-06
[SPEC][ONNX] Handle negative axis for TopK:v1, add dynamic shape support for ONNX Arg Min/Max ops

mbencer:
- according to spec `TopK:v1` should handle negative axis which was fixed (to serialization origin value is taken)
- ONNX `ArgMin` and `ArgMax` implementation was changed in order to support dynamic shapes


src/ngraph/op/topk.hpp
```diff
@@ -161,8 +161,13 @@ namespace ngraph
...
                 virtual size_t get_version() const override { return 1; }
-                size_t get_axis() const { return m_axis; }
-                void set_axis(const size_t axis) { m_axis = axis; }
+                size_t get_axis() const { return m_normalized_axis; }
+                int64_t get_origin_axis() const { return m_axis; }
+                void set_axis(const size_t axis)
+                {
+                    m_normalized_axis = axis;
+                    m_axis = axis;
+                }
```
tsocha: We should consider change `set_axis` to accept `int64_t` instead of `size_t` here.

src/ngraph/op/topk.hpp
```diff
@@ -161,8 +161,13 @@ namespace ngraph
                     copy_with_new_args(const NodeVector& new_args) const override;
 
                 virtual size_t get_version() const override { return 1; }
-                size_t get_axis() const { return m_axis; }
-                void set_axis(const size_t axis) { m_axis = axis; }
+                size_t get_axis() const { return m_normalized_axis; }
+                int64_t get_origin_axis() const { return m_axis; }
```
postrational: Consider a different name.

diyessi: `get_specified_axis` ?

src/ngraph/op/topk.hpp
diyessi: Replace the `size_t`s with `int64_t`.
size_t is very platform-specific.

src/ngraph/op/topk.cpp
```diff
@@ -277,20 +280,26 @@ void op::v1::TopK::validate_and_infer_types()
                                       get_input_element_type(1));
     }
 
+    if (m_axis >= 0)
+    {
+        m_normalized_axis = m_axis;
+    }
+    else
+    {
+        NODE_VALIDATION_CHECK(this,
```
tomdol: Just wondering - can we postpone the normalization until the shape is known? Perhaps we can move it further to the `if` that checks if the `output_shape` rank is static? If we perform this check here, we won't be able to construct a graph that might still be valid when static shapes are passed in.

src/ngraph/op/topk.hpp
tomdol: Or turn it around - return `m_axis` from `get_axis()` and `m_normalized_axis` from `get_normalized_axis()`

arogowie-intel: @tomdol's conventions seems to me the most clear.

src/ngraph/op/topk.hpp
arogowie-intel: If changing to `int64_t` call `normalize_axis()`.

test/onnx/onnx_import_dyn_shapes.in.cpp
```diff
@@ -175,3 +175,45 @@ NGRAPH_TEST(onnx_dyn_shapes_${BACKEND_NAME}, dynamic_rank_input_inference)
         test_case.run();
     }
 }
+
+NGRAPH_TEST(onnx_dyn_shapes_${BACKEND_NAME}, arg_max_dyn_shape)
+{
+    // the model contains a single Add operation that takes a fully dynamic input and a scalar
```
arogowie-intel: Please update/remove this comment.

test/onnx/onnx_import_dyn_shapes.in.cpp
```diff
@@ -175,3 +175,45 @@ NGRAPH_TEST(onnx_dyn_shapes_${BACKEND_NAME}, dynamic_rank_input_inference)
...
+
+    std::vector<int64_t> expected_values{1, 1, 1, 1, 1, 1};
+    test_case.add_expected_output<int64_t>(Shape{3, 1, 2}, expected_values);
+
+    test_case.run();
+}
+
+NGRAPH_TEST(onnx_dyn_shapes_${BACKEND_NAME}, arg_min_no_keep_dims_dyn_shape)
+{
+    // the model contains a single Add operation that takes a fully dynamic input and a scalar
```
arogowie-intel: Please update/remove this comment.

src/ngraph/op/topk.hpp
mbencer: @arogowie-intel @tomdol I agree that `get_normalized_axis()` is the most clear, but current backends use `get_axis` method to read axis. I think that such change requires spec change.
My proposition is `get_raw_axis`

src/ngraph/op/topk.hpp
mbencer: Done, changed to `int64_t`. setter call normalization too.
It will be refactored after https://github.com/NervanaSystems/ngraph/pull/4299 merge.

src/ngraph/op/topk.cpp
mbencer: I agree with you. Axis is used to calculate output shape only for static rank case, so restrictions can be reduced. As a result `get_axis` and `set_axis` must be modified (in current version exception can be thrown in `get_axis`)

test/onnx/onnx_import_dyn_shapes.in.cpp
mbencer: done

test/onnx/onnx_import_dyn_shapes.in.cpp
mbencer: done

src/ngraph/op/topk.hpp
```diff
@@ -161,13 +161,9 @@ namespace ngraph
...
                 virtual size_t get_version() const override { return 1; }
-                size_t get_axis() const { return m_normalized_axis; }
-                int64_t get_origin_axis() const { return m_axis; }
-                void set_axis(const size_t axis)
-                {
-                    m_normalized_axis = axis;
-                    m_axis = axis;
-                }
+                size_t get_axis() const;
+                int64_t get_raw_axis() const { return m_axis; }
```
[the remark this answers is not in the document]
mbencer: Name `get_provided_axis`, documentation added

src/ngraph/op/topk.cpp
```diff
@@ -235,13 +236,16 @@ op::v1::TopK::TopK(const Output<Node>& data,
...
     , m_axis{axis}
+    , m_normalized_axis{0}
     , m_mode{mode_from_string(mode)}
     , m_sort{sort_type_from_string(sort)}
     , m_index_element_type{index_element_type}
 {
     constructor_validate_and_infer_types();
 }
 
+static const std::size_t UNKNOWN_NORMALIZED_AXIS = std::numeric_limits<size_t>::max();
```
ilyachur: Can you change `size_t` to `uint64_t`?

src/ngraph/op/topk.hpp
```diff
@@ -182,6 +187,7 @@ namespace ngraph
 
             protected:
                 int64_t m_axis;
+                size_t m_normalized_axis;
```
ilyachur: Please change type to `uint64_t`

src/ngraph/op/topk.cpp
mbencer: done

src/ngraph/op/topk.hpp
mbencer: done

src/ngraph/op/topk.cpp
```diff
@@ -235,13 +236,16 @@ op::v1::TopK::TopK(const Output<Node>& data,
                    const element::Type& index_element_type)
     : Op{{data, k}}
     , m_axis{axis}
+    , m_normalized_axis{0}
```
diyessi: This needs to be a `op::v2::TopK` because the semantics have changed; even though it is backwards compatible for frontends, it is not backwards compatible for users of the op like passes and backends.

mbencer: I see your point but according to spec `TopK:v1` should handle negative axis. For me it was base to introduce this change.  Can we change something in order to provide compatibility?

diyessi: So it's bug-fix. Okay.

PR #4294, opened 2020-02-06
Fix some warnings

dcaballe:
Control reaches end of non-void function


src/ngraph/op/constant.hpp
```diff
@@ -260,9 +260,7 @@ namespace ngraph
                         auto vector = get_vector<uint64_t>();
                         return std::vector<T>(vector.begin(), vector.end());
                     }
-                    case element::Type_t::u1:
-                    case element::Type_t::undefined:
-                    case element::Type_t::dynamic: throw std::runtime_error("unsupported type");
+                    default: throw std::runtime_error("unsupported type");
```
rkimballn1: There should be a case for each type so if a new type is added it must be handled. Is the error related to case fallthrough? If so then just add throw to each line, and you can make the throw message unique per type too.

dcaballe: Thanks, Bob. The warning is `Control reaches end of non-void function`. IIRC, there some discrepancies wrt switch statements between clang and gcc. Clang will warn about unhandled cases but gcc doesn't seem to be that strict in this regard, as we can see here. I could keep the original code and add a throw after the switch statement but I'm not sure if clang will complain about unreachable code. Any other idea?

diyessi: You need all the enums values to keep clang happy and the throw to keep gcc happy.

PR #4296, opened 2020-02-06
Disable -Werror by default in nGraph build

dcaballe:
This PR disables treating warnings as errors by default so that
nGraph can be built with compiler versions that are not currently
tested in CI. We could enable -Werror in CI for those compiler versions
we want to target for warning-free builds.


CMakeLists.txt
```diff
@@ -143,7 +143,7 @@ option(NGRAPH_PLAIDML_STATIC_LIB_ENABLE "Enable build PlaidML backend as a stati
 option(NGRAPH_DYNAMIC_COMPONENTS_ENABLE "Enable dynamic loading of components" TRUE)
 option(NGRAPH_NATIVE_ARCH_ENABLE "Enable build for native archtecture" TRUE)
 option(NGRAPH_EXPORT_TARGETS_ENABLE "Enable exporting nGraph cmake export targets" TRUE)
-option(NGRAPH_WARNINGS_AS_ERRORS "Make all nGraph compile-time warnings into errors" TRUE)
+option(NGRAPH_WARNINGS_AS_ERRORS "Make all nGraph compile-time warnings into errors" FALSE)
```
rkimballn1: Does CI set warnings as errors for at least some of the runs? If not we need to get that enabled before we merge this.

rkimballn1:
I added Blocked label because I want to make sure that CI is setting WARNINGS_AS_ERRORS for at least some of the runs

rkimballn1:
I just checked the log files and warnings as errors is enabled for at least the one run I look at so we are all good. Thanks for fixing this.

aslepko:
TESTNOW

PR #4309, opened 2020-02-11
fix gcc issue non-void return type

gf712:
I tried to compile ngraph using gcc (8.3.0) and got the following error when compiling `constant_folding_one_hot.cpp`: `error: control reaches end of non-void function [-Werror=return-type]`.

The issue is that unlike clang, gcc's static analyser doesn't detect that all enum values are handled in the switch statement. However, it is possible to tell gcc that the code after the switch is unreachable.

I am not sure if this is the ideal fix, but it is enough to be able compile the unit-tests target with my gcc version. I think this is better than adding a default statement. (I suspect this issue occurs with other gcc versions, but the gcc version in the CI doesn't complain)


diyessi:
Thanks for the PR -- there are some others already addressing the problem. 
This approach is actually not a good one since it tells gcc that things are okay even if future changes that fail to remove the hint to gcc make them not okay.

gf712:
I see, good to know! Thanks for the quick feedback.

PR #4310, opened 2020-02-11
[ONNX] ONNX importer tests refactor

mitruska:
Rewrote all onnx tests in ngraph to use `test_case` helper class.


test/onnx/onnx_import.in.cpp
```diff
@@ -1348,11 +1377,29 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_softplus_infinity)
...
+                                       std::numeric_limits<float>::infinity(),
+                                       std::numeric_limits<float>::infinity(),
+                                       std::numeric_limits<float>::infinity(),
+                                       std::numeric_limits<float>::infinity(),
+                                       std::numeric_limits<float>::infinity(),
+                                       std::numeric_limits<float>::infinity(),
+                                       std::numeric_limits<float>::infinity(),
+                                       std::numeric_limits<float>::infinity(),
+                                       std::numeric_limits<float>::infinity(),
+                                       std::numeric_limits<float>::infinity()};
```
arogowie-intel: I'm not sure whether this is the right way to check for infinity. @rkimballn1 Do you think this is OK?

test/onnx/onnx_import_reshape.in.cpp
```diff
@@ -421,17 +444,22 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_split_variable_parts_2d)
...
-    for (std::size_t i = 0; i < expected_outputs.size(); ++i)
-    {
-        EXPECT_EQ(outputs[i].size(), expected_outputs[i].size());
-        EXPECT_TRUE(test::all_close_f(outputs[i], expected_outputs[i]));
-    }
+    // for (std::size_t i = 0; i < expected_outputs.size(); ++i)
+    // {
+    //     EXPECT_EQ(outputs[i].size(), expected_outputs[i].size());
+    //     EXPECT_TRUE(test::all_close_f(outputs[i], expected_outputs[i]));
+    // }
```
arogowie-intel: Please remove commented code.

test/onnx/onnx_import_reshape.in.cpp
```diff
@@ -402,17 +420,22 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_split_equal_parts_2d)
...
```
arogowie-intel: Please remove commented code.

test/onnx/onnx_import.in.cpp
mitruska: I'm not sure too, this is what I actually wanted to dicsuss on the review.
Maybe this test doesn't have to check so many infinities.
As we can see on the diff it was previously check in loop with `std::isinf()`
  ```
for (float v : outputs.front())
{
        EXPECT_TRUE(std::isinf(v));
}
```

tomdol: It looks like there are 13 elements in both input and output because the model used for this test was originally designed for another UT -  the `model_softplus` test which is one above.

To make this one cleaner you can replace the creation of both inputs and expected outputs with this: `std::vector<float>(13, std::numeric_limits<float>::infinity());`.

PR #4322, opened 2020-02-11
Undo GetOutputElement creations during cloning

diyessi:
Because `copy_with_new_args` has a `NodeVector` argument, `GetOutputElement` can be created by `copy_with_new_inputs`. This replaces those `GetOutputElement`s. This should be removed when there is time to change `copy_with_new_args` to take an `OutputVector` instead of a `NodeVector`.


diyessi:
Will replace with improved version.

PR #4324, opened 2020-02-12
Change attributes to int64_t

diyessi:


src/ngraph/op/deformable_convolution.hpp
```diff
@@ -63,8 +63,8 @@ namespace ngraph
                                       const CoordinateDiff& pads_end,
                                       const Strides& dilations,
                                       const PadType& auto_pad = PadType::EXPLICIT,
-                                      const size_t group = 1,
-                                      const size_t deformable_group = 1);
+                                      const int64_t group = 1,
```
ilyachur: In general group and deformable_group is unsigned fields and cannot be less than 1. Can we use uint64_t?

diyessi: We might want negative as the uninitialized value.

PR #4326, opened 2020-02-12
AlgebraicSimplification simplification

rkimballn1:
Change Multiply and Add to use simple node comparisons instead of the pattern matcher. With my test model BERT Large Squad the pass used to take 32,335ms and now takes 250ms or 130X faster.


silee2:
I see the error ```/home/dockuser/ngraph/src/ngraph/pass/algebraic_simplification.cpp:67:39: error: unused function 'get_broadcast_label' [-Werror,-Wunused-function]
``` for distributed mlsl

PR #4330, opened 2020-02-13
Start finishing cloning with outputs

diyessi:
Undo GOEs added by the by `copy_with_new_inputs`
Create `clone_with_new_inputs` helper. When all `copy_with_new_args` are converted to `clone_with_new_inputs` we will no longer need GOEs when cloning.

cyphers/ck has two additional converted cloners that will go in when that PR is finished.


src/ngraph/node.cpp
```diff
@@ -104,48 +104,36 @@ std::shared_ptr<Node>
...
-        clone = make_shared<op::GetOutputElement>(value.get_node_shared_ptr(), value.get_index());
+        clone->add_control_dependency(cdep);
     }
-    else
+    return clone;
+}
+
+std::shared_ptr<Node> Node::copy_with_new_args(const NodeVector& args) const
+{
+    NODE_VALIDATION_CHECK(
```
rkimballn1: This check is a NOP since the string is the second arg thus it is the condition.

src/ngraph/node.cpp
```diff
@@ -104,48 +104,36 @@ std::shared_ptr<Node>
...
+        this, "Internal error: copy_with_new_args not replaced by clone_with_new_inputs");
```
rkimballn1:
```suggestion
        this, false, "Internal error: copy_with_new_args not replaced by clone_with_new_inputs");
```

src/ngraph/node.hpp
```diff
@@ -405,8 +405,12 @@ namespace ngraph
         std::shared_ptr<Node> get_input_node_shared_ptr(size_t index) const;
 
     protected:
-        // Will be replaced with an OutputVector version
-        virtual std::shared_ptr<Node> copy_with_new_args(const NodeVector& new_args) const = 0;
+        // Will be replaced with clone_with_new_inputs
```
rkimballn1: put DEPRECATED on this

PR #4331, opened 2020-02-13
Set output shape of Deconv with dynamic input

jane-intel:


src/ngraph/op/convolution.cpp
```diff
@@ -404,6 +404,10 @@ void op::v1::ConvolutionBackpropData::validate_and_infer_types()
             output_shape.insert(output_shape.begin(), data_shape.at(0));
             output_pshape = output_shape;
         }
+        else
```
arogowie-intel: I'd just change it to 
```suggestion
        else if (data_pshape.rank().is_static())
```

And add this in  389 line:
```
output_pshape = PartialShape::dynamic();
```

jane-intel: It is not needed due to this check: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/partial_shape.cpp#L83

PR #4332, opened 2020-02-13
Use default_opset in ONNX importer code

postrational:
Fix remaining places where we're not using `default_opset` namespace for ops.


aslepko:
TESTNOW_DLDT_E2E

PR #4333, opened 2020-02-13
Update ONNX OneHot to produce v1 ops only

postrational:
Update ONNX OneHot operator to produce only v1 operators by replacing `v0::Slice` with `v1::Split`.


arogowie-intel:
Please just consider adding `const` to all variables which values we're not changing in this operator.

PR #4335, opened 2020-02-13
[MLIR] Update MLIR repo

dcaballe:


src/contrib/mlir/utils.cpp
```diff
@@ -35,15 +46,47 @@ static llvm::cl::opt<bool> clPrintIRAfterAll(
...
+        registerDialect<loop::LoopOpsDialect>();
+        registerDialect<StandardOpsDialect>();
+        registerDialect<vector::VectorOpsDialect>();
+        
+        // nGraph dialects. 
+        registerDialect<mlir::NGraphOpsDialect>();
+
+        // In-tree passes.
+        // NO-OPT to avoid DCE on the following pass initializations.
+        if (std::getenv("bar") != (char*)-1)
```
nmostafa: what is this for ?

src/contrib/mlir/utils.cpp
```diff
@@ -35,15 +46,47 @@ static llvm::cl::opt<bool> clPrintIRAfterAll(
...
-    }
+    // Initialize MLIR dialects and passes only once.
+    static bool init_once = []() {
+        // In-tree Dialects.
+        registerDialect<AffineOpsDialect>();
+        registerDialect<LLVM::LLVMDialect>();
+        registerDialect<loop::LoopOpsDialect>();
+        registerDialect<StandardOpsDialect>();
+        registerDialect<vector::VectorOpsDialect>();
+
```
nmostafa: nit: Remove extra spaces.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -249,6 +248,7 @@ namespace
         inline size_t insertAttrs(opAttrs attrs);
 
         MemoryAnalysis* getMemAnalysis() const { return m_memAnalysis; }
+
```
nmostafa: extra space.

src/contrib/mlir/backend/cpu/cpu_backend.cpp
```diff
@@ -194,7 +194,8 @@ void MLIRCPUBackend::lowerNgDialect()
 void MLIRCPUBackend::lowerStandardDialect()
 {
     mlir::PassManager pm(&m_context);
-    pm.addPass(mlir::createLowerToLLVMPass());
+    pm.addPass(mlir::createLowerToLLVMPass(
+        false /*useAlloca*/, false /*useBarePtrCallConv*/, true /*emitCWrappers*/));
```
dcaballe: Do you think we could use the default calling convention (flattened memref descriptors) instead to avoid the extra function call? It could be a follow-up patch but at least here we have to make sure the these can be easily integrated with https://github.com/NervanaSystems/ngraph/pull/4336. Do you foresee any issue?

src/contrib/mlir/utils.cpp
dcaballe: Copy-paste from `mlir-opt` to avoid DCE of the code below. Fixing typo: NO-OPT -> No-op

src/contrib/mlir/backend/pass/affine_lowerer.cpp
dcaballe: Let me know if it's OK now. It seems a clang-format thing.

src/contrib/mlir/backend/cpu/cpu_backend.cpp
nmostafa: nit: better use clang-tidy's format ? `/*useAlloca=*/ false`

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -317,7 +344,8 @@ namespace
 
             // TODO: Encode no alias attribute as part of the function signature conversion or as a
             // separate rewrite pattern. Retrieve new function after signature conversion.
-            insertNoAliasArgAttrs();
+            // TODO: To be enabled in follow-up commit.
```
nmostafa: Is disabling this needed for the MLIR update ? If not, then we should keep it.

src/contrib/mlir/runtime/cpu/cpu_callbacks.cpp
```diff
@@ -719,7 +719,7 @@ static void __mlir_cblas_sgemm_with_bias(StaticMemRef* memRefmatA,
     }
 }
 
-extern "C" void __mlir_callback_1_input(void* input, void* output, size_t index, OpType type)
+extern "C" void _mlir_ciface_callback_1_input(void* input, void* output, size_t index, OpType type)
```
nmostafa: I suggest adding a macro to append the prefix to the func name for better readability. Something like
```
#define MLIR_CALLBACK(NAME) _mlir_ciface_ ## NAME
extern "C" void MLIR_CALLBACK(callback_1_input) ...
```

src/contrib/mlir/backend/pass/affine_lowerer.cpp
dcaballe: Yes, it's needed due to the new calling convention. Otherwise tests will fail since we are adding `noalias` to arguments that are not pointers.

src/contrib/mlir/runtime/cpu/cpu_callbacks.cpp
dcaballe: I guess this is more like a personal opinion but I think macros make the code more obscure. I would second this if the macro was replacing more code. However, in this case it's just a prefix with kind of the same number of characters.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
dcaballe: I'm enabling this in a follow-up patch for the bare ptr calling convention.

src/contrib/mlir/runtime/cpu/cpu_callbacks.cpp
nmostafa: If the *ciface* naming convention changes, we would have to go and rename all callbacks.

PR #4336, opened 2020-02-13
[MLIR] Enable bare ptr calling convention

dcaballe:
This PR adds a flag to enable the bare pointer calling convention
in LLVM lowering pass.


src/contrib/mlir/runtime/cpu/cpu_runtime.cpp
```diff
@@ -142,33 +157,57 @@ void MLIRCPURuntime::cleanup()
...
+// The default call ABI takes a single arg pointer (argPtr) pointing to a list of args.
 // Each arg is a  pointer to a StaticMemRef which contains a data pointer
 //
 // The args are laid out as follows
 // argPtr-> arg[0]-> StaticMemRef -> <data>
 //          arg[1]-> StaticMemRef -> <data>
 //          ...
+//
+// The bare pointer ABI takes a single arg pointer pointing to data for that MemRef. Not extra
+// information about the MemRef is passed at the moment.
```
nmostafa: nit: can you provide an example countering the one above ?

src/contrib/mlir/runtime/cpu/cpu_runtime.cpp
```diff
@@ -53,6 +53,11 @@ static llvm::cl::opt<std::string>
     clObjectFilename("ngraph-mlir-object-filename",
                      llvm::cl::desc("Dump MLIR JITted-compiled object to file jitted_mlir.o"));
 
+llvm::cl::opt<bool> clEnableBarePtrMemRefLowering(
```
nmostafa: nit: add comment here

src/contrib/mlir/backend/cpu/cpu_backend.cpp
```diff
@@ -194,7 +198,10 @@ void MLIRCPUBackend::lowerNgDialect()
 void MLIRCPUBackend::lowerStandardDialect()
 {
     mlir::PassManager pm(&m_context);
-    pm.addPass(mlir::createLowerToLLVMPass());
+    // We lower memrefs to StaticMemRef descriptors by default. If 'clEnableBarePtrMemRefLowering'
```
nmostafa: nit: avoid using variable names in comment in case we rename it later

src/contrib/mlir/runtime/cpu/cpu_runtime.cpp
dcaballe: Is the example above still valid? Don't we pass `arg0`, `arg1`, etc., directly, instead of through `argPtr`

PR #4339, opened 2020-02-14
update code owner of distributed part

wenzhe-nrv:


diyessi:
Picked up in Sang Ik's PR.

PR #4340, opened 2020-02-14
Add attribute visitor for ops R

tsocha:
- Range op
* Reduce ops
-ReduceLogicalAnd
-ReduceLogicalOr
-ReduceMax
-ReduceMean
-ReduceMin
-ReduceProd
-ReduceSum
- RegionYolo op
- Reshape op
- Result op
- Reverse op
- ReverseSequence op
- RNNCell op


src/ngraph/op/fused/rnn_cell.hpp
```diff
@@ -132,6 +132,7 @@ namespace ngraph
                     const std::vector<float>& activations_beta = {},
                     float clip = 0.f);
 
+                using RNNCellBase::visit_attributes;
```
ilyachur: Why do you use using? It also will work without this.

src/ngraph/op/min.hpp
```diff
@@ -73,6 +73,7 @@ namespace ngraph
                           const Output<Node>& reduction_axes,
                           bool keep_dims = false);
 
+                using util::ArithmeticReductionKeepDims::visit_attributes;
```
ilyachur: Why do you use using? It also will work without this.

tsocha: Yes, you are correct.
I removed using from all reduce ops.

src/ngraph/op/fused/rnn_cell.hpp
tsocha: `RNNCell` operator have multiple inheritance and because of that `visit_attributes` will be ambiguous(`op::visit_attributes` and `RNNCellBase::visit_attributes`).

ilyachur:
Could you please add unit tests?

PR #4343, opened 2020-02-17
avoid error tautological-constant-out-of-range-compare

dkurt:
During compilation for Android `armeabi-v7a` fixes error:

```
ngraph/src/ngraph/op/gather.cpp:134:40: error: result of comparison of constant 9223372036854775807 with expression of type 'unsigned int' is always true
      [-Werror,-Wtautological-constant-out-of-range-compare]
    if (input_rank.is_static() && axis != AXIS_NOT_SET_VALUE)
                                  ~~~~ ^  ~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/ngraph/op/gather.cpp
+++ b/src/ngraph/op/gather.cpp
@@ -130,7 +130,7 @@ void op::v1::Gather::validate_and_infer_types()
                               ").");
     }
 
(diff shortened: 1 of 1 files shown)


silee2:
Created a fork #4352 for merging PR

PR #4345, opened 2020-02-18
[ONNX] Change constants' target shape to scalar shape

Ewa21:


src/ngraph/frontend/onnx_import/op/leaky_relu.cpp
```diff
@@ -39,10 +39,10 @@ namespace ngraph
                         << " alpha value should be in range (0,1)";
 
                     std::shared_ptr<ngraph::Node> alpha_node =
-                        std::make_shared<default_opset::Constant>(data->get_element_type(),
-                                                                  data->get_shape(),
-                                                                  std::vector<double>{alpha});
-                    return {std::make_shared<default_opset::Maximum>(data * alpha_node, data)};
+                        std::make_shared<default_opset::Constant>(
+                            data->get_element_type(), Shape{}, std::vector<double>{alpha});
```
tomdol: Can you try removing the explicit `std::vector<...>` instances in the whole PR and change them to `{...}` instead? It's not critical but I don't see any reason to keep explicit constructors and since those files are modified already - we could clean this up.

tomdol: Just to make it clear, I suggest changing `std::vector<double>{alpha}` to `{alpha}` in this particular line.

Ewa21: Done

PR #4353, opened 2020-02-19
[Py] Python wheel building support for Windows

tomdol:
This is still under development but in general it works. To build the wheel on Windows run the following sequence in `ngraph/build`:

```
cmake -G"Visual Studio 15 2017 Win64" -DCMAKE_BUILD_TYPE=Release -DNGRAPH_DEX_ONLY=TRUE -DNGRAPH_USE_LEGACY_MKLDNN=FALSE -DNGRAPH_PYTHON_BUILD_ENABLE=TRUE -DNGRAPH_ONNX_IMPORT_ENABLE=TRUE -DCMAKE_INSTALL_PREFIX="C:\temp_install_dir" ..
cmake --build . --config Release -j 8 --target install
cmake --build . --config Release -j 8 --target python_wheel
```

The `install` target has to be executed separately - any other target (`python_wheel` in particular) can't depend on it.

Any suggestions welcome.


python/CMakeLists.txt
```diff
@@ -30,19 +30,33 @@ ExternalProject_Add(
...
-configure_file(${BUILD_SH_IN} ${BUILD_SH} @ONLY)
-
-add_custom_command(
-        OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/dist/
-        POST_BUILD
-        WORKING_DIR ${CMAKE_CURRENT_BINARY_DIR}
-        COMMAND ${CMAKE_BUILD_TOOL} -C ../ DESTDIR=python/_install install && bash build_wheel.sh
-)
+if (WIN32)
+        set(BUILD_PY_IN "${CMAKE_CURRENT_SOURCE_DIR}/build_wheel.py.in")
```
ilyachur: Why do you want to use python script only for Windows? Looks like it can be used for other platforms too.

tomdol: There's an offline discussion about changes in the wheel files layout. Unification and switching to python build script for all systems could be included in that activity.

python/setup.py
```diff
@@ -142,6 +148,8 @@ def cpp_flag(compiler):
     """Check and return the -std=c++11 compiler flag."""
     if has_flag(compiler, '-std=c++11'):
         return '-std=c++11'
+    elif sys.platform == 'win32':
```
arogowie-intel: If MSVC has '-std=c++11' flag then control logic would never enter into this branch.

python/setup.py
```diff
@@ -371,14 +381,33 @@ def _add_extra_compile_arg(self, flag, compile_args):
...
+    def _add_win_compiler_flags(self, ext):
+        self._add_extra_compile_arg('/GL', ext.extra_compile_args)  # Whole Program Optimization
+        self._add_extra_compile_arg('/analyze', ext.extra_compile_args)
+
+    def _add_unix_compiler_flags(self, ext):
+        if not self._add_extra_compile_arg('-fstack-protector-strong', ext.extra_compile_args):
+            self._add_extra_compile_arg('-fstack-protector', ext.extra_compile_args)
+
+        self._add_extra_compile_arg('-fvisibility=hidden', ext.extra_compile_args)
+        self._add_extra_compile_arg('-flto', ext.extra_compile_args)
```
arogowie-intel: It would be great if you would find equivalent flag to `-flto` (aka _link time optimizations_) on MSVC. I remember that the overall whl size was reduced about 30-40% just by using this flag.

python/setup.py
tomdol: Yep, good catch. I already have it fixed, just haven't pushed it yet.

python/setup.py
tomdol: I've found `/GL` which is whole program optimization. I don't know the difference in sizes of the package with or without this flag though.

python/BUILDING.md
```diff
@@ -16,13 +16,50 @@ important differences:
...
+      -DCMAKE_INSTALL_PREFIX="C:\temporary_install_dir"
+      -DNGRAPH_DEX_ONLY=TRUE
+      -DNGRAPH_PYTHON_BUILD_ENABLE=TRUE
+      -DNGRAPH_ONNX_IMPORT_ENABLE=TRUE
+      -DCMAKE_CXX_COMPILER=C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\bin\Hostx64\x64
+~~~~
+There are a couple of things to notice here. One is that the full path to the x64 version of
+MSVC compiler has to be specified. This is because DNNL requires a 64-bit version and cmake might
+not detect it correctly by itself.
+On Windows only the DEX mode is supported - hence the `NGRAPH_DEX_ONLY` flag.
```
indie: Only the DEX mode is supported on Windows systems, hence the `NGRAPH_DEX_ONLY` flag.

python/BUILDING.md
```diff
@@ -16,13 +16,50 @@ important differences:
...
+The last but equally important thing is that you should specify a temporary directory where the
```
indie: The other equally important thing to note is that the temporary directory where the build is to be installed can be specified.  This examples uses `C:\temporary_install_dir`,

python/BUILDING.md
```diff
@@ -16,13 +16,50 @@ important differences:
...
+build of ngraph will be installed. In this example it's `C:\temporary_install_dir` but you can pass
+a subdirectory of `ngraph\build` as well. The final python will will contain the contents of this
```
indie: a subdirectory of `ngraph\build`works as well.

indie: > The final python will will contain the contents ...

This needs rewritten

python/BUILDING.md
```diff
@@ -16,13 +16,49 @@ important differences:
...
 
     $ ls python/dist/
-    ngraph-core-0.21.0rc0.tar.gz  
-    ngraph_core-0.21.0rc0-cp37-cp37m-linux_x86_64.whl  
+    ngraph-core-0.21.0rc0.tar.gz
+    ngraph_core-0.21.0rc0-cp37-cp37m-linux_x86_64.whl
+
+## Building nGraph Python Wheels on Windows
+
+It's now possible to build the python wheel on Windows too. The build process consists of 3 steps:
```
postrational: > It's now possible to build the python wheel on Windows too.

This sentence is not future proof. Let's not make the docs sound as news.
I would recommend removing the sentence.

python/BUILDING.md
```diff
@@ -16,13 +16,49 @@ important differences:
...
+      -DCMAKE_BUILD_TYPE=Release
+      -DCMAKE_INSTALL_PREFIX="C:\temporary_install_dir"
+      -DNGRAPH_DEX_ONLY=TRUE
+      -DNGRAPH_PYTHON_BUILD_ENABLE=TRUE
+      -DNGRAPH_ONNX_IMPORT_ENABLE=TRUE
+      -DCMAKE_CXX_COMPILER=C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\bin\Hostx64\x64
+~~~~
+There are a couple of things to notice here. One is that the full path to the x64 version of
+MSVC compiler has to be specified. This is because DNNL requires a 64-bit version and cmake might
+not detect it correctly by itself.
```
postrational:
```suggestion
may fail to detect it correctly.
```

python/BUILDING.md
```diff
@@ -16,13 +16,49 @@ important differences:
...
+      -G"Visual Studio 15 2017 Win64"
+      -DCMAKE_BUILD_TYPE=Release
+      -DCMAKE_INSTALL_PREFIX="C:\temporary_install_dir"
+      -DNGRAPH_DEX_ONLY=TRUE
+      -DNGRAPH_PYTHON_BUILD_ENABLE=TRUE
+      -DNGRAPH_ONNX_IMPORT_ENABLE=TRUE
+      -DCMAKE_CXX_COMPILER=C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\bin\Hostx64\x64
+~~~~
+There are a couple of things to notice here. One is that the full path to the x64 version of
+MSVC compiler has to be specified. This is because DNNL requires a 64-bit version and cmake might
```
postrational:
```suggestion
MSVC compiler has to be specified. This is because DNNL requires a 64-bit version and cmake
```

python/BUILDING.md
```diff
@@ -16,13 +16,49 @@ important differences:
...
+not detect it correctly by itself.
+Only the DEX mode is supported on Windows systems, hence the `NGRAPH_DEX_ONLY` flag.
+The other equally important thing to note is that the temporary directory where the build is to be installed can be specified.
+This examples uses `C:\temporary_install_dir` however, a subdirectory of `ngraph\build` works as well.
+The final python will contain the contents of this temporary directory so it's very important to set it.
```
postrational: Please capitalize Python throughout. 
```suggestion
The final Python wheel package will contain the contents of this temporary directory so it's very important to set it.
```

python/ngraph/impl/__init__.py
```diff
@@ -20,17 +20,24 @@
...
+import os
 import sys
 import six
 
-# workaround to load the libngraph.so with RTLD_GLOBAL
-if six.PY3:
-    import os
-    flags = os.RTLD_NOW | os.RTLD_GLOBAL
+if sys.platform == 'win32':
+    # ngraph.dll is currently located 3 directories above
```
postrational: The user may choose to have a custom installation. The word `currently` would be misleading.
```suggestion
    # ngraph.dll is installed 3 directories above by default
```

python/requirements_build_wheel.txt
```diff
@@ -0,0 +1 @@
+virtualenv
```
postrational: If there is only one requirement, do we need a file?
Especially, that this is not really required to build a wheel. It's only required **if** you're using the make target.
I would recommend not adding this file.

python/ngraph/impl/__init__.py
tomdol: It's a code comment for ng developers, not documentation for the users. Also if the user changes location of ngraph.dll this code will stop working.

postrational: Let's document that too.
> If you're using a custom installation of nGraph, add the location of ngraph.dll to your system PATH.

This is the type of documentation that will be found by someone debugging their problems with an nGraph installation. That includes users. I have been this user of other libs many times.

python/BUILDING.md
postrational: Please review this fragment: 
> The final Python will contain

postrational:
Let's make sure this still works on other OSes.

tomdol:
> Let's make sure this still works on other OSes.

I've just build a wheel on linux using this branch and executed the python tests: `259 passed, 2 warnings in 1.42s`. Seems that it still works.

postrational:
@rkimballn1 Let's wait for the final review from Ilya. I'm hoping he will run tests in more environments.

ilyachur:
I will aprove this PR, when Python build will be integrated to CI.

tomdol:
Just FYI, this PR is not about the integration with any CI. It just lets the python wheel be built on Windows. The CI integration is a separate task and it won't touch the ngraph sources. Whether you approve it now or postpone it - it's up to you. I think this PR is complete and it has been tested on my side quite extensively, so I'm quite confident that it can already be merged.

ilyachur:
> Just FYI, this PR is not about the integration with any CI. It just lets the python wheel be built on Windows. The CI integration is a separate task and it won't touch the ngraph sources. Whether you approve it now or postpone it - it's up to you. I think this PR is complete and it has been tested on my side quite extensively, so I'm quite confident that it can already be merged.

Ok, in this case, let's merge it. If we face with issues in the process of enabling Python build for Windows in CI, their can be fixed in other PRs.

ilyachur:
@tomdol I reseted my approve and removed Full Reviewed label, because I thought about this PR and I have several questions.

- Do you have a regular build of Python build for Windows?
- Is it integrated in the nGraph CI?

tomdol:
No the python API is not tested in CI on Windows AFAIK. The API is tested in each PR in the "python_ci" check, like this one for example https://aipg-jenkins-algo.intel.com/job/ngraph-python-merge-trigger/16079/

ilyachur:
From my point of view, we cannot merge a new functionality if we haven't a regular CI. Because we cannot guarantee that this feature works, while we haven't validation for this feature.

tomdol:
@postrational @diyessi ^^ any opinions?

diyessi:
> From my point of view, we cannot merge a new functionality if we haven't a regular CI. Because we cannot guarantee that this feature works, while we haven't validation for this feature.

I agree.

raramer01:
TESTNOW_DLDT_E2E

raramer01:
There is still no CI in place for this feature...

PR #4358, opened 2020-02-19
Improve Provenance tests

postrational:
Includes:

* Disable adding ONNX Provenance tags if Provenance globally disabled


src/ngraph/frontend/onnx_import/core/graph.cpp
```diff
@@ -106,7 +107,11 @@ namespace ngraph
 
                     // For each initializer, create a Constant node and store in cache
                     auto ng_constant = tensor.get_ng_constant();
-                    add_provenance_tag_to_initializer(tensor, ng_constant);
+                    // Add provenance information
+                    if (ngraph::get_provenance_enabled())
```
diyessi: Would it make sense to put the `ngraph::get_provenance_enabled` test in `add_provenance_tag_to_initializer`?

postrational: I actually had it like that initially: https://github.com/NervanaSystems/ngraph/blob/b61eca356b0e90162fa9453b58df002183a2d9cb/src/ngraph/frontend/onnx_import/core/graph.cpp#L211-L214

Then changed it in this commit https://github.com/NervanaSystems/ngraph/pull/4358/commits/6ca4856a4ac97f153ed5102e273942a8c093a412

The idea was to be more explicit in the calling code, but I can change it back.

PR #4362, opened 2020-02-19
[MLIR] Add a flag to control ngraph dialect fusion pass.

ayzhuang:
NgDialectFusedOpsPass is disabled by default. Use "-ngraph-kernel-lib-fusion" to enable it.


src/contrib/mlir/core/compiler.cpp
```diff
@@ -60,6 +60,10 @@
 // Defines a new LLVM debug type for this file to be used by LLVM_DEBUG macro.
 #define DEBUG_TYPE "mlir-compiler"
 
+static llvm::cl::opt<bool> clEnableNgOpFusion("ngraph-op-fusion",
```
dcaballe: I'm adding exactly the same flag for the generic op fusion pass :). Could we rename this to be more specific? Something like `ngraph-mkldnn-op-fusion` or `ngraph-kernel-lib-fusion` or similar...

src/contrib/mlir/core/compiler.cpp
```diff
@@ -60,6 +60,10 @@
...
+                                              llvm::cl::init(false),
+                                              llvm::cl::desc("Enable ngraph op fusion pass"));
```
dcaballe: Please, update this comment to be more specific and update flag variable accordingly.

src/contrib/mlir/core/compiler.cpp
```diff
@@ -124,6 +128,11 @@ void MLIRCompiler::buildNgDialectModule()
 
 void MLIRCompiler::optimizeNgDialect()
 {
+    if (!clEnableNgOpFusion)
+    {
+        return;
```
dcaballe: Since we are adding more passes here, rather do something like:
```
if (flag) {
    pm.addPass(...);
}
```

PR #4365, opened 2020-02-20
[MLIR] Use sane utilities to handle environment variables.

ayzhuang:


src/ngraph/runtime/cache.cpp
```diff
@@ -15,21 +15,22 @@
...
 using namespace ngraph;
 using namespace std;
 
 // Constructor
 runtime::LRUCache::LRUCache()
 {
-    char* cache_size = getenv("NGRAPH_CACHE_SIZE");
-    if (cache_size == nullptr)
+    int32_t cache_size = getenv_int("NGRAPH_CACHE_SIZE");
+    if (cache_size <= 0)
```
nbpatel: shouldn't this be < 0  and not <=0? ....lets assume the user wants to set cache size of 0 and not use the cache...in this case it will set it to 1024

src/ngraph/runtime/cpu/cpu_call_frame.cpp
```diff
@@ -37,14 +38,14 @@ runtime::cpu::CPU_CallFrame::CPU_CallFrame(std::shared_ptr<CPU_ExternalFunction>
     , m_compiled_destroy_ctx_func(compiled_destroy_ctx_func)
     , m_compiled_function(compiled_function)
 {
-    const auto envConcurrency = std::getenv("NGRAPH_CPU_CONCURRENCY");
-    m_num_ctx = envConcurrency == nullptr ? 1 : std::atoi(envConcurrency);
+    const auto envConcurrency = getenv_int("NGRAPH_CPU_CONCURRENCY");
+    m_num_ctx = envConcurrency <= 0 ? 1 : envConcurrency;
```
nbpatel: same as above

src/ngraph/runtime/cpu/cpu_call_frame.cpp
```diff
@@ -234,14 +235,13 @@ void runtime::cpu::CPU_CallFrame::setup_runtime_context(Allocator* allocator)
...
-            std::getenv("NGRAPH_CPU_USE_TBB") != nullptr)
+        if (m_external_function->is_direct_execution() && getenv_bool("NGRAPH_CPU_USE_TBB"))
         {
             // For codegen mode, graph and global control are now part of the code generated
             // CPURuntimeContextCG class.
             ctx->G = new tbb::flow::graph;
-            const auto envParallelism = std::getenv("NGRAPH_INTER_OP_PARALLELISM");
-            const auto parallelism = envParallelism == nullptr ? 1 : std::atoi(envParallelism);
+            const auto envParallelism = getenv_int("NGRAPH_INTER_OP_PARALLELISM");
+            const auto parallelism = envParallelism <= 0 ? 1 : envParallelism;
```
nbpatel: same

src/ngraph/runtime/cpu/cpu_call_frame.cpp
ayzhuang: m_num_ctx can't be 0. We need at least 1 cpu runtime context. Should have checked it before.

src/ngraph/runtime/cpu/cpu_call_frame.cpp
ayzhuang: We are setting max_allowed_parallelism to parallelism. It can't be 0.

src/ngraph/runtime/cache.cpp
ayzhuang: @nbpatel I feel runtime::LRUCache::add_entry needs to be modified if we allow cache_size to be 0, WDYT?

PR #4377, opened 2020-02-24
Move ops in ngraph::op namespace to ngraph::op::v0

rkimballn1:
Moved all v0 ops to the v0 namespace
Please set `Hide whitespace  changes` when doing this review since most of the changes are indenting.
image


src/ngraph/op/op_version_tbl.hpp
```diff
@@ -70,25 +70,25 @@ NGRAPH_OP(ConvolutionBackpropFilters, ngraph::op::v1, 1)
...
+NGRAPH_OP(Cosh, ngraph::op::v0, 0)
+NGRAPH_OP(CropAndResize, ngraph::op::v0, 0)
+NGRAPH_OP(CrossEntropy, ngraph::op::v0, 0)
+NGRAPH_OP(CrossEntropyBackprop, ngraph::op::v0, 0)
 NGRAPH_OP(CumSum, ngraph::op::v0, 0)
 NGRAPH_OP(DeformableConvolution, ngraph::op::v1, 1)
 NGRAPH_OP(DeformablePSROIPooling, ngraph::op::v1, 1)
-NGRAPH_OP(DepthToSpace, ngraph::op::v0, 0)
-NGRAPH_OP(Dequantize, ngraph::op, 0)
+NGRAPH_OP(DepthToSpace, ngraph::op::v0, 1)
```
csullivan: Looks like this doesn't have your previously merged fix for the DepthToSpace op version. After changing this, there shouldn't be a change needed in gpu/gpu_emitter.cpp

rkimballn1: Looks like a merge error, fixed.

PR #4379, opened 2020-02-25
[ONNX] Add dynamic shapes support for expand op

mbencer:


src/ngraph/runtime/gpu/unit_test.manifest
```diff
@@ -258,6 +258,13 @@ model_split_equal_parts_default
 model_split_equal_parts_2d
 model_split_variable_parts_2d
 model_expand_static_shape
+expand_1_dyn_shape
+expand_2_dyn_shape
+expand_3_dyn_shape
+expand_4_dyn_shape
+expand_5_dyn_shape
+expand_6_dyn_shape
+expand_uint16_dyn_shape
```
tsocha: I'm surprised that these tests does not work on GPU.
`Expand` op use only `Constant` and `Multiply` ops could you explain it?

mbencer: On GPU tests fail as following:
```
[ RUN      ] onnx_dyn_shapes_GPU.expand_2d_1_dyn_shape
/home/dockuser/ngraph/test/util/test_case.cpp:76: Failure
Value of: values_match(expected_result_constant, result_tensor)
  Actual: false (1 is not close to 2 at index 1
1 is not close to 3 at index 2
1 is not close to 0 at index 3
1 is not close to 0 at index 4
1 is not close to 0 at index 5
diff count: 35 out of 36
```

src/ngraph/frontend/onnx_import/op/expand.cpp
```diff
@@ -42,18 +35,11 @@ namespace ngraph
...
-                        ngraph::as_type_ptr<default_opset::Constant>(shape)
-                            ->get_vector<std::size_t>();
-
-                    const ngraph::Shape shape_shape{shape_vector};
-                    return {std::make_shared<default_opset::Broadcast>(
-                        data,
-                        default_opset::Constant::create(
-                            element::i64, Shape{shape_shape.size()}, shape_shape))};
+                    return {
+                        std::make_shared<default_opset::Multiply>(data, const_filled_with_ones)};
```
arogowie-intel: Why not just using `v1::Broadcast` ??? 
https://github.com/NervanaSystems/ngraph/blob/1f2ae36f44d48f9ea341bd5156726896ef920353/src/ngraph/op/broadcast.hpp#L150-L153

test/onnx/onnx_import_dyn_shapes.in.cpp
```diff
@@ -392,3 +392,121 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_constant_of_shape_int_ones)
...
+}
+
+NGRAPH_TEST(onnx_dyn_shapes_${BACKEND_NAME}, expand_2_dyn_shape)
+{
+    const auto function = onnx_import::import_onnx_model(
+        file_util::path_join(SERIALIZED_ZOO, "onnx/dynamic_shapes/expand_dyn.prototxt"));
+
+    auto test_case = NgraphTestCase(function, "${BACKEND_NAME}", BackendMode::DYNAMIC);
+
+    test_case.add_input<float>(Shape{3, 1}, std::vector<float>{1, 2, 3});
```
arogowie-intel: Please be consistent with using `.f` suffix for floating point values.

test/onnx/onnx_import_dyn_shapes.in.cpp
```diff
@@ -392,3 +392,121 @@ NGRAPH_TEST(onnx_${BACKEND_NAME}, model_constant_of_shape_int_ones)
...
+    test_case.add_input<int64_t>(Shape{3}, std::vector<int64_t>{2, 3, 4});
+
+    std::vector<float> expected_values{1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3,
```
arogowie-intel: Same as above.

test/onnx/onnx_import_dyn_shapes.in.cpp
mbencer: done

test/onnx/onnx_import_dyn_shapes.in.cpp
mbencer: done

PR #4391, opened 2020-02-28
Allow for opset extensions

diyessi:


src/ngraph/opsets/opset.hpp
```diff
@@ -32,6 +32,12 @@ namespace ngraph
         static std::mutex& get_mutex();
 
     public:
+        OpSet(ngraph::FactoryRegistry<ngraph::Node>* factory_registry =
```
ilyachur: Can we add constructor for OpSet which will create OpSet with empty factory?

ilyachur: Why do you use the raw pointer? Can we use the smart pointer?

diyessi: > Can we add constructor for OpSet which will create OpSet with empty factory?

I was thinking you were making an extension of an existing opset, instead of a brand new opset with just one or two ops in it.

diyessi: For something like Opset1, a shared pointer be a shared pointer to a statically allocated object that shouldn't be deleted. I think it is safer to have the opset and registry as members of an extension object and let that take care of deleting them. But you know what your extension mechanism looks like, so you can adjust this PR a little to make it match better.

ilyachur: I created a PR with proposal: https://github.com/NervanaSystems/ngraph/pull/4395

PR #4394, opened 2020-02-29
CMAKE flag for DLDT build

pruthviIntel:
- Add a build flag for DLDT
- eliminate GOE only for DLDT build


silee2:
This is not going to work. Some part of CMake has to pass a compiler flag to define the build macro. You are just adding a CMake option only. If the CMake option is ON/TRUE, macro has to be passed to the actual couple flag.

PR #4396, opened 2020-03-02
[ONNX] Make unique result node friendly names.

arogowie-intel:


arogowie-intel:
Not needed because of changes in IE.

PR #4401, opened 2020-03-02
Add get_ops and get_ordered_ops to PyBind wrapper of Function

postrational:


ilyachur:
Can we add new tests for those changes?

PR #4405, opened 2020-03-02
Peg virtualenv to version 16.7.10

ashahba:
With `virtualenv 20.x` and newer we are getting:
```
ImportError: cannot import name 'ContextManager'
```
while creating the environment. This is an attempt to fix it.


raramer01:
Sorry, we don't use these dockerfiles for validation...

PR #4409, opened 2020-03-03
Move CPU executable to a separate file

rkimballn1:


raramer01:
looks like failed ci is due to lack of memory, general-ci failed only on the 'gpu' system, while the python-ci failed trying to clone the ngraph repo...will re-trigger the ci hoping for better result...

PR #4412, opened 2020-03-04
Fixed GEMM conversion

GlebKazantaev:


src/ngraph/frontend/onnx_import/op/gemm.cpp
```diff
@@ -73,15 +71,22 @@ namespace ngraph
                     input_a = ngraph::builder::opset1::flatten(input_a, 1);
                     input_b = ngraph::builder::opset1::flatten(input_b, 1);
 
-                    auto matmul_node = std::make_shared<default_opset::MatMul>(input_a, input_b);
+                    std::shared_ptr<ngraph::Node> last_node =
```
postrational: This code is now less clearly readable because of the changed variable names. Let's use the names we had before.
```suggestion
                    std::shared_ptr<ngraph::Node> matmul_node =
```

src/ngraph/frontend/onnx_import/op/gemm.cpp
```diff
@@ -73,15 +71,22 @@ namespace ngraph
...
+                        std::make_shared<default_opset::MatMul>(input_a, input_b);
+
+                    if (alpha != 1)
+                    {
+                        last_node = std::make_shared<default_opset::Multiply>(
+                            last_node,
```
postrational:
```suggestion
                            matmul_node,
```

src/ngraph/frontend/onnx_import/op/gemm.cpp
```diff
@@ -73,15 +71,22 @@ namespace ngraph
                     input_a = ngraph::builder::opset1::flatten(input_a, 1);
                     input_b = ngraph::builder::opset1::flatten(input_b, 1);
 
-                    auto matmul_node = std::make_shared<default_opset::MatMul>(input_a, input_b);
+                    std::shared_ptr<ngraph::Node> last_node =
+                        std::make_shared<default_opset::MatMul>(input_a, input_b);
+
+                    if (alpha != 1)
+                    {
+                        last_node = std::make_shared<default_opset::Multiply>(
```
postrational:
```suggestion
                        matmul_node = std::make_shared<default_opset::Multiply>(
```

src/ngraph/frontend/onnx_import/op/gemm.cpp
```diff
@@ -73,15 +71,22 @@ namespace ngraph
...
+                            last_node,
+                            default_opset::Constant::create(
```
postrational: Let's keep this Constant assigned to `alpha_node`, a compiler will optimize the name away, but it will be clearer for the next programmer what's going on.

src/ngraph/frontend/onnx_import/op/gemm.cpp
```diff
@@ -108,24 +113,28 @@ namespace ngraph
...
-                    const auto alpha_node = default_opset::Constant::create(
-                        element::Type_t::f32, Shape{}, std::vector<float>{alpha});
                     const auto beta_node = default_opset::Constant::create(
                         element::Type_t::f32, Shape{}, std::vector<float>{beta});
 
                     const bool trans_a = node.get_attribute_value<int64_t>("transA", 0);
                     const bool trans_b = node.get_attribute_value<int64_t>("transB", 0);
 
-                    auto matmul_node =
+                    std::shared_ptr<ngraph::Node> last_node =
```
postrational: Same notes here as above.

src/ngraph/frontend/onnx_import/op/gemm.cpp
GlebKazantaev: Done.

src/ngraph/frontend/onnx_import/op/gemm.cpp
GlebKazantaev: Done.

src/ngraph/frontend/onnx_import/op/gemm.cpp
GlebKazantaev: Done.

src/ngraph/frontend/onnx_import/op/gemm.cpp
GlebKazantaev: Done.

src/ngraph/frontend/onnx_import/op/gemm.cpp
GlebKazantaev: Done.

tsocha:
I wonder if this logic should also be applied to the beta attribute.

postrational:
> I wonder if this logic should also be applied to the beta attribute.

In the future this should be handled by an optimization pass. This is a workaround for a specific model while the optimization passes are not yet in place.

PR #4419, opened 2020-03-06
Inference Engine backend for nGraph tests

arogowie-intel:
First version of Inference Engine backend used to run nGraph tests.


CMakeLists.txt
```diff
@@ -133,12 +133,14 @@ option(NGRAPH_CODE_COVERAGE_ENABLE "Enable code coverage data collection" FALSE)
...
 option(NGRAPH_PYTHON_BUILD_ENABLE "Enable build nGraph python package wheel" FALSE)
 option(NGRAPH_PLAIDML_ENABLE "Enable the PlaidML backend" ${PLAIDML_FOUND})
+option(NGRAPH_INFERENCE_ENGINE_ENABLE "Enable the Inference Engine backend" FALSE)
 option(NGRAPH_FAST_MATH_ENABLE "Enable fast math" ON)
 option(NGRAPH_JSON_ENABLE "Enable JSON based serialization and tracing features" TRUE)
 option(NGRAPH_STATIC_LIB_ENABLE "Enable build nGraph as a static library" FALSE)
 option(NGRAPH_INTERPRETER_STATIC_LIB_ENABLE "Enable build INTERPRETER backend as a static library" FALSE)
 option(NGRAPH_CPU_STATIC_LIB_ENABLE "Enable build CPU backend as a static library" FALSE)
 option(NGRAPH_PLAIDML_STATIC_LIB_ENABLE "Enable build PlaidML backend as a static library" FALSE)
+option(NGRAPH_IE_STATIC_LIB_ENABLE "Enable build Inference Engine backend as a static library" FALSE)
```
rkimballn1: I know the name would be really long but it really should be NGRAPH_INFERENCE_ENGINE_STATIC_LIB_ENABLE, or we could rename the other variable as NGRAPH_IE_ENABLE, but either way they should match.

CMakeLists.txt
```diff
@@ -241,6 +244,8 @@ message(STATUS "NGRAPH_MLIR_ENABLE:                   ${NGRAPH_MLIR_ENABLE}")
 message(STATUS "NGRAPH_NATIVE_ARCH_ENABLE:            ${NGRAPH_NATIVE_ARCH_ENABLE}")
 message(STATUS "NGRAPH_NOP_ENABLE:                    ${NGRAPH_NOP_ENABLE}")
 message(STATUS "NGRAPH_ONNX_IMPORT_ENABLE:            ${NGRAPH_ONNX_IMPORT_ENABLE}")
+message(STATUS "NGRAPH_INFERENCE_ENGINE_ENABLE:       ${NGRAPH_INFERENCE_ENGINE_ENABLE}")
```
rkimballn1: This list is in alphabetical order. Please move these lines to the correct position in the list.

src/ngraph/runtime/inference_engine/ie_tensor_view.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "ngraph/runtime/tensor.hpp"
+#include "ngraph/type/element_type.hpp"
+
+namespace ngraph
+{
+    namespace runtime
+    {
+        namespace inference_engine
+        {
+            class IETensorView : public ngraph::runtime::Tensor
```
rkimballn1: the term TensorView is long dead. Please call this IETensor and rename the files to ie_tensor.?pp

src/ngraph/runtime/inference_engine/ie_backend.hpp
```diff
@@ -0,0 +1,88 @@
...
+                    {
+                        size *= x;
+                    }
+                    std::vector<T> v(data, data + size);
+                    tensor->write(data, size * sizeof(T));
+                    return tensor;
+                }
+
+            private:
+                std::string device;
```
rkimballn1:
```suggestion
                std::string m_device;
```

src/ngraph/runtime/inference_engine/ie_backend.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+shared_ptr<ngraph::runtime::Executable>
+    ngraph::runtime::inference_engine::IE_Backend::compile(shared_ptr<Function> func,
+                                                           bool /* enable_performance_data */)
+{
+    return make_shared<IE_Executable>(func, device);
+}
+
+bool ngraph::runtime::inference_engine::IE_Backend::is_supported(const Node& node) const
```
rkimballn1: In other decls you have commented out unused variable yet here `node` is unused. be consistent. Personally if a variable is not used in a cpp file I just omit it altogether since the header file is the documentation.

src/ngraph/runtime/inference_engine/ie_backend.hpp
```diff
@@ -0,0 +1,88 @@
...
+                template <typename T>
+                std::shared_ptr<ngraph::runtime::Tensor>
+                    create_tensor(ngraph::element::Type type, ngraph::Shape shape, T* data)
+                {
+                    auto tensor = std::make_shared<IETensorView>(type, shape);
+                    size_t size = 1;
+                    for (auto x : shape)
+                    {
+                        size *= x;
+                    }
```
rkimballn1:
```suggestion
                    size_t size = shape_size(shape);
```

src/ngraph/runtime/inference_engine/ie_backend.hpp
```diff
@@ -0,0 +1,88 @@
...
+                    std::vector<T> v(data, data + size);
```
rkimballn1: I don't see `v` used anywhere. Is this line needed?

src/ngraph/runtime/inference_engine/ie_executable.hpp
```diff
@@ -0,0 +1,50 @@
...
+            class IE_Executable final : public Executable
+            {
+            public:
+                IE_Executable(std::shared_ptr<Function> func, std::string device);
+                virtual ~IE_Executable() {}
+                bool call(const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
+                          const std::vector<std::shared_ptr<runtime::Tensor>>& inputs) final;
+
+            private:
+                InferenceEngine::CNNNetwork network;
```
rkimballn1:
```suggestion
                InferenceEngine::CNNNetwork m_network;
```

src/ngraph/runtime/inference_engine/ie_executable.hpp
```diff
@@ -0,0 +1,50 @@
...
+                std::string device;
```
rkimballn1:
```suggestion
                std::string m_device;
```

src/ngraph/runtime/inference_engine/ie_executable.hpp
```diff
@@ -0,0 +1,50 @@
...
+            };
+
+            InferenceEngine::Blob::Ptr fill_blob(InferenceEngine::SizeVector shape,
+                                                 std::vector<float> data);
```
rkimballn1: Passing `data` by value seems expensive. Can you use `const std::vector<float>& data` instead?

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+    default: THROW_IE_EXCEPTION << "Can't convert dims " << shape.size() << " to Layout!";
+    }
+    InferenceEngine::MemoryBlob::Ptr blob(
+        new InferenceEngine::TBlob<float>({InferenceEngine::Precision::FP32, shape, layout}));
+    blob->allocate();
+    float* blob_ptr = blob->rwmap().as<float*>();
+    for (int i = 0; i < data.size(); i++)
+    {
+        blob_ptr[i] = data[i];
+    }
```
rkimballn1: memcpy would be more efficient.

src/ngraph/runtime/inference_engine/ie_tensor_view.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+    m_descriptor->set_tensor_layout(
+        std::make_shared<descriptor::layout::DenseTensorLayout>(*m_descriptor));
+}
+
+void runtime::inference_engine::IETensorView::write(const void* p, size_t n)
+{
+    const int8_t* v = (const int8_t*)p;
+    if (v == nullptr)
+        return;
```
rkimballn1:
```suggestion
    if (v == nullptr)
    {
        return;
    }
```

src/ngraph/runtime/inference_engine/ie_tensor_view.cpp
```diff
@@ -0,0 +1,62 @@
...
+    data.resize(n);
+    std::copy(v, v + n, data.begin());
+}
+
+void runtime::inference_engine::IETensorView::read(void* p, size_t n) const
+{
+    int8_t* v = (int8_t*)p;
+    if (v == nullptr)
+        return;
```
rkimballn1:
```suggestion
    if (v == nullptr)
    {
        return;
    }
```

src/ngraph/runtime/inference_engine/ie_tensor_view.cpp
```diff
@@ -0,0 +1,62 @@
...
+    if (n > data.size())
+        n = data.size();
```
rkimballn1:
```suggestion
    if (n > data.size())
    {
        n = data.size();
    }
```

src/ngraph/runtime/inference_engine/ie_tensor_view.cpp
```diff
@@ -0,0 +1,62 @@
...
+                                                      const Shape& shape)
+    : runtime::Tensor(std::make_shared<descriptor::Tensor>(element_type, shape, ""))
+{
+    m_descriptor->set_tensor_layout(
+        std::make_shared<descriptor::layout::DenseTensorLayout>(*m_descriptor));
+}
+
+void runtime::inference_engine::IETensorView::write(const void* p, size_t n)
+{
+    const int8_t* v = (const int8_t*)p;
```
rkimballn1:
```suggestion
    const int8_t* v = static_cast<const int8_t*>(p);
```

src/ngraph/runtime/inference_engine/ie_tensor_view.cpp
```diff
@@ -0,0 +1,62 @@
...
+    if (v == nullptr)
+        return;
+    data.resize(n);
+    std::copy(v, v + n, data.begin());
+}
+
+void runtime::inference_engine::IETensorView::read(void* p, size_t n) const
+{
+    int8_t* v = (int8_t*)p;
```
rkimballn1:
```suggestion
    int8_t* v = static_cast<int8_t*>(p);
```

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+#include "ngraph/runtime/inference_engine/ie_executable.hpp"
+#include "ngraph/opsets/opset.hpp"
+#include "ngraph/op/get_output_element.hpp"
+#include "ngraph/pass/manager.hpp"
+#include "ngraph/pass/opset1_upgrade.hpp"
+#include "ngraph/runtime/inference_engine/ie_tensor_view.hpp"
+
+ngraph::runtime::inference_engine::IE_Executable::IE_Executable(std::shared_ptr<Function> func,
+                                                                std::string _device)
```
rkimballn1: We don't start variables with underscores, please remove it.

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+        {
+            if (node->get_type_info() == op::GetOutputElement::type_info)
+            {
+                // IE currently can handle GetOutuputElement op;
+                continue;
+            }
+            else
+            {
+                std::cout << "UNSUPPORTED OP DETECTED: " << node->get_type_info().name << std::endl;
+                THROW_IE_EXCEPTION << "Detected op not belonging to opset1!";
```
rkimballn1: Oooo, I like the use of THROW_IE_EXCEPTION with the `<<` operator. I think ngraph needs something like this too. very nice!

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+bool ngraph::runtime::inference_engine::IE_Executable::call(
+    const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
+    const std::vector<std::shared_ptr<runtime::Tensor>>& inputs)
+{
+    InferenceEngine::Core ie;
+
+    //  Loading model to the plugin (BACKEND_NAME)
+    InferenceEngine::ExecutableNetwork exeNetwork = ie.LoadNetwork(network, device);
```
rkimballn1: ngraph local var names are all snake_case so this should be exec_network, but it is also customary to use the naming convention of the library you are using so, up for discussion. Same with other variables below.

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+        THROW_IE_EXCEPTION << "Function inputs number differ from number of given inputs";
+    }
+
+    size_t i = 0;
+    for (auto& it : inputInfo)
+    {
+        std::shared_ptr<runtime::inference_engine::IETensorView> tv =
+            std::static_pointer_cast<runtime::inference_engine::IETensorView>(inputs[i]);
+        size_t size = tv->data.size() / sizeof(float);
+        float* orig_data = (float*)tv->data.data();
```
rkimballn1:
```suggestion
        float* orig_data = static_cast<float*>(tv->data.data());
```

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+    {
+        THROW_IE_EXCEPTION << "Function inputs number differ from number of given inputs";
+    }
+
+    size_t i = 0;
+    for (auto& it : inputInfo)
+    {
+        std::shared_ptr<runtime::inference_engine::IETensorView> tv =
+            std::static_pointer_cast<runtime::inference_engine::IETensorView>(inputs[i]);
+        size_t size = tv->data.size() / sizeof(float);
```
rkimballn1:
```suggestion
        size_t size = tv->get_element_count();
```

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+        float* orig_data = (float*)tv->data.data();
+        std::vector<float> data(orig_data, orig_data + size);
```
rkimballn1: This data copy is not needed. You can change fill_blob to take float* and size and copy that way. Copies are very expensive.

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+    InferenceEngine::MemoryBlob::Ptr moutput =
+        InferenceEngine::as<InferenceEngine::MemoryBlob>(output);
+    if (!moutput)
+    {
+        THROW_IE_EXCEPTION << "Cannot get output MemoryBlob in call_with_validate()";
+    }
+
+    auto lm = moutput->rmap();
+    float* output_ptr = lm.as<float*>();
+    // TODO: how to get size without explicit calculation?
```
rkimballn1: shape_size does this for you.

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+    size_t size = 1;
+    for (const auto& dim : output->getTensorDesc().getDims())
+    {
+        size *= dim;
+    }
+    //  Vector initialization from pointer
+    std::vector<float> result(output_ptr, output_ptr + size);
```
rkimballn1: There is no need to make a copy of the data here, just use output_ptr in the write call below.

src/ngraph/runtime/inference_engine/ie_tensor_view.hpp
```diff
@@ -0,0 +1,62 @@
...
+                /// \param n Number of bytes to write, must be integral number of elements.
+                void write(const void* p, size_t n) override;
+
+                /// \brief Read bytes directly from the tensor
+                /// \param p Pointer to destination for data
+                /// \param n Number of bytes to read, must be integral number of elements.
+                void read(void* p, size_t n) const override;
+
+                /// \brief copy bytes directly from source to this tensor
+                /// \param source The source tensor
```
rkimballn1: This comment should be removed

src/ngraph/runtime/inference_engine/ie_tensor_view.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+            private:
+                IETensorView(const IETensorView&) = delete;
+                IETensorView(IETensorView&&) = delete;
+                IETensorView& operator=(const IETensorView&) = delete;
+
+            public:
+                std::vector<int8_t> data;
```
rkimballn1: I like using runtime::AlignedBuffer for this since it gives better memory alignment.

rkimballn1:
Failing style check

CMakeLists.txt
diyessi: I don't want to see `NGRAPH_CENTRAL_PROCESSING_UNIT_STATIC_LIB_ENABLE` so let's consistently go with `IE` instead of `INFERENCE_ENGINE`

src/ngraph/runtime/inference_engine/ie_executable.cpp
```diff
@@ -0,0 +1,143 @@
...
+    std::cout << "Nodes in test: ";
+    for (auto& node : func->get_ops())
+    {
+        std::cout << node << std::endl;
+    }
+    std::cout << std::endl;
+#endif
+
+    network = InferenceEngine::CNNNetwork(func);
+    device = _device == "INFERENCE_ENGINE" ? "CPU" : _device;
```
tomdol: When can we expect that `INFERENCE_ENGINE` will be passed in and why do we need to do this translation?

src/ngraph/runtime/inference_engine/ie_backend.hpp
```diff
@@ -0,0 +1,88 @@
...
+{
+    namespace runtime
+    {
+        namespace inference_engine
+        {
+            class IE_Backend : public runtime::Backend
+            {
+            public:
+                IE_Backend(const std::string& configuration_string);
+                ~IE_Backend() {}
```
tomdol:
```suggestion
                virtual ~IE_Backend() = default;
```

CMakeLists.txt
arogowie-intel: Good to hear that `IE` is enough :) I wasn't sure whether this would be sufficient and self-explanatory or whether introduce confusion what kind of backend it is.

src/ngraph/runtime/inference_engine/ie_executable.cpp
arogowie-intel: This is ugly I know :) However At this moment I don't know how to pass the device name for Inference Engine. I could pass "IE:CPU" which would work if I create backend manually. However in multi-source backend unit test files the variable `BACKEND_NAME` is substituted in multiple places, where the name "IE:CPU" causes compilation errors. (ie. you can't have a unit test case name with "IE:CPU" part). Thus I decided that at this moment the _default_ device is "CPU" if you use "IE" backend.

src/ngraph/runtime/inference_engine/ie_backend.hpp
```diff
@@ -0,0 +1,88 @@
...
+                    return std::make_shared<IETensorView>(type, shape);
+                }
+
+                std::shared_ptr<ngraph::runtime::Tensor>
+                    create_tensor(const ngraph::element::Type& element_type, const Shape& shape);
+
+                std::shared_ptr<ngraph::runtime::Tensor>
+                    create_tensor(const ngraph::element::Type& element_type,
+                                  const Shape& shape,
+                                  void* memory_pointer)
```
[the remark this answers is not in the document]
arogowie-intel: Of course not :)

src/ngraph/runtime/ie/ie_executable.cpp
```diff
@@ -0,0 +1,165 @@
...
+        memcpy(blob_ptr, data, data_size * elem_type.size());
+        return blob;
+    }
+}
+
+runtime::ie::IE_Executable::IE_Executable(shared_ptr<Function> func, string device)
+{
+    const auto& opset = get_opset1();
+    pass::Manager passes;
+    passes.register_pass<pass::Opset1Upgrade>();
```
mbencer: Can we use `GetOutputElementElimination` pass here and skip checking if node is `op::GetOutputElement::type`?

arogowie-intel: Currently IE is running `GetOutputElementElimination` pass anyway, so I thought it's not necessary.

postrational:
Do not merge until 2020.2 brach is forked off.

src/ngraph/runtime/ie/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    ie_tensor.cpp
+    ie_tensor.hpp
+    ie_executable.cpp
+    ie_executable.hpp
+)
+
+add_library(ie_backend ${LIBRARY_TYPE} ${SRC})
+target_compile_definitions(ie_backend PRIVATE IE_BACKEND_DLL_EXPORTS)
+target_include_directories(ie_backend PUBLIC ${IE_MAIN_SOURCE_DIR}/include)
+target_link_libraries(ie_backend PUBLIC ngraph inference_engine)
```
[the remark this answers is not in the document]
arogowie-intel: I haven't been testing this yet. Actually now the only way to have this `IE` backend build is to build nGraph _inside_ DLDT repository with `NGRAPH_IE_ENABLE` set to `TRUE`. This way, we would have inference engine build already.

test/CMakeLists.txt
```diff
@@ -216,14 +216,11 @@ if (NGRAPH_UNIT_TEST_OPENVINO_ENABLE)
     message(STATUS "NGRAPH_UNIT_TEST_OPENVINO_ENABLE = TRUE")
     add_definitions(-DNGRAPH_UNIT_TEST_OPENVINO_ENABLE)
 
-    if (ENABLE_MKL_DNN)
-        message(STATUS "NGRAPH_TESTS: CPU enabled")
```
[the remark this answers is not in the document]
arogowie-intel: I'm aware that this might not be the best solution. However at the moment of writing this I couldn't add to the nGraph another backend with the same name. We already had backends named `CPU` and `GPU`, thus that would conflict. I thought of one solution like `IE:<device_name>` ie: `IE:GPU`, but that appeared to cause some compilation errors (I have explained this in one of the comments above). 

After the recent work of one of our colleagues on enabling this backend in CI it turned out that there is much of work needed to enable building simultaneously multiple backends in nGraph along with `IE` backend (ie: `CPU`, `PlaidML` and `IE`) inside DLDT repository. Thus we would have to come with another solution in the future.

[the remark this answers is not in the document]
arogowie-intel: The other point that was behind this change was to somehow differentiate this backend from other nGraph backends. That was needed to be able to run unit tests from ONNX library.

src/ngraph/runtime/ie/unit_test.manifest
```diff
@@ -0,0 +1,1026 @@
...
+split_var_len_parts
+floor_int32
+strided_slice_0
+strided_slice_1
+strided_slice_stride_optional
+divide_int32
+divide_cpp_rounding_int32
+divide_python_rounding_int32
+convert_int32_bool
+lesseq_int32
```
AnastasiaKazantaeva, line 1026: Who will track this file later and keep it up to date?

tomdol: Anyone working on ngraph code who introduces a new test, breaking change or anything else that makes a unit test on IE backend should add this test name here. There are similar manifests for other backends and for various reasons we have to exclude certain tests in them.

AnastasiaKazantaeva: I am not 100% sure that all of developers/testers know why some test was added here and when it is possible to remove it. This is so long file and not the easiest way to filter test on backend. However, as far as I know this is the only way.

src/ngraph/runtime/ie/CMakeLists.txt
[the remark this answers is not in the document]
ilyachur: It is a nitpick.

Just to think what happens if I build nGraph separately  from OpenVINO with `NGRAPH_IE_ENABLE`?
I will get an error that inference_engine library was not found. Maybe we can use find_package command.

CMakeLists.txt
```diff
@@ -133,12 +133,14 @@ option(NGRAPH_CODE_COVERAGE_ENABLE "Enable code coverage data collection" FALSE)
 option(NGRAPH_LIB_VERSIONING_ENABLE "Enable shared library versioning" FALSE)
 option(NGRAPH_PYTHON_BUILD_ENABLE "Enable build nGraph python package wheel" FALSE)
 option(NGRAPH_PLAIDML_ENABLE "Enable the PlaidML backend" ${PLAIDML_FOUND})
+option(NGRAPH_IE_ENABLE "Enable the Inference Engine backend" FALSE)
```
ilyachur, line 136: Could you use `NGRAPH_UNIT_TEST_OPENVINO_ENABLE` to enable IE backend?

or the second variant is removing `NGRAPH_UNIT_TEST_OPENVINO_ENABLE` option and replace this to `NGRAPH_IE_ENABLE`

Also please notify if cmake options should be updated (for example NGRAPH_IE_ENABLE ON for the Inference Engine).

arogowie-intel: @ilyachur Hi, I've remove the `NGRAPH_UNIT_TEST_OPENVINO_ENABLE` flag from our cMake. The `NGRAPH_IE_ENABLE` which is used to enable this backend should be enough. Please update your CMake.

ilyachur: Hi @arogowie-intel,

Thank you for the notification, but we will be able to update CMake files only when update nGraph repository to new master which will contain these changes. Because in other case CMake will ignore this option and old tests also will be disabled.

arogowie-intel: OK, this shouldn't be a problem. We simply pass `-DNGRAPH_IE_ENABLE` flag when running cmake command in DLDT and it works.

PR #4420, opened 2020-03-06
[Py] Add support for testing Wheels through tox

postrational:
As part of this PR, I needed to move the sources of nGraph's Python API into an `src` subdirectory. 

This is a recommended practice, [rationale here](https://hynek.me/articles/testing-packaging/).

Moved directories:
`ngraph/python/ngraph -> ngraph/python/src/ngraph`
`ngraph/python/pyngraph -> ngraph/python/src/pyngraph`

It is necessary in our case if we want to run tests using code installed in the `tox` environment `site-packages` rather than on the sources in the root directory.

Main change in `tox` is going from:

```
python setup.py develop
```

to

```
python setup.py bdist_wheel
pip install --no-index --pre --find-links=dist/ ngraph-core
```


python/setup.py
```diff
@@ -420,7 +409,7 @@ def build_extensions(self):
     long_description=open(os.path.join(PYNGRAPH_ROOT_DIR, 'README.md')).read(),
     long_description_content_type='text/markdown',
     ext_modules=ext_modules,
-    package_dir=package_dir,
+    package_dir={'': 'src'},
```
arogowie-intel: Does this `'' ` empty quote mean no name for package?  Earlier there was `'ngraph.impl.onnx_import'`

raramer01:
I cannot approve this PR until cmake files that can be used to build ngraph on windows as documented in PR 4353 are included.

raramer01:
At this time, I am getting errors when running tox with the interpreter backend...The tox environment assumes the location of the interpreter.dll library to be .tox/py3/lib but it is not there, even though ngraph was built with -DNGRAPH_INTERPRETER_ENABLE=TRUE...will also try with cpu as those libraries are present...

raramer01:
Latest status: getting 2 test failures with the cpu backend, email has been sent to Tomasz and Michal with details...

PR #4422, opened 2020-03-06
Use GetOutputElement for ONNX Identity op

postrational:


postrational:
We should select this option or https://github.com/NervanaSystems/ngraph/pull/4423

PR #4423, opened 2020-03-06
Use Add (input+0) for ONNX Identity op

postrational:


postrational:
We should select this option or https://github.com/NervanaSystems/ngraph/pull/4422

postrational:
In order to use Transpose we would need to know at least the rank of the input, to calculate the transpose input_order. Let's stick to this for now, as it's the most dynamic solution.

PR #4425, opened 2020-03-06
GetOutputElement removal preparation

diyessi:
Not all outputs are used so don't force them to be connected in replace
Add pattern that matches on any output
Remove GOEs by default, allow to disable
Fix failing core passes/tests with GOE dependency
Disable GOE-removal on cpu-specific tests that don't create a transformer before building the graph
Convert a few CPU passes to be value-friendly


src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -2010,7 +2010,7 @@ namespace
             auto allIVs = ValueHandle::makeIndexHandles(vArg.rank());
             auto pAllIVs = makeHandlePointers(allIVs);
             auto steps = vArg.getSteps();
-            SmallVector<ValueHandle, 8> nonRedIVs;
+            SmallVector<ValueHandle, 8> nonRedIVs, tempIVs;
```
rkimballn1: Single decl per line, no commas, or is this ok here because it is in mlir?
```suggestion
            SmallVector<ValueHandle, 8> nonRedIVs;
            SmallVector<ValueHandle, 8> tempIVs;
```

src/ngraph/node.hpp
```diff
@@ -60,6 +60,12 @@ namespace ngraph
 
     class Function;
 
+    // Intermal, controls whether GetOutputElement nodes are elided
+    // Defaults to being elided. Transformer should set to false if
+    // it has passes that depend on GetOutputElement.
+    NGRAPH_API void set_remove_goe(bool value);
```
rkimballn1: These fuctions should be deprecated.

src/ngraph/node_output.hpp
```diff
@@ -39,6 +39,8 @@ namespace ngraph
     template <>
     class NGRAPH_API Output<Node>
     {
+        void eliminate_goe();
```
rkimballn1: should be deprecated

src/contrib/mlir/backend/pass/affine_lowerer.cpp
diyessi: Borrowed from another PR, but I fixed it there and here.

abhiram-krishnan:
TESTNOW_DLDT_E2E

src/ngraph/pattern/op/any_output.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include "ngraph/pattern/op/pattern.hpp"
+
+namespace ngraph
+{
+    namespace pattern
+    {
+        namespace op
+        {
+            /// Matches any output of a node
+            class NGRAPH_API AnyOutput : public Pattern
```
ilyachur: Why do we require this op? We don't contain this op in any opsets.

src/ngraph/node_output.cpp
```diff
@@ -15,21 +15,34 @@
...
 #include "ngraph/node_output.hpp"
+#include "ngraph/log.hpp"
 #include "ngraph/node.hpp"
 #include "ngraph/op/get_output_element.hpp"
 
 namespace ngraph
 {
+    namespace
+    {
+        static bool remove_goe = true;
```
ilyachur: Why do we need remove_goe as a global object? It can be a problem if user works with several nGraph functions.

test/util/test_tools.hpp
```diff
@@ -45,6 +45,19 @@ namespace ngraph
     class Function;
 }
 
+class DisableRemoveGOE
```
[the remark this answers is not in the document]
diyessi: We still need IATransformer to work, and it still needs many changes to work without GetOutputElement.

src/ngraph/pattern/op/any_output.hpp
diyessi: Pattern "ops" are for pattern matching patterns; they are in the Node hierarchy so they can use sub-graphs as pattern constants.

src/ngraph/node_output.cpp
diyessi: In normal (non-test) use, it is set once during initialization of a transformer that doesn't work when GetOutputElements are missing. To fix those transformers would take at least another month (they do need to be fixed).

ilyachur: Could you add a comment that it is a temporary solution? And that it will be removed after the transformers fix.

ilyachur: Can we plan the activity to fix those transformers in order to don't forget it?

PR #4427, opened 2020-03-09
Update README.md

rkimballn1:
Get rid of ugly python syntax highlight on what is bash code, not python code


indie:
If this was my mistake, apologies; pip as a python module is what usually works. Markdown is messy.

PR #4428, opened 2020-03-09
replace copy_with_new_args with clone_with_new_inputs

rkimballn1:
Mostly regex replace but:
* copy_with_new_args was protected in Node and that need to change to public since passes call copy_with_new_args
* CompiledKernel and TensorIterator needed a little extra care


ilyachur:
Does this change break backward compatibility?

diyessi:
It breaks incorrect use, i.e. calling a protected member that happened to be defined non-protected in a derived class. Without this PR, we will cease work on removing GOEs in ngraph.

ilyachur:
> It breaks incorrect use, i.e. calling a protected member that happened to be defined non-protected in a derived class. Without this PR, we will cease work on removing GOEs in ngraph.

But If user create a new nGraph operation and doesn't specify `clone_with_new_inputs` method, will it work? or user should implement this method?

diyessi:
New ops should implement this operation instead of copy_with_new_args. I have caught a few not using it in review (this will greatly reduce that, since people copy existing ops to make new ones.

ilyachur:
Ok, but if customer has own custom operation (which was developed before this change), he MUST  replace `copy_with_new_args` to `clone_with_new_inputs`, right?

diyessi:
> Ok, but if customer has own custom operation (which was developed before this change), he MUST replace `copy_with_new_args` to `clone_with_new_inputs`, right?

Yes. And GetOutputElement must exist if they define `copy_with_new_args`

diyessi:
> > Ok, but if customer has own custom operation (which was developed before this change), he MUST replace `copy_with_new_args` to `clone_with_new_inputs`, right?
> 
> Yes. And GetOutputElement must exist if they define `copy_with_new_args`

It looks like this PR doesn't drop support for keeping `copy_with_new_args`, it just updates all the ops we already have, so the customer's op would continue to work, but only until GOEs were removed.

PR #4431, opened 2020-03-10
[ NonZero ] operator with constant folding and onnx_importer support

mozga-intel:
This PR adds the ONNX NonZero op support https://github.com/onnx/onnx/blob/master/docs/Operators.md#NonZero
THe behavior is similar to numpy's NonZero https://docs.scipy.org/doc/numpy/reference/generated/numpy.nonzero.html


src/ngraph/op/non_zero.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#pragma once
+
+#include "ngraph/op/op.hpp"
+
+namespace ngraph
+{
+    namespace op
+    {
+        namespace v0
```
postrational: This is a v2 op, I believe it should be in the `ngraph::op::v2` namespace to keep things tidy.

postrational: @diyessi What do you think?

diyessi: Doesn't matter to me as long as the next version is a bigger number.

src/ngraph/op/non_zero.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+bool ngraph::op::v0::NonZero::visit_attributes(AttributeVisitor& visitor)
+{
+    return true;
+}
+
+void op::v0::NonZero::validate_and_infer_types()
+{
+    auto args_et_pshape = validate_and_infer_elementwise_args();
```
postrational: The output shape of this op should be `[rank(data), num_non_zero]` where `num_non_zero` is dependent on input data, so it must be dynamic. I think `validate_and_infer_elementwise_args` will not produce this shape.

src/ngraph/op/non_zero.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    element::Type& args_et = std::get<0>(args_et_pshape);
+    PartialShape& args_pshape = std::get<1>(args_et_pshape);
+
+    set_output_type(0, args_et, args_pshape);
```
postrational: Please also use `set_input_is_relevant_to_shape` to indicate that shape is dependent on input.

src/ngraph/op/non_zero.hpp
postrational: I think we should stick to giving ops names indicating the opset they were introduced with. This will make it easier to recreate their history just by looking at the namespaces in the op definition.

src/ngraph/frontend/onnx_import/op/non_zero.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+    namespace onnx_import
+    {
+        namespace op
+        {
+            namespace set_1
+            {
+                NodeVector non_zero(const Node& node)
+                {
+                    const Output<ngraph::Node> data = node.get_ng_inputs().at(0);
```
tomdol: This isn't required but I'd suggest changing `Output<ngraph::Node>` to `auto`. The `data` type will then be deduced to `shared_ptr<Node>` and this will be forwarded to the `NonZero` constructor. This will be more future-proof if for any reason the way ops are connected in the graph changes.

postrational:
We need constant folding for this operation
We need unit tests for shape inference functions
Operation should belong to opset3.

src/ngraph/pass/constant_folding_non_zero.cpp
```diff
@@ -0,0 +1,198 @@
...
+    auto constant_non_zero_callback = [data_label](pattern::Matcher& m) {
+        auto pattern_map = m.get_pattern_map();
+
+        const auto data = static_pointer_cast<op::Constant>(pattern_map[data_label]);
+
+        // const auto found_nz_node = m.get_match_root();
+        // this fails because the output of NonZero is still dynamic - is it ok to remove this line?
+        // or should the output shape of this op be calculated as the maximum possible number
+        // of non-zero indices it can return if the input is a constant?
+        // NGRAPH_CHECK(revalidate_and_ensure_static(found_nz_node));
```
tomdol: @diyessi I think this line should be removed for an op that is dynamic until the end of its inference. Is this correct? Please have a look at the comment I left in this code above.

src/ngraph/pass/constant_folding_non_zero.cpp
```diff
@@ -0,0 +1,198 @@
...
+        {
+            if (input_values[i] != zero_value)
+            {
+                indices.push_back(i);
+            }
+        }
+
+        indices.shrink_to_fit();
+        return op::Constant::create(element::i64, Shape{1, indices.size()}, indices);
+    }
```
arogowie-intel: Have you tried 1D data with you helper structure`NonZeroElements`? I suppose it might work.

tomdol: Yes, it does work but the point of this specialization for vectors is to have an even faster implementation. In this `else if` branch there's less allocation, less copying and less processing of indices.

jane-intel:
You could take some constant folding tests from my closed review: https://github.com/NervanaSystems/ngraph/pull/4467/files#diff-db47459bf7a016ad45057ea93b04578f

src/ngraph/pass/constant_folding_non_zero.cpp
```diff
@@ -0,0 +1,198 @@
...
+        case element::Type_t::i8: replacement = fold_constant_non_zero<int8_t>(data); break;
+        case element::Type_t::i16: replacement = fold_constant_non_zero<int16_t>(data); break;
+        case element::Type_t::i32: replacement = fold_constant_non_zero<int32_t>(data); break;
+        case element::Type_t::i64: replacement = fold_constant_non_zero<int64_t>(data); break;
+        case element::Type_t::u8: replacement = fold_constant_non_zero<uint8_t>(data); break;
+        case element::Type_t::u16: replacement = fold_constant_non_zero<uint16_t>(data); break;
+        case element::Type_t::u32: replacement = fold_constant_non_zero<uint32_t>(data); break;
+        case element::Type_t::u64: replacement = fold_constant_non_zero<uint64_t>(data); break;
+        case element::Type_t::u1:
+        case element::Type_t::boolean:
```
tsocha: Just wonder if NonZero operator could return `indicies` of all true values for boolean type. It could be helpful in some cases.

tomdol: The spec only mentions numerical types for the input data, this is why I disabled bool in here.

tomdol:
> 
> 
> You could take some constant folding tests from my closed review: https://github.com/NervanaSystems/ngraph/pull/4467/files#diff-db47459bf7a016ad45057ea93b04578f

Done: https://github.com/NervanaSystems/ngraph/pull/4431/commits/2b7f98de613b6e51944c1b324fde085bcb7ad606

src/ngraph/frontend/onnx_import/op/non_zero.hpp
```diff
@@ -0,0 +1,46 @@
...
+        namespace op
+        {
+            namespace set_1
+            {
+                ///
+                /// \brief Convert ONNX NonZero operation to an nGraph node.
+                ///
+                /// \param node   The ONNX node object representing this operation.
+                ///
+                /// \return The vector containing Ngraph nodes producing output of ONNX NonZero
```
rkimballn1:
```suggestion
                /// \return The vector containing nGraph nodes producing output of ONNX NonZero
```

src/ngraph/op/op_version_tbl.hpp
```diff
@@ -153,6 +153,7 @@ NGRAPH_OP(Multiply, ngraph::op::v0, 0)
 NGRAPH_OP(Multiply, ngraph::op::v1, 1)
 NGRAPH_OP(Negative, ngraph::op::v0, 0)
 NGRAPH_OP(NonMaxSuppression, ngraph::op::v1, 1)
+NGRAPH_OP(NonZero, ngraph::op::v3, 0)
```
rkimballn1:
```suggestion
NGRAPH_OP(NonZero, ngraph::op::v3, 3)
```
If this is in the v3 namespace then it is a member of opset 3, not opset 0

rkimballn1:
You should put a link to the onnx non zero doc or document the op. I have no idea what this is trying to do. Please put the url to the doc in the description

src/ngraph/pass/constant_folding_non_zero.cpp
```diff
@@ -0,0 +1,198 @@
...
+        NonZeroElements(const Shape& input_shape)
+            : m_input_shape{input_shape}
+            , m_results{Results_t(input_shape.size())}
+        {
+        }
+
+        template <typename T>
+        const Results_t& find_indices(const T* values)
+        {
+            m_current_index = Shape(m_input_shape.size(), 0UL);
```
rkimballn1: There is a trivial function Constant::get_all_data_elements_bitwise_identical() which returns if all elements are identical. This is computed as the constant is constructed so it simply returns a stored value. You should check if you need to go any further in this routine since it iterates through all values of every constant in a model which is expensive.

src/ngraph/pass/constant_folding_non_zero.cpp
```diff
@@ -0,0 +1,198 @@
...
+            const T zero_value = T{0};
+            for (size_t i = 0; i < values_count; ++i)
+            {
+                if (values[i] != zero_value)
+                {
+                    add_to_results(m_current_index);
+                }
+
+                // don't generate the next index when the last value has been processed
+                if (i < values_count - 1)
```
lazarevevgeny: It will be more efficient if we change the "for" loop condition to "i + 1 < values_count" and remove the "if" statement in this line. Then after the "for" loop we check that "values_count != 0" and add if the last value is not equal to 0 then call "add_to_results(m_current_index)"

src/ngraph/pass/constant_folding_non_zero.cpp
```diff
@@ -0,0 +1,198 @@
...
+            {
+                m_results.at(dim).push_back(index[dim]);
+            }
+        }
+
+        // Generates an index pointing to the next element in the flattened tensor
+        // It behaves similar to flipping bits when incrementing a binary number
+        void next_index()
+        {
+            for (size_t dim = m_current_index.size() - 1; dim >= 0; --dim)
```
lazarevevgeny: lets insert an assert here that m_current_index.size() != 0

test/constant_folding.cpp
```diff
@@ -2376,3 +2376,73 @@ TEST(constant_folding, pass_property)
     ASSERT_FALSE(pass->get_property(pass::PassProperty::REQUIRE_STATIC_SHAPE));
     ASSERT_TRUE(pass->get_property(pass::PassProperty::CHANGE_DYNAMIC_STATE));
 }
+
+TEST(constant_folding, constant_non_zero_1D)
```
lazarevevgeny: Please, add a test case for 0D input.

src/ngraph/pass/constant_folding_non_zero.cpp
lazarevevgeny: Let's add support for boolean data types also.

src/ngraph/pass/constant_folding_non_zero.cpp
tomdol: I'd rather do it once after this line https://github.com/NervanaSystems/ngraph/pull/4431/files?file-filters%5B%5D=.cpp#diff-7079b532eda2d11cc6adcf43a0a6d972R38
We don't have to check it every time `next_index()` is called, do we?

src/ngraph/pass/constant_folding_non_zero.cpp
tomdol: Thanks for this hint, I've used this method to short circuit here https://github.com/NervanaSystems/ngraph/pull/4431/files?file-filters%5B%5D=.cpp#diff-7079b532eda2d11cc6adcf43a0a6d972R119-R124

src/ngraph/pass/constant_folding_non_zero.cpp
tomdol: Done, thanks for the tip

test/constant_folding.cpp
tomdol: It was already tested via an onnx model but I've added a test here too.

src/ngraph/pass/constant_folding_non_zero.cpp
diyessi: By `assert` I assume you mean `throw`.

src/ngraph/pass/constant_folding_non_zero.cpp
```diff
@@ -0,0 +1,235 @@
...
+{
+    const auto input_shape = data->get_shape();
+    const auto* input_values = data->get_data_ptr<T>();
+    const bool identical_elems_in_data = data->get_all_data_elements_bitwise_identical();
+
+    if (identical_elems_in_data)
+    {
+        NGRAPH_CHECK(input_values[0] != T{0},
+                     "It's not possible to constant fold a NonZero op with an input containing "
+                     "only zeros.");
```
diyessi: You can't make not being able to fold a constant an error -- if you can't fold, you just don't fold.

src/ngraph/pass/constant_folding_non_zero.cpp
tomdol: I've added an NGRAPH_CHECK in this line https://github.com/NervanaSystems/ngraph/pull/4431/files?file-filters%5B%5D=.cpp#diff-7079b532eda2d11cc6adcf43a0a6d972R35

src/ngraph/pass/constant_folding_non_zero.cpp
tomdol: Done with a test to make sure that CF doesn't happen in this case.

src/ngraph/pass/constant_folding_non_zero.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    const auto input_shape = data->get_shape();
+    const auto* input_values = data->get_data_ptr<T>();
+    const bool identical_elems_in_data = data->get_all_data_elements_bitwise_identical();
+
+    if (identical_elems_in_data && input_values[0] == T{0})
+    {
+        NGRAPH_WARN << "It's not possible to constant fold a NonZero op with an input containing "
+                       "only zeros.";
+
```
rkimballn1, line 125:
```suggestion
        NGRAPH_WARN << "It's not possible to constant fold a NonZero op with an input containing "
                       "only zeros.";

```

rkimballn1: This should not be a warning. It is normal for Constants to contain zeros.

PR #4441, opened 2020-03-13
Update Python API to produce v1 ops

postrational:


python/pyngraph/node_factory.cpp
```diff
@@ -0,0 +1,177 @@
...
+    virtual void on_attribute(const std::string& name, std::string& value) override
+    {
+        if (m_attributes.contains(name))
+        {
+            value = m_attributes[name.c_str()].cast<std::string>();
+        }
+    }
+    virtual void on_attribute(const std::string& name, bool& value) override
+    {
+        if (m_attributes.contains(name))
```
tsocha: Why this if is needed every where?
What happen when name is absent from `m_attributes`?

python/setup.py
```diff
@@ -102,6 +102,16 @@ def parallelCCompile(
     macros, objects, extra_postargs, pp_opts, build = self._setup_compile(
         output_dir, macros, include_dirs, sources, depends, extra_postargs)
     cc_args = self._get_cc_args(pp_opts, debug, extra_preargs)
+
+    if NGRAPH_PYTHON_DEBUG in ['TRUE', 'ON', True]:
+        try:
+            # pybind11 is much more verbose without -DNDEBUG
```
tsocha: That's weird.

python/pyngraph/node_factory.cpp
arogowie-intel: We have to check whether the requested attribute name is present in the dictionary we received. If the attribute name is absent we just don't update it.

python/setup.py
arogowie-intel: But unfortunately that's true. When I made sure that this flag is present, then I much faster managed to find out what the problem was. That's just debugging convenience.

PR #4450, opened 2020-03-17
[ONNX] Enable cross-compiling and ninja build for onnx_importer

tsocha:
- Change minimum cmake version to 3.11
- Enable cross-compilation of protobuf
- Rewrite ONNX external project into FetchContent
- Fix some issues with ninja build on windows


cmake/external_onnx.cmake
```diff
@@ -31,70 +30,31 @@ set(ONNX_GIT_REPO_URL https://github.com/onnx/onnx.git)
...
 
-ExternalProject_Add(
+if (WIN32)
+    string(REPLACE "/W3" "/W0" CMAKE_ORIGINAL_CXX_FLAGS "${CMAKE_ORIGINAL_CXX_FLAGS}")
+else()
+    string(REPLACE "-Werror" "" CMAKE_CXX_FLAGS "${CMAKE_ORIGINAL_CXX_FLAGS}")
+endif()
+
+FetchContent_Declare(
     ext_onnx
```
arogowie-intel: Is it the project name or build library name? You have earlier changed the `ext_protobuf` dependency to `onnx`. This is just not clear to me.

tsocha: This is a name of FetchContent declaration AFAIK.
Library name will be known in line 53 when subdirectory will be added and onnx configuration step will finish.

cmake/external_onnx.cmake
```diff
@@ -31,70 +30,31 @@ set(ONNX_GIT_REPO_URL https://github.com/onnx/onnx.git)
...
 
 add_definitions(-DONNX_BUILD_SHARED_LIBS=ON)
-add_definitions(-DONNX_ML=ON)
 
-ExternalProject_Add(
+if (WIN32)
+    string(REPLACE "/W3" "/W0" CMAKE_ORIGINAL_CXX_FLAGS "${CMAKE_ORIGINAL_CXX_FLAGS}")
+else()
+    string(REPLACE "-Werror" "" CMAKE_CXX_FLAGS "${CMAKE_ORIGINAL_CXX_FLAGS}")
+endif()
```
rkimballn1: This is what is causing the failure with GCPU. What you are doing here is completely overwriting the CMAKE_CXX_FLAGS for the entire ngraph. Don't do this.
By overwriting CMAKE_CXX_FLAGS for ngraph you are disabling dynamic loading of shared backends. Since this is disabled then the GCPU backend is not looked for as it should be. You need to find a way to disable -Werror without overwriting CMAKE_CXX_FLAGS. This file is add to the cmake file with `include` so you are running at the scope of the main cmake file.

rkimballn1: I removed this code

test/CMakeLists.txt
```diff
@@ -591,6 +591,10 @@ if (NGRAPH_INTERPRETER_ENABLE)
     target_link_libraries(unit-test PRIVATE interpreter_backend)
 endif()
 
+if (NGRAPH_GENERIC_CPU_ENABLE)
+    target_link_libraries(unit-test PRIVATE gcpu_backend)
+endif()
+
```
silee2: @rkimballn1 I previously added this code to see if it helps the build issue. Do we still need it or would it be better to remove it?  Not sure if leaving it there will do any harm.

silee2: @rkimballn1 If GCPU backend is always dynamically loaded, build time linking isn't needed. Right (no reference to GCPU related symbols from anywhere in unit-test)?

silee2:
general_ci: `CPU_gcc_ubuntu1804_MLIR_prebuilt` is still CMake 3.10.2. Need to be updated to 3.11+

raramer01:
TESTNOW_DLDT_E2E

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 # ******************************************************************************
 
-cmake_minimum_required (VERSION 3.4)
+cmake_minimum_required (VERSION 3.11)
```
dkurt, line 17: Ubuntu 18.04 has CMake 3.10.2: https://packages.ubuntu.com/bionic/cmake  
Ubuntu 16.04 has 3.5.1: https://packages.ubuntu.com/xenial/cmake

Please do not force users to look for non-default CMake versions. Which feature is not covered by previous 3.4 requirement?

/cc @alalek @mshabunin @ilyachur

tomdol: The minimum cmake version bump was required because of this https://cliutils.gitlab.io/modern-cmake/chapters/projects/fetch.html which is supported starting from 3.11

dkurt: Is there alternatives for it?

/cc @ilya-lavrenov

tomdol: Not that I'm aware of. FetchContent was used for the ONNX dependency to be able to cross compile it for ARM platforms that OV has to support.

PR #4451, opened 2020-03-17
Squeeze shape inference for dynamic input shape

jane-intel:


src/ngraph/op/fused/squeeze.cpp
```diff
@@ -38,25 +38,33 @@ void op::Squeeze::pre_validate_and_infer_types()
...
+    bool axes_is_empty_constant = (axes_is_constant) ? as_type_ptr<op::Constant>(axes_node)->cast_vector<int64_t>().empty() : false;
+
+    if (dynamic_data_rank || !axes_is_constant || (dynamic_data_shape && axes_is_empty_constant))
     {
         set_output_type(0, get_input_element_type(0), PartialShape::dynamic());
         return;
     }
 
-    auto data_shape = data.get_shape();
+    auto data_rank = static_cast<int64_t>(data.get_partial_shape().rank());
```
diyessi: Use `get_length()` instead of the static cast, which will have problems on 32-bit platforms.

src/ngraph/op/fused/squeeze.cpp
```diff
@@ -73,24 +81,27 @@ void op::Squeeze::pre_validate_and_infer_types()
...
+                        "provided axis value is invalid. Only axes of size 1 may be removed.");
+            }
             axes_to_squeeze.at(axis) = 1;
         }
     }
 
-    Shape output_data_shape;
-    for (size_t idx = 0; idx < data_shape.size(); ++idx)
+    std::vector<Dimension> output_data_shape;
+    for (size_t idx = 0; idx < data_rank; ++idx)
```
diyessi: This `size_t` will probably need to be `uint64_t` when you make the other change.

jane-intel: Thanks! Done

src/ngraph/op/fused/squeeze.cpp
```diff
@@ -38,25 +38,35 @@ void op::Squeeze::pre_validate_and_infer_types()
     auto data = input_value(0);
     auto axes_node = input_value(1).get_node_shared_ptr();
 
-    if (data.get_partial_shape().is_dynamic() || !axes_node->is_constant())
+    bool dynamic_data_rank = data.get_partial_shape().rank().is_dynamic();
+    bool dynamic_data_shape = data.get_partial_shape().is_dynamic();
+
+    bool axes_is_constant = axes_node->is_constant();
```
diyessi: Might be simpler if you said `axes_constant = as_type_ptr<axes_node>`. Then `axes_constant` will be null if it's not a constant, which might make the next lines simpler.

src/ngraph/op/fused/squeeze.cpp
```diff
@@ -38,25 +38,35 @@ void op::Squeeze::pre_validate_and_infer_types()
...
+                           : false;
+
+    if (dynamic_data_rank || !axes_is_constant || (dynamic_data_shape && axes_is_empty_constant))
     {
         set_output_type(0, get_input_element_type(0), PartialShape::dynamic());
         return;
     }
 
-    auto data_shape = data.get_shape();
+    uint64_t data_rank = data.get_partial_shape().rank().get_length();
```
diyessi: Switch this and the next line so you get the rank from `data_partial_shape`.

src/ngraph/op/fused/squeeze.cpp
```diff
@@ -38,25 +38,35 @@ void op::Squeeze::pre_validate_and_infer_types()
...
+    auto data_partial_shape = data.get_partial_shape();
 
     // Get value of axes from Constant
     auto axes_constant = as_type_ptr<op::Constant>(axes_node);
```
diyessi: This line would be up above with my suggested refactor.

src/ngraph/op/fused/squeeze.cpp
```diff
@@ -73,24 +83,27 @@ void op::Squeeze::pre_validate_and_infer_types()
...
-                "provided axis value is invalid. Only axes of size 1 may be removed.");
+            if (!dynamic_data_shape)
+            {
+                auto data_shape = data.get_shape();
+                NODE_VALIDATION_CHECK(
+                    this,
+                    (data_shape.at(axis) == 1),
+                    "provided axis value is invalid. Only axes of size 1 may be removed.");
+            }
             axes_to_squeeze.at(axis) = 1;
```
diyessi: Add a test for this check.

src/ngraph/op/fused/squeeze.cpp
Ewa21: Done

src/ngraph/op/fused/squeeze.cpp
Ewa21: Done

src/ngraph/op/fused/squeeze.cpp
Ewa21: Done

src/ngraph/op/fused/squeeze.cpp
Ewa21: Done

src/ngraph/op/fused/squeeze.cpp
```diff
@@ -38,25 +38,33 @@ void op::Squeeze::pre_validate_and_infer_types()
     auto data = input_value(0);
     auto axes_node = input_value(1).get_node_shared_ptr();
 
-    if (data.get_partial_shape().is_dynamic() || !axes_node->is_constant())
+    bool dynamic_data_rank = data.get_partial_shape().rank().is_dynamic();
```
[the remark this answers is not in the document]
Ewa21: Done

PR #4452, opened 2020-03-17
Update calls to copy_with_new_args to clone_with_new_inputs

rkimballn1:
This PR just changes all calls to the function copy_with_new_args and does not address changing the op implementations which will be a large follow-on PR


test/copy.cpp
```diff
@@ -114,16 +114,16 @@ TEST(copy, concat)
...
     size_t axis = 0;
     auto node = make_shared<op::Concat>(NodeVector{arg0, arg1}, axis);
-    auto new_node = node->copy_with_new_args(new_args);
+    auto new_node = node->clone_with_new_inputs(new_args);
     auto node_cast = as_type_ptr<op::Concat>(new_node);
     ASSERT_NE(node_cast, nullptr);
 
     ASSERT_TRUE(nullptr != new_node);
-    ASSERT_TRUE(new_args == new_node->get_arguments());
+    ASSERT_TRUE(new_args == as_output_vector(new_node->get_arguments()));
```
diyessi: Why not `new_args == new_node->input_values()` ?

test/copy.cpp
```diff
@@ -289,17 +289,17 @@ TEST(copy, select)
...
 
     auto node = make_shared<op::Select>(arg0, arg1, arg2);
-    auto new_node = node->copy_with_new_args(new_args);
+    auto new_node = node->clone_with_new_inputs(new_args);
     auto node_cast = as_type_ptr<op::Select>(new_node);
     ASSERT_NE(node_cast, nullptr);
 
     ASSERT_TRUE(nullptr != new_node);
-    ASSERT_TRUE(new_args == new_node->get_arguments());
+    ASSERT_TRUE(new_args == as_output_vector(new_node->get_arguments()));
```
diyessi: Same question as above.

test/copy.cpp
rkimballn1: why not put that in a suggestion so I can just click it to accept?

PR #4455, opened 2020-03-18
[ ShapeOf Const Folding ] Dynamic ShapeOf value propagation

jane-intel:


src/ngraph/pass/constant_folding_shape_of.cpp
```diff
@@ -14,6 +14,7 @@
 // limitations under the License.
 //*****************************************************************************
 
+#include <ops.hpp>
```
diyessi: "ngraph/ops.hpp" and sort/run style check

src/ngraph/pass/constant_folding_shape_of.cpp
```diff
@@ -47,6 +49,34 @@ void pass::ConstantFolding::construct_constant_shape_of()
...
+        {
+            auto shape_of = make_shared<op::ShapeOf>(arg_match);
+            auto dimensions = NodeVector{};
+            auto output_dimensions = vector<Dimension>(partial_shape);
+            for (size_t i = 0; i < output_dimensions.size(); ++i)
+            {
+                if (output_dimensions[i].is_static())
+                {
+                    auto value = std::vector<int64_t>{output_dimensions[i].get_length()};
+                    dimensions.push_back(make_shared<op::Constant>(element::i64, Shape{1}, value));
```
diyessi: Here and below, make a temp for the constant and set its friendly name and then do the push_back. This also lets you iterate over the output dimensions; no need for `i`.

src/ngraph/pass/constant_folding_shape_of.cpp
```diff
@@ -47,6 +49,34 @@ void pass::ConstantFolding::construct_constant_shape_of()
 
             return true;
         }
+        else if (partial_shape.rank().is_static())
+        {
+            auto shape_of = make_shared<op::ShapeOf>(arg_match);
+            auto dimensions = NodeVector{};
```
diyessi: Since you are collecting values to pass to `Concat`, use `OutputVector`.

src/ngraph/pass/constant_folding_shape_of.cpp
```diff
@@ -35,7 +36,8 @@ void pass::ConstantFolding::construct_constant_shape_of()
 
         auto arg_match = pattern_value_map[arg_label];
 
-        if (arg_match.get_partial_shape().is_static())
+        auto partial_shape = arg_match.get_partial_shape().;
```
mitruska: Little dot artifact
```suggestion
        auto partial_shape = arg_match.get_partial_shape();
```

src/ngraph/pass/constant_folding_shape_of.cpp
```diff
@@ -47,6 +49,34 @@ void pass::ConstantFolding::construct_constant_shape_of()
...
+        else if (partial_shape.rank().is_static())
+        {
+            auto shape_of = make_shared<op::ShapeOf>(arg_match);
+            auto dimensions = NodeVector{};
+            auto output_dimensions = vector<Dimension>(partial_shape);
+            for (size_t i = 0; i < output_dimensions.size(); ++i)
+            {
+                if (output_dimensions[i].is_static())
+                {
+                    auto value = std::vector<int64_t>{output_dimensions[i].get_length()};
```
rkimballn1:
```suggestion
                    auto value = std::vector<int64_t>{static_cast<int64_t>(output_dimensions[i].get_length())};
```

src/ngraph/pass/constant_folding_shape_of.cpp
rkimballn1: There is a constructor that takes an initializer list as the 3rd argument. That might be easier than creating a vector. Same comment for below

src/ngraph/pass/constant_folding_shape_of.cpp
```diff
@@ -35,7 +36,13 @@ void pass::ConstantFolding::construct_constant_shape_of()
 
         auto arg_match = pattern_value_map[arg_label];
 
-        if (arg_match.get_partial_shape().is_static())
+        auto partial_shape = arg_match.get_partial_shape();
+        auto original_shape_of_node = as_type_ptr<op::ShapeOf>(m.get_match_root());
+        if (!original_shape_of_node->is_foldable())
+        {
+            return true;
```
tomdol: I'm not sure but I think we should return false if constant folding didnt occur.

src/ngraph/op/experimental/shape_of.hpp
```diff
@@ -39,6 +39,11 @@ namespace ngraph
                     copy_with_new_args(const NodeVector& new_args) const override;
 
                 void validate_and_infer_types() override;
+
```
diyessi: It should be `set_is_foldable(bool value)`
This needs documentation.  What is it for? It looks like it is being used in such a way that it should be preserved in a cloning operation, which makes it like an attribute.

arogowie-intel: Maybe rename it like `is_recursively_folded` which would has meaning that recursive folding has happened.

jane-intel:
**General observation**

Let's say that we executed ConstantFolding, while having dynamic ShapeOf. 
It can not be folded to `Constant`, so we extend it with `Gather`s -> `Concat`.

After that some magical transformations were executed.

And we call ConstantFolding again. `ShapeOf` now takes input with static shape, and it can be folded, but we already marked it as "not foldable". 

We should take "not foldable" mark into account only for dynamicly-inputed `ShapeOf`.

tsocha:
> 
> 
> @jane-intel 
> **General observation**
> 
> Let's say that we executed ConstantFolding, while having dynamic ShapeOf.
> It can not be folded to `Constant`, so we extend it with `Gather`s -> `Concat`.
> 
> After that some magical transformations were executed.
> 
> And we call ConstantFolding again. `ShapeOf` now takes input with static shape, and it can be folded, but we already marked it as "not foldable".
> 
> We should take "not foldable" mark into account only for dynamicly-inputed `ShapeOf`.

Done.

src/ngraph/op/experimental/shape_of.hpp
tsocha: Renamed `block_constant_folding` => `block_recurent_constant_folding`.
I would prefer to avoid introduce any new attributes because this will force me to upgrade the spec as well.
I added a comment to these two methods, is it enough for you?

src/ngraph/op/experimental/shape_of.hpp
```diff
@@ -39,6 +39,17 @@ namespace ngraph
...
+                // FOR CONSTANT FOLDING INTERNAL USAGE ONLY
+                // Constant folding for cases with static rank but dynamic shape create a subgraph
+                // which contains a Shape of.
+                // In this case we need to prevent constant folding from endless creation of these
+                // subgraphs.
+                // These metods should be removed if better solution will be designed.
+                void block_recurent_constant_folding() { m_is_foldable = false; }
```
arogowie-intel, line 49: typo: 
```suggestion
                void block_recurrent_constant_folding() { m_is_foldable = false; }
```

src/ngraph/op/experimental/shape_of.hpp
diyessi: It's not an attribute; it doesn't affect the semantics of the op instance.

src/ngraph/op/experimental/shape_of.hpp
```diff
@@ -39,6 +39,17 @@ namespace ngraph
...
                 void validate_and_infer_types() override;
+
+                // FOR CONSTANT FOLDING INTERNAL USAGE ONLY
+                // Constant folding for cases with static rank but dynamic shape create a subgraph
+                // which contains a Shape of.
+                // In this case we need to prevent constant folding from endless creation of these
+                // subgraphs.
+                // These metods should be removed if better solution will be designed.
+                void block_recurrent_constant_folding() { m_is_foldable = false; }
+                bool is_foldable() { return m_is_foldable; }
```
diyessi, line 50: Alternate solution proposal: Cache the sub-graph in the op instance and have constant-folding use the cached value if  it there.

tsocha: Caching the sub-graph in the op instance could make an unnecessary overhaul if model would have more than one `shape_of`.

I recreated these two methods as getter and setter.
I also preserve this attribute in the cloning operation.

PR #4458, opened 2020-03-18
[ Broadcast ] Dynamic shape infer

jane-intel:


src/ngraph/op/broadcast.cpp
```diff
@@ -21,6 +21,7 @@
 #include "ngraph/partial_shape.hpp"
 
 #include <numeric>
+#include <ops.hpp>
```
diyessi: `#include <system files>` separated by a line and then `#include "ngraph/ngraph_files.hpp"`

src/ngraph/op/broadcast.cpp
```diff
@@ -144,6 +145,31 @@ void op::v1::Broadcast::validate_and_infer_types()
         result_shape = static_pointer_cast<op::Constant>(input_value(1).get_node_shared_ptr())
                            ->get_shape_val();
     }
+    else if (dynamic_pointer_cast<op::Concat>(input_value(1).get_node_shared_ptr()))
```
diyessi: Specify the op version by op version or opset. `op::Concat` could change over time.
Use `else if (auto concat = as_type_ptr<opset1::Concat>(...)`

src/ngraph/op/broadcast.cpp
```diff
@@ -144,6 +145,31 @@ void op::v1::Broadcast::validate_and_infer_types()
...
+        if (concat->get_output_partial_shape(0).is_static() && concat->get_shape().size() == 1 &&
+            concat_inputs.size() == shape_size(concat->get_shape()))
+        {
+            auto output_partial_shape = vector<Dimension>{};
+            for (const auto& concat_input : concat_inputs)
+            {
+                auto source_node_ptr = concat_input.get_source_output().get_node_shared_ptr();
+                if (source_node_ptr->is_constant())
+                {
+                    auto source_const_ptr = dynamic_pointer_cast<op::Constant>(source_node_ptr);
```
diyessi: `if (auto source_const_ptr = as_type_ptr<Constant>(source_node_ptr)` does all three lines at once. Also, use a specific op version of `Constant`.

src/ngraph/op/broadcast.cpp
Ewa21: Done

src/ngraph/op/broadcast.cpp
Ewa21: Done

src/ngraph/op/broadcast.cpp
Ewa21: Done

src/ngraph/op/broadcast.cpp
diyessi: As you suggested in Slack, let's use `op::v0::Concat` here.

src/ngraph/op/broadcast.cpp
```diff
@@ -141,9 +142,33 @@ void op::v1::Broadcast::validate_and_infer_types()
     PartialShape result_shape{PartialShape::dynamic()};
     if (input_value(1).get_node_shared_ptr()->is_constant())
```
diyessi: `if (auto constant_val = as_type_ptr<op::v0::Constant>(input_value(1)...`

src/ngraph/op/broadcast.cpp
```diff
@@ -141,9 +142,33 @@ void op::v1::Broadcast::validate_and_infer_types()
...
     {
-        result_shape = static_pointer_cast<op::Constant>(input_value(1).get_node_shared_ptr())
+        result_shape = static_pointer_cast<op::v0::Constant>(input_value(1).get_node_shared_ptr())
                            ->get_shape_val();
     }
+    else if (dynamic_pointer_cast<op::v0::Concat>(input_value(1).get_node_shared_ptr()))
```
diyessi: This test will never be true since you caught all the constants in the previous test.

src/ngraph/op/broadcast.cpp
```diff
@@ -166,10 +191,10 @@ void op::v1::Broadcast::validate_and_infer_types()
                 input_value(2).get_node_shared_ptr()->is_constant())
             {
                 auto target_shape =
-                    static_pointer_cast<op::Constant>(input_value(1).get_node_shared_ptr())
+                    static_pointer_cast<op::v0::Constant>(input_value(1).get_node_shared_ptr())
```
diyessi: `static_pointer_cast` means you are absolutely sure this is a Constant; if it isn't, it will give you garbage.
It looks like there are a lot of tests for the same value being a constant. Why not do one `as_type_ptr` at the beginning and just use that everywhere if it's not null?

src/ngraph/op/broadcast.cpp
Ewa21: Done

src/ngraph/op/broadcast.cpp
Ewa21: Done

src/ngraph/op/broadcast.cpp
Ewa21: Done

src/ngraph/op/broadcast.cpp
```diff
@@ -65,7 +66,7 @@ std::pair<bool, AxisSet> op::v1::Broadcast::get_broadcast_axes() const
             auto target_shape = get_input_shape(1);
             NGRAPH_CHECK(target_shape.size() == 1);
             auto axes_mapping_val =
-                static_pointer_cast<op::Constant>(input_value(2).get_node_shared_ptr())
+                static_pointer_cast<op::v0::Constant>(input_value(2).get_node_shared_ptr())
```
diyessi, line 69: If there were an `op::v3::Constant`, `is_constant()` would return `true` and `axes_mapping_val` would be garbage (`static_pointer_cast` means you know you're correct). Use `as_type_ptr<op::v0::Constant>(get_input_node_shared_ptr(2))` and then that won't happen. You should be able to replace the `is_constant()` with a little reorganizing.

Ewa21: Done

src/ngraph/op/broadcast.cpp
```diff
@@ -19,6 +19,7 @@
 #include "ngraph/op/constant.hpp"
 #include "ngraph/op/sum.hpp"
 #include "ngraph/partial_shape.hpp"
+#include "ops.hpp"
```
mbencer, line 22: only `Concat` header can be included instead of all ops

Ewa21: Done

raramer01:
TESTNOW_DLDT_E2E

PR #4459, opened 2020-03-18
[ NonMaxSuppression ] Fixed shape inference

jane-intel:


tomdol:
Should we make the output element type configurable via a setter?

PR #4460, opened 2020-03-18
[ ConvolutionBackPropData ] Dynamic shape inference

jane-intel:


src/ngraph/validation_util.cpp
```diff
@@ -869,14 +869,15 @@ int64_t ngraph::normalize_axis(const std::string& node_description,
...
 
-void ngraph::opset1::infer_conv_backprop_output_spatial_shape(const Shape& input_data_shape,
-                                                              const Shape& filters_shape,
-                                                              const Strides& strides,
-                                                              const Strides& dilations,
-                                                              const CoordinateDiff& pads_begin,
-                                                              const CoordinateDiff& pads_end,
-                                                              const CoordinateDiff& output_padding,
-                                                              Shape& output_spatial_shape)
+void ngraph::opset1::infer_conv_backprop_output_spatial_shape(
```
diyessi: I think it will be better to tie this to the op version.

src/ngraph/op/convolution.cpp
```diff
@@ -315,7 +313,7 @@ void op::v1::ConvolutionBackpropData::validate_and_infer_types()
             m_dilations = conv_default_strides(this, data_pshape, filters_pshape);
         }
 
-        const size_t num_spatial_dims = data_shape.size() - 2;
+        const size_t num_spatial_dims = data_pshape.rank().get_length() - 2;
```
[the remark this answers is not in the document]
Ewa21: Done

src/ngraph/op/convolution.cpp
```diff
@@ -382,15 +411,13 @@ void op::v1::ConvolutionBackpropData::validate_and_infer_types()
...
-            const Shape& data_shape = data_pshape.to_shape();
+            vector<Dimension> data_shape{data_pshape}, filters_shape{filters_pshape}, output_shape;
 
-            Shape output_shape;
-            opset1::infer_conv_backprop_output_spatial_shape(
-                Shape{std::next(data_shape.begin(), 2), std::end(data_shape)},
-                Shape{std::next(filters_shape.begin(), 2), std::end(filters_shape)},
+            infer_conv_backprop_output_spatial_shape(
+                vector<Dimension>{std::next(data_shape.begin(), 2), std::end(data_shape)},
+                vector<Dimension>{std::next(filters_shape.begin(), 2), std::end(filters_shape)},
```
tomdol: @jane-intel Should these 2 vectors contain a range `< begin() + 2; end() )` or two values `begin() + 2` and `end()`? I think the way they are constructed up here will use a constructor that takes an initializer_list and thus both arguments to `infer_conv_backprop_output_spatial_shape` will contain 2 elements.

src/ngraph/validation_util.cpp
Ewa21: Done

src/ngraph/op/convolution.cpp
```diff
@@ -270,6 +270,37 @@ void op::v1::ConvolutionBackpropData::set_output_shape(const Shape& shape)
...
+    const vector<Dimension>& filters_shape,
+    const Strides& strides,
+    const Strides& dilations,
+    const CoordinateDiff& pads_begin,
+    const CoordinateDiff& pads_end,
+    const CoordinateDiff& output_padding,
+    vector<Dimension>& output_spatial_shape)
+{
+    size_t num_spatial_dims = input_data_shape.size();
+    NGRAPH_CHECK(filters_shape.size() == num_spatial_dims && strides.size() == num_spatial_dims &&
```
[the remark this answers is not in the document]
Ewa21: dne

PR #4462, opened 2020-03-18
[ VariadicSplit ] Dynamic shape inference

jane-intel:


raramer01:
TESTNOW_DLDT_E2E

raramer01:
TESTNOW_DLDT_E2E

src/ngraph/op/variadic_split.cpp
```diff
@@ -99,25 +99,31 @@ void ngraph::op::v1::VariadicSplit::validate_and_infer_types()
                     sum_of_splits += split_lengths[i];
                 }
             }
+            auto data_shape_dims = vector<Dimension>{data.get_partial_shape()};
+            auto dimension_at_axis = data_shape_dims.at(axis);
 
-            if (negative_one > 0)
+            if (negative_one > 0 && dimension_at_axis.is_static())
```
mitruska: I just wonder, what if we get `-1` as a first element, so at the `[0]` index of the `split_lengths` vector
```suggestion
            if (negative_one >= 0 && dimension_at_axis.is_static())
```

postrational: Well spotted!

postrational: I added a test to cover this.

PR #4464, opened 2020-03-19
Update includes in reference runtime headers

pavel-rodionov:
Include additional headers to reference runtime files to make them self-sufficient


raramer01:
TESTNOW_DLDT_E2E

raramer01:
TESTNOW_DLDT_E2E

rkimballn1:
It was failing style check so I fixed it. Hope you don't mind.

rkimballn1:
@diyessi I agree that a test would be great but I think that is beyond the scope of this PR. Perhaps a separate PR?

pavel-rodionov:
Can we merge this MR as is and then in few days I will create another MR with proposed test?

diyessi:
Certainly, once @GlebKazantaev has reviewed.

PR #4465, opened 2020-03-20
Embed python interpreter in unit-test

silee2:


test/backend/numpy.in.cpp
```diff
@@ -0,0 +1,116 @@
...
+    auto locals = py::dict("a"_a = a, "b"_a = b);
+    py::exec(R"(
+        c = a + b
+    )",
+             py::globals(),
+             locals);
+
+    auto sum = locals["c"].cast<int>();
+
+    EXPECT_TRUE(sum == a + b);
```
gaurides, line 68: Should "a+b" go through the backend? if not, then do we need the "backend" in line 55?

test/CMakeLists.txt
```diff
@@ -14,6 +14,36 @@
...
 # ******************************************************************************
 
+if(NGRAPH_UNIT_TEST_NUMPY_ENABLE)
+    include(ExternalProject)
+
+    ExternalProject_Add(
+        ext_pybind11
+        PREFIX ut_pybind11
+        GIT_REPOSITORY "https://github.com/pybind/pybind11.git"
+        GIT_TAG "v2.5.0"
```
gaurides, line 24: Is there a way to make sure the TAG here and in python/CMakeLists.txt is same?

rkimballn1: Yes there is, defined it in the top-level cmake file and then reference it in the two places

test/backend/numpy.in.cpp
silee2: Will remove line 55.

silee2: Removed.

test/CMakeLists.txt
```diff
@@ -14,6 +14,36 @@
 # limitations under the License.
 # ******************************************************************************
 
+if(NGRAPH_UNIT_TEST_NUMPY_ENABLE)
```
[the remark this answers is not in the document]
silee2: Created cmake/external_pybind11.cmake

PR #4466, opened 2020-03-21
Interval-based dimension

diyessi:
Track valid min/max intervals in `Dimension` and `PartialShape`


diyessi:
> @diyessi I want to review this and trying to go through the min set of files to give me a high level picture of dimensions with respect to shape. What do you suggest I read? Files and Adam Procter's tome?

We can simplify the tome now. A PartialShape is now the set of possible shapes; if static, the set has exactly one element. The set of possible shapes is always the enclosing box of possible shapes.

src/ngraph/partial_shape.cpp
```diff
@@ -36,6 +36,65 @@ bool ngraph::PartialShape::is_static() const
                                            [](const Dimension& d) { return d.is_static(); });
 }
 
+bool ngraph::PartialShape::operator==(const PartialShape& partial_shape) const
+{
+    if (rank() != partial_shape.rank())
+    {
+        return false;
+    }
+    if (rank().is_dynamic())
```
silee2, line 45: Is this "If" condition correct? I think you are trying to say "If one of the two are dynamic".
The "If" is to only allow "the two are both static with same rank or both dynamic" to fall to this condition.
The "If" below assumes "both are static rank".

diyessi: The if at 41 ensures both ranks are equal, so the one in 45 implies both are dynamic.

src/ngraph/interval.hpp
```diff
@@ -0,0 +1,120 @@
...
+    ///
+    /// An interval is the set of integers from m_min_val through m_max_val.
+    /// The value s_max acts like infinity. The
+    /// addition, subtraction, or multiplication of intervals is the smallest interval
+    /// containing the sums, differences, or products of elements of the two intervals. An empty
+    /// interval is canonicalized to [s_max, s_max].
+    class NGRAPH_API Interval
+    {
+    public:
+        using value_type = std::int64_t;
```
rkimballn1, line 39: This is the second definition of value_type, the first being in dimension. This should be defined in a single location so they don't get out of sync.

src/ngraph/op/experimental/range.cpp
```diff
@@ -24,8 +24,12 @@ using namespace ngraph;
 
 constexpr NodeTypeInfo op::Range::type_info;
 
-op::Range::Range(const Output<Node>& start, const Output<Node>& stop, const Output<Node>& step)
+op::Range::Range(const Output<Node>& start,
+                 const Output<Node>& stop,
+                 const Output<Node>& step,
+                 int64_t max_output_length)
```
rkimballn1, line 30: sizes should be size_t

src/ngraph/op/experimental/range.hpp
```diff
@@ -44,13 +44,25 @@ namespace ngraph
                 ///             element type, and same element type as `start` and `stop`.
                 Range(const Output<Node>& start,
                       const Output<Node>& stop,
-                      const Output<Node>& step);
+                      const Output<Node>& step,
+                      int64_t max_output_length = -1);
```
rkimballn1, line 48: Why not define a constant called something like `UNBOUNDED` = -1 and then use it here and on line 65 below? That would keep them in sync and remove the need for documenting what -1 means.

src/ngraph/dimension.cpp
```diff
@@ -153,9 +157,10 @@ Dimension::operator size_t() const
     {
         throw std::invalid_argument("Cannot convert dynamic dimension to size_t");
     }
-    if (m_dimension < 0)
+    auto result = m_dimension.get_min_val();
+    if (result > std::numeric_limits<size_t>::max())
```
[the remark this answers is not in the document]
diyessi: This has been a deprecated method for a few months (see header).
`result < 0` only if `is_dynamic()` is true, so the negative values aren't an issue.

src/ngraph/interval.hpp
diyessi: No need for them to be in sync.

test/intervals.cpp
```diff
@@ -0,0 +1,165 @@
...
+    Interval::value_type min_minus = numeric_limits<Interval::value_type>::max();
+    Interval::value_type max_minus = numeric_limits<Interval::value_type>::min();
+    auto a_times_b = a * b;
+    Interval a_times = a;
+    a_times *= b;
+    EXPECT_TRUE(a_times_b == a_times);
+    Interval::value_type min_times = numeric_limits<Interval::value_type>::max();
+    Interval::value_type max_times = numeric_limits<Interval::value_type>::min();
+
+    for (auto a_i = a.get_min_val(); a_i <= a.get_max_val(); ++a_i)
```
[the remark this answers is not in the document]
diyessi: It's checking that the bounds of all the sums, products, etc. are exactly the bounds of the interval and that each is in their interval. Added comment. Same for sets test below.

src/ngraph/interval.hpp
```diff
@@ -0,0 +1,120 @@
...
+        using size_type = std::uint64_t;
```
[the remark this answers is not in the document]
rkimballn1: why do you need 64 bits? If you are on a 32 bit machine you only have 4G max RAM and can't have sizes larger than fit in a 32-bit number.  Things like vector use size_t and you should too.

PR #4481, opened 2020-03-26
Merge convert ops sequence 

swapna-intel:
JIRA: https://jira.devtools.intel.com/browse/CVS-28831
Plan:

See graph in test file


swapna-intel:
I dont have access to JIRA yet. My questions:
1. Should this be an algebraic_simplification pass or a new one by itself?
2. what all conversions are to be targeted?
3. Does shapes have any impact?

ashokei:
@swapna-intel one pattern i notice is char -> convert(uint8) -> convert(f32). I think the middle convert can be removed?

PR #4486, opened 2020-03-26
Add value_type to Dimension, deprecate int64_t cast

diyessi:


src/ngraph/validation_util.cpp
```diff
@@ -866,7 +866,7 @@ int64_t ngraph::normalize_axis(const std::string& node_description,
         axis = axis + tensor_rank;
     }
 
-    return static_cast<int64_t>(axis);
+    return int64_t(axis);
```
mbencer: cast not needed

PR #4489, opened 2020-03-27
Cyphers/topkconst

diyessi:


silee2:
The evaluator caches node evaluation result. We would have to state that evaluator needs to be reconstructed after graph modification.

PR #4494, opened 2020-03-30
Use Squeeze instead of Reshape in interpret_as_scalar

tomdol:


src/ngraph/frontend/onnx_import/utils/reshape.cpp
```diff
@@ -110,7 +110,10 @@ namespace ngraph
                         node->get_element_type(), ngraph::Shape{}, value);
                 }
 
-                return ngraph::builder::opset1::reshape(node, Shape{});
+                const auto axis =
+                    default_opset::Constant::create(element::i64, ngraph::Shape{}, {0});
+
+                return std::make_shared<ngraph::opset1::Squeeze>(node, axis);
```
arogowie-intel: This doesn't solve the real problem. We should fix `builder::opset1::reshape` to handle case when we want to reshape into scalar value.

tomdol: It does solve the problem by replacing a regular op with a fused op which will not be decomposed to Reshape in IE. This way the IE will not receive a constant with a scalar shape that it doesn't support at the moment.
I've initially fixed the builder but it still produced a constant with a scalar and an empty shape, so the problem was still reproducible.

arogowie-intel: Decided to finish in separate PR.

PR #4495, opened 2020-03-30
Eliminate mismatch between opset1/opset2 and our real intentions

slyalin:
Modify opset2 based on our real intentions in forming opset1. Opset1 has some mismatches with what we have in OpenVINO spec. As we cannot modify opset1 as it had been already released, so we apply fixed in opset2.


slyalin:
We have removed @diyessi from the review as it stuff is not architectural and affects mostly OpenVINO stack. The PR was created yesterday, there hasn't been any feedback from his side since then. We would like to merge it quickly and avoid waiting another day.

PR #4497, opened 2020-03-31
update concat_elimination pass for openvino use

ashokei:


src/ngraph/pass/concat_fusion.cpp
```diff
@@ -99,22 +99,31 @@ void pass::ConcatElimination::construct_concat_elimination()
...
         return false;
     };
 
     auto m = std::make_shared<pattern::Matcher>(concat_label, "ConcatElimination");
-    this->add_matcher(m, callback, PassProperty::REQUIRE_STATIC_SHAPE);
+    this->add_matcher(m, callback);
+}
+
+bool ngraph::pass::ConcatElimination::run_on_function(std::shared_ptr<Function> function)
+{
```
diyessi: What problem is this working around?

ashokei: they have CI / functional tests with the graph having just concat op node (with single input in some cases). This pass was eliminates even these, resulting graph "parameter->result" (no op node). This fix avoids the pass if the graph is just one op node.

src/ngraph/pass/concat_fusion.cpp
```diff
@@ -99,22 +99,31 @@ void pass::ConcatElimination::construct_concat_elimination()
...
+        auto concat = as_type_ptr<op::Concat>(m.get_match_root());
+        // eliminate concat with 1 input
+        if (concat && (concat->get_input_size() == 1))
         {
-            NGRAPH_DEBUG << " eliminated " << m.get_match_root() << "\n";
-            replace_node(m.get_match_root(), op);
-
+            NGRAPH_DEBUG << " eliminated " << concat << "\n";
+            concat_input_op->set_friendly_name(concat->get_friendly_name());
+            replace_node(concat, concat_input_op);
```
GlebKazantaev, line 110: Code snippet below describes when we can remove operation and when we need to preserve the concat name.
```cpp
bool has_result_output = false;
for (const auto & output : concat->output(0).get_target_inputs()) {
    if (dynamic_cast<ngraph::op::Result*>(output.get_node())) {
        has_result_output = true;
    }
}

size_t parent_consumers_count = 0;
for (const auto &output : concat_input_op->outputs()) {
    parent_consumers_count += output.get_target_inputs().size();
}
 
if (!has_result_output || parent_consumers_count == 1) {
    if (has_result_output && !std::dynamic_pointer_cast<ngraph::op::Parameter>(concat_input_op)) {
        concat_input_op->set_friendly_name(concat->get_friendly_name());
    }
    // And only then you can replace concat op

}
```

diyessi: Why does a Result output cause a problem for you? They only exist to prevent splicing problems.
Why do multiple outputs cause a problem for you? replace_output and replace_node know what to do for those cases. If splicing in these cases is a problem, a lot of other things aren't going to work properly.

ashokei: it seems this is for preserving their friendly_name. 

If parent_consumers_count > 1 and has_result_output, why cant we remove the null/identity op?

diyessi: @GlebKazantaev do you really want to disable an optimization because of the friendly name?

ashokei:
since openvino has special handling for removing ops / friendly_name handling, atleast for this concat null op case i created these changes in below PR
https://github.com/NervanaSystems/ngraph/pull/4536

src/ngraph/pass/concat_fusion.cpp
GlebKazantaev: @diyessi we can't remove nodes that has Result as an output because if we remove it, Result will be reconnected to other node output and when we finally convert nGraph to CNNNetwork we remove Result ops (current limitation) and CNNNetwork start produce wrong output names.

In other words we can't remove because we will loose original output names.

@ashokei I'm going to make this check common for all transformations that removes ops (useless ops) and when we finish work with transition to nGraph we remove this constraints.

src/ngraph/pass/concat_fusion.cpp
```diff
@@ -99,22 +99,31 @@ void pass::ConcatElimination::construct_concat_elimination()
             << "concat_elimination: In callback for construct_concat_elimination against node = "
             << m.get_match_root()->get_name();
         auto pattern_map = m.get_pattern_map();
-        auto op = pattern_map[op_label];
+        auto concat_input_op = pattern_map[op_label];
```
diyessi, line 102: You want `pattern_value_map` here; `pattern_map` is for back-compatibility with GOE-based patterns, so 
`auto concat_input = m.get_pattern_value_map()[op_label];`
`pattern_value_map` will give you the correct output corresponding to the label. And then you'll want
`concat.output(0).replace(concat_input);`

PR #4503, opened 2020-03-31
Fixes for v1::ConvolutionBackpropData and v1:Convolution ops

itikhono:
Convolution:
1. Nullify pads_begin and pads_end attributes in case of auto_pad == "VALID".

ConvolutionBackpropData:
1. Support a case when output_shape is omitted, but auto_pad is specified. 
This case is not specified in the onnx spec, so [issue ](https://github.com/onnx/onnx/issues/2586) has been created.
Added implementation according to the onnxruntime.
Implemented tests on dldt side to cover this case.
2. Fix for auto_pad calculation according to the onnx spec formulas.
onnx spec and onnx runtime implementation:
> If (auto_pads != SAME_UPPER): pads[start_i] = total_padding[i]/2; pads[end_i] = total_padding[i] - (total_padding[i]/2)
> Else: pads[start_i] = total_padding[i] - (total_padding[i]/2); pads[end_i] = (total_padding[i]/2).

But there is inconsistency in ONNX specification: [issue](https://github.com/onnx/onnx/issues/1609)


itikhono:
> Has it been verified that no none is currently calling the constructor with arguments that will get errors with this PR? @ilyachur

These changes were verified by all dldt pre-commits. Also, it fixes inference for several winML models. I’ll run additional dldt tests in order to be more confident.
The specification was updated according to the changes.

itikhono:
dldt testing passed. PR is ready to merge.

PR #4508, opened 2020-03-31
add reshape_elimination_v1 pass

ashokei:
currently just handles identity removal


src/ngraph/pass/reshape_elimination_v1.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+void pass::ReshapeEliminationV1::construct_identity_reshape_pattern()
+{
+    auto op = make_shared<pattern::op::Label>(element::f32, Shape{1, 3});
+    auto rpattern = make_shared<pattern::op::Label>(element::i64, Shape{1});
+    auto reshape1 = make_shared<op::v1::Reshape>(op, rpattern, false);
+
+    auto callback = [op](pattern::Matcher& m) {
+        NGRAPH_DEBUG << "In callback for construct_identity_reshape_pattern against node = "
+                     << m.get_match_root()->get_name();
```
diyessi, line 42: Just use `m.get_match_root()`

src/ngraph/pass/reshape_elimination_v1.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+        if (gop->get_output_partial_shape(0).is_dynamic() ||
+            r1->get_output_partial_shape(0).is_dynamic() ||
+            gop->get_output_shape(0) != r1->get_output_shape(0))
+        {
+            NGRAPH_DEBUG << "Not a no-op; Shapes are different!";
+            return false;
+        }
+
+        gop->set_friendly_name(r1->get_friendly_name());
```
diyessi, line 56: Not sure if this is right; whose friendly name is right here?

ashokei: this is the approach openvino uses in their passes : if they replace a node, they have to use the friendly name of node being replaced , set to the new node that replaces it.

diyessi: Okay. Would be fixed by provenance II, which would keep both names.

ashokei:
looks like openvino has special friendly_name handling needs, will revisit this PR with more reshape_v1 fusions in future. Identity removal is added to below
https://github.com/NervanaSystems/ngraph/pull/4536

PR #4510, opened 2020-04-01
Add doxygen description for Validate pass

yimeisun123:


yimeisun123:
validate

rkimballn1:
@gaurides The easiest way for both you and Yimei to suggest a change such as rewording is to use the "insert suggestion" feature in github
image

yimeisun123:
> Add that it also re-runs shape inference and is supposed to be safe to run multiple times.
> Connected with that, can you replace the `validate_and_infer_types` doc in nodes.hpp with "Verifies that attributes and inputs are consistent and computes output shapes and element types. Must be implemented so that it can be run any number of times."

@diyessi - is the following wording update ok in the validation pass?
The Validate pass performs sanity check on the input shape and data type ```as well as re-runs shape inference``` for all computation nodes in a given computation graph.

yimeisun123:
Here is what the 2nd commit looks like.
image

yimeisun123:
TESTNOW

raramer01:
TESTNOW_DLDT_E2E

src/ngraph/pass/validate.hpp
```diff
@@ -22,6 +22,18 @@ namespace ngraph
 {
     namespace pass
     {
+        /// \brief The Validate pass performs santify check on attributes and inputs, and
```
gaurides, line 25:
```suggestion
        /// \brief The Validate pass performs sanity check on attributes and inputs, and
```

src/ngraph/pass/validate.hpp
```diff
@@ -22,6 +22,18 @@ namespace ngraph
...
+        /// computation graph.
+        ///
+        /// \details The verification and inference is done via invoking each node's specific
+        /// implementation of \link ngraph::Node::validate_and_infer_types() \endlink function.
+        ///
+        /// By default, the \ref ngraph::pass::Manager runs this pass after executing every
+        /// optimization pass. This is to ensure that any update to the graph by an optimization
+        /// pass will not break the shape and data type requirement on a compuatation node.
+        /// This default validation run can be changed via calling the
+        /// \link ngraph::pass::Manager::set_per_pass_validation(bool) \endlink function.
```
rkimballn1, line 36:
```suggestion
        /// \brief The Validate pass performs sanity checks on attributes and inputs, and
        /// computes output shapes and element types for all computation nodes in a given
        /// computation graph.
        ///
        /// \details The verification and inference is done via invoking each node's specific
        /// implementation of \link ngraph::Node::validate_and_infer_types() \endlink function.
        ///
        /// By default, the \ref ngraph::pass::Manager runs this pass after executing every
        /// optimization pass. This is to ensure that any update to the graph by an optimization
        /// pass does not break the shape and data type requirement on a computation node.
        /// This default validation run can be changed via calling the
        /// \link ngraph::pass::Manager::set_per_pass_validation(bool) \endlink function.
```

PR #4470, opened 2020-03-23
[ONNX] Add Slice-10 to ONNX importer

mbencer:

PR #4513, opened 2020-04-02
[ONNX] Add Slice-10

mbencer:


src/ngraph/frontend/onnx_import/op/slice.cpp
```diff
@@ -36,6 +38,99 @@ namespace ngraph
...
+                    {
+                        steps = inputs.at(4);
+                    }
+
+                    const auto axes_const = as_type_ptr<default_opset::Constant>(axes);
+                    const auto axes_vec = axes_const->cast_vector<int64_t>();
+
+                    const auto data_rank = data->get_output_partial_shape(0).rank();
+                    NGRAPH_CHECK(data_rank.is_static(), "Data rank of input must be static");
+                    NGRAPH_CHECK(data_rank.get_length() == axes_vec.size(),
```
GlebKazantaev, line 94: Why axes size should be equal to rank size? axes input can have any positive size <= data_rank.get_length().

mbencer: @GlebKazantaev  In PR to master branch https://github.com/NervanaSystems/ngraph/pull/4470 such limitations do not exist but it requires some changes to `StridedSlice` operator which are not allowed on release branch:
1.  default `strides` length is calculated in incorrect way
2. cases such as: we have 4D input data, 2-elements `begins` and `ends` and we want to apply it to 2, 3 axes are not supported
In order to provide version which support some expected model such limitations are acceptable but generally you are right

diyessi:
@mbencer should this be switched to master?

mbencer:
> @mbencer should this be switched to master?

@diyessi Version to master is here (the only difference is that it doesn't modify `StridedSlice:v1`): https://github.com/NervanaSystems/ngraph/pull/4470. 
`Slice-10` enables a model which is expected for `release 2020.2` by some customers but final decision has not been made.

PR #4516, opened 2020-04-02
oneDNN: Update to 1.4

silee2:


cmake/mkldnn_v1.patch
```diff
@@ -51,10 +51,10 @@ index 0bc6ea57a..4fa0a22d1 100644
 +    endif()
 +endmacro()
 +
- if(WIN32 AND ${CMAKE_CXX_COMPILER_ID} STREQUAL MSVC)
-     add_definitions(/Qpar)
-     add_definitions(/openmp)
-@@ -79,6 +115,7 @@ if (DNNL_CPU_THREADING_RUNTIME MATCHES "OMP")
+ macro(set_openmp_values_for_old_cmake)
```
gaurides, line 54: where is `set_openmp_values_for_old_cmake` used? and don't we need to check for WIN32

silee2: @gaurides That part is a diff for the patch.  Not the diff for the file itself. `set_openmp_values_for_old_cmake` is not our code. It would be easier to look at the the file `cmake/mkldnn_v1.patch` instead of the diff. I haven't changed anything in the patch compared to the current patch. I understand it is a bit confusing.

gaurides: ah, ok. thanks for the reminder

PR #4517, opened 2020-04-03
add pass to remove useless transpose op 

wenzhe-nrv:
~the pattern will match a transpose op. If it's input's shape's axis 0 is 1 (e.g. shape {1, ?} or shape{1, 4}), and the permutation is {1, 0}, which is to swap the axis 0 and 1, then this transpose will be removed.~
Two patterns to identify uesless transpose op:
1. transpose op that takes an input of shape{1, ?} and outputs it to shape{?,1}. Then reshape.
2. transpose op that takes an input of shape{1, ?} (or {?,?}), and outputs it to {?,1}, then squeeze axis 1. This transpose could be removed and change axis input of the squeeze op to 0.


pruthviIntel:
Do we need a separate graph pass for this, just wondering https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/pass/algebraic_simplification.hpp, registering the pattern here should do?

diyessi:
It should be in algebraic_simplification.

wenzhe-nrv:
sounds good. Will move it there.

test/algebraic_simplification.cpp
```diff
@@ -596,3 +597,52 @@ TEST(algebraic_simplification, pass_property)
...
+}
+
+TEST(algebraic_simplification, transpose_partial_shape)
+{
+    // add a simple graph with transpose
+    auto ps = PartialShape{1, Dimension::dynamic()};
+    auto param = make_shared<op::Parameter>(element::boolean, ps);
+    auto constant_perm = make_shared<op::Constant>(element::i64, Shape{2}, vector<int64_t>{1, 0});
+    auto transpose = make_shared<op::Transpose>(param, constant_perm);
+    auto f = make_shared<Function>(transpose, ParameterVector{param});
```
wenzhe-nrv, line 624: @diyessi I added the change in algebraic_simplification pass. And this test with partial shape fails now, because `f->is_dynamic() == true` and `algebraic_simplification` pass only works on static functions with `set_property(PassProperty::REQUIRE_STATIC_SHAPE, true);`. 
The test could pass with setting `algebraic_simplification` pass `PassProperty::CHANGE_DYNAMIC_STATE`. Is this ok or we want to keep it for static shape only?

diyessi: Probably nothing in the pass tests that shapes are static before doing things to the ops.

ashokei:
please align with https://github.com/NervanaSystems/ngraph/pull/4566

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +546,77 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+    auto cnst_perm_lable =
+        make_shared<pattern::op::Label>(element::i64, Shape{2}, pattern::has_class<op::Constant>());
+    auto t = make_shared<op::Transpose>(input, cnst_perm_lable);
+
+    if (squeeze)
+    {
+        auto cnst_axis_lable = make_shared<pattern::op::Label>(
+            element::i64, Shape{1}, pattern::has_class<op::Constant>());
+        auto s = make_shared<op::Squeeze>(t, cnst_axis_lable);
+        auto t_s_matcher = make_shared<pattern::Matcher>(s, "transpose_squeeze_pattern");
```
diyessi, line 568: You don't need to use the pattern matcher here -- just check if the inputs correct.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +546,77 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+        {
+            NGRAPH_DEBUG << n << "Doesn't match pattern" << t_s_matcher->get_name();
+            return false;
+        }
+        NGRAPH_DEBUG << "Match pattern " << t_s_matcher->get_name();
+        auto t_s_pattern_value_map = t_s_matcher->get_pattern_map();
+        auto cnst_perm_op = as_type_ptr<op::Constant>(t_s_pattern_value_map[cnst_perm_lable]);
+        auto cnst_axis_op = as_type_ptr<op::Constant>(t_s_pattern_value_map[cnst_axis_lable]);
+
+        auto replace_gather_op = make_shared<op::v0::Gather>(cnst_perm_op, cnst_axis_op);
```
diyessi, line 579: Use `op::v1::Gather`.

jane-intel:
Having transformation replacing `Transpose` of you interest with `Reshape` operation (https://github.com/NervanaSystems/ngraph/pull/4566) and `Reshape`-`Reshape` / `Reshape`-`Squeeze` fusion to single `Reshape` op (https://github.com/NervanaSystems/ngraph/pull/4602) the task will be completed. 

@wenzhe-nrv, please share your experience and test cases with @ashokei and @nbpatel. 

In case there are no objections we may close current PR.

wenzhe-nrv:
@jane-intel agree. I'm closing it.

PR #4518, opened 2020-04-03
ScatterAdd and ScatterUpdate opset3 ops

arogowie-intel:
These ops behavior is similar to TF ones: [ScatterUpdate](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/scatter-update) and [ScatterAdd](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/scatter-add) with addition that these are able to do updates/add on any axis.


src/ngraph/op/util/scatter_base.cpp
```diff
@@ -0,0 +1,149 @@
...
+                          compatible,
+                          "Updates shape must have appropriate dimensions equal to indices and "
+                          "data dimensions. Updates shape:",
+                          updates_shape.to_shape(),
+                          ", data shape: ",
+                          data_shape.to_shape(),
+                          ", indices_shape: ",
+                          indices_shape.to_shape(),
+                          ", axis: ",
+                          axis,
```
mbencer, line 131: - `to_shape` can thrown additonal exception during `NODE_VALIDATION_CHECK` assert
- Value of `axis` can be not defined

arogowie-intel: I haven't updated this message after changing above `if` condition. This message is evaluated only when `compatible` is false, thus (earlier shapes where known) now only ranks are static for sure. However the axis value should also be known at this point.

src/ngraph/op/scatter_add.cpp
```diff
@@ -78,3 +78,26 @@ void op::ScatterAdd::validate_and_infer_types()
 
     set_output_type(0, inputs_et, inputs_shape);
 }
+
+//------------------------------------------------------------------------------
+//
+//          Opset version 3
```
tomdol, line 84: In fact it's not a different opset here but a different op version.

src/ngraph/opsets/opset.cpp
```diff
@@ -85,3 +85,22 @@ const ngraph::OpSet& ngraph::get_opset2()
     }
     return opset;
 }
+
+const ngraph::OpSet& ngraph::get_opset3()
+{
+    static std::mutex init_mutex;
+    static bool opset_is_initialized = false;
```
tomdol, line 92: I'm not sure that this extra bool is needed. Since C++11 the Meyer's singleton is guaranteed to be safe with just the object being static.

src/ngraph/serializer.cpp
```diff
@@ -2669,6 +2669,11 @@ shared_ptr<Node> JSONDeserializer::deserialize_node(json node_js)
             node = make_shared<op::ScatterAdd>(args[0], args[1], args[2]);
```
diyessi, line 2669: Don't need any of the serializer changes now.

src/ngraph/op/scatter_add.cpp
diyessi: How about "Introduced in Opset 3"?

src/ngraph/op/scatter_update.hpp
```diff
@@ -0,0 +1,57 @@
...
+                ScatterUpdate(const Output<Node>& data,
+                              const Output<Node>& indices,
+                              const Output<Node>& updates,
+                              const Output<Node>& axis);
+
+                virtual std::shared_ptr<Node>
+                    clone_with_new_inputs(const OutputVector& inputs) const override;
+            };
+        }
+        // TODO: Is it correct?
```
diyessi, line 54: Leave this out. These are put in for back-compatibility with code that imports from `ngraph::op`; there are no current users of this op so no back-compatibility.

src/ngraph/op/util/scatter_base.cpp
```diff
@@ -0,0 +1,149 @@
...
+//*****************************************************************************
+
+#include "ngraph/op/util/scatter_base.hpp"
+#include "ngraph/shape.hpp"
+#include "ngraph/validation_util.hpp"
+
+using namespace std;
+using namespace ngraph;
+
+static constexpr int DATA = 0;
```
diyessi, line 24: Why don't you make these static members of the class?

src/ngraph/opsets/opset.cpp
arogowie-intel: Me neither. I've followed convention from: https://github.com/NervanaSystems/ngraph/pull/4483

src/ngraph/opsets/opset3_tbl.hpp
```diff
@@ -22,4 +22,6 @@
 #include "opset2_tbl.hpp"
 NGRAPH_OP(NonZero, ngraph::op::v3)
 NGRAPH_OP(ROIAlign, ngraph::op::v3)
+NGRAPH_OP(ScatterAdd, ngraph::op::v3)
```
jane-intel, line 25: Please remove this operation from opset3 table. There is no such operation in opset3 documentation yet.
```suggestion
```

PR #4527, opened 2020-04-07
Fused ops need to defuse

diyessi:
Move two non-fused ops out of fused (should move all eventually)


silee2:
@diyessi What do you mean by `non-fused` op? I still see `decompose_op` in the updated code along with including `fused_op.hpp`

diyessi:
> @diyessi What do you mean by `non-fused` op? I still see `decompose_op` in the updated code along with including `fused_op.hpp`

The v0 versions are fused ops; the v1 versions do not have defusers; rather than just use op, they used FusedOp and overrode the virtual says it is fused.

PR #4529, opened 2020-04-08
Equality of SlicePlans

jane-intel:


src/ngraph/slice_plan.cpp
```diff
@@ -226,3 +226,16 @@ SlicePlan ngraph::make_slice_plan(const Shape& input_shape,
     }
     return p;
 }
+
```
diyessi, line 229: You should also add a `!=`.

PR #4532, opened 2020-04-08
Expand onnx importer API to accept model proto.

tsocha:
This PR should resolve a problem with model proto de-serialization issue when binary package is used.
Custom namespace have to be used for onnx library.


tomdol:
This PR can be closed since it's not going to be included in 2020.2

PR #4534, opened 2020-04-08
[ONNX] Dynamic builder for Transpose

arogowie-intel:


arogowie-intel:
Removed Michał from reviewers, since he is on sick leave.

src/ngraph/builder/reshape.cpp
```diff
@@ -195,10 +195,25 @@ shared_ptr<Node> builder::opset1::reorder_axes(const Output<Node>& value, vector
...
-    vector<size_t> axes_order(value.get_shape().size());
-    iota(begin(axes_order), end(axes_order), 0);
-    reverse(begin(axes_order), end(axes_order));
-    return builder::opset1::reorder_axes(value, axes_order);
+    // This part is left to preserve backward compatibility and ensure passing ONNX tests.
+    if (value.get_partial_shape().is_static())
+    {
+        vector<size_t> axes_order(value.get_shape().size());
+        iota(begin(axes_order), end(axes_order), 0);
+        reverse(begin(axes_order), end(axes_order));
```
tomdol, line 203: Not sure here, but perhaps you can pass `rbegin` and `rend` to `iota` and avoid the need to call reverse.

PR #4535, opened 2020-04-08
[MLIR] Convert to MLIR integer type

adstraw:


src/contrib/mlir/core/ngraph_dialect/dialect.cpp
```diff
@@ -98,16 +97,15 @@ mlir::Type NGraphOpsDialect::parseEltType(mlir::DialectAsmParser& parser) const
...
         {
-        case 8:
-            return isSigned ? NGIntegerType::getInt8(context) : NGIntegerType::getUInt8(context);
-        case 16:
-            return isSigned ? NGIntegerType::getInt16(context) : NGIntegerType::getUInt16(context);
-        case 32:
-            return isSigned ? NGIntegerType::getInt32(context) : NGIntegerType::getUInt32(context);
-        case 64:
-            return isSigned ? NGIntegerType::getInt64(context) : NGIntegerType::getUInt64(context);
+        case 8: return NGIntegerType::get(8, signedness, context);
```
nmostafa, line 105: Remove the switch. Do `NGIntegerType::get(width, signedness, context)`

src/contrib/mlir/core/ngraph_dialect/ops.cpp
```diff
@@ -172,7 +172,8 @@ static mlir::LogicalResult verifyCmpOp(T op)
 
     // result of same shape as input and has bool type
     if (!resType.isCompatibleShape(opType0) ||
-        !resType.getElementType().cast<NGIntegerType>().isUInt8())
+        !(resType.getElementType().cast<NGIntegerType>().isUnsigned() &&
+          resType.getElementType().cast<NGIntegerType>().getWidth() == 8))
```
nmostafa, line 176: Perhaps better to have NGIntegerType as a wrapper around mlir Integer type. To keep helpers like isUInt8() and other checks instead of using literals.

cmake/external_mlir.cmake
```diff
@@ -19,7 +19,7 @@ include(ExternalProject)
 set(MLIR_LLVM_REPO_URL https://github.com/llvm/llvm-project.git)
 
 # Change these commit IDs to move to latest stable versions
-set(MLIR_LLVM_COMMIT_ID 376c6853)
+set(MLIR_LLVM_COMMIT_ID 35b685270b410f6a1351c2a527021f22330c25b9)
```
nmostafa, line 22: I would split upgrading MLIR and IntegerType change to two separate PRs.

adstraw: This was necessary to pick up the MLIR IntegerType change which added signedness.  This does not move MLIR pointer all the way to tip of master.  Only picks up the signedness change.

src/contrib/mlir/core/ngraph_dialect/dialect.cpp
adstraw: Good call.  Will do.

src/contrib/mlir/core/ngraph_dialect/ops.cpp
adstraw: Fixed by using combined signedness/type is_type getters.  See latest commit.  Resolving.

PR #4536, opened 2020-04-08
update nop_elimination pass for openvino use

ashokei:


src/ngraph/pass/nop_elimination.cpp
```diff
@@ -35,13 +39,37 @@ using namespace ngraph;
 
 #define TI(x) std::type_index(typeid(x))
 
+static bool remove_update_name(const std::shared_ptr<Node>& node,
+                               const std::shared_ptr<Node>& node_input)
+{
+    bool has_result_output = false;
+    for (auto& output : node->output(0).get_target_inputs())
+    {
+        if (dynamic_cast<op::Result*>(output.get_node()))
```
diyessi, line 48: Use `as_type<op::Result>(output.get_node())` since `dynamic_cast` has problems on some platforms.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -35,13 +39,37 @@ using namespace ngraph;
 
 #define TI(x) std::type_index(typeid(x))
 
+static bool remove_update_name(const std::shared_ptr<Node>& node,
```
diyessi, line 42: Since you're in this file anyways, could you change the map key. to `Node::type_info_t` and change `TI` to be `x::type_info` ? The table lookup then uses `node->get_type_info()` to look up the handler instead of `TI(node)`.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -33,76 +38,156 @@
...
-    auto convert = std::static_pointer_cast<op::Convert>(node);
-    if (convert->get_convert_element_type() == convert->get_argument(0)->get_element_type())
+    bool is_out_type_agnostic = false;
+    static const std::set<NodeTypeInfo> type_agnostic{TI(op::v3::NonZero)};
+    if (node->output(0).get_target_inputs().size() == 1)
     {
-        replace_node(node, node->get_argument(0));
-        return true;
+        auto& out = *node->output(0).get_target_inputs().begin();
+        is_out_type_agnostic = type_agnostic.count(out.get_node()->get_type_info()) == 1;
```
swapna-intel, line 102: for an expanding convert, why do we need to care if user node is type_agnostic? Is it because this is a nop_elimination pass? If we have convert(char) -> convert(int16) -> convert(int32)->{ any set of output ops}, can't we do away with the first 2 converts? For type_agnostic, any convert input node can be eliminated, if that's the only user of the convert?

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -33,76 +38,156 @@
...
     return false;
 }
 
 static bool eliminate_convert(const std::shared_ptr<Node>& node)
 {
-    auto convert = std::static_pointer_cast<op::Convert>(node);
-    if (convert->get_convert_element_type() == convert->get_argument(0)->get_element_type())
+    bool is_out_type_agnostic = false;
+    static const std::set<NodeTypeInfo> type_agnostic{TI(op::v3::NonZero)};
+    if (node->output(0).get_target_inputs().size() == 1)
```
swapna-intel, line 99: we could do this for any number of inputs, if they are all the same op? I saw an use case of gather op.

src/ngraph/pass/nop_elimination.cpp
ashokei: this change only handles convert cases where the following op accepts any type (eg: NonZero).

src/ngraph/pass/nop_elimination.cpp
ashokei: yes ,the static lookup table `type_agnostic` allows us to add more ops (currently just have NonZero, but we could add others, whose compute-logic/result is independent of type of input)

swapna-intel: what I meant is, even if we have more than 1 consumers of the convert, if they are all `type_agnostic`, we can eliminate the convert

src/ngraph/pass/nop_elimination.cpp
swapna-intel: I meant that, we can also add the case I mentioned above, here

src/ngraph/pass/nop_elimination.cpp
ashokei: this PR tries to handle limited cases, along with other op eliminations. may be we could do this as another PR, also could you add a unit-test so in new PR we can test this use-case.

swapna-intel: yes!

swapna-intel: this use case will be a different PR

src/ngraph/pass/nop_elimination.cpp
swapna-intel: will be addressed in follow-up PR

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -33,76 +38,156 @@
...
 using namespace ngraph;
 
-#define TI(x) std::type_index(typeid(x))
+#define TI(x) x::type_info
+
+static bool remove_update_name(const std::shared_ptr<Node>& node,
+                               const std::shared_ptr<Node>& node_input)
+{
+    bool has_result_output = false;
+    for (auto& output : node->output(0).get_target_inputs())
```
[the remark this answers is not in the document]
ashokei: yes, just checks if any of one the outputs is a `Result`,

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -33,76 +38,156 @@
...
+    }
+    // ignore trivial elimination
+    if (has_result_output && as_type_ptr<ngraph::op::Parameter>(node_input))
+    {
+        return false;
+    }
+    if (!has_result_output || node_input->get_users().size() == 1)
+    {
+        node_input->set_friendly_name(node->get_friendly_name());
+        replace_node(node, node_input);
```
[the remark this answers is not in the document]
ashokei: this is only for op elimination, but we could potentially make it part of graph_util.hpp.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -33,76 +38,156 @@
...
+        }
+    }
+    // ignore trivial elimination
+    if (has_result_output && as_type_ptr<ngraph::op::Parameter>(node_input))
+    {
+        return false;
+    }
+    if (!has_result_output || node_input->get_users().size() == 1)
+    {
+        node_input->set_friendly_name(node->get_friendly_name());
```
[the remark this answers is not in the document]
ashokei: this is openvino limitation/requirement, they need this friendly_name to be retained like this.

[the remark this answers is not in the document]
swapna-intel: ok. do we know if that breaks anything for ngraph tests? Or add a test to make sure we will not break existing use case? I am not sure if we know the impact!

src/ngraph/pass/nop_elimination.cpp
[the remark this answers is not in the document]
swapna-intel: but we are checking only output at 0? Result is always at 0? and we can have only 1 result?

src/ngraph/pass/nop_elimination.cpp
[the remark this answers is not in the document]
swapna-intel: If we are going to change the friendly_name usage - it should be generic ti all replace_node cases?

src/ngraph/pass/nop_elimination.cpp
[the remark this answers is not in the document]
ashokei: ngraph itself doesnt use this in its core, this mainly a util for users.

src/ngraph/pass/nop_elimination.cpp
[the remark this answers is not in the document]
ashokei: i will make it available in graph_util; yes , all passes that OV uses from ngraph will have to retain name.  (they do not use any other passes currently, except for cf)

src/ngraph/pass/nop_elimination.cpp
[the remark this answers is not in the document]
ashokei: can be multiple results, we just want to ensure we handle friendly_name differently if/when any one of the output is a "Result".

src/ngraph/graph_util.cpp
```diff
@@ -881,3 +881,28 @@ void ngraph::traverse_functions(std::shared_ptr<Function> p,
 {
     f(p);
 }
+
+bool ngraph::remove_update_name(const std::shared_ptr<Node>& node,
```
swapna-intel, line 885: Can we call this replace_node_update_name

src/ngraph/graph_util.cpp
```diff
@@ -881,3 +881,28 @@ void ngraph::traverse_functions(std::shared_ptr<Function> p,
...
+        }
+    }
+    // ignore trivial elimination
+    if (has_result_output && as_type_ptr<ngraph::op::Parameter>(node_input))
+    {
+        return false;
+    }
+    if (!has_result_output || node_input->get_users().size() == 1)
+    {
+        node_input->set_friendly_name(node->get_friendly_name());
```
GlebKazantaev, line 903: We should set friendly_name to input node only if input is not Parameter and has_result_output = True
`if (has_result_output && !std::dynamic_pointer_cast<ngraph::op::Parameter>(node_input))`

PR #4541, opened 2020-04-09
Constant_folding method

diyessi:
- Refactor graph rewrite to take handlers other than pattern matchers
- Add `constant_fold_default` method to node


src/ngraph/node.hpp
```diff
@@ -189,6 +189,8 @@ namespace ngraph
         virtual const op::AutoBroadcastSpec& get_autob() const;
         /// \returns true if the node can decompose
         virtual bool supports_decompose() const { return false; }
+        /// \returns true replaces this node with a constant
```
slyalin, line 192: @diyessi, please provide more details on what this function should do and what is the expectations (constant inputs etc.). Could you also provide explicit confirmation that replacing it directly in a graph by a constant is more solid approach in comparison to just returning a vector of Constants without doing any in-graph manipulations? Can you make sure that it won't lead to too much extra code when we just want to evaluate some node in a graph without folding it to a Constant (some kind of partial graph evaluation in addition to shape and type propagation)? For example, in MO we are propagating values through the graph without folding parts of a graph into Constants. This apprach helps us to keep shape sub-expression in a graph untouched but do shape inference for entire graph in the same time. I would like to have a solution that doesn't block or complicate this path.

src/ngraph/op/add.cpp
```diff
@@ -59,6 +59,20 @@ void op::v0::Add::generate_adjoints(autodiff::Adjoints& adjoints, const OutputVe
     adjoints.add_delta(y, delta);
 }
 
+namespace
+{
+    bool constant_fold_add(Node* add, const Output<Node> in0, const Output<Node> in1)
+    {
+        cerr << "Calling constand_fold_add" << endl;
```
slyalin, line 66: Please provide a real example of evaluation for a node with multiple _outputs_, for example opset1::Split.

src/ngraph/pass/graph_rewrite.cpp
```diff
@@ -138,30 +130,30 @@ static vector<regex> initialize_fusion_regexes()
     return regexes;
 }
 
-bool pass::GraphRewrite::is_enabled(const shared_ptr<pattern::Matcher>& m) const
+bool pass::GraphRewriteBase::is_enabled(const std::string& name) const
```
slyalin, line 133: @diyessi, please provide details (in the PR description or in the ticket) why we need all these changes here and below for GraphRewrite.

src/ngraph/node.hpp
diyessi: #4489 sketches partial evaluation. 
The request was for a way for custom ops to add have their own constant folders. If they also need to participate in partial evaluation, then output vector results make more sense; that wasn't part of the request.

src/ngraph/pass/graph_rewrite.cpp
diyessi: The code was a mess, and using the pattern matcher as a way to look up a handler is based on op type is not a very good way to do things.

src/ngraph/node.hpp
slyalin: @diyessi, sure it was not a part of that particular request but I want to unlock the possibility to re-use it for something related.

slyalin: @diyessi, can we finalize this interface by the end of this week just to unblock others? If you think that returning a vector of constants is a better approach considering the other usages that I mentioned, please change the signature. Can you provide more examples with implementing this function for multi-port operations as I asked in another comment?

diyessi: Example using `ShapeOf`.

GlebKazantaev:
@diyessi @slyalin I like this approach because constant folding is a Node class method and cf implementation belongs to an operation. Also as I understand right this changes in GraphRewrite helps not to use patterns when we want to match single op that's more convenient and helps to avoid this workaround I did in my POC.
So I accept this approach!

src/ngraph/node.hpp
```diff
@@ -193,6 +193,9 @@ namespace ngraph
         virtual const op::AutoBroadcastSpec& get_autob() const;
         /// \returns true if the node can decompose
         virtual bool supports_decompose() const { return false; }
+        /// \brief Can replace a node with a constant during constant folding.
+        /// \returns vector of outputs to replace node outputs. Empty outputs will not be replace.
+        virtual OutputVector constant_fold_default() { return {}; }
```
slyalin, line 198: Please rename to `constant_fold`, without `_default`. I don't know what `default` means in this case. This is _only_ place where we define constant folding for an operation. We are going to migrate all code for constant folding under this API. With this renaming the rest of the code looks good to me.

PR #4543, opened 2020-04-10
[MLIR] Use OutputVector for compiled kernel inputs/outputs

diyessi:
Remove deprecated initialization in `CompiledKernel`


src/ngraph/op/experimental/compiled_kernel.cpp
```diff
@@ -34,13 +34,13 @@ shared_ptr<Node>
     }
 
     // map inputs
-    unordered_map<Node*, Output<Node>> nm;
+    map<Output<Node>, Output<Node>> nm;
```
adstraw, line 37: Why change to `map` from `unordered_map`?  Do you rely on keys being ordered somehow?  Looks like we are only accessing below by single element using [] or at().  This should be faster with an unordered_map, right?

PR #4544, opened 2020-04-10
[DynamicShape] Add dynamic shape support for GlobalLpPool

mbencer:
- dynamic version of `builder::opset1::get_monotonic_range`
- changed normalization builders to use Node


src/ngraph/frontend/onnx_import/op/lp_pool.cpp
```diff
@@ -39,7 +39,16 @@ namespace ngraph
...
                     const std::shared_ptr<ngraph::Node> data{node.get_ng_inputs().at(0)};
                     const std::size_t channel_axis{1};
-                    const std::size_t channels_count = data->get_shape().at(channel_axis);
+
+                    const auto data_shape = data->get_output_partial_shape(0);
+                    NGRAPH_CHECK(data_shape.rank().is_static(),
+                                 "Rank of input data must be static");
+                    NGRAPH_CHECK(data_shape[0].is_static(),
+                                 "First dimension of input data must be static");
+                    NGRAPH_CHECK(data_shape[channel_axis].is_static(),
```
tomdol, line 48: Not sure but for safety, should we also check that rank is at least(or exactly equal to) 2?

silee2:
The PR replaces a static attribute with an input. That changes the API and would have backward compatibility issue. Any reason why you couldn't add additional API that supports dynamic shape instead of replacing the old one?

mbencer:
> The PR replaces a static attribute with an input (Node) for builders. That changes the builder API and would have backward compatibility issue. Any reason why you couldn't add additional builder API that supports dynamic shape instead of replacing the old one? Or nobody is using the old builders?

@silee2 
I am not aware of any user of this `non-static` `mean` builder API.
Due to the fact that it is NOT part of `ngraph.hpp` main header I've thought about it as internal helper.
However, it is not problem to keep both version (in separate files such as `builder/mean` and `builder/dynamic_mean` or in the same file???)

@diyessi What do you think about it?

src/ngraph/frontend/onnx_import/op/lp_pool.cpp
mbencer: done

PR #4549, opened 2020-04-10
[ Gather ] Constant Folding for scalar case

jane-intel:


src/ngraph/pass/constant_folding_gather.cpp
```diff
@@ -236,9 +237,16 @@ void pass::ConstantFolding::construct_constant_gather_with_subgraph()
         const int64_t raw_index = indices->cast_vector<int64_t>()[0];
         const int64_t positive_index = raw_index < 0 ? rank + raw_index : raw_index;
         // gather takes exactly one element out of the Concat output
-        const auto gathered =
+        const auto gathered_concat_input =
```
diyessi, line 240: Check that `positive_index` < `rank` before `const_inputs[positive_index]`

PR #4553, opened 2020-04-13
Introduced mechanism to propagate rt info attributes through transformations

GlebKazantaev:
Introduction: we have a request to preserve connection between names in IR and names in nGraph function that comes to plugins. And it's possible only if we propagate this names in through nGraph transformations. And for this purpose I introduce this approach for original names propagation via rt info attributes.  So on OV side I have created a FusedNames class to store set of names and VariantWrapper for it.  And as the next step I had to somehow propagate this attribute from nodes that I replace to the new nodes. 

Within this PR I extended Variant class with two virtual methods: init and merge. This helps to set the  rules for rt info attribute how it should be merged. For example: Conv->Bias fusion transformation, both of this ops has FusedNames attribute. So first we need to merge this attributes and then set it to the new ConvBias operation. To simplify this process and not handle any type of attribute I have copy_runtime_info function that should be used in transformation.


src/ngraph/rt_info.cpp
```diff
@@ -0,0 +1,80 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include <ngraph/node.hpp>
+#include <ngraph/variant.hpp>
+
```
rkimballn1, line 19:
```suggestion
#include "ngraph/node.hpp"
#include "ngraph/variant.hpp"
```

src/ngraph/rt_info.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "rt_info.hpp"
+
+namespace ngraph
+{
```
rkimballn1, line 23: Convention for cpp files is to `prepend ngraph::` to the name of the function and not wrap the whole file in the ngraph namespace. This saves one level of indenting.

src/ngraph/rt_info.cpp
```diff
@@ -0,0 +1,80 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#include <ngraph/node.hpp>
+#include <ngraph/variant.hpp>
+
+#include "rt_info.hpp"
```
rkimballn1, line 20:
```suggestion
#include "ngraph/rt_info.hpp"
```

src/ngraph/rt_info.cpp
GlebKazantaev: Done.

GlebKazantaev:
Smth went worng with commits in this MR. So I have to close it and reopen new PR: https://github.com/NervanaSystems/ngraph/pull/4565

PR #4561, opened 2020-04-13
Move nop backend

rkimballn1:
A start just to get everyone aligned on where the backends will land


raramer01:
TESTNOW_DLDT_E2E

raramer01:
TESTNOW

PR #4562, opened 2020-04-13
Remove obsolete cmake

rkimballn1:


silee2:
This seems to depend on #4563

PR #4563, opened 2020-04-13
Remove support for old version of MKLDNN

rkimballn1:
There are two methods of conditional compile here
`#if MKLDNN_VERSION_MAJOR < 1`
and
`#if MKLDNN_VERSION_MAJOR >= 1`
making the PR extra fun. Please make sure I did this all correctly.


src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1226,23 +1226,11 @@ void runtime::cpu::CPU_ExternalFunction::register_common_passes(
...
-#else
-            // TODO: will be supported in mkldnn v1.1
-            return false;
-#endif
         }
         else if (typeid(ngraph::op::Gelu) == typeid(node))
         {
-#if MKLDNN_VERSION_MAJOR < 1
-            // TODO: (gauri): need to differentiate which implementation : erf vs tanh
             return false;
```
gaurides, line 1233: Same here about to-do comment.

src/ngraph/runtime/cpu/op/lstm.hpp
```diff
@@ -32,41 +32,29 @@ namespace ngraph
...
+            // [0] - {Xt} input tensor of layout TNC, Shape{sequence length*batch_size,
+            //       feature_size}
+            // [1] - recurrent state tensors {ht_1 | ct_1} of Shape{sequence length*batch_size,
+            //       feature_size}
+            // [2] - initializer for the input weights matrix, used for the linear transformation of
+            //       the inputs.
+            // [3] - initializer for the recurrent weights matrix, used for the linear
+            //       transformation of the recurrent state.
+            // [4] - Initializer for the bias vector w.r.to inputs + hidden state (ibh_bias +
+            //       hbh_bias)
```
gaurides, line 45: not in scope of this review but we will need to check with Pruthvi about the documentation here. the new Lstm op has more args

PR #4564, opened 2020-04-13
Optimize gather

pruthviIntel:
Optimize gather, when gather is Nop i.e if gather is fetching the input tensor as it is.


src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,88 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+
+        if (axis == std::numeric_limits<int64_t>::max())
+        {
+            NGRAPH_DEBUG << "axis value not set";
+            return false;
+        }
+
+        // lamba function to remove the gather op and rewire the
+        // input data of gather to its users
+        auto replace_gather = [&]() {
```
ashokei, line 378: can we use replace_node similar to other optimizations in this pass?

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,88 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+    auto gather = as_type_ptr<op::Gather>(node);
+
+    if (gather)
+    {
+        // check if we are gathering the whole input
+        auto data = node->get_argument(0);
+        auto indices = node->get_argument(1);
+
+        // if rank of data and gather output dont match, we will skip
+        if (data->get_shape().size() != node->get_shape().size())
```
ashokei, line 363: can you handle/check dynamic shape case?

src/ngraph/pass/algebraic_simplification.cpp
ashokei: https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/graph_util.hpp#L508

ashokei:
@pruthviIntel  couple of things : we need to handle v1 gather, so please update this PR accordingly. Also, i think for static shape you could just look at input and output shape and eliminate if it doesnt' change? i will push a commit to this PR in nop_elimination please check.

pruthviIntel:
@ashokei inferring gather to be Nop just be shape is not right. You can get the same shape in output with different permutation of indices.  It needs to be checked explicitly with indices ordering as I have done in my changes. The changes you have made will fail for the below case


```
>>> data 
array([[[0, 1],
        [2, 3]],

       [[4, 5],
        [6, 7]]])
>>> data.shape
(2, 2, 2)
>>> gather = tf.gather(data, [1, 0], axis = 1)
>>> gather      
<tf.Tensor 'GatherV2_40:0' shape=(2, 2, 2) dtype=int64>
>>> eval(gather)
array([[[2, 3],
        [0, 1]],

       [[6, 7],
        [4, 5]]])
>>> data == eval
array([[[False, False],
        [False, False]],

       [[False, False],
        [False, False]]])
>>> data.shape
(2, 2, 2)
>>> gather.shape
TensorShape([Dimension(2), Dimension(2), Dimension(2)])
>>> eval(gather).shape
(2, 2, 2)
>>> 
```

ashokei:
@pruthviIntel can you please add this test, i removed the nop_elimination changes, also updated tests to use v1 (i think cpu backend does op downgrade to v0, but need a way to test v1 gather). 
could you update this PR to use v1 gather, checks dynamic shapes, handle partial dnamic where possible? (as openvino doesnt use cpu backend).

pruthviIntel:
sure, will do.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,91 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+        auto indices = node->get_argument(1);
+
+        // we need to know data and indices shape to infer if gather is Nop
+        if (data->get_output_partial_shape(0).is_dynamic() ||
+            indices->get_output_partial_shape(0).is_dynamic())
+        {
+            return false;
+        }
+        // if rank of data and gather output dont match, we will skip
+        if (data->get_shape().size() != node->get_shape().size())
```
ashokei, line 369: do we need to check if node->get_shape maybe dynamic (as 3rd input,  axis , could be dynamic?)

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,91 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+        if (axis == std::numeric_limits<int64_t>::max())
+        {
+            NGRAPH_DEBUG << "axis value not set";
+            return false;
+        }
+
+        // lamba function to remove the gather op and rewire the
+        // input data of gather to its users
+        auto replace_gather = [&]() {
+            remove_node_update_name(node, node->get_argument(0));
```
ashokei, line 385: return the result of this function. also, why not directly call this func instead of lambda

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,91 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+        {
+            replace_gather();
+        }
+
+// case_2 : if the input tensor is of shape (4, 3, 4)
+// we need to check the contents of indices, if indices
+// is 1D tensor of value {0, 1, 2}, we can optimize this
+// op has Nop
+
+#define COMPARE_AND_REPLACE_GATHER(T)                                                              \
```
ashokei, line 404: i wonder if there is a way to not use macro? not a big issue, just curious.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,58 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+    auto gather = as_type_ptr<op::Gather>(node);
+
+    if (gather)
+    {
+        // check if we are gathering the whole input
+        auto data = node->get_argument(0);
+        auto indices = node->get_argument(1);
+
+        // check if the indices is constant
+        if (!std::dynamic_pointer_cast<ngraph::op::Constant>(node->get_argument(1)))
```
ashokei: use as_type_ptr

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,91 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+        {
+            return false;
+        }
+        // if rank of data and gather output dont match, we will skip
+        if (data->get_shape().size() != node->get_shape().size())
+        {
+            return false;
+        }
+
+        auto axis = gather->get_axis();
```
ashokei, line 374: dynamic case?

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,91 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+#define COMPARE_AND_REPLACE_GATHER(T)                                                              \
+    std::vector<T> ref_indices(data->get_shape()[axis], 0);                                        \
+    std::iota(ref_indices.begin(), ref_indices.end(), 0);                                          \
+    if (ref_indices == indices_node->get_vector<T>())                                              \
+    {                                                                                              \
+        replace_gather();                                                                          \
+    }
+
+        // check if the indices is constant
+        auto indices_node = std::dynamic_pointer_cast<ngraph::op::Constant>(node->get_argument(1));
```
ashokei, line 413: use `as_type_ptr`

src/ngraph/pass/algebraic_simplification.cpp
pruthviIntel: if its dynamic, we are doing that check in the  Gather Op definition  and axis will be set to INT_MAX; I am checking the same and  and skip optimization if its true.

src/ngraph/pass/algebraic_simplification.cpp
pruthviIntel: I though of using template function, but since its a small patch of code, I ended up using macro

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,86 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
     return rc;
 }
 
+//`simplify_gather`, optimizes gather if Gather is gathering the
+// whole input tensor
+static bool simplify_gather(std::shared_ptr<Node> node)
+{
+    auto gather = as_type_ptr<op::v1::Gather>(node);
+
+    if (gather)
```
diyessi, line 356: `if (auto gather=as_type_ptr<op::v1::Gather>(node))`

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,86 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+    {
+        // check if we are gathering the whole input
+        auto data = node->get_argument(0);
```
diyessi, line 359: These should be `input_value(0)`, `input_value(1)` since `get_argument` doesn't work for multi-output nodes.

src/ngraph/pass/algebraic_simplification.cpp
diyessi: Use a template.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,86 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+        }
+        // if rank of data and gather output dont match, we will skip
+        if (data->get_shape().size() != node->get_shape().size())
+        {
+            return false;
+        }
+
+        auto axis = gather->get_axis();
+
+        if (axis == std::numeric_limits<int64_t>::max())
```
jane-intel, line 376: Should we replace numeric limits with AXIS_NOT_SET_VALUE from Gather op?

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,86 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+    auto gather = as_type_ptr<op::v1::Gather>(node);
+
+    if (gather)
+    {
+        // check if we are gathering the whole input
+        auto data = node->get_argument(0);
+        auto indices = node->get_argument(1);
+
+        // we need to know data and indices shape to infer if gather is Nop
+        if (data->get_output_partial_shape(0).is_dynamic() ||
```
jane-intel, line 363: Algorithm should be the following:
1. get_axis() != AXIS_NOT_SET_VALUE && axis > 0
2. data_partial_shape[axis] is static
3. indices are 1D constant and contain range of values 0...len(data_partial_shape[axis]). case with single 1D index doesn't require it to be constant.

I'm I missing smth or we do not need such restrictions?

src/ngraph/pass/algebraic_simplification.cpp
pruthviIntel: @jane-intel `AXIS_NOT_SET_VALUE` is a const variable in gather Op , whose value is set to `std::numeric_limits<int64_t>::max()`

src/ngraph/pass/algebraic_simplification.cpp
pruthviIntel: 1. when I query `gather->get_axis()`, the member function would do the check for point 1 and point 2 as you explained and infer the axis value. if its dynamic the axis value will be set to `AXIS_NOT_SET_VALUE`. 

2. For point 3, I am doing this check here, 
https://github.com/NervanaSystems/ngraph/blob/8edab564148392b14200ac0e906e2b7b46ec8121/src/ngraph/pass/algebraic_simplification.cpp#L398-#L405
if the indices are 1D and whose value values equal to {0..data_partial_shape[axis]}. then that would fetch the entire input tensor

Please let me know if i resolved your questions :)

src/ngraph/pass/algebraic_simplification.cpp
ashokei: @pruthviIntel you could move AXIS_NOT_SET_VALUE inside header, see https://github.com/NervanaSystems/ngraph/pull/4596

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,73 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+        }
+        else
+        {
+            // if ref_inidices == indices, we are capturing the
+            // entire input tensor
+            std::vector<int64_t> ref_indices(data.get_shape()[axis], 0);
+            std::iota(ref_indices.begin(), ref_indices.end(), 0);
+            if (ref_indices == constant_indices->get_vector<int64_t>())
+            {
+                remove_node_update_name(node, node->get_argument(0));
```
ashokei, line 409: return the output on this.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -345,6 +347,73 @@ static bool is_input_uniform_constant(shared_ptr<Node> op,
...
+            return true;
+        }
+
+        // case_2 : if the input tensor is of shape (4, 3, 4)
+        // we need to check the contents of indices, if indices
+        // is 1D tensor of value {0, 1, 2}, we can optimize this
+        // op has Nop
+
+        // check if the indices is constant
+        auto constant_indices = as_type_ptr<op::Constant>(gather->get_argument(1));
```
jane-intel, line 396: Please use `input_value(1).get_node_shared_ptr()` instead of `get_argument(1)` here and everywhere in the code.

src/ngraph/op/gather.hpp
```diff
@@ -58,6 +58,7 @@ namespace ngraph
             {
             public:
                 static constexpr NodeTypeInfo type_info{"Gather", 1};
+                static const int64_t AXIS_NOT_SET_VALUE = std::numeric_limits<int64_t>::max();
```
diyessi, line 61: Might need `const int64_t Gather::AXIS_NOT_SET_VALUE;` in the .cpp file for some compilers.

PR #4565, opened 2020-04-13
Introduced mechanism to propagate rt info attributes through transformations

GlebKazantaev:
Introduction: we have a request to preserve connection between names in IR and names in nGraph function that comes to plugins. And it's possible only if we propagate this names in through nGraph transformations. And for this purpose I introduce this approach for original names propagation via rt info attributes. So on OV side I have created a FusedNames class to store set of names and VariantWrapper for it. And as the next step I had to somehow propagate this attribute from nodes that I replace to the new nodes.

Within this PR I extended Variant class with two virtual methods: init and merge. This helps to set the rules for rt info attribute how it should be merged. For example: Conv->Bias fusion transformation, both of this ops has FusedNames attribute. So first we need to merge this attributes and then set it to the new ConvBias operation. To simplify this process and not handle any type of attribute I have copy_runtime_info function that should be used in transformation.


GlebKazantaev:
@diyessi I had to close previous PR due to smth went wrong with commit history after rebase... 
So if you have no concerns can we merge this PR?

PR #4566, opened 2020-04-14
[OV] Replace transpose with reshape 

nbpatel:


src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,73 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+
+    if (transpose)
+    {
+        auto perm = as_type_ptr<op::Constant>(n->get_argument(1));
+        auto perm_value = perm->get_vector<int64_t>();
+        auto data = n->get_argument(0);
+        auto data_shape = n->input_value(0).get_partial_shape().to_shape();
+
+        // Data shape should be >= 3
+        if (data_shape.size() < 3 || data_shape.size() > 4)
```
lazarevevgeny, line 560: Why do you have such a limitations? What is the reason do not support  2D and 5D+ tensors?

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,73 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+            return false;
+
+        // Check if input shape contains a Dim of value 1
+        int index_of_one = 0; // index of dim 1 in data_shape
```
lazarevevgeny, line 564: It may happen that there are multiple dimensions equal to 1. You should remove **all** dimensions equal to 1 in the code below

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,73 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+{
+    bool rc = false;
+    auto transpose = as_type_ptr<op::Transpose>(n);
+
+    if (transpose)
+    {
+        auto perm = as_type_ptr<op::Constant>(n->get_argument(1));
+        auto perm_value = perm->get_vector<int64_t>();
+        auto data = n->get_argument(0);
+        auto data_shape = n->input_value(0).get_partial_shape().to_shape();
```
lazarevevgeny, line 557: You should check that the rank is static. Then you should iterate over the partial shape which may have dynamic dimensions. The dynamic dimensions should not be a blocker to perform this optimization. If the dimension is dynamic then we should consider it as not equal to 1.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,73 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+        }
+
+        if (!index_of_one)
+            return false;
+
+        // Remove index of 1 from perm_value and check if the other
+        // dimensions haven't interchanged
+        vector<int64_t> perm_value_copy = perm_value;
+
+        for (int i = 0; i < perm_value_copy.size(); i++)
```
lazarevevgeny, line 581: Since multiple 1s should be supported this for loop must be rewritten because you iterate over the vector which you modify (remove element). I would recommend another approach. Let's iterate over all elements in the perm_value. If the corresponding shape element is not one then we add the current perm_value element to the output vector. In the end we will have an output vector with indices of dimensions not equal to 1.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,73 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+
+        Shape out_shape{};
+        AxisVector input_order{};
+        for (int i = 0; i < perm_value.size(); i++)
+        {
+            out_shape.push_back(data_shape[perm_value[i]]);
+            input_order.push_back(perm_value[i]);
+        }
+
+        auto reshape_op = make_shared<op::Reshape>(data, input_order, out_shape);
```
lazarevevgeny, line 607: The output shape for the Reshape is just the output shape from Transpose. So can we just use `n->output_value(0).get_partial_shape()` ?

test/algebraic_simplification.cpp
```diff
@@ -596,3 +596,37 @@ TEST(algebraic_simplification, pass_property)
...
+
+    pass::Manager pass_manager;
+    pass_manager.register_pass<pass::AlgebraicSimplification>();
+    pass_manager.run_passes(f);
+
+    ASSERT_EQ(count_ops_of_type<op::Transpose>(f), 0);
+    ASSERT_EQ(count_ops_of_type<op::Reshape>(f), 1);
+}
+
+TEST(algebraic_simplification, replace_transpose_with_reshape_3d)
```
lazarevevgeny, line 617: Need add much more unit tests:
- shape [10,20,1,1], permute_order = [0,2,3,1]
- shape [10,1,1,20], permute_order = [0,3,1,2]
- shape [10,20,1,2], permute_order = [0,2,1,3]
- shape [10,1,1,1,20], permute_order = [0,4,1,2,3]
- shape [10,20,1,1,1], permute_order = [0,2,3,4,1]
- shape [10,1,1,1,1], permute_order = [1,4,2,3,0]
- shape [10,1,1,1,1], permute_order = [4,2,0,1,3]
- shape [1,1,1,1,1], permute_order = [4,2,0,1,3]

Also add tests where the transformation should not be applied:
- shape [10,20,1,2], permute_order = [0,2,3,1]
- shape [10,20,1,2], permute_order = [0,3,1,2]
- shape [10,20], permute_order = [1,0]

Also need some tests where some dimensions are dynamic.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,73 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+            {
+                perm_value_copy.erase(perm_value_copy.begin() + i);
+                break;
+            }
+        }
+
+        // If all the indices are in increasing order,
+        // then the other dims are unchanged and only dim of 1
+        // has changed its position
+        for (int i = 0; i < perm_value_copy.size() - 1; i++)
```
lazarevevgeny, line 593: This line will result in an infinite while loop if the size of the perm_value_copy is 0. The better solution is to rewrite this line the following way:

`for (size_t i = 0; i + 1 < perm_value_copy.size(); ++i)`

src/ngraph/pass/algebraic_simplification.cpp
lazarevevgeny: After some time of reviewing the code it seems that the dynamic input shape may be a problem. If there is just one dynamic dimension then we can insert value "-1" for this dimension in the Reshape second input, but if there are more than two dynamic dimensions then we cannot perform this transformation. We need to cover this cases in unit-tests.

lazarevevgeny: But the general solution will be the following: Insert ShapeOf operation which gets input from the input to Transpose. Then pass output from ShapeOf to operation Gather as first input. The second input to Gather will be the permute_order from Transpose. So the result of Gather operation should go as second input to Reshape operation.

src/ngraph/pass/algebraic_simplification.cpp
wenzhe-nrv, line 564: setting to 0 may cause an issue if the first dim is 1 (e.g. shape{1, ?}). But It will return false in the line 575.  Maybe initiate with `-1` or use an iterator instead.

src/ngraph/pass/algebraic_simplification.cpp
nbpatel: good catch

src/ngraph/pass/algebraic_simplification.cpp
nbpatel: because we dont deal with 5D tensor and @wenzhe-nrv had a optimization to remove out useless transpose for 2D vectors, hence I had it as < 3 and > 4..
anyway, removed it and now its generic and can work for n dims

src/ngraph/pass/algebraic_simplification.cpp
nbpatel: generalized it

src/ngraph/pass/algebraic_simplification.cpp
nbpatel: Was doing the same thing indirectly ...changed it though

test/algebraic_simplification.cpp
nbpatel: Tested all the cases above manually, not sure if we need a test for every permuatation @diyessi ? , although I wouldn't mind,  and added a test for a failing case (in cases it shouldn't replace)

rkimballn1: Not sure what "manually" testing things is. I would add unit tests for all cases listed above. Unit tests for cases that pass makes sure they don't fail in the future when someone else edits the code.

nbpatel: By manually, I meant I copied the shapes and made sure everything mentioned above passed. I added those cases now.

lazarevevgeny:
We have discussed that this transformation must be implemented in the DLDT repository @GlebKazantaev  can provide more details how and where.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,62 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+static bool replace_transpose_with_reshape(shared_ptr<Node> n)
+{
+    bool rc = false;
+    auto transpose = as_type_ptr<op::Transpose>(n);
+
+    if (transpose)
+    {
+        auto perm = as_type_ptr<op::Constant>(n->get_argument(1));
+        auto perm_value = perm->get_vector<int64_t>();
+        auto data = n->get_argument(0);
```
diyessi, line 556: This should be `n->input_value(0)` and then use `get_partial_shape()` on that. You need to check that the partial shape is static before calling `to_shape()` or you'll get an error.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,62 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+        auto data_shape = n->input_value(0).get_partial_shape().to_shape();
+
+        // Check if input shape contains a Dim of value 1
```
diyessi, line 559: Why not just run through the perm axes and if those dimensions are all static of length 1 then you can do the reshape.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,62 @@ static bool simplify_reduction(shared_ptr<Node> n)
     return true;
 }
 
+static bool replace_transpose_with_reshape(shared_ptr<Node> n)
+{
+    bool rc = false;
+    auto transpose = as_type_ptr<op::Transpose>(n);
+
+    if (transpose)
```
jane-intel, line 552: reduce indentation by `if(!transpose) return false;` or rewrite like `if (const auto transpose = as_type_ptr<op::Transpose>(n))`

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,62 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+        // If all the indices are in increasing order,
+        // then the other dims are unchanged and only dim of 1
+        // has changed its position
+        for (int i = 0; i + 1 < perm_value_copy.size(); i++)
+        {
+            if (perm_value_copy[i] > perm_value_copy[i + 1])
+                return false; // not a valid permutation
+        }
+
+        auto reshape_op = make_shared<op::Reshape>(data,
```
ashokei, line 594: has to be V1 reshape

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,83 @@ static bool simplify_reduction(shared_ptr<Node> n)
     return true;
 }
 
+static bool replace_transpose_with_reshape(shared_ptr<Node> n)
+{
+    bool rc = false;
+    auto transpose = as_type_ptr<op::Transpose>(n);
```
ashokei, line 550: use explicit op version, v1::Transpose, helps clarity , and avoid future conflicts if any

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,83 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+
+    if (!transpose)
```
ashokei, line 552: may have to apply style / format?

nbpatel: I used it

src/ngraph/pass/algebraic_simplification.cpp
nbpatel: It supports 1 unknown dim

src/ngraph/pass/algebraic_simplification.cpp
ashokei: i mean for braces, not sure CI will pass on style

nbpatel: it did, but I'll add to be consistent

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,93 @@ static bool simplify_reduction(shared_ptr<Node> n)
     return true;
 }
 
+static bool replace_transpose_with_reshape(shared_ptr<Node> n)
+{
+    bool rc = false;
+    auto transpose = as_type_ptr<op::v1::Transpose>(n);
+
+    if (!transpose)
```
ashokei, line 552: this check is probably redundant as this func only gets called if the node is transpose.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,93 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+            }
+        }
+
+        if (count_dynamic_dims > 1)
+        {
+            return false; // only works on one dynamic dim
+        }
+    }
+
+    auto perm = as_type_ptr<op::Constant>(n->get_argument(1));
```
ashokei, line 581: why assume constant

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,93 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+    auto perm_value = perm->get_vector<int64_t>();
+    auto data = n->get_argument(0);
```
ashokei, line 583: get_argument may have issues with multioutput nodes, prefer `input_value(0).get_node_shared_ptr()`

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,93 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+        if (perm_value_copy[i] > perm_value_copy[i + 1])
+        {
+            return false; // not a valid permutation
+        }
+    }
+
+    auto constant_node =
+        ngraph::op::Constant::create(element::i64, Shape{data_shape.size()}, data_shape);
+    auto reshape_op = make_shared<op::v1::Reshape>(data, constant_node, false);
+    replace_node(n, reshape_op);
```
ashokei, line 628: use remove_node_update_name

nbpatel: > use remove_node_update_name

whats that? never used it, it is something recently added and how is that better than replace_node?

ashokei: handles openvino friendly_name

test/algebraic_simplification.cpp
```diff
@@ -593,6 +593,89 @@ TEST(algebraic_simplification, pass_property)
...
+
+        pass::Manager pass_manager;
+        pass_manager.register_pass<pass::AlgebraicSimplification>();
+        pass_manager.run_passes(f);
+
+        ASSERT_EQ(count_ops_of_type<op::v1::Transpose>(f), 0);
+        ASSERT_EQ(count_ops_of_type<op::v1::Reshape>(f), 1);
+    };
+
+    check_usecase(Shape{2, 3, 1}, vector<int64_t>{2, 0, 1});
```
wenzhe-nrv, line 616: will a 2D test case be considered?

nbpatel: yes

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,104 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+
+        if (count_dynamic_dims > 1)
+        {
+            return false; // only works on one dynamic dim
+        }
+
+        is_dyn_dim = true;
+    }
+
+    auto perm = as_type_ptr<op::Constant>(n->get_argument(1));
```
jane-intel, line 580: Please get rid of `get_argument`

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -543,6 +544,104 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
+            {
+                count_dynamic_dims++;
+                shape[i] = 2; // any non 1 number
+                dyn_dim_index = i;
+            }
+        }
+
+        if (count_dynamic_dims > 1)
+        {
+            return false; // only works on one dynamic dim
```
jane-intel, line 574: Why it works for only one dyn dim?

jane-intel: In OV we usually get ShapeOf of that node, than use Split(num_splits=rank) and than Concatenate it in the order we like with additional dimensions equal to one. Does it seem sane?

jane-intel:
Consider this is a pseudocode -- i haven't chance to compile it yet. Please rewrite transform according to it.
```
auto transpose = as_type_ptr<op::v1::Transpose>(n);

PartialShape shape = n->input_value(0).get_partial_shape();
if (shape.rank().is_dynamic())
{
    return false;
}

auto perm = as_type_ptr<op::Constant>(n->input_value(1).get_node_shared_ptr());
if (!perm)
{
    return false;
}

auto perm_value = perm->get_vector<int64_t>();
for (auto i = shape.rank().get_length() - 1; i >= 0; --i)
{
    if (shape[i].is_static() && shape[i] == 1)
        perm_value.erase(perm_value.begin() +  i);
}

if (std::is_sorted(perm_value)) 
{
    auto shape_of = make_shared<op::v3::ShapeOf>(data);
    auto gather = make_shared<op::Gather>(shape_of, perm);
    auto reshape_op = make_shared<op::v1::Reshape>(data, gather, false);
    return remove_node_update_name(n, reshape_op);
}
return false;
```

PR #4568, opened 2020-04-14
[ONNX] Provide support for function in ONNX models [Approach 2]

Ewa21:


src/ngraph/frontend/onnx_import/op/dynamic_quantize_linear.cpp
```diff
@@ -0,0 +1,96 @@
...
+                        for (auto& input : nodes.at(i)->inputs())
+                        {
+                            if (input.get_shape() == node.get_ng_inputs().at(0)->get_shape())
+                            {
+                                input.replace_source_output(node.get_ng_inputs().at(0));
+                            }
+                        }
+                    }
+                    // Quantze, Divde, Convert
+                    return NodeVector{nodes.at(23), nodes.at(10), nodes.at(16)};
```
tomdol, line 88: Is it guaranteed that the nodes returned here will always be at those indices?

src/ngraph/frontend/onnx_import/utils/common.cpp
```diff
@@ -66,6 +73,53 @@ namespace ngraph
...
+            {
+                const auto* schema = ONNX_NAMESPACE::OpSchemaRegistry::Schema(
+                    new_node->op_type(), opset_version, "");
+                const ONNX_NAMESPACE::FunctionProto* func = schema->GetFunction();
+
+                FunctionExpandHelper(*new_node, *func, graph);
+
+                graph.mutable_node()->erase(graph.node().begin());
+
+                // Save graph to file
```
tomdol, line 113: This comment seems to be obsolete, you're not saving to a file.

Ewa21: Done

src/ngraph/frontend/onnx_import/op/dynamic_quantize_linear.cpp
Ewa21: Done

Ewa21:
Moved to the  OpenVino repository.

PR #4572, opened 2020-04-14
Import ONNX models directly from ModelProto

tomdol:
This PR extends the onnx_importer API with a function that takes a reference to onnx::ModelProto and imports a model that is already in memory. This new API saves time an memory because the serialization/deserialization step between UEP and onnx_importer is not required any more.

An alternative approach with custom ONNX namespace is here https://github.com/NervanaSystems/ngraph/pull/4530 but with the "ngraph_onnx" default value it does not work under UEP(linker doesn't find the new function from onnx_importer API)


src/ngraph/frontend/onnx_import/onnx.cpp
```diff
@@ -52,6 +52,19 @@ namespace ngraph
             } // namespace error
         }     // namespace detail
 
+        std::shared_ptr<Function> import_onnx_model(const onnx::ModelProto& model_proto)
```
tsocha, line 55: I wonder what will happen if there will be a missmatch between caller and our versions of onnx library

tomdol: It won't work, just like it currently doesn't if we choose to use "ngraph_onnx" (https://github.com/NervanaSystems/ngraph/pull/4530) and try to link against it in UEP.

PR #4573, opened 2020-04-14
Expand attribute visitor types

diyessi:


ilyachur:
Can we cover those changes by new tests?

diyessi:
> Can we cover those changes by new tests?

Added.

We're doing all scalar int types as `int64_t` and `float` as `double`. This was intentional to keep serialization simpler, but I can change it so `[u]int_dd_t` and `float` are themselves. Let me know before I forget how these work again.

PR #4432, opened 2020-03-10
RoiAlign operator - ONNX importer + core

tomdol:

PR #4577, opened 2020-04-15
Adjust NonZero to the new specs

Ewa21:


tomdol:
Can you add a type prop test for i32 output type? A test with i32 for constant folding would also be nice to see if the pattern matcher can find NonZero nodes with i32 attribute value.

src/ngraph/pass/constant_folding_non_zero.cpp
```diff
@@ -193,23 +194,51 @@ void pass::ConstantFolding::construct_constant_non_zero()
         auto pattern_map = m.get_pattern_map();
 
         const auto data = static_pointer_cast<op::Constant>(pattern_map[data_label]);
+        const auto non_zero_matched = as_type_ptr<op::v3::NonZero>(m.get_match_root());
+        auto outpu_type = non_zero_matched->get_output_type();
```
tomdol: Just a typo:
```suggestion
        auto output_type = non_zero_matched->get_output_type();
```

test/constant_folding.cpp
```diff
@@ -2647,6 +2647,29 @@ TEST(constant_folding, constant_non_zero_1D)
...
+
+    pass::Manager pass_manager;
+    pass_manager.register_pass<pass::ConstantFolding>();
+    pass_manager.run_passes(f);
+
+    ASSERT_EQ(count_ops_of_type<op::v3::NonZero>(f), 0);
+    ASSERT_EQ(count_ops_of_type<op::Constant>(f), 1);
+
+    const auto new_const = as_type_ptr<op::Constant>(f->get_results().at(0)->get_argument(0));
+    ASSERT_TRUE(new_const);
```
tomdol, line 2665: Can you also assert that the `new_const`'s element type is i32?

Ewa21: Done

src/ngraph/op/non_zero.hpp
```diff
@@ -42,13 +42,26 @@ namespace ngraph
...
+                element::Type get_output_type() const { return m_index_element_type; }
+                void set_output_type(element::Type output_type)
+                {
+                    m_index_element_type = output_type;
+                }
+                // Overload collision with method on Node
+                using Node::set_output_type;
+
+            protected:
+                element::Type m_index_element_type;
```
tomdol, line 64: Just noticed in the specs that this should be a string. The class member can still be an enum but we should add a constructor that takes a string and converts it to enum. Refer to this PR for more details on how to do it https://github.com/NervanaSystems/ngraph/pull/4432/files?file-filters%5B%5D=.cpp#diff-b358c2a0e091411efef86dea664fb6cdR24

test/type_prop/non_zero.cpp
```diff
@@ -24,7 +24,7 @@ using namespace ngraph;
 TEST(type_prop, non_zero)
 {
     auto data = make_shared<op::Parameter>(element::f32, Shape{3, 3, 224, 224});
-    auto non_zero = make_shared<op::NonZero>(data);
+    auto non_zero = make_shared<op::v3::NonZero>(data, element::i64);
```
lazarevevgeny, line 27: I think that we need to keep existing tests and add new tests for v3 version of NonZero

Ewa21: There is no version of NonZero other than the v3::NonZero I have modified. It was necessary to adjust the existing tests to the new implementation.

src/ngraph/op/non_zero.hpp
Ewa21: Done

src/ngraph/op/non_zero.cpp
```diff
@@ -22,14 +22,23 @@ using namespace std;
...
+op::v3::NonZero::NonZero(const Output<Node>& arg, const std::string& index_element_type)
     : Op({arg})
+    , m_index_element_type(EnumNames<element::Type_t>::as_enum(index_element_type))
+{
+    constructor_validate_and_infer_types();
+}
+
+op::v3::NonZero::NonZero(const Output<Node>& arg, const element::Type& index_element_type)
+    : Op({arg})
+    , m_index_element_type(index_element_type)
```
lazarevevgeny, line 34: The attribute is called "output_type" in the specification.

src/ngraph/op/non_zero.cpp
```diff
@@ -22,14 +22,23 @@ using namespace std;
 
 constexpr NodeTypeInfo op::v3::NonZero::type_info;
 
-op::v3::NonZero::NonZero(const Output<Node>& arg)
+op::v3::NonZero::NonZero(const Output<Node>& arg, const std::string& index_element_type)
```
jane-intel, line 25: Spec says, output_type is optional with default i64 value -- we can avoid massive code changes.

src/ngraph/op/non_zero.cpp
Ewa21: Done

src/ngraph/op/non_zero.cpp
Ewa21: Done

PR #4579, opened 2020-04-15
Add support for v3 TopK operator

tsocha:
Changes:
1. Expand opset2_tbl in opset3_tbl
2. Add a new v3 version of TopK operator
3. Make v1::TopK::Mode be visible for v3::TopK as v1::TopKMode in attr_types.hpp

Changes in v1::TopK..v3::TopK (spec):
+ +attribute `output_indices_type`
+ ~K input: can have any integer type.


src/ngraph/op/topk.hpp
```diff
@@ -119,12 +119,7 @@ namespace ngraph
             {
             public:
                 using SortType = TopKSortType;
-
-                enum class Mode
-                {
-                    MAX,
-                    MIN
-                };
+                using Mode = v1::TopKMode;
```
lazarevevgeny, line 122: Is this a backward compatible change?

src/ngraph/op/topk.cpp
```diff
@@ -443,21 +443,218 @@ void op::v1::TopK::set_k(size_t k)
...
+                   const Output<Node>& k,
+                   const int64_t axis,
+                   const std::string& mode,
+                   const std::string& sort,
+                   const element::Type& output_indices_type)
+    : Op{{data, k}}
+    , m_axis{axis}
+    , m_normalized_axis{0}
+    , m_mode{mode_from_string(mode)}
+    , m_sort{sort_type_from_string(sort)}
```
tomdol, line 459: You can use this in the constructor instead https://github.com/NervanaSystems/ngraph/pull/4432/files#diff-b358c2a0e091411efef86dea664fb6cdR37 since you have a template specialization for the enum already

src/ngraph/op/topk.hpp
tsocha: Since visibility and values of this enum are the same it's compatible in my opinion.

src/ngraph/op/topk.cpp
tsocha: Done

tsocha:
> 
> 
> Overall looks good... since it is in 99% the same as v1 version. Personally I really don't like it. Maybe we should create some base class with common functionality?
> 
> It would be very useful If you could note in PR description what has changed in operator specification.

I would create a base class for it but this operator is too simple for that IMO.
The main problem was that v1::TopK is de-facto a v3::TopK with different name of attribute(which is absent in v1::TopK spec).

src/ngraph/op/topk.cpp
```diff
@@ -443,21 +443,197 @@ void op::v1::TopK::set_k(size_t k)
         op::Constant::create(element::i64, Shape{}, {k})->output(0));
 }
 
-namespace ngraph
+// v3 version starts
+constexpr NodeTypeInfo op::v3::TopK::type_info;
+
+op::v3::TopK::TopK(const Output<Node>& data,
```
lazarevevgeny, line 449: Since the TopK op already supported the output indices type I will instead update the specification and use  'index_element_type' as an attribute name. Then let's use inheritance to create a version3 op. This will significantly reduce the amount of code.

lazarevevgeny: The [specification](https://gitlab-icv.inn.intel.com/inference-engine/dldt/merge_requests/7010/) has just been updated.

tsocha: Done.

diyessi:
@tsocha see #4601

src/ngraph/op/topk.hpp
```diff
@@ -196,34 +191,60 @@ namespace ngraph
...
+                                                         const element::Type& k_element_type) const;
 
                 template <typename T>
                 size_t validate_and_get_k(const std::shared_ptr<op::Constant>& k_constant) const;
             };
         } // namespace v1
 
-        using v0::TopK;
-    } // op
+        namespace v3
```
mbencer, line 202: In "Adding a new Op" instruction there is: "If the opset3 op is a new version of an existing op, put the new definition earlier in the file, so that you see the current version of the op first". In this case `TopK::v1` foward declaration will be needed

src/ngraph/op/topk.cpp
```diff
@@ -443,21 +421,75 @@ void op::v1::TopK::set_k(size_t k)
...
+    visitor.on_attribute("sort", m_sort);
+    visitor.on_attribute("index_element_type", m_index_element_type);
+    return true;
+}
+
+void op::v3::TopK::validate_and_infer_types()
+{
+    NODE_VALIDATION_CHECK(this,
+                          get_input_element_type(1).is_integral_number(),
+                          "K input have to be an integer type. Which not match the provided one:",
```
postrational, line 462:
```suggestion
                          "K input has to be an integer type, which does match the provided one:",
```

postrational:
Please run the OV test suite on this PR and add `Fully Reviewed` label when they pass.

mbencer:
OV tests: merge_requests/7345

PR #4583, opened 2020-04-16
Interpreter/Constant folding with same kernels

diyessi:
- [x] Add interpreter based on Evaluator with universal op handler
- [x] Copy interpreter changes into GCPU
- [x] Allocation on demand
- [x] Lambda callable on tensors


src/ngraph/node.hpp
```diff
@@ -188,7 +188,7 @@ namespace ngraph
         virtual bool supports_decompose() const { return false; }
         /// \brief Can replace a node with a constant during constant folding.
         /// \returns vector of outputs to replace node outputs. Empty outputs will not be replaced.
-        virtual OutputVector constant_fold() { return {}; }
+        virtual OutputVector constant_fold(const OutputVector& input_values) { return {}; }
```
slyalin, line 191: @diyessi, as we are moving away from using a local graph context for a node when evaluating value (completely or partially), I would suggest to rename this function to something like partial_infer or partial_evaluate or just evaluate (because partial becomes everywhere with more and more dynamic stuff involved). Probably a combination with `reference` word is also required to avoid confusion with real performance-oriented inference/evaluation in a HW plug-in.

Also I would suggest to keep an original method which will be a helper for case if we really want to evaluate a node based on the in-graph arguments. It will just call this new method with explicit `input_values` (as you already did later in this PR).

src/ngraph/validation_util.hpp
```diff
@@ -185,6 +185,14 @@ namespace ngraph
     /// \return (true, max_value) if can be determined, or (false, 0) if not.
     NGRAPH_API std::pair<bool, int64_t> maximum_value(const Output<Node>& value);
 
+    /// \brief Evaluates values, treating values in value_map as already computed. value_map is
+    /// updated.
+    /// \param value_map Key is RawNodeOutput in graph, value is the computed value. Updated by the
+    /// function.
+    /// \param values Root set of values to try to compute
+    NGRAPH_API void evaluate_nodes(std::map<RawNodeOutput, Output<Node>>& value_map,
```
slyalin, line 193: ngraph::Node has `constant_fold` (or better `evaluate`) method. Why not to attach the evaluation capability to ngraph::Lambda or ngraph::Function directly as a class method?

src/ngraph/node.hpp
diyessi: `evaluate` seems fine to me. The actual "reference kernel" tends to be a separate template so that it can deal with all the needed element types in one place, so the `evaluate` method would be glue around instantiations of the reference kernel. That also allows a reference kernel to be used in more than one place, since it's just a template definition that executes on void*s.

src/ngraph/validation_util.hpp
diyessi: If you do that, then you can only use it on the entire graph. This way, there could be a method on Lambda that calls the generic evaluator, but it's still available for use in other transformations.
Note that custom ops can participate in evaluation.

test/backend/shape_of.in.cpp
```diff
@@ -88,13 +140,39 @@ NGRAPH_TEST(${BACKEND_NAME}, shape_of_matrix)
...
+
+    auto handle = backend->compile(f);
+    handle->call_with_validate({result64, result32}, {a});
+    vector<int64_t> expected64{2, 4};
+    vector<int32_t> expected32{2, 4};
+    EXPECT_EQ(expected64, read_vector<int64_t>(result64));
+    EXPECT_EQ(expected32, read_vector<int32_t>(result32));
+}
+
+NGRAPH_TEST(${BACKEND_NAME}, shape_of_5d_v0)
```
slyalin, line 169: @diyessi, I hope you will remove all such stuff before final merge. Is it for experiment only? opset3::ShapeOf should be committed and reviewed in a separate PR. Here we are massaging evaluate method and its automatic calling for all new ops in constant_folding pass and interpreter, right? And if we have all ingredients (except solution for separate/conditional compilation of reference kernel) I would like to finalize it and commit.

src/ngraph/validation_util.hpp
slyalin: We discussed it during our meeting. If it is easy to do, please provide a method. If it requires more time, please postpone to some other PR. I want to finalize this PR with core stuff ASAP.

src/ngraph/op/op_version_tbl.hpp
```diff
@@ -219,6 +219,7 @@ NGRAPH_OP(Select, ngraph::op::v1, 1)
 NGRAPH_OP(Selu, ngraph::op::v0, 0)
 NGRAPH_OP(Send, ngraph::op::v0, 0)
 NGRAPH_OP(ShapeOf, ngraph::op::v0, 0)
+NGRAPH_OP(ShapeOf, ngraph::op::v3, 3)
```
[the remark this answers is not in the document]
diyessi: It opset 2 with the ops superseded in 3 removed and the new ones on 3 added.

src/ngraph/validation_util.hpp
diyessi: Lambda is in now, but it won't work until all the ops in the lambda have evaluate methods.

src/ngraph/op/op_version_tbl.hpp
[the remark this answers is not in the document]
diyessi: This PR had the ShapeOf-3 PR in it to avoid merge issues. That is merged now so these are no longer present in this PR.

src/ngraph/opsets/opset3_tbl.hpp
```diff
@@ -19,8 +19,144 @@
...
+NGRAPH_OP(VariadicSplit, ngraph::op::v1)
+
+// Moved out of opset2, it was added to opset1 by mistake
+// NGRAPH_OP(Xor, ngraph::op::v0)
+
+// New operations added in opset2
+NGRAPH_OP(Gelu, ngraph::op::v0)
+NGRAPH_OP(BatchToSpace, ngraph::op::v1)
+NGRAPH_OP(SpaceToBatch, ngraph::op::v1)
+
```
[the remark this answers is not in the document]
diyessi: Part of ShapeOf-3 PR. No longer in this PR.

test/backend/shape_of.in.cpp
diyessi: More of the already merged ShapeOf3 PR.

src/ngraph/node.hpp
rkimballn1: reference kernels are in general too slow to be used for constant folding. An example is the reference Broadcast kernel which took 114 seconds on BERT Large. Replacing the reference kernel with a more optimized kernel reduced this time to 2.5 seconds. Broadcast is a simple example, there are other reference kernels that can't be used in constant folding due to performance problems.

diyessi: We'll want to use the opt kernels when they exist. They fall back on reference kernels for the non-opt cases anyways.

src/ngraph/evaluator_tensor.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+size_t ngraph::EvaluatorTensor::get_element_count()
+{
+    return shape_size(get_partial_shape().get_shape());
+}
+
+size_t ngraph::EvaluatorTensor::get_size_in_bytes()
+{
+    return get_element_type().size() * get_element_count();
+}
```
silee2, line 60: No newline at end of file

src/ngraph/pass/constant_folding_binary.cpp
```diff
@@ -561,20 +535,20 @@ shared_ptr<op::Constant> fold_constant_binary_helper(shared_ptr<op::Constant> a,
...
+            is_type<op::v0::Greater>(n) || is_type<op::v1::Greater>(n) ||
+            is_type<op::v0::GreaterEq>(n) || is_type<op::v1::GreaterEqual>(n) ||
+            is_type<op::v0::Less>(n) || is_type<op::v1::Less>(n) || is_type<op::v0::LessEq>(n) ||
+            is_type<op::v1::LessEqual>(n) || is_type<op::v0::Maximum>(n) ||
+            is_type<op::v1::Maximum>(n) || is_type<op::v0::Minimum>(n) ||
+            is_type<op::v1::Minimum>(n) || is_type<op::v0::And>(n) ||
+            is_type<op::v1::LogicalAnd>(n) || is_type<op::v0::Or>(n) ||
+            is_type<op::v1::LogicalOr>(n) || is_type<op::v0::Xor>(n) ||
+            is_type<op::v1::LogicalXor>(n) || is_type<op::v0::Subtract>(n) ||
+            is_type<op::v1::Subtract>(n));
```
slyalin, line 551: Just a note, I understand this was already in the code: Why do we need this function? We will face the same list when will do real constant folding in `fold_constant_binary_arithmetic`.

src/ngraph/op/add.cpp
```diff
@@ -64,6 +65,47 @@ shared_ptr<Node> ngraph::operator+(const Output<Node>& arg0, const Output<Node>&
...
+    {
+        return try_evaluate_add<element::Type_t::i8>(arg0, arg1, out, broadcast_spec) ||
+               try_evaluate_add<element::Type_t::i16>(arg0, arg1, out, broadcast_spec) ||
+               try_evaluate_add<element::Type_t::i32>(arg0, arg1, out, broadcast_spec) ||
+               try_evaluate_add<element::Type_t::i64>(arg0, arg1, out, broadcast_spec) ||
+               try_evaluate_add<element::Type_t::u8>(arg0, arg1, out, broadcast_spec) ||
+               try_evaluate_add<element::Type_t::u16>(arg0, arg1, out, broadcast_spec) ||
+               try_evaluate_add<element::Type_t::u32>(arg0, arg1, out, broadcast_spec) ||
+               try_evaluate_add<element::Type_t::u64>(arg0, arg1, out, broadcast_spec) ||
+               try_evaluate_add<element::Type_t::f32>(arg0, arg1, out, broadcast_spec) ||
```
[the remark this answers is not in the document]
diyessi: If you want to adjust the ordering, you can reorder. If you want to save space, you can add compiler conditionals between lines.

src/ngraph/op/shape_of.cpp
```diff
@@ -112,15 +137,22 @@ namespace
...
+bool op::v3::ShapeOf::evaluate(const EvaluatorTensorVector& output_values,
+                               const EvaluatorTensorVector& input_values)
+{
+    return evaluate_shape_of(output_values[0], input_values[0]);
+}
+
+bool op::v3::ShapeOf::constant_fold(OutputVector& output_values, const OutputVector& input_values)
 {
-    return constant_fold_shape_of(shared_from_this(), m_output_type, m_is_foldable);
+    return constant_fold_shape_of(this, output_values[0], input_values[0], m_is_foldable);
```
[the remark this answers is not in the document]
slyalin: More I think, the more I tend to agree with your design. But let's view it as follow: `evaluate` is a real reference kernel (for education purposes and validation), while `constant_fold` is for real-time execution in the application while graph transformation. Does it fit your intentions? Does it look fine with you?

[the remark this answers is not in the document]
diyessi: Note that I have not changed what the constant folder for ShapeOf does, just changed where it does it.

src/ngraph/pass/constant_folding_binary.cpp
diyessi: I wouldn't have written it this way, but it is what it is until it's replace with the new way.

src/ngraph/node.cpp
```diff
@@ -1102,3 +1104,50 @@ vector<Output<const Node>> Node::outputs() const
...
+    vector<op::Constant::ConstantEvaluatorTensorPtr> output_constant_tensors;
+    EvaluatorTensorVector output_tensors;
+    OutputVector output_constants;
+    for (auto output : outputs())
+    {
+        auto element_type = output.get_element_type();
+        auto partial_shape = output.get_partial_shape();
+        if (partial_shape.is_dynamic() || element_type.is_dynamic())
+        {
+            return false;
```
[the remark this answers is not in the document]
diyessi: You might be right, but it's also highly unlikely you will be constant-folding something that has dynamic shape outputs.

src/ngraph/op/shape_of.cpp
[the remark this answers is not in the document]
diyessi: One difference with constant-fold, exemplified by the one in ShapeOf, is that it might do partial evaluation, i.e. replace a node with a sub-graph that is simpler to evaluate at runtime. Evaluate has to come up with actual tensors.

src/ngraph/op/add.cpp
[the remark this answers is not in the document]
slyalin: Nitpick: convert to i64/f64 do the addition and convert back the result? Unlikely gives anything for Add, but is it something that can be applied as a pattern?

src/ngraph/node.cpp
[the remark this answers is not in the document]
slyalin: @diyessi, please fix it. If you call constant_fold without really getting values from the graph it can be anything in the outputs.

src/ngraph/node.cpp
```diff
@@ -1102,3 +1104,45 @@ vector<Output<const Node>> Node::outputs() const
 
     return result;
 }
+
+bool Node::evaluate(const EvaluatorTensorVector& output_values,
+                    const EvaluatorTensorVector& input_values)
```
slyalin, line 1109: I disagree with this signature. This function should accept a vector of inputs and returns (as a return value or as non-constant argument by a reference) a vector of outputs without any pre-requisites on calling validate_and_infer_types with constants. It does't look solid.

@diyessi if you don't have any quick options, please commit as-is we will find better solution (if any) later.

diyessi:
Will merge with missing reviewers when CI finishes at request from @slyalin

rkimballn1:
I did not get a chance to review this PR and I do not approve of it. It should be changed.

PR #4593, opened 2020-04-17
Removed axis normalization for Concat op in Onnx importer. 

Ewa21:
Since Concat op can handle negative axes, the axis normalization is removed from ONNX importer. 
I have provided a test for this change in this PR: https://github.com/NervanaSystems/ngraph/pull/4453


tomdol:
Since v0::Concat can already handle negative axes, we don't need a new v3 version of this op. The static-input-rank-only normalization was removed from the ONNX importer in this PR.

Please check the spec MR - it was updated after those findings.

PR #4596, opened 2020-04-17
simplify gather->shapeof into shapeof->gather for 0D indices

ashokei:


src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -419,6 +432,32 @@ static bool simplify_log(shared_ptr<Node> n)
...
+static bool simplify_gather_shapeof(shared_ptr<Node> node)
+{
+    auto shapeof = as_type_ptr<op::v0::ShapeOf>(node);
+    auto gather = as_type_ptr<op::v1::Gather>(shapeof->input_value(0).get_node_shared_ptr());
+    if (!gather)
+    {
+        return false;
+    }
+    auto indices = as_type_ptr<op::Constant>(gather->input_value(1).get_node_shared_ptr());
+    if (!indices || indices->get_shape() != Shape{} ||
```
jane-intel, line 445: 1. We could support `Shape{1}` case too.
2. Why do we require indices to be constant and have static shape? Static rank check is enough, isn't it?
3. Why we use numeric_limits instead of AXIS_NOT_SET_VALUE?

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -419,6 +432,32 @@ static bool simplify_log(shared_ptr<Node> n)
...
+        gather->get_axis() == std::numeric_limits<int64_t>::max())
+    {
+        NGRAPH_DEBUG << gather << " cannot simplify shapeof->gather";
+        return false;
+    }
+    auto new_shapeof = make_shared<op::v0::ShapeOf>(gather->input_value(0).get_node_shared_ptr());
+    auto new_axis = op::Constant::create<int64_t>(element::i64, Shape{}, {0});
+    std::vector<int64_t> vi(gather->get_shape().size() + 1);
```
jane-intel, line 453: Why we have static shape restrictions?

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -419,6 +432,32 @@ static bool simplify_log(shared_ptr<Node> n)
...
+    std::iota(vi.begin(), vi.end(), 0);
+    vi.erase(vi.begin() + gather->get_axis());
```
jane-intel, line 455: After dynamic shape enablement we should check for negative axis.

src/ngraph/pass/algebraic_simplification.cpp
ashokei: only 0D indices ( Shape{} ) result in axis dimension being removed.  Will try make other changes.

src/ngraph/pass/algebraic_simplification.cpp
ashokei: if input rank is not dynamic, get_axis will adjust negative axis.

src/ngraph/pass/algebraic_simplification.cpp
ashokei: will make change

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -419,6 +432,35 @@ static bool simplify_log(shared_ptr<Node> n)
...
+    if (!gather)
+    {
+        return false;
+    }
+    auto indices = gather->input_value(1).get_node_shared_ptr();
+    auto gather_in_ps = gather->get_input_partial_shape(0);
+    auto indices_ps = indices->get_output_partial_shape(0);
+    auto axis = gather->get_axis();
+    if (gather_in_ps.rank().is_dynamic() || indices_ps.rank().is_dynamic() ||
+        indices_ps.rank().get_length() != 0 || axis == op::v1::Gather::AXIS_NOT_SET_VALUE)
```
jane-intel, line 449: shouldn't we transform `Gather->ShapeOf` to `Concat(ShapeOf(data)[:axis], ShapeOf(indices) + ShapeOf(data)[axis + 1:]`? and as an exceptional case leave only `indices.shape`? does it sounds sane?

ashokei: this would be `Concat(Gather(ShapeOf(data),[:axis]), ShapeOf(indices) + Gather(ShapeOf(data),[axis + 1:])` right ? resulting in several ops. If this is needed, we could do it.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -487,6 +488,66 @@ static bool simplify_log(shared_ptr<Node> n)
...
+    if (gather_in_rank.is_dynamic() || indices_rank.is_dynamic() ||
+        axis == op::v1::Gather::AXIS_NOT_SET_VALUE)
+    {
+        NGRAPH_DEBUG << gather << " cannot simplify gather->shapeof";
+        return false;
+    }
+
+    auto zero_axis = op::Constant::create<int64_t>(element::i64, Shape{}, {0});
+    shared_ptr<Node> replace_node;
+    auto new_shapeof = make_shared<op::v0::ShapeOf>(gather->input_value(0).get_node_shared_ptr());
```
GlebKazantaev, line 513: As we are going to use this transformation in our common optimizations pass we need to be sure that all operations produced by this transformation belongs to opset1 or higher. 
So using v0/v1/vXXX namespace complicates transformation review and I can't 100% say that all this ops are from opset1 or higher. That's why opsetXXX namespace was introduced. In case of new transformation we use opset3 now.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -614,15 +681,18 @@ static bool simplify_reduction(shared_ptr<Node> n)
...
-         {op::Concat::type_info, simplify_concat},
+        {{op::v0::Add::type_info, simplify_add},
+         {op::v0::Multiply::type_info, simplify_multiply},
          {op::v1::Gather::type_info, simplify_gather},
-         {op::Sum::type_info,
-          function<bool(shared_ptr<Node>)>{simplify_reduction<op::Sum, get_sum_constant>}},
-         {op::Product::type_info,
-          function<bool(shared_ptr<Node>)>{simplify_reduction<op::Product, get_prod_constant>}},
-         {op::Log::type_info, simplify_log}});
+         {op::v1::Add::type_info, simplify_add},
```
GlebKazantaev, line 687: We cant apply this transformation for v1::Add and v1::Multiply.
You need to update simplify_add/simplfy_mul to use remove_node_update_name otherwise it 100% will brake some existing models tests.

ashokei: sure, made these changes.

src/ngraph/pass/algebraic_simplification.cpp
ashokei: technically opset3::xx will resolve to same underlying op, anyways i made the changes to use opset3 namespace.

jane-intel:
@GlebKazantaev please review and estimate how much time will it take to rewrite internal OV transformations preserving names to make this merge happen.

ashokei:
@jane-intel @GlebKazantaev this PR has topk unittest and @diyessi has a fix for it, we are testing this as MR on OV repo.

PR #4597, opened 2020-04-17
Add output_type attribute to NonMaxSuppression for OV opset3

adstraw:


src/ngraph/op/non_max_suppression.hpp
```diff
@@ -88,6 +90,7 @@ namespace ngraph
             protected:
                 BoxEncodingType m_box_encoding = BoxEncodingType::CORNER;
                 bool m_sort_result_descending = true;
+                std::string m_output_type = "i64";
```
tomdol, line 93: LGTM overall but I'd suggest to adjust the implementation in the following way:
- switch the member from `string` to `element::Type`
- implement 2 constructors, one that takes an enum and one that takes a string
- in the second constructor use a helper that converts a string to an enum https://github.com/NervanaSystems/ngraph/pull/4432/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.hpp#diff-b358c2a0e091411efef86dea664fb6cdR37
- this helper automatically checks if the string's value passed to the constructor is allowed

src/ngraph/op/non_max_suppression.cpp
```diff
@@ -30,10 +30,12 @@ op::v1::NonMaxSuppression::NonMaxSuppression(
     const Output<Node>& iou_threshold,
     const Output<Node>& score_threshold,
     const op::v1::NonMaxSuppression::BoxEncodingType box_encoding,
-    const bool sort_result_descending)
+    const bool sort_result_descending,
```
lazarevevgeny, line 33: Why do you modify the existing operation? We need to **add** new version of the operation! I recommend to use inheritance since the core functionality of the operation will be re-used.

src/ngraph/op/non_max_suppression.hpp
adstraw: Switch from `string` to `element::Type` is done.  

I did not implement the 2nd constructor which converts from `string` to `enum` as you suggested.  I am following the `ShapeOf` example mentioned by @diyessi above which has no such constructor.  Let me know if you think it is required.

src/ngraph/op/non_max_suppression.cpp
adstraw: My opinion is that a brand new operation is overkill in this case.  The simple addition of a default parameter for `output_type` to the v1 op should suffice.

diyessi: It needs to be a new op. Any backend already implementing the old op will not know about the new attribute and do the wrong thing.

src/ngraph/op/non_max_suppression.hpp
tomdol: This double constructor approach(one with steing and one with enum) comes from the times when we were working on opset1. Now with the visitor API it might not be needed anymore. @diyessi @lazarevevgeny any thoughts?

lazarevevgeny: Yes, with the visitor API the constructor with string attribute does not look like very important. I don't have strong preference in this case.

src/ngraph/op/non_max_suppression.hpp
```diff
@@ -92,8 +94,95 @@ namespace ngraph
...
+            class NGRAPH_API NonMaxSuppression : public Op
+            {
+            public:
+                enum class BoxEncodingType
+                {
+                    CORNER,
+                    CENTER
+                };
+
+                static constexpr NodeTypeInfo type_info{"NonMaxSuppression", 1};
```
diyessi, line 112: There will be trouble in River City if you don't change the version to 3.

adstraw: Missed that one.  Will correct.

PR #4600, opened 2020-04-20
Add Broadcast:v3 op

mbencer:
- `Broadcast:v3` introduces support for `BIDIRECTIONAL` broadcast.
- Constant folding (reference implementation) + using `Broadcast:v3` in ONNX `Exapnd` will be implemented in separate PR


src/ngraph/node.cpp
```diff
@@ -919,6 +919,13 @@ std::tuple<element::Type, PartialShape>
                     PartialShape::broadcast_merge_into(pshape, get_input_partial_shape(i), autob),
                     "Argument shapes are inconsistent.");
             }
+            else if (autob.m_type == op::AutoBroadcastType::BIDIRECTIONAL)
+            {
+                NODE_VALIDATION_CHECK(
+                    this,
+                    false,
+                    "BIDIRECTIONAL auto broadcast is not supported for elementwise operations");
```
mbencer, line 927: Numpy elementwise broadcasting is de facto bidederectional. I am not sure if the exception is needed.
@diyessi What do you think about it?

src/ngraph/op/broadcast.hpp
```diff
@@ -24,92 +24,70 @@ namespace ngraph
...
+                /// \param arg            The input tensor to be broadcast.
+                /// \param target_shape   The shape of the output tensor.
+                /// \param broadcast_spec Broadcast specification to use for determining broadcast
+                ///                       axes
+                Broadcast(const Output<Node>& arg,
+                          const Output<Node>& target_shape,
+                          const AutoBroadcastSpec& broadcast_spec =
+                              AutoBroadcastSpec(AutoBroadcastType::NUMPY));
                 bool visit_attributes(AttributeVisitor& visitor) override;
+                size_t get_version() const override { return 1; }
```
yimeisun123, line 68: What is the reason to override the Node class implementation for this method? If use Node's implementation, this call will return 3. But here the code returns 1, is this intensional?

mbencer: right, thanks it was copy-paste artifact ;)

src/ngraph/runtime/dynamic/dynamic_backend.cpp
```diff
@@ -93,7 +93,7 @@ bool is_dynamic_op(const std::shared_ptr<Node>& op)
            is_type<op::v1::ConvolutionBackpropData>(op) ||
            is_type<op::v1::ConvolutionBackpropFilters>(op) ||
            is_type<op::v1::AvgPoolBackprop>(op) || is_type<op::v1::Broadcast>(op) ||
-           is_type<op::v1::GenerateMask>(op);
+           is_type<op::v3::Broadcast>(op) || is_type<op::v1::GenerateMask>(op);
```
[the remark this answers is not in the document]
diyessi: dynamic backend intercepts a transformer for functions with dynamic ops and runs specialize function on them and this test makes sure it has gotten rid of all the dynamic behavior. There's a cache from param shapes to specialized functions.

src/ngraph/node.cpp
tsocha: If this check is unnecessary I would vote for removing it.
Treating  BIDIRECTIONAL type as a numpy one in this case should be safe IMO.

src/ngraph/op/broadcast.hpp
```diff
@@ -24,92 +24,70 @@ namespace ngraph
 {
     namespace op
     {
-        namespace v0
+        namespace v3
```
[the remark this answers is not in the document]
tsocha: This approach is working for v3::TopK so far.
It would be nice to try it here too.
#4579

src/ngraph/op/util/attr_types.hpp
```diff
@@ -166,7 +166,8 @@ namespace ngraph
             NONE = 0,
             EXPLICIT = NONE,
             NUMPY,
-            PDPD
+            PDPD,
+            BIDIRECTIONAL
```
tsocha, line 170: What will happen if someone will use BIDIRECTIONAL mode for v1::Broadcast?
We should probably detect this situation to avoid undocumented behavior.

test/type_prop/broadcast.cpp
```diff
@@ -465,3 +472,172 @@ TEST(type_prop, broadcast_v1_axes_et_wrong)
...
+    const auto broadcast_spec = op::AutoBroadcastType::BIDIRECTIONAL;
+
+    const auto broadcast_v3 = make_shared<op::v3::Broadcast>(arg, shape, broadcast_spec);
+
+    ASSERT_EQ(broadcast_v3->get_element_type(), element::f32);
+    ASSERT_EQ(broadcast_v3->get_shape(), (Shape{3, 3, 3}));
+    ASSERT_EQ(broadcast_v3->get_broadcast_axes(), (make_pair<bool, AxisSet>(true, AxisSet{0, 2})));
+}
+
+TEST(type_prop, broadcast_v3_shape_6_type_infer)
```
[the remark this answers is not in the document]
mbencer: done, negative tests introduced

src/ngraph/op/broadcast.hpp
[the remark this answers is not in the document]
mbencer: Inheritance approach was introduced. However, inheritance from v1 is problematic in this case. I introduced `BroadcastBase` which contains common code

src/ngraph/op/util/attr_types.hpp
mbencer: done, check added

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -0,0 +1,283 @@
...
+        const auto axes_mapping_constant =
+            as_type_ptr<op::v0::Constant>(input_value(2).get_node_shared_ptr());
+        if (get_input_partial_shape(1).is_static() && axes_mapping_constant)
+        {
+            auto target_shape = get_input_shape(1);
+            NGRAPH_CHECK(target_shape.size() == 1);
+            auto axes_mapping_val = axes_mapping_constant->get_axis_vector_val();
+
+            std::vector<size_t> axes(target_shape[0]);
+            std::iota(axes.begin(), axes.end(), 0);
```
lazarevevgeny, line 223: I got a compilation failure on mac:

/Users/elazarev/Documents/git/ie_dldt/ngraph/src/ngraph/op/util/broadcast_base.cpp:223:18: error: no member named 'iota' in namespace 'std'
            std::iota(axes.begin(), axes.end(), 0);

src/ngraph/op/broadcast.hpp
```diff
@@ -19,11 +19,111 @@
...
+                /// \brief Constructs a broadcast operation.
+                Broadcast() = default;
+                /// \brief Constructs a broadcast operation.
+                ///
+                /// \param arg            The input tensor to be broadcast.
+                /// \param target_shape   The shape of the output tensor.
+                /// \param axes_mapping   The axis positions (0-based) in the result that correspond
+                ///                       to input axes. 'Arg' tensor is broadcast along the
+                ///                       remaining
+                ///                       axes.
```
arogowie-intel, line 97: Formatting:
```suggestion
                ///                       remaining axes.```

File: src/ngraph/op/util/broadcast_base.cpp [hunk 9]
```diff
@@ -0,0 +1,283 @@
...
+#include "ngraph/op/constant.hpp"
+#include "ngraph/op/sum.hpp"
+#include "ngraph/partial_shape.hpp"
+
+using namespace std;
+using namespace ngraph;
+/*
+op::util::BroadcastBase::BroadcastBase()
+{
+}*/
```
arogowie-intel, line 29: Please remove commented code.

test/type_prop/broadcast.cpp
```diff
@@ -199,39 +199,68 @@ TEST(type_prop, broadcast_partial_rank_static_dynamic_shape_mismatch_wrong_size)
     }
 }
 
-TEST(type_prop, broadcast_v1_numpy)
+TEST(type_prop, broadcast_v1_biderectional_not_supported)
```
arogowie-intel, line 202: Typo:
```suggestion
TEST(type_prop, broadcast_v1_bidirectional_not_supported)
```

src/ngraph/op/util/attr_types.hpp
lazarevevgeny: Our original specification for Broadcast-1 was misleading:

 "If `mode` is set to `numpy`, it     means that the standard one-directional numpy broadcasting rules are applied. They are similar to rules that applied in all binary element-wise operations in case when `auto_broadcasting` attribute is set to `numpy`, and are similar to rules described at [here](https:/    /docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules), when only one-directional broadcasting is applied: input tensor `data` is broadcasted to `target_shape` but not vice-versa."

I think that both UNIDIRECTIONAL and BIDIRECTIONAL will have the same level of confusion for people who are familiar with broadcasting rules.

What I currently see in the nGraph op code is that the broadcasting with NUMPY mode is a kind of mix of uni- and bi-directional broadcasting. If the length of the target_shape is less than the length of then input shape we raise an exception. At the same time when we compare respective dimensions of the input tensor and the target shape we check that they are equal or one of them is equal to 1. This is the behaviour of bi-directional broadcasting. This means that

lazarevevgeny: Oh, I now see the awkwardness of my comment: we cannot broadcast the tensor to smaller shape.

lazarevevgeny: Now I see that the shape infer function of Broadcast with NUMPY mode works not a described in the documentation. According to the specification if the input tensor shape is (10,1) and target shape is (1,20) then error should be produced, but the implementation will produce output shape (10,20). But the output shape (10,20) should be produced only for BIDIRECTIONAL model.

@arogowie-intel, do you agree?

mbencer: @arogowie-intel To be clear, as I understand numpy broadcasting is bidirectional but only for element-wise operations, Numpy mode for Broadcast operator is one-directional. Probably more clear will be using separate `AutoBroadcastType` enum for element-wise operations and for Broadcast op. @lazarevevgeny What do you think about it?

src/ngraph/pass/implicit_broadcast_elimination.cpp
```diff
@@ -60,6 +60,10 @@ NodeVector ngraph::pass::explicit_broadcast(std::shared_ptr<Node>& node)
         {
             rc = as_node_vector(builder::pdpd_broadcast(node->input_values(), autob.m_axis));
         }
+        else if (autob.m_type == op::AutoBroadcastType::BIDIRECTIONAL)
+        {
+            throw ngraph_error("BIDIRECTIONAL mode is not supported");
```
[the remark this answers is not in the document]
mbencer: As you said it is misleading with `NUMPY`, but anyway done

src/ngraph/runtime/reference/autobroadcast_binop.hpp
```diff
@@ -218,7 +218,14 @@ namespace ngraph
                                 elementwise_functor(arg0[arg0_transform.index(output_coord)],
                                                     arg1[arg1_transform.index(arg1_coord)]);
                         }
+                        break;
                     }
+                case op::AutoBroadcastType::BIDIRECTIONAL:
+                {
+                    throw std::runtime_error(
```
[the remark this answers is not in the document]
mbencer: It is discussed now

src/ngraph/runtime/reference/autobroadcast_binop.hpp
```diff
@@ -423,6 +430,13 @@ namespace ngraph
                                                 arg1[arg1_transform.index(output_coord)],
                                                 arg2[arg2_transform.index(arg2_coord)]);
                     }
+                    break;
+                }
+                case op::AutoBroadcastType::BIDIRECTIONAL:
+                {
+                    throw std::runtime_error(
```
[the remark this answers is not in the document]
mbencer: as above

src/ngraph/op/util/broadcast_base.cpp
mbencer: fixed

src/ngraph/op/util/attr_types.hpp
lazarevevgeny: Yes, this makes sence.

lazarevevgeny: But we should keep backward compatibility for the Broadcast op.

lazarevevgeny: And emphasise once again the difference that is also not covered with NUMPY mode currently is that the target shape rank for BIDIRECTIONAL model can be less than rank of input tensor.

arogowie-intel: Probably these all discrepancies comes from the fact, as Mateusz already mentioned, originally nGraph autobroadcasting rules where developed for element-wise operations. @lazarevevgeny You're right, current nGraph NUMPY `AutoBroadcastType` works like bidirectional, with limitations you pointed. I guess that separating modes for autobroadcasting (element-wise ops) from `Broadcast` modes may be good solution.

mbencer: @diyessi any objections to separate `AutoBroadcastType` for `Broadcast` and element-wise operations?

gaurides: if current mode NUMPY works as bidirectional, can we treat that as bidirectional and add a separate UNIDIRECTIONAL mode? 
Also, by unidirectional, do you mean the numpy method `broadcast_to()` ?

mbencer: @gaurides
To sum up. As I understand:
The main difference between BIDIRECTIONAL and Numpy for broadcast op is that for BIDIRECTIONAL target_shape.size() can be less than input.size()
Please look to `Expand` spec which is exactly broadcast in BIDIRECTIONAL mode (https://github.com/onnx/onnx/blob/master/docs/Operators.md#Expand).
1. Numpy is de facto BIDIRECTIONAL but only for element-wise operations
2. Numpy used in Broadcast operator is not BIDIRECTIONAL according to specification
3. Numpy `broadcast_to` is not BIDIRECTIONAL, `numpy.array(input) * numpy.ones(shape)` is BIDIRECTIONAL

mbencer: I think that introduction `UNIDIRECTIONAL` ensure lack of `Broadcast:v3`-`Broadcast:v1` backward compatibility of need of changing `Broadcast:v1` spec

mbencer: @arogowie-intel @lazarevevgeny I've introduced separate type class for `Broadcast;v3`, please to re-review

src/ngraph/op/util/attr_types.hpp
```diff
@@ -298,4 +326,52 @@ namespace ngraph
         static constexpr DiscreteTypeInfo type_info{"AttributeAdapter<op::AutoBroadcastSpec>", 0};
         const DiscreteTypeInfo& get_type_info() const override { return type_info; }
     };
+
+    namespace op
+    {
+        /// \brief Implicit broadcast specification
+        struct NGRAPH_API BroadcastModeSpec
```
mbencer, line 333: @diyessi @lazarevevgeny  Can we break `PDPD` support for `Broadcast:v3`? (it will significantly simplify the code but break backward compatibility)

lazarevevgeny: We should not do it. I understand that the code complexity may increase, but we must keep backward compatibility

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -315,6 +315,8 @@ static shared_ptr<op::Constant> get_constant(shared_ptr<Node> op)
 {
     set<Node::type_info_t> nomath = {op::v0::Broadcast::type_info,
                                      op::v0::Reshape::type_info,
+                                     op::v1::Broadcast::type_info,
+                                     op::v1::Broadcast::type_info,
```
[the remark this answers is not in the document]
mbencer: right, redundant line removed

src/ngraph/node.cpp
diyessi: If numpy and bidirectional are the same, there should only be one.

src/ngraph/op/util/attr_types.hpp
lazarevevgeny: Done

mbencer:
OV tests: merge_requests/7343

src/ngraph/op/util/attr_types.hpp
```diff
@@ -187,6 +187,34 @@ namespace ngraph
         const DiscreteTypeInfo& get_type_info() const override { return type_info; }
     };
 
+    namespace op
+    {
+        enum class BroadcastType
+        {
+            NONE,
```
arogowie-intel, line 194: Cherry on top of this cake would be at least a small note, that the meanings of all those values are similar to ones in `AutoBroadcastType`.

mbencer: done, location changed, documentation added

src/ngraph/op/util/attr_types.hpp
```diff
@@ -166,12 +166,37 @@ namespace ngraph
...
-            PDPD
+            PDPD,
         };
 
         NGRAPH_API
         std::ostream& operator<<(std::ostream& s, const AutoBroadcastType& type);
     }
+    namespace op
+    {
+        /// \brief BroadcastType specifies rules used for mapping of input tensor axes to output
```
gaurides, line 177: It would be nice to have some notes here for the need to have 2 similar enums, how they differ and where to use? Else it will be confusing to a user. There's a lot of discussion in the PR but once it is merged, it will not be easily accessible.

mbencer: I added short comment about it

PR #4601, opened 2020-04-20
Fix TopK output shape when k is not known

diyessi:
TopK validation left the "k" axis unchanged when k wasn't known


diyessi:
@tomdol I'm not sure how to make `k` a scalar in the python test. Could you fix it? Thanks.

PR #4602, opened 2020-04-20
add redundant reshape, squeeze, unsqueeze elimination

ashokei:


src/ngraph/pass/nop_elimination.cpp
```diff
@@ -55,7 +55,7 @@ static bool eliminate_pad(const std::shared_ptr<Node>& node)
 
     if (pad->get_input_shape(0) == pad->get_output_shape(0))
     {
-        return remove_node_update_name(node, node->get_argument(0));
+        return remove_node_update_name(node, node->input_value(0).get_node_shared_ptr());
```
diyessi, line 58: Up above, this is for `op::v0::Pad` instead of `op::v1::Pad` The caller of `eliminate_pad` is also the wrong version of `Pad`.

ashokei: made it generic, so this can be called for other cases.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -138,69 +106,89 @@ static bool eliminate_concat(const std::shared_ptr<Node>& node)
...
+    // eliminate redundant reshape, squeeze, or unsqueeze
+    if (as_type_ptr<op::v0::Squeeze>(input) || as_type_ptr<op::v0::Unsqueeze>(input) ||
+        as_type_ptr<op::v1::Reshape>(input))
+    {
+        auto shape = node->get_shape();
+        std::vector<int64_t> vi;
+        vi.assign(shape.begin(), shape.end());
+        auto pat = op::Constant::create<int64_t>(element::i64, Shape{vi.size()}, vi);
+        auto new_reshape =
+            make_shared<op::v1::Reshape>(input->input_value(0).get_node_shared_ptr(), pat, false);
```
jane-intel, line 131: Could we use `decompose_op` method here since `Squeeze` and `Unsqueeze` are fused ops?

PR #4603, opened 2020-04-20
Migrate a few kernels/folders to using methods

diyessi:


src/ngraph/op/divide.cpp
```diff
@@ -78,6 +79,60 @@ shared_ptr<Node> ngraph::operator/(const Output<Node>& arg0, const Output<Node>&
...
+               try_evaluate_divide<element::Type_t::u16>(
+                   arg0, arg1, out, broadcast_spec, pythondiv) ||
+               try_evaluate_divide<element::Type_t::u32>(
+                   arg0, arg1, out, broadcast_spec, pythondiv) ||
+               try_evaluate_divide<element::Type_t::u64>(
+                   arg0, arg1, out, broadcast_spec, pythondiv) ||
+               try_evaluate_divide<element::Type_t::f32>(
+                   arg0, arg1, out, broadcast_spec, pythondiv) ||
+               try_evaluate_divide<element::Type_t::f64>(
+                   arg0, arg1, out, broadcast_spec, pythondiv);
```
rkimballn1, line 126: Use a switch statement here because
1) It is easier to read and understand
2) It allows for compile-time checking that all types are supported
3) Allows a more informative error message if something goes wrong, currently any failure returns false. It might be nice for more information why something failed

This statement applies to all ops below

src/ngraph/op/divide.cpp
```diff
@@ -78,6 +79,60 @@ shared_ptr<Node> ngraph::operator/(const Output<Node>& arg0, const Output<Node>&
...
+                             bool pythondiv)
+    {
+        return (ET == arg0->get_element_type()) && (runtime::reference::divide(arg0->get_ptr<ET>(),
+                                                                               arg1->get_ptr<ET>(),
+                                                                               out->get_ptr<ET>(),
+                                                                               arg0->get_shape(),
+                                                                               arg1->get_shape(),
+                                                                               broadcast_spec,
+                                                                               pythondiv),
+                                                    true);
```
rkimballn1, line 98: I'm not a fan of tying the op definition so to a reference kernel. It seems to me that the op is just a definition of the op and that a separate class could be used to execute nodes thus separating the definition of the op from an particular implementation. Embedding an implementation here blurs the line between components.

src/ngraph/op/divide.cpp
```diff
@@ -129,3 +184,9 @@ void op::v1::Divide::generate_adjoints(autodiff::Adjoints& adjoints, const Outpu
     adjoints.add_delta(x, delta / y);
     adjoints.add_delta(y, -delta * shared_from_this() / y);
 }
+
+bool op::v1::Divide::evaluate(const EvaluatorTensorVector& outputs,
+                              const EvaluatorTensorVector& inputs)
```
rkimballn1, line 189: Why a new class for Tensor? What is wrong with runtime/tensor.hpp and runtime/host_tensor.hpp? Those already exist for this specific purpose and should be used instead of creating new classes.

PR #4609, opened 2020-04-22
Add reference kernel for non_zero

yimeisun123:


rkimballn1:
NonZero ref kernel and eval already merged to master

PR #4610, opened 2020-04-22
Add dynamic shape evaluation support

diyessi:
- Add support for dynamic shape evaluation to `EvalutatorTensor`
 - Enable dynamic inputs on `Add`
 - `evaluate` method for `Range`
 - test `f(start, stop, step, p)=Range(start, stop, step)+p`
 - Modify interpreter to be able to run dynamic shape graphs if dynamic ops have `evaluate`


src/ngraph/evaluator_tensor.hpp
```diff
@@ -29,26 +29,85 @@ namespace ngraph
...
+        size_t get_size_in_bytes() const;
+        /// \return True if storage has been allocated
+        bool get_is_allocated() const;
+        /// \brief Set the element type. Must be compatible with the current element type.
+        /// \param node The node being evaluated, for context
+        /// \param element_type The element type
+        void set_element_type(Node* node, const element::Type& element_type);
+        /// \brief Set the shape for allocation. Must be compatible with the current partial shape.
+        void set_allocation_shape(Node* context, const Shape& shape);
+        /// \brief Set the actual shape of the tensor. Dimension 0 can be less than the allocation
```
yimeisun123, line 90: Why only Dimension 0 can be < allocation but not other dimensions? In the case of NonZero op, the output shape is {input_shape_rank, num_non_zero_elem}. Dim 0 might be dynamic, Dim 1 is most likely less then the allocation shape. How would this work if only Dim 0 can be less?

diyessi: Only dim 0 can be shortened without shuffling elements. If you need to shorten something else, your kernel will need to build a temporary result and then copy.

diyessi: I took out this shortening capability.

src/ngraph/runtime/executable.hpp
```diff
@@ -44,6 +44,16 @@ class NGRAPH_API ngraph::runtime::Executable
     virtual bool call(const std::vector<std::shared_ptr<runtime::Tensor>>& outputs,
                       const std::vector<std::shared_ptr<runtime::Tensor>>& inputs) = 0;
 
+    /// \param outputs vector of runtime::Tensor used as outputs. Null elements will be replaced
```
rkimballn1, line 47: Why do we need a new call method? What is different than the other method? Why would I call this? Please document in the code

src/ngraph/runtime/interpreter/int_executable.cpp
```diff
@@ -100,38 +98,77 @@ runtime::interpreter::INTExecutable::INTExecutable(const std::string& model_stri
 
 bool runtime::interpreter::INTExecutable::call(const vector<shared_ptr<runtime::Tensor>>& outputs,
                                                const vector<shared_ptr<runtime::Tensor>>& inputs)
+{
+    HostTensor::HostEvaluatorTensorVector tensor_outputs;
```
rkimballn1, line 102: Looks like you really just want HostTensor to be a HostEvaluatorTensor. Can we do that?

diyessi: I agree; with some changes to Constant I think I can. And that solves the dynamic_call issue as well.

raramer01:
TESTNOW

raramer01:
TESTNOW_DLDT_E2E

PR #4614, opened 2020-04-23
[rov.2020.3] Add ONNX Slice-10 operator

mbencer:


src/ngraph/op/strided_slice.cpp
```diff
@@ -56,9 +91,7 @@ op::v1::StridedSlice::StridedSlice(const Output<Node>& data,
     : StridedSlice(data,
                    begin,
                    end,
-                   op::Constant::create(element::i64,
```
mbencer, line 59: It was bug found during working on ONNX Slice-10. However, in current implementation of Slice-10 this version of constructor is not used so StridedSlice changes can be reverted in this PR

lazarevevgeny:
Do we have an OpenVINO e2e tests passed with this version of nGraph?

tomdol:
> Do we have an OpenVINO e2e tests passed with this version of nGraph?

I've just created a MR #7284 - can you please have a look and run the CIs that didn't start automatically?

PR #4619, opened 2020-04-23
Constant/HostTensor interroperability

diyessi:


src/ngraph/op/constant.cpp
```diff
@@ -48,14 +48,18 @@ string to_cpp_string(T value)
 
 constexpr NodeTypeInfo op::Constant::type_info;
 
+op::Constant::Constant(const shared_ptr<runtime::HostTensor>& tensor)
+    : Constant(tensor->get_element_type(), tensor->get_shape())
+{
+    m_tensor = tensor;
```
rkimballn1, line 54: Should copy the data since we can't guarantee that tensor won't change its stored value

src/ngraph/op/constant.cpp
```diff
@@ -48,14 +48,18 @@ string to_cpp_string(T value)
...
+}
```
rkimballn1, line 55: You need to set `m_all_elements_bitwise_identical = are_all_data_elements_bitwise_identical();`

src/ngraph/op/constant.hpp
```diff
@@ -39,6 +41,9 @@ namespace ngraph
                 static constexpr NodeTypeInfo type_info{"Constant", 0};
                 const NodeTypeInfo& get_type_info() const override { return type_info; }
                 Constant() = default;
+
+                Constant(const std::shared_ptr<runtime::HostTensor>& tensor);
```
rkimballn1, line 45: If you plan on sharing the tensor data you really need to document that else someone will change tensor

raramer01:
TESTNOW_DLDT_E2E

diyessi:
#4632

PR #4621, opened 2020-04-24
Add EvaluatorTensor functionality into HostTensor

diyessi:
- Constructors for partial shape
- Set compatible shape/element type
- Change EvaluatorTensor::get_ptr to get_data_ptr for compatibility
Will require #4619 before EvaluatorTensor can be replaced


rkimballn1:
Blocked temporarily

PR #4623, opened 2020-04-24
[ONNX] ONNX Importer as shared library.

tsocha:
- Change `onnx_importer` into shared library.
- ONNX importer symbols have been removed from ngraph library.
- From now to use onnx importer you have to link against onnx_importer library.

image


src/ngraph/frontend/onnx_import/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ add_library(onnx_import_interface OBJECT
         onnx_utils.hpp
         onnx_utils.cpp)
 
-add_library(onnx_import STATIC
+add_library(onnx_import_core STATIC
```
ilyachur, line 28: Why do you create a static library `onnx_import_core`? Looks like you can create only `onnx_importer` shared library.

The `onnx_import_core` is linking only with the `onnx_importer` shared library and doesn't use anymore.

tsocha: We have a separation between interface and implementation in onnx importer.
Since onnx_import_core is a `PRIVATE` library for onnx_importer its symbols don't pollute the main library interface.

ilyachur: But in general in the line 266 I see that we buid the library with visibility hidden flag, in this case the library interface contains only expoeted methods.

ilyachur: I mean that visibility hidden allows to hide all internal symbols and you don't need to create `onnx_import_core` library. With visibility hidden only exported symbols will be exists in the interface of `onnx_importer` library.

tsocha: Done

python/setup.py
```diff
@@ -87,6 +87,10 @@ def find_pybind_headers_dir():
 if len([fn for fn in os.listdir(NGRAPH_CPP_LIBRARY_DIR) if re.search('ngraphd', fn)]):
     NGRAPH_CPP_LIBRARY_NAME = 'ngraphd'
 
+ONNX_IMPORTER_CPP_LIBRARY_NAME = 'onnx_importer'
+if len([fn for fn in os.listdir(NGRAPH_CPP_LIBRARY_DIR) if re.search('onnx_importerd', fn)]):
```
ilyachur, line 91: @tomdol @tsocha @mbencer  Do we need onnx_importer dependency for the Python API? As we discussed user can use OpenVINO Reader API to load ONNX model and this API can be hided from Python.

tomdol: The onnx_importer is still exposed in `ngraph.impl` python module. We can probably remove it but let's discuss it offline and create a separate story for that.

postrational: We need the ONNX importer API for testing the ONNX spec.

ilyachur: Good! Thank you for the clarification.

PR #4624, opened 2020-04-24
Add visitor and default axis to CumSum for opset3

adstraw:


src/ngraph/op/cum_sum.cpp
```diff
@@ -35,6 +36,20 @@ op::v0::CumSum::CumSum(const Output<Node>& arg,
     constructor_validate_and_infer_types();
 }
 
+op::v0::CumSum::CumSum(const Output<Node>& arg, const bool exclusive, const bool reverse)
+    : Op({arg, op::Constant::create(element::i32, Shape{}, {0})})
+    , m_exclusive(exclusive)
```
diyessi, line 41: Needs to be a new op version.

lazarevevgeny: I always vote for a version change for each change in the operation unless it an obvious bug fix. But for this particular case we may keep the version.

raramer01:
TESTNOW

raramer01:
TESTNOW_DLDT_E2E

raramer01:
TESTNOW

adstraw:
TESTNOW

raramer01:
TESTNOW_DLDT_E2E

PR #4625, opened 2020-04-24
Add Interpolate v3 Shell Op

silee2:


src/ngraph/op/interpolate.cpp
```diff
@@ -70,3 +70,55 @@ shared_ptr<Node> op::Interpolate::clone_with_new_inputs(const OutputVector& new_
...
+
+    if (auto const_shape = as_type_ptr<op::Constant>(input_value(1).get_node_shared_ptr()))
+    {
+        auto out_shape = const_shape->cast_vector<int64_t>();
+        size_t i = 0;
+        for (auto axis : m_attrs.axes)
+        {
+            output_shape[axis] = Dimension(out_shape[i++]);
+        }
+        set_output_type(0, get_input_element_type(0), output_shape);
```
diyessi, line 112: I think you want to move this out of the if and get rid of the else.

src/ngraph/serializer.cpp
```diff
@@ -1693,7 +1693,33 @@ shared_ptr<Node> JSONDeserializer::deserialize_node(json node_js)
             node = make_shared<op::HardSigmoid>(args[0], args[1], args[2]);
             break;
         }
-        case OP_TYPEID::Interpolate: { break;
+        case OP_TYPEID::Interpolate:
```
diyessi, line 1696: No serialize changes; need to use the attribute visitor.

src/ngraph/op/interpolate.hpp
```diff
@@ -22,18 +22,18 @@ namespace ngraph
...
-            std::vector<size_t> pads_begin;
-            std::vector<size_t> pads_end;
-        } InterpolateAttrs;
-
         namespace v0
         {
+            typedef struct
+            {
+                AxisSet axes;
+                std::string mode;
```
rkimballn1, line 30: Can we use an enum here? Using string is very unsafe.
This struct could use some documentation since it is a parameter to a function.

src/ngraph/op/interpolate.hpp
```diff
@@ -43,7 +43,46 @@ namespace ngraph
...
+            };
+        }
+
+        namespace v3
+        {
+            typedef struct
+            {
+                AxisSet axes;
+                std::string mode;
+                std::string coordinate_transformation_mode;
```
rkimballn1, line 70: Enum?

src/ngraph/op/interpolate.hpp
silee2: I agree enum is better but I followed how v0 done.

src/ngraph/op/interpolate.hpp
```diff
@@ -22,18 +22,18 @@ namespace ngraph
...
-            std::string mode;
-            bool align_corners = true;
-            bool antialias = false;
-            std::vector<size_t> pads_begin;
-            std::vector<size_t> pads_end;
-        } InterpolateAttrs;
-
         namespace v0
         {
+            typedef struct
```
diyessi, line 27: Why not `struct InterpolateAttrs` ?

src/ngraph/op/interpolate.cpp
silee2: Done

src/ngraph/serializer.cpp
silee2: Done

src/ngraph/op/interpolate.hpp
silee2: Added an Enum. The string will be check against the Enum in the constructor.

src/ngraph/op/interpolate.hpp
silee2: Added an Enum to check against.

src/ngraph/op/interpolate.hpp
silee2: Done

src/ngraph/op/interpolate.hpp
silee2: v0 already has users based on the this struct. so could not change it.

silee2: Also added unit-test to check for invalid string.

src/ngraph/op/interpolate.hpp
```diff
@@ -17,38 +17,132 @@
...
+                // This addition of pixels is done before interpolation calculation.
+                std::vector<size_t> pads_end;
+            };
+
             /// \brief Layer which performs bilinear interpolation
             class NGRAPH_API Interpolate : public Op
             {
             public:
+                enum class InterpolateMode
+                {
```
diyessi, line 56: Make the enum case the same as in the spec, i.e. "nearest" instead of "NEAREST".

src/ngraph/attribute_adapter.cpp
```diff
@@ -337,6 +337,28 @@ namespace ngraph
         m_buffer_valid = false;
     }
 
+#ifdef __APPLE__
+    // size_t is not uint64_t on OSX
```
rkimballn1, line 341: The comment mentions uint64_t but the code references int64_t. Is the comment or the code wrong?
Same comment for the header file below.

silee2: The comment is correct. It is trying to say that the existing adapter for uint64_t (size_t on other platforms) does not work on OS X. So a new adapter is needed. int64_t is in the implementation since get/set method for all integer type is based on int64_t.

silee2: The extra code is not needed once we have #4666 merged. The extra code is a temporary solution to remove build failures on macOS.

src/ngraph/op/interpolate.hpp
silee2: Changed the enums to lowercase.

src/ngraph/op/interpolate.hpp
silee2: Changed to enums for v3. v0 is still strings.

src/ngraph/attribute_adapter.cpp
diyessi: To clarify, on Macos, `size_t` is not one of the `uintx_t` types.

src/ngraph/op/interpolate.hpp
diyessi: Okay. Best not to mess with the previous version of the op.

src/ngraph/op/interpolate.cpp
```diff
@@ -32,8 +31,21 @@ op::Interpolate::Interpolate(const Output<Node>& image,
...
+    visitor.on_attribute("attrs.align_corners", m_attrs.align_corners);
+    visitor.on_attribute("attrs.antialias", m_attrs.antialias);
+    visitor.on_attribute("attrs.pads_begin", m_attrs.pads_begin);
+    visitor.on_attribute("attrs.pads_end", m_attrs.pads_end);
+    return true;
+}
+
 void op::Interpolate::validate_and_infer_types()
 {
+    // Check if string attribute is valid by converting to and back from valid enum type
```
diyessi, line 47: Take out this check. Something will be passing the string incorrectly and count on it not getting an error and this will be considered breaking API.

src/ngraph/op/interpolate.cpp
```diff
@@ -32,8 +31,21 @@ op::Interpolate::Interpolate(const Output<Node>& image,
     constructor_validate_and_infer_types();
 }
 
+bool op::Interpolate::visit_attributes(AttributeVisitor& visitor)
```
diyessi, line 34: Please change `op::Interpolate*` to `op::v0::Interpolate*` in this file so it's easy to tell what version everything in.

src/ngraph/op/interpolate.cpp
silee2: Done.

PR #4627, opened 2020-04-24
Add build rules to find OpenVINO inference engine

adk9:
Add cmake rules to find and link with Inference Engine when the nGraph IE
backend is enabled.


ilyachur:
Why do you need this PR? Why don't you use Inference Engine CMake configs? It should be enough.

cmake/Modules/InferenceEngineConfig.cmake
```diff
@@ -0,0 +1,171 @@
+# Copyright (C) 2018-2020 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
```
rkimballn1, line 2: Use the correct license header. Copy from one of the other files

adk9:
> Why do you need this PR? Why don't you use Inference Engine CMake configs? It should be enough.

Hi @ilyachur, this is to be able to build the Inference Engine backend outside of the `dldt` repo.

ilyachur:
> > Why do you need this PR? Why don't you use Inference Engine CMake configs? It should be enough.
> 
> Hi @ilyachur, this is to be able to build the Inference Engine backend outside of the `dldt` repo.

@adk9 From my point of view it is a work around. Tomorrow the Inference Engine will introduce a new library and it will cause CI fails and changes in this file

If you build Inference Engine separately the CMake scripts generates CMake configs and you can use them. The OpenVINO package also contains the CMake configs. I don't see any case when you need this custom CMake config.

arogowie-intel:
I agree with @ilyachur. If there are already available tools for this in IE, we should use them. Maybe we could use `find_package` mechanism?

cmake/Modules/InferenceEngineConfig.cmake
```diff
@@ -0,0 +1,184 @@
+# ******************************************************************************
+# Copyright 2018-2020 Intel Corporation
```
rkimballn1, line 2:
```suggestion
# Copyright 2017-2020 Intel Corporation
```
The years much match all files in the repo

adk9:
> > > Why do you need this PR? Why don't you use Inference Engine CMake configs? It should be enough.
> > 
> > 
> > Hi @ilyachur, this is to be able to build the Inference Engine backend outside of the `dldt` repo.
> 
> @adk9 From my point of view it is a work around. Tomorrow the Inference Engine will introduce a new library and it will cause CI fails and changes in this file
> 
> If you build Inference Engine separately the CMake scripts generates CMake configs and you can use them. The OpenVINO package also contains the CMake configs. I don't see any case when you need this custom CMake config.

Fair enough. I just use `find_package` now to locate the CMake configs distributed with the Inference Engine release.

src/ngraph/runtime/ie/CMakeLists.txt
```diff
@@ -30,11 +30,11 @@ set(SRC
     ie_executable.hpp
 )
 
+find_package(InferenceEngine REQUIRED)
 add_library(ie_backend ${LIBRARY_TYPE} ${SRC})
-add_dependencies(ie_backend inference_engine)
 target_compile_definitions(ie_backend PRIVATE IE_BACKEND_DLL_EXPORTS)
-target_include_directories(ie_backend PUBLIC ${IE_MAIN_SOURCE_DIR}/include)
-target_link_libraries(ie_backend PUBLIC ngraph inference_engine)
+target_include_directories(ie_backend PUBLIC ${InferenceEngine_INCLUDE_DIRS})
```
ilyachur, line 36: I guess you can remove this line and don't add `${InferenceEngine_INCLUDE_DIRS}` CMake automatically adds this folders from `${InferenceEngine_LIBRARIES}`

adk9:
TESTNOW_DLDT_E2E

adk9:
TESTNOW

PR #4628, opened 2020-04-24
Bucketize op shell implementation

gaurides:


src/ngraph/CMakeLists.txt
```diff
@@ -145,6 +145,7 @@ set (SRC
     op/binary_convolution.hpp
     op/broadcast.cpp
     op/broadcast_distributed.cpp
+    op/bucketize.cpp
```
yimeisun123, line 148: also add op/bucketize.hpp here as for the others. (I don't know why the hpp need to be added here though, but just to follow the same behavior).

src/ngraph/serializer.cpp
```diff
@@ -1120,6 +1120,13 @@ shared_ptr<Node> JSONDeserializer::deserialize_node(json node_js)
             node = make_shared<op::BroadcastLike>(args[0], args[1], initial_axes);
             break;
         }
+        case OP_TYPEID::Bucketize_v3:
+        {
+            element::Type output_type = read_element_type(node_js.at("output_type"));
+            bool with_right_bound = get_or_default<bool>(node_js, "with_right_bound", true);
+            node = make_shared<op::v3::Bucketize>(args[0], args[1], output_type, with_right_bound);
+            break;
+        }
```
silee2, line 1129: Don't need this since the op has attribute_visitor.

src/ngraph/serializer.cpp
```diff
@@ -2894,6 +2901,13 @@ json JSONSerializer::serialize_node(const Node& n)
         node["initial_axes"] = serialize_axis_set(tmp->get_initial_broadcast_axes());
         break;
     }
+    case OP_TYPEID::Bucketize_v3:
+    {
+        auto tmp = static_cast<const op::v3::Bucketize*>(&n);
+        node["output_type"] = write_element_type(tmp->get_output_type());
+        node["with_right_bound"] = tmp->get_with_right_bound();
+        break;
+    }
```
silee2, line 2910: Don't need this since the op has attribute_visitor.

src/ngraph/serializer.cpp
gaurides: thanks, do i need the serializer tests ?

src/ngraph/op/bucketize.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+    const PartialShape& data_pshape = get_input_partial_shape(0);
+    const PartialShape& buckets_pshape = get_input_partial_shape(1);
+
+    NODE_VALIDATION_CHECK(this,
+                          m_output_type == element::i64 || m_output_type == element::i32,
+                          "Output type must be i32 or i64. Default is i64");
+
+    NODE_VALIDATION_CHECK(
+        this, buckets_pshape.rank().compatible(1), "buckets input must be a 1D tensor");
```
diyessi, line 52: Only run this test if the rank is static. If something is computing the bucket shape, we might not know anything about the shape while constructing the graph, even though we know during execution it needs to be 1-dimensional.

src/ngraph/serializer.cpp
diyessi: No.

src/ngraph/serializer.cpp
diyessi: No changes should be in serializer.

raramer01:
TESTNOW_DLDT_E2E

gaurides:
TESTNOW

gaurides:
TESTNOW_DLDT_E2E

src/ngraph/op/bucketize.hpp
```diff
@@ -0,0 +1,70 @@
...
+    namespace op
+    {
+        namespace v3
+        {
+            /// \brief Operation that bucketizes the input based on boundaries
+            class NGRAPH_API Bucketize : public Op
+            {
+            public:
+                static constexpr NodeTypeInfo type_info{"Bucketize", 3};
+                const NodeTypeInfo& get_type_info() const override { return type_info; }
```
yimeisun123, line 32: Do you need to override the Node's class implementation for this method?

PR #4630, opened 2020-04-25
Add evaluate to convert

swapna-intel:
Need to add test


src/ngraph/op/convert.cpp
```diff
@@ -55,3 +56,38 @@ void op::Convert::generate_adjoints(autodiff::Adjoints& adjoints, const OutputVe
...
+                              size_t count)
+
+    {
+        return (ET == arg->get_element_type()) &&
+               (runtime::reference::convert(arg->get_ptr<ET>(), out->get_ptr<ET>(), count), true);
+    }
+
+    bool
+        evaluate_convert(const EvaluatorTensorPtr& arg, const EvaluatorTensorPtr& out, size_t count)
+    {
```
diyessi, line 74: You will need to set the output element type for `out` here, and copy the output shape from `arg` to `out`

src/ngraph/op/convert.cpp
```diff
@@ -55,3 +56,38 @@ void op::Convert::generate_adjoints(autodiff::Adjoints& adjoints, const OutputVe
 
     adjoints.add_delta(x, make_shared<op::Convert>(delta, x.get_element_type()));
 }
+
+namespace
+{
+    template <element::Type_t ET>
+    bool try_evaluate_convert(const EvaluatorTensorPtr& arg,
+                              const EvaluatorTensorPtr& out,
+                              size_t count)
```
diyessi, line 65: You need a separate expansion of element types for the `out` argument types, i.e. `try_evaluate_convert` needs to do the same thing `evaluate_convert` does, but for the output type instead of the input type.

src/ngraph/op/convert.cpp
swapna-intel: done

src/ngraph/op/convert.cpp
swapna-intel: done

swapna-intel:
> Please make sure all tests pass with this change.

done

src/ngraph/op/convert.cpp
swapna-intel: done

src/ngraph/op/convert.hpp
```diff
@@ -53,6 +54,10 @@ namespace ngraph
                     m_destination_type = destination_type;
                 }
 
+                size_t get_version() const override { return 0; }
```
diyessi, line 57: Get rid of this. `Node::get_version()` gets the version from the type_info.

swapna-intel: done

PR #4647, opened 2020-04-28
Custom ONNX namespace for nGraph

tomdol:
This PR adds the ability to set a custom namespace for ONNX objects. If not passed explicitly to cmake, the default value is now `ngraph_onnx`. This prevents the segfaults and other memory corruption problems when ngraph and onnx importer are used with UEP in the same binary.


ilyachur:
Does this PR solve the issue with import from ModelProto?

src/ngraph/frontend/onnx_import/onnx.hpp
```diff
@@ -61,17 +56,6 @@ namespace ngraph
...
-        /// \brief      Imports an in-memory ONNX model and converts it to an nGraph Function.
-        ///
-        /// \note       If the ONNX model contains unsupported ops, the function throws
-        ///             an ngraph_error exception.
-        ///
-        /// \param[in]  model_proto    The in-memory ONNX model reference
-        ///
-        /// \return     An nGraph function that represents a single output from the created graph.
-        NGRAPH_API
-        std::shared_ptr<Function> import_onnx_model(const onnx::ModelProto& model_proto);
```
ilyachur, line 73: Why did you remove this method? Don't we need it anymore?

tomdol: Yes, I've removed it because we went back to the `std::stringstream` approach. Now with a custom namespace for ONNX in nGraph, the import problems are finally fixed.

PR #4651, opened 2020-04-28
Cyphers/constant

diyessi:
- Disable constant_fold for Constant
- Add evaluate for Constant
- Add attribute visitor for Constant
- Reduce need for layout in Tensor since it isn't set up until storage is allocated


src/ngraph/op/constant.cpp
```diff
@@ -613,6 +614,27 @@ bool op::Constant::are_all_data_elements_bitwise_identical() const
...
+bool op::v0::Constant::visit_attributes(AttributeVisitor& visitor)
+{
+    visitor.on_attribute("element_type", m_element_type);
+    visitor.on_attribute("shape", m_shape);
+    if (m_data == nullptr)
+    {
+        // Filling in a fresh constant
+        allocate_buffer();
+    }
+    DataHandle handle(get_data_ptr_nc(), shape_size(m_shape) * m_element_type.size());
```
ilyachur, line 626: Who is an owner of this data? As I understand `DataHandle` doesn't control the data.

src/ngraph/op/constant.hpp
```diff
@@ -244,6 +244,17 @@ namespace ngraph
...
 
+                bool visit_attributes(AttributeVisitor& visitor) override;
+
+                bool evaluate(const HostTensorVector& outputs,
+                              const HostTensorVector& inputs) override;
+
+                // Don't constant fold a constant; it would make a copy
+                bool constant_fold(OutputVector& outputs, const OutputVector& inputs) override
+                {
+                    return false;
```
ilyachur, line 255: Could you explain cases when constant folding can create a copy?

src/ngraph/op/constant.cpp
diyessi: Constant owns the data. DataHandle provides access.

src/ngraph/op/constant.hpp
diyessi: It's already a constant. If we were to "fold it" it would be making a copy of the constant and replacing it in the graph, which doesn't make any sense.

src/ngraph/op/util/attr_types.hpp
```diff
@@ -368,6 +368,48 @@ namespace ngraph
         const DiscreteTypeInfo& get_type_info() const override { return type_info; }
     };
 
+    class NGRAPH_API DataHandle
```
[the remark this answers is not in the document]
diyessi: This provides access for a generic deserializer to initialize the data void* such as from a binary file. It provides the void* and it's length. Likewise, a deserializer has the void* and length of the data.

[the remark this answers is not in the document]
diyessi: I think I will get rid of DataHandle and just have a visitor method that gets passed the void* and size_t.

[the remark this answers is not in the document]
arogowie-intel: Although the usage of `DataHandle` looks quite convenient it might be more explicit and readable that we're dealing with raw data if we just have another overload `on_attribute(const string& name, void* data_ptr, size_t size_in_bytes)` ?

src/ngraph/op/constant.cpp
diyessi: Got rid of DataHandle and made things simpler.

src/ngraph/op/util/attr_types.hpp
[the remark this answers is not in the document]
diyessi: That's exactly what I just did.

raramer01:
TESTNOW_DLDT_E2E

PR #4654, opened 2020-04-29
NonZero evaluate support

yimeisun123:


src/ngraph/op/non_zero.cpp
```diff
@@ -70,3 +75,72 @@ shared_ptr<Node> op::v3::NonZero::clone_with_new_inputs(const OutputVector& new_
...
+        else
+        {
+            return evaluate_nonzero_execute<INPUT_ET, element::Type_t::i32>(input, output);
+        }
+    }
+}
+
+bool op::v3::NonZero::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
+    return evaluate_nonzero<element::Type_t::i8>(inputs[0], outputs[0]) ||
```
diyessi, line 135: Bob will want you to switch this to use his TYPECASE macro. See, for example, Add.

src/ngraph/op/non_zero.cpp
```diff
@@ -70,3 +75,72 @@ shared_ptr<Node> op::v3::NonZero::clone_with_new_inputs(const OutputVector& new_
...
+        else if (input_rank == 0)
+        {
+            out_shape = Shape{1, 1};
+        }
+        else
+        {
+            out_shape = Shape{input_rank, non_zero_count};
+        }
+
+        output->set_shape(out_shape);
```
diyessi, line 107: Excellent.

src/ngraph/op/non_zero.cpp
yimeisun123: I chatted with him on this. I can use the TYPECASE, it will still need additional function as the input and output are different element type for NonZero op.

rkimballn1, line 135: Scott will want you to switch this to use the TYPE_CASE macro. See, for example, Add.

src/ngraph/runtime/reference/non_zero.hpp
```diff
@@ -0,0 +1,146 @@
...
+    {
+        namespace reference
+        {
+            /// \brief Return number of non-zero entries in the input argument.
+            ///
+            /// \param arg Input tensor
+            /// \param arg_shape Input tensor shape
+            /// Output number of non-zero entries in arg
+            template <typename T>
+            size_t non_zero_get_count(const T* arg, const Shape& arg_shape)
```
swapna-intel, line 34: @yimeisun123 if we break the non_zero op into 2 calls, will that work for any user trying to call non_zero reference kernel directly? Or is it that the interpreter will be the sole user for this op?

test/eval.cpp
```diff
@@ -155,3 +156,21 @@ TEST(eval, interpret_dynamic_range_sum)
...
+{
+    auto input = make_shared<op::Parameter>(element::f32, PartialShape{});
+    auto non_zero = make_shared<op::v3::NonZero>(input, element::i64);
+    auto fun = make_shared<Function>(OutputVector{non_zero}, ParameterVector{input});
+    auto backend = runtime::Backend::create("INTERPRETER");
+    auto input_val = backend->create_tensor(element::f32, Shape{});
+    copy_data(input_val, vector<float>{1.0f});
+    auto result = make_shared<HostTensor>();
+    auto cfun = backend->compile(fun);
+    cfun->call({result}, {input_val});
```
swapna-intel, line 170: dont we want a test just for evaluate without going through compile? I see both for dynamic_range_sum. Trying to understand why/why not.

yimeisun123: See the commit with more test cases and also creating op_eval test folder.

src/ngraph/runtime/reference/non_zero.hpp
yimeisun123: Any caller can call. As with other reference kernels, the caller needs to allocate sufficient memory for output.

yimeisun123:
@rkimballn1 @diyessi 
1) I created a op_eval dir to put op evaluate tests so we don't put everything in the eval.cpp
2) For the backend test, currently only interpreter backend supports this evaluate trigger and only HostTensor has the ability of delayed allocation and shape setting. In the test case, I use check for backend name and the test cases only execute on interpreter backend.

src/ngraph/op/non_zero.cpp
```diff
@@ -70,3 +74,98 @@ shared_ptr<Node> op::v3::NonZero::clone_with_new_inputs(const OutputVector& new_
     check_new_args_count(this, new_args);
     return make_shared<v3::NonZero>(new_args.at(0), m_output_type);
 }
+
+namespace
+{
+    template <element::Type_t INPUT_ET, element::Type_t OUT_ET>
+    bool evaluate_nonzero_execute(const HostTensorPtr& input, const HostTensorPtr& output)
+    {
+        using T = typename element_type_traits<INPUT_ET>::value_type;
```
diyessi, line 83: How about IN_T and OUT_T instead of T and U?

src/ngraph/op/non_zero.cpp
```diff
@@ -70,3 +74,98 @@ shared_ptr<Node> op::v3::NonZero::clone_with_new_inputs(const OutputVector& new_
...
+    }
+
+    template <element::Type_t INPUT_ET>
+    bool evaluate(const HostTensorPtr& input, const HostTensorPtr& output)
+    {
+        if (output->get_element_type() == element::i64)
+        {
+            return evaluate_nonzero_execute<INPUT_ET, element::Type_t::i64>(input, output);
+        }
+        else
```
diyessi, line 120: Don't assume output element type will always be one of i64 and i32. They are already talking about adding more types. So use a switch and return false if it's not a case you handle.
I see you checked for other output types below, but it's good to be careful here too.

src/ngraph/op/non_zero.cpp
```diff
@@ -70,3 +74,98 @@ shared_ptr<Node> op::v3::NonZero::clone_with_new_inputs(const OutputVector& new_
...
+            break;
+        default: rc = false; break;
+        }
+        return rc;
+    }
+}
+
+bool op::v3::NonZero::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
+    if (outputs[0]->get_element_type() != element::i64 &&
```
diyessi, line 162: Why don't we do this one as a switch with default returning `false` so it's symmetric with the inputs selection switches, and it will be simpler to add new types later.

src/ngraph/runtime/gcpu/unit_test.manifest
```diff
@@ -7,6 +7,11 @@ convert_bf16_float32
 GCPU.onnx_model_quant_conv_linear
 GCPU.onnx_top_k_opset_10
```
diyessi, line 9: GCPU should work. It also calls evaluate.

src/ngraph/runtime/interpreter/opset_int_tbl.hpp
```diff
@@ -28,4 +28,5 @@ NGRAPH_OP(LogicalNot, op::v1)
 
 #define ID_SUFFIX(NAME) NAME##_v3
 NGRAPH_OP(ShapeOf, op::v3)
```
diyessi, line 30: Let's take this little section our and remove the _v3 enums from the switch cases. They were only added because we needed a way to get the interpreter to run a couple new ops.

test/backend/non_zero.in.cpp
```diff
@@ -0,0 +1,105 @@
...
+    auto backend = runtime::Backend::create("${BACKEND_NAME}");
+    auto cfun = backend->compile(fun);
+
+    auto input = backend->create_tensor(element::f32, Shape{3, 2});
+    copy_data(input, vector<float>{0.0f, 0.0f, 0.0f, 0.0f, 1.0f, 3.0f});
+
+    std::vector<int32_t> expected_result{2, 2, 0, 1};
+    Shape expected_output_shape{2, 2};
+
+    auto result = make_shared<HostTensor>();
```
diyessi, line 45: This won't work for non-CPU devices since you didn't use the backend to create the tensor. @rkimballn1 what is the best way to deal with this?

rkimballn1: Scott is correct, the CPU backend will cast this to a CPU Tensor which it is not.
We need to call backend->create_tensor() with no arguments. The backend needs to support this which none of the backends currently support. 
For now put this test in the CPU manifest and we can do a separate PR to add the needed backend API for creating tensors of unknown size/type

src/ngraph/runtime/gcpu/unit_test.manifest
yimeisun123: GCPU doesn't work, I already tried because of the result tensor.

src/ngraph/runtime/interpreter/opset_int_tbl.hpp
yimeisun123: Without adding this, it has build error. This is the same as ShapeOf v3.

src/ngraph/op/non_zero.cpp
yimeisun123: Done

src/ngraph/op/non_zero.cpp
yimeisun123: Use switch now.

src/ngraph/op/non_zero.cpp
yimeisun123: sure

PR #4655, opened 2020-04-29
Switch ONNX Importer to produce ops from opset3

mbencer:
- `Opset1Downgrade` pass is temporary solution to preserve ONNX ops tests


mbencer:
OV tests: https://gitlab-icv.inn.intel.com/inference-engine/dldt/merge_requests/7368

lazarevevgeny:
@GlebKazantaev could you take a look at this change? I see here a downgrade pass from Opset3 to opset1 operations. Do we plan to execute it on the OV side?

lazarevevgeny:
Please, do not merge this until you get a confirmation from @jane-intel that it does not break the onnx model conversion.

src/ngraph/frontend/onnx_import/op/expand.cpp
```diff
@@ -35,11 +35,8 @@ namespace ngraph
                     const std::shared_ptr<ngraph::Node> data{node.get_ng_inputs().at(0)};
                     const std::shared_ptr<ngraph::Node> shape{node.get_ng_inputs().at(1)};
 
-                    const auto const_filled_with_ones = std::make_shared<default_opset::Broadcast>(
-                        default_opset::Constant::create(data->get_element_type(), {}, {1}), shape);
-
-                    return {
-                        std::make_shared<default_opset::Multiply>(data, const_filled_with_ones)};
+                    return {std::make_shared<default_opset::Broadcast>(
```
jane-intel, line 38: `Broadcast` from opset other than `opset3` would not work. Does it mean we have to have some if-else logic based on the opset version?

tomdol: `default_opset` is just an alias that we use for onnx_importer. In this PR it was switched to point to opset3 https://github.com/NervanaSystems/ngraph/pull/4655/files?file-filters%5B%5D=.hpp#diff-1ca2b8cb1f7660163e9568e1ae2f2601R7

jane-intel: I get it, but should we have a comment here or smth that points out difference between opsets? What if someone change default_opset and forget about Broadcasts?

src/ngraph/pass/opset1_downgrade.cpp
```diff
@@ -0,0 +1,125 @@
...
+        }
+        case op::BroadcastType::PDPD:
+        {
+            op::AutoBroadcastSpec broadcast_spec;
+            broadcast_spec.m_type = op::AutoBroadcastType::PDPD;
+            broadcast_spec.m_axis = node->get_broadcast_spec().m_axis;
+            replacement_node = make_shared<op::v1::Broadcast>(data, target_shape, broadcast_spec);
+            break;
+        }
+        }
```
arogowie-intel, line 66: IMHO It is a good practice to add `default` statement to catch unhandled enum values. Moreover lack of it would come up as an issue in any tool used to check the code during SDL process.

src/ngraph/frontend/onnx_import/op/expand.cpp
tomdol: I don't think so TBH. The `default_opset` alias should always point to the most recent ngraph-core opset. It was introduced so that all onnx_importer files reference the core ops in an uniform way. If and only if there's a need to reference any of the previous opsets in a particular ONNX op implementation, the developer should use `opsetX::OpName` explicitly.

The alias should not be changed by anyone but our team. If we switch to `opset4` at some point, this line we're discussing now will automatically use `Broadcast` from `opset4`. If this will be a new version of the operator and the constructor changes its signature, the person updating the alias will have to correct the usage and fix all of the compilation errors (and tests).

src/ngraph/pass/opset1_downgrade.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+namespace
+{
+    shared_ptr<Node> op_cast(shared_ptr<op::v3::Broadcast> node)
+    {
+        const auto data = node->input_value(0).get_node_shared_ptr();
+        const auto target_shape = node->input_value(1).get_node_shared_ptr();
+
+        shared_ptr<Node> replacement_node;
+        switch (node->get_broadcast_spec().m_type)
```
lazarevevgeny, line 36: Do we have any tests for these changes? I don't see them in the review.

src/ngraph/runtime/cpu/cpu_external_function.cpp
```diff
@@ -1263,6 +1264,7 @@ void runtime::cpu::CPU_ExternalFunction::register_common_passes(
 
     REGISTER_KNOBBED_PASS(LikeReplacement, true, ngraph::pass)
     REGISTER_KNOBBED_PASS_WITH_ARGS(FusedOpDecomposition, true, ngraph::pass, is_supported)
+    REGISTER_KNOBBED_PASS(Opset1Downgrade, true, ngraph::pass)
```
lazarevevgeny, line 1267: Here we unconditionally convert Opset-3 operations to opser-1. But what if some backend supports opset3 operation? In this case we should not replace it and potentially we loose some performance boost.

PR #4659, opened 2020-04-29
Make get_registry_mutex protected and not exported

rkimballn1:


src/ngraph/factory.hpp
```diff
@@ -95,6 +93,12 @@ namespace ngraph
         static FactoryRegistry<BASE_TYPE>& get();
 
     protected:
+        static std::mutex& get_registry_mutex()
```
diyessi, line 96: I don't think this will help. If CLR's `<mutex>` doesn't define `std::mutex` it doesn't matter where it appears in the header, and even if you buried mutex in the .cpp file and used a wrapper in the header, mutex support is probably going to be missing in the runtime library. On the other hand, this is a cleaner place for the mutex to live than outside of the class.

PR #4661, opened 2020-04-29
Provide case-insensitive access to opset

diyessi:
Add `create_insensitive` method


raramer01:
TESTNOW

raramer01:
TESTNOW_DLDT_E2E

PR #4662, opened 2020-04-29
Swapna/eliminate convert into reduce min

swapna-intel:
https://jira.devtools.intel.com/browse/CVS-28831


src/ngraph/pass/nop_elimination.cpp
```diff
@@ -69,25 +69,54 @@ static bool eliminate_sum(const std::shared_ptr<Node>& node)
...
+    auto destination_type = convert->get_destination_type();
+    auto input_op = convert->input_value(0);
+
+    // case 1
+    if (is_type<opset3::ReduceMin>(input_op.get_node()))
+    {
+        auto convert_2 = input_op.get_node()->input_value(0).get_node();
+        if (is_type<opset3::Convert>(convert_2) && convert_2->get_users().size() == 1 &&
+            destination_type == convert_2->get_input_element_type(0))
+        {
```
diyessi, line 95: Not valid for all type combinations. (A, B) as (float, double), (bf16|fp16, float|double), (intx, int y>=x), (uintx, uinty>=x), (uintx, inty>x), and (A,A) are valid.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -69,25 +69,54 @@ static bool eliminate_sum(const std::shared_ptr<Node>& node)
...
+    }
     bool is_out_type_agnostic = false;
     static const std::set<NodeTypeInfo> type_agnostic{TI(opset3::NonZero)};
-    if (node->output(0).get_target_inputs().size() == 1)
+    // case 2 & 3
+    if (convert->get_users().size() == 1)
     {
-        auto& out = *node->output(0).get_target_inputs().begin();
-        is_out_type_agnostic = type_agnostic.count(out.get_node()->get_type_info()) == 1;
+        is_out_type_agnostic = type_agnostic.count(convert->get_users()[0]->get_type_info()) == 1;
```
diyessi, line 110: Again, not all conversions are valid. uint16->uint8 will turn multiples of 256 into 0. In this case intx->inty>=x is okay, unitx->uinty>=x, uintx->inty>x, int->float, int->double, float->double are okay. double->float could get extra 0s due to underflow.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -69,25 +69,54 @@ static bool eliminate_sum(const std::shared_ptr<Node>& node)
...
     }
-    auto convert = as_type_ptr<opset3::Convert>(node);
-    auto input = convert->input_value(0);
-    if (convert->get_convert_element_type() == input.get_element_type() || is_out_type_agnostic)
+
+    if (destination_type == input_op.get_element_type() || is_out_type_agnostic)
```
diyessi, line 113: Same issue as case 1.

src/ngraph/pass/nop_elimination.cpp
swapna-intel: Should we just limit to the following types, which is just the MaskRCNN use cases:
For case 1: ReduceMin -> i64 ->i32->i64

src/ngraph/pass/nop_elimination.cpp
swapna-intel: In this case, we are removing a convert that feeds into and only into a NonZero which returns only the indices of non-zero elements. It could be a narrowing convert and the intent is to make more elements zero. I could add check to remove only the ones MaskRCNN wants. @jane-intel what would be the exact type conversions cases we need ?

src/ngraph/pass/nop_elimination.cpp
swapna-intel: @jane-intel could you please let me know the exact types in MaskRCNN we need to eliminate?
The corresponding graph is case 3. op1-> convert1->convert2-> op2

src/ngraph/pass/nop_elimination.cpp
swapna-intel: updated.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -69,25 +69,57 @@ static bool eliminate_sum(const std::shared_ptr<Node>& node)
...
+ * Eliminates 3 types of spines with converts:
+ *
+ * 1. [ op1--> convert2(type A->B)--> ReduceMin--> convert1(type B->A)--> op2 ] to
+ *                    [ op1--> ReduceMin--> op2 ]
+ * 2. [ op3--> convert1-->  NonZero--> op4 ] to
+ *                    [ op3--> NonZero--> op4 ]
+ * 3. [ op5--> convert1--> convert2--> op6 ] to
+ *                    [op5--> convert2--> op6 ]
+ *
+ **************************/
```
rkimballn1, line 82: Use line comments `//`

swapna-intel:
Split the convert elimination across 2 passes. 1. Original pass nop_elimination eliminates convert with destination_type same as element_type. 2. New pass convert_elimination for MASKRCNN use contains specific eliminations for MASKRCNN

src/ngraph/pass/convert_elimination.cpp
```diff
@@ -0,0 +1,93 @@
...
+//************************
+// Eliminates 2 types of spines with converts:
+//
+// 1. [ op1--> convert2(type i64->i32)--> ReduceMin--> convert1(type i32->i64)--> op2 ] to
+//                    [ op1--> ReduceMin--> op2 ]
+// 2. [ op3--> convert1-->  NonZero--> op4 ] to
+//                    [ op3--> NonZero--> op4 ]
+//
+//*************************
+static bool eliminate_convert(const std::shared_ptr<Node>& node)
```
rkimballn1, line 43: make this a private static member function of ConvertElimination to reduce its scope.

swapna-intel: done

src/ngraph/pass/convert_elimination.cpp
```diff
@@ -0,0 +1,91 @@
...
+    if (is_type<opset3::ReduceMin>(input_op.get_node()))
+    {
+        auto convert_2 = input_op.get_node()->input_value(0).get_node();
+        if (is_type<opset3::Convert>(convert_2) && convert_2->get_users().size() == 1 &&
+            destination_type == convert_2->get_input_element_type(0))
+        {
+            if (destination_type == element::i64 && convert_2->get_element_type() == element::i32)
+            {
+                // replace ReplaceMin as convert's output
+                if (replace_output_update_name(convert->output(0), input_op))
```
GlebKazantaev, line 57: `replace_output_update_name` can be used only in case when you eliminate operation. But here you replace sub-graph convert-reduce-convert with reduce op. 
Please use two different patterns for detecting convert-reduce-convert and convert-nonzero. And when you replace subgraph with node use replace_node() and do not forget about copy_rt_info() and friendly names.

test/nop_elimination.cpp
```diff
@@ -85,10 +86,75 @@ TEST(nop_elimination, convert_type_agnostic)
...
+{
+    auto param = make_shared<op::Parameter>(element::boolean, Shape{3, 2, 4});
+    auto shape_of = make_shared<op::v0::ShapeOf>(param);
+    auto concat = make_shared<op::v0::Concat>(NodeVector{shape_of}, 0);
+    auto convert_1 = make_shared<op::v0::Convert>(concat, element::f32);
+    Shape axes_shape{1};
+    vector<int32_t> values_axes{0};
+    auto constant_axes = op::Constant::create(element::i64, axes_shape, values_axes);
+    auto reduce_min = make_shared<op::v1::ReduceMin>(convert_1, constant_axes);
+    auto convert_2 = make_shared<op::v0::Convert>(reduce_min, element::i32);
```
GlebKazantaev, line 137: please use opset3 namespace instead of internal versioning

src/ngraph/pass/convert_elimination.cpp
swapna-intel: Not sure what you mean by `Please use two different patterns for detecting convert-reduce-convert and convert-nonzero.` We already have an if - else enclosure.

test/nop_elimination.cpp
swapna-intel: Done

PR #4663, opened 2020-04-29
add Clamp reference kernel

adstraw:


src/ngraph/op/fused/clamp.cpp
```diff
@@ -18,12 +18,46 @@
...
+        return true;
+    }
+
+    bool evaluate_clamp(
+        const HostTensorPtr& arg, const HostTensorPtr& out, double min, double max, size_t count)
+    {
+        bool rc = true;
+        switch (arg->get_element_type())
+        {
+            TYPE_CASE(f32)(arg, out, min, max, count);
```
diyessi, line 45: The spec doesn't restrict the input types, just the min/max types.

src/ngraph/runtime/interpreter/int_executable.cpp
```diff
@@ -63,6 +63,15 @@ runtime::interpreter::INTExecutable::INTExecutable(const shared_ptr<Function>& f
     : m_is_compiled{true}
     , m_performance_counters_enabled{enable_performance_collection}
 {
+    auto is_supported = [](const Node& node) {
```
diyessi, line 66: Why is this here? Also, it can be simplified to `is_type<op::v0::Clamp>`

adstraw: This allows the interpreter to run reference kernels for the fused ops as opposed to running the decomposed reference kernels.  Should I remove it or simplify it?

src/ngraph/op/fused/clamp.cpp
adstraw: OK, will fix.  This will force me to static_cast the argument to a `double` for comparison with min/max.  The semantics of this op are strange.  Why not have min/max be of the same type as the argument?

src/ngraph/runtime/reference/clamp.hpp
```diff
@@ -0,0 +1,52 @@
...
+{
+    namespace runtime
+    {
+        namespace reference
+        {
+            template <typename T>
+            void clamp(const T* arg, T* out, double min, double max, size_t count)
+            {
+                T min_t = static_cast<T>(min);
+                T max_t = static_cast<T>(max);
```
adstraw, line 32: Here I am casting min/max (type `double`) to the argument type T.  Information can be lost here.  For example, if min or max are negative and T is unsigned.  Or, if T is less than the size of min/max mantissa e.g. an 8 or 16 bit integer.  I can instead cast the argument to a `double` for comparison, but this doesn't solve the problem as I will still need to case min/max to type T in lines 38 and 42 below.

src/ngraph/op/fused/clamp.cpp
diyessi: They are attributes; we don't have variant types that they could be.

src/ngraph/runtime/reference/clamp.hpp
diyessi: Right -- you need to floor/ceiling for the int types. There's a template way to do that .

src/ngraph/op/fused/clamp.cpp
adstraw: Added support for all types.

src/ngraph/runtime/reference/clamp.hpp
```diff
@@ -0,0 +1,66 @@
...
+                return static_cast<T>(x);
+            }
+
+            template <typename T>
+            void clamp(const T* arg, T* out, double min, double max, size_t count)
+            {
+                for (size_t i = 0; i < count; i++)
+                {
+                    T min_t = double_to_T<T>(min, [](double x) { return std::ceil(x); });
+                    T max_t = double_to_T<T>(max, [](double x) { return std::floor(x); });
```
rkimballn1, line 48: It should be an error if the value of min/max can't exactly convert to integers.

src/ngraph/runtime/reference/clamp.hpp
```diff
@@ -0,0 +1,66 @@
...
+                    double max_t = static_cast<double>(std::numeric_limits<T>::max());
+                    x = std::max(x, min_t);
+                    x = std::min(x, max_t);
+                    x = float_to_int_converter(x);
+                }
+                return static_cast<T>(x);
+            }
+
+            template <typename T>
+            void clamp(const T* arg, T* out, double min, double max, size_t count)
```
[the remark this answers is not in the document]
yimeisun123: Agree with Bob, if T is double, doing ceil/floor adjustment on min/max will potentially give wrong result.

src/ngraph/op/fused/clamp.cpp
```diff
@@ -18,12 +18,62 @@
...
+        default: rc = false; break;
+        }
+        return rc;
+    }
+}
+
+bool op::v0::Clamp::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
+    return evaluate_clamp(
+        inputs[0], outputs[0], get_min(), get_max(), shape_size(get_output_shape(0)));
```
yimeisun123, line 74: output shape may still be partial shape here. I think you need to use the input shape for getting the shape_size here.

src/ngraph/runtime/reference/clamp.hpp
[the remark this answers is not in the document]
adstraw: Fixed.  Reference kernel takes min and max as type T.

src/ngraph/runtime/reference/clamp.hpp
```diff
@@ -0,0 +1,66 @@
...
+        namespace reference
+        {
+            template <typename T>
+            T double_to_T(double x, double float_to_int_converter(double))
+            {
+                if (std::is_integral<T>())
+                {
+                    double min_t = static_cast<double>(std::numeric_limits<T>::min());
+                    double max_t = static_cast<double>(std::numeric_limits<T>::max());
+                    x = std::max(x, min_t);
```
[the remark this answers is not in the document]
adstraw: These conversions are gone given that min/max in the reference kernel are type T.  I will resolve this conversation.  Let's move the question of how to handle "invalid" min/max values to the next conversation.

src/ngraph/runtime/reference/clamp.hpp
adstraw: There are two potential issues when converting from min/max attributes (type double) to type T (matching the type of the argument) when T is an integer.
1. The attribute my not have an integral value
2. The attribute value may be greater than the maximum or less than the minimum value that can be represented by the type T

My solutions:
1. Provide a lambda to convert the attribute to an integral value
2. Return numeric_limits<T>::max() or min() respectively

No errors.  Note that this conversion (call to double_to_int) is now made from src/ngraph/op/fused/clamp.cpp instead of the reference kernel.

src/ngraph/op/fused/clamp.cpp
adstraw: Input shape == output shape so I don't think I can rely on input shape to solve dynamic shape problems, but I take your point that I need to handle a dynamic shape scenario.

yimeisun123: When evaluate is called, the input tensor has the static info, the input shape on input tensor is static at this point.

adstraw: Ahh, thanks.  I learned something new today.  I was going to dig into the code and figure all that out.

adstraw: @yimeisun123 I added some dynamic tests cases and left `get_output_shape` here expecting it to fail.  It didn't.  Not sure what this means.

adstraw: UPDATE:  Had some tests fail with commit 7232e30 that do not fail for me locally.  Not sure if that is due to my build config or what.  Experimenting with get_intput_shape() in commit a52d5b6 while I am debugging.

adstraw: Fixed.

src/ngraph/op/fused/clamp.cpp
```diff
@@ -18,12 +18,105 @@
...
+    template <element::Type_t ET, typename T>
+    bool evaluate(const HostTensorPtr& arg, const HostTensorPtr& out, T min, T max, size_t count)
+    {
+        runtime::reference::clamp<T>(
+            arg->get_data_ptr<ET>(), out->get_data_ptr<ET>(), min, max, count);
+        return true;
+    }
+
+    bool evaluate_clamp(
+        const HostTensorPtr& arg, const HostTensorPtr& out, double min, double max, size_t count)
```
rkimballn1, line 40: min and max should be EITHER type T in a templated function or perhaps better be HostTensor which must be scalar.
If you use type T then T must be the same type as HostTensorPtr.
If you use HostTensor then add a check that all types match.

src/ngraph/op/fused/clamp.cpp
```diff
@@ -18,12 +18,105 @@
...
+    {
+        auto ceil_func = [](double x) { return std::ceil(x); };
+        auto floor_func = [](double x) { return std::floor(x); };
```
rkimballn1, line 43: Why ceil/floor? I have not seen a description of clamp that says you can't do something like 0.5 as min or max. Get rid of ceil/floor.

src/ngraph/op/fused/clamp.cpp
```diff
@@ -18,12 +18,105 @@
...
+
+        bool rc = true;
+        switch (arg->get_element_type())
+        {
+            TYPE_CASE(i8)
+            (arg,
+             out,
+             double_to_int<int8_t>(min, ceil_func),
+             double_to_int<int8_t>(max, floor_func),
```
rkimballn1, line 52: Just pass in min and max directly

src/ngraph/op/fused/clamp.cpp
```diff
@@ -42,13 +135,81 @@ void op::Clamp::pre_validate_and_infer_types()
 NodeVector op::Clamp::decompose_op() const
 {
     const auto data = input_value(0);
-    const auto data_shape = data.get_shape();
+    const auto type = data.get_element_type();
+    const auto shape = data.get_shape();
+
+    auto ceil_func = [](double x) { return std::ceil(x); };
+    auto floor_func = [](double x) { return std::floor(x); };
```
rkimballn1, line 142: Remove these

src/ngraph/op/fused/clamp.cpp
```diff
@@ -42,13 +135,81 @@ void op::Clamp::pre_validate_and_infer_types()
...
+
+    shared_ptr<Node> clamp_min, clamp_max;
 
-    const auto clamp_min = builder::make_constant(data.get_element_type(), data_shape, m_min);
-    const auto clamp_max = builder::make_constant(data.get_element_type(), data_shape, m_max);
+    switch (type)
+    {
+    case element::Type_t::i8:
+    {
+        clamp_min = builder::make_constant(type, shape, double_to_int<int8_t>(m_min, ceil_func));
```
rkimballn1, line 150: Don't call make_constant, just do a make_share<Constant> and then you can pass min/max directly as the correct type. There is a ctor for Constant that takes a scalar as the 3rd arg.

src/ngraph/util.hpp
```diff
@@ -394,6 +394,32 @@ namespace ngraph
...
+        }
+
+        double max_t = static_cast<double>(std::numeric_limits<T>::max());
+        if (x > max_t)
+        {
+            return std::numeric_limits<T>::max();
+        }
+
+        return static_cast<T>(x);
+    }
```
rkimballn1, line 422: Remove this function

test/backend/fused_op.in.cpp
```diff
@@ -1149,49 +1149,724 @@ NGRAPH_TEST(${BACKEND_NAME}, gemm_broadcast_axes_1_input_C)
...
+
+    auto min = numeric_limits<ctype>::min();
+    auto max = numeric_limits<ctype>::max();
+    auto pinf = numeric_limits<double>::infinity();
+    auto ninf = -numeric_limits<double>::infinity();
+
+    vector<ctype> input{min, max, ninf, pinf, 9.99999, 10.0, 10.000001, 19.999999, 20.0, 20.000001};
+
+    // static shape
+    clamp_test<ctype>("${BACKEND_NAME}",
```
rkimballn1, line 1196: I want to see one test with input 0.1, 0.2, ... 1.0 with a min of 0.2 and a max of 0.8 or something like that. That should work.

src/ngraph/op/fused/clamp.cpp
rkimballn1, line 43: You could just do a simple cast instead of this. For integer types the value stored in the double will be an integer.

src/ngraph/op/fused/clamp.cpp
```diff
@@ -42,13 +135,81 @@ void op::Clamp::pre_validate_and_infer_types()
...
 {
     const auto data = input_value(0);
-    const auto data_shape = data.get_shape();
+    const auto type = data.get_element_type();
+    const auto shape = data.get_shape();
+
+    auto ceil_func = [](double x) { return std::ceil(x); };
+    auto floor_func = [](double x) { return std::floor(x); };
+
+    shared_ptr<Node> clamp_min, clamp_max;
```
rkimballn1, line 144:
```suggestion
    shared_ptr<Node> clamp_min;
    shared_ptr<Node> clamp_max;
```

test/util.cpp
```diff
@@ -712,3 +712,56 @@ TEST(util, topological_sort_replace)
...
+    double x = -1.5;
+    EXPECT_TRUE(double_to_int<int32_t>(x, ceil_func) == -1);
+    EXPECT_TRUE(double_to_int<int32_t>(x, floor_func) == -2);
+    EXPECT_TRUE(double_to_int<int32_t>(x, round_func) == -2);
+
+    x = 1.5;
+    EXPECT_TRUE(double_to_int<int32_t>(x, ceil_func) == 2);
+    EXPECT_TRUE(double_to_int<int32_t>(x, floor_func) == 1);
+    EXPECT_TRUE(double_to_int<int32_t>(x, round_func) == 2);
+}
```
rkimballn1, line 767:
```suggestion
}

```

src/ngraph/op/fused/clamp.cpp
adstraw: There is nothing preventing the user from providing non-integral values for min/max when the input is an integer type.  The fact that the op is specified using doubles (attributes) for min/max leaves us with a decision to make in terms of the policy we will use when converting min/max to type T.  Here, by using ceil for min and floor for max, we select the nearest integer values "between" min and max for both min and max.  This may not be the best policy, but it's the one that @diyessi and I discussed and agreed on.

src/ngraph/op/fused/clamp.cpp
adstraw: OK, will fix.

test/backend/fused_op.in.cpp
adstraw: OK, will add.

src/ngraph/op/fused/clamp.cpp
adstraw: Done

test/backend/fused_op.in.cpp
adstraw: Done.

src/ngraph/op/fused/clamp.cpp
adstraw: I added comments to clamp.cpp documenting the policy we take toward converting min/max to integer types.

PR #4664, opened 2020-04-29
add more squeeze unsqueeze optimizations and check gather usecase

ashokei:


src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -726,6 +726,7 @@ static unordered_map<NodeTypeInfo, function<bool(shared_ptr<Node>)>> initialize_
          {op::v0::Multiply::type_info, simplify_multiply},
          {opset3::Gather::type_info, simplify_gather},
          {op::v0::Concat::type_info, simplify_concat},
+         {opset2::ShapeOf::type_info, simplify_gather_shapeof},
```
pruthviIntel, line 729: if we have the same optimization needed for particular Op belonging to different `Opsets` Isn't it , we end up implementing and registering same pass to make it work for different Opsets ?

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -495,8 +496,7 @@ static bool simplify_log(shared_ptr<Node> n)
 // other cases into Concat of shapeof/gather(data) + shapeof(indices)
 static bool simplify_gather_shapeof(shared_ptr<Node> node)
 {
-    auto shapeof = as_type_ptr<opset3::ShapeOf>(node);
-    auto gather = as_type_ptr<opset3::Gather>(shapeof->input_value(0).get_node_shared_ptr());
+    auto gather = as_type_ptr<opset3::Gather>(node->input_value(0).get_node_shared_ptr());
```
pruthviIntel, line 499: Is this Optimization intended to apply only for `gather::v3` ? , what if we need apply the same optimization for `gather::v2`? with `ShapeOf::v2/v3` ?

ashokei: these resolve to underlying op version; This pass does not do any opset upgrade/downgrade, just handles given op.

All new code will try to use latest opset, at the same time support optimizing older opsets (as much as possible). As part of rewriting/creating subgraph we use new opset, even if we are replacing older opset.

pruthviIntel: okay, but if their is new Opset and If we need apply to the same Optimization, we will end up reimplementing isn't it?, since this one will only work for the specific Opset. I understand for the current requirement this works.

ashokei: no, new opset does not mean op version will change. so we may not need to redo this.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -410,7 +411,7 @@ static bool simplify_gather(std::shared_ptr<Node> node)
             // entire input tensor
             std::vector<int64_t> ref_indices(data.get_shape()[axis], 0);
             std::iota(ref_indices.begin(), ref_indices.end(), 0);
-            if (ref_indices == constant_indices->get_vector<int64_t>())
+            if (ref_indices == constant_indices->cast_vector<int64_t>())
```
yimeisun123, line 414: Curious, is there an issue with using get_vector?

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    std::vector<int64_t> axes;
+    size_t i = 0;
+    for (auto o = 0; o < out_shape.rank().get_length(); o++)
+    {
+        if (i < data_shape.rank().get_length() && data_shape[i].same_scheme(out_shape[o]))
+        {
+            i += 1;
+            continue;
+        }
+        if (out_shape[o].is_static() && out_shape[o] == 1)
```
yimeisun123, line 185: What if out_shape[o] is non-static, the axes won't be correct. Does this case exist?

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    std::vector<int64_t> axes;
+    size_t o = 0;
+    for (auto i = 0; i < data_shape.rank().get_length(); i++)
+    {
+        if (o < out_shape.rank().get_length() && data_shape[i].same_scheme(out_shape[o]))
+        {
+            o += 1;
+            continue;
+        }
+        if (data_shape[i].is_static() && data_shape[i] == 1)
```
yimeisun123, line 205: Same questio here. What if data_shape[o] is non-static, the axes won't be correct. Does this case exist?

src/ngraph/pass/nop_elimination.cpp
ashokei: we treat dynamic (non-static) dimension as if they are not "1". So squeeze/unsqueeze will not be done for them.

src/ngraph/pass/algebraic_simplification.cpp
ashokei: if indices is passed as i32, get_vector will conflict as we declare ref_indices as i64, so we explicitly cast

src/ngraph/pass/algebraic_simplification.cpp
diyessi: And if the op does change, the optimization may need to be changed as well since the current optimization may be incorrect for the new semantics.

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -513,7 +513,8 @@ static bool simplify_gather_shapeof(shared_ptr<Node> node)
 
     auto zero_axis = op::Constant::create<int64_t>(element::i64, Shape{}, {0});
     shared_ptr<Node> replace_node;
-    auto new_shapeof = make_shared<opset3::ShapeOf>(gather->input_value(0).get_node_shared_ptr());
+    auto new_shapeof =
+        make_shared<opset3::ShapeOf>(gather->input_value(0).get_node_shared_ptr()->output(0));
```
diyessi, line 517: This should be `gather->input_value(0)`. The input to `gather` isn't necessarily coming from output 0.

diyessi: I am pretty sure the case in lines 520-524 is wrong. The axis isn't removed from the shape, it is a 0 in the shape, and the shape can be a constant, no need for gather,

src/ngraph/pass/algebraic_simplification.cpp
```diff
@@ -534,7 +535,7 @@ static bool simplify_gather_shapeof(shared_ptr<Node> node)
             concat_inputs.push_back(gather);
         }
         auto shapeof_indices =
-            make_shared<opset3::ShapeOf>(gather->input_value(1).get_node_shared_ptr());
+            make_shared<opset3::ShapeOf>(gather->input_value(1).get_node_shared_ptr()->output(0));
```
diyessi, line 538: Again, should just be `input_valie(1)`.

ashokei: we need the first output of the node , this node is the first input of gather. Wasn't sure input_value(1) represents this, will take a look.

src/ngraph/pass/algebraic_simplification.cpp
ashokei: this turns gather->shapeof into shapeof->gather (as shapes can be dynamic, we use shapeof , and gather helps with removing axis dimension ) .

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
 }
 
+static bool replace_squeeze_unsqueeze(const std::shared_ptr<Node>& node)
+{
+    auto out_shape = node->get_output_partial_shape(0);
+    if (!out_shape.is_dynamic() && shape_size(out_shape.get_shape()) != 0)
+    {
+        shared_ptr<Node> reshape;
+        auto input = node->input_value(0).get_node_shared_ptr();
+        auto shape = node->get_shape();
```
diyessi, line 144: `get_shape()` gets the shape from output 0 and gives an error if there's more than one. Also, `get_shape()` will get an error if it's a partial shape. So use `input_value(0).get_partial_shape()` and check that it's static and then get the shape from the partial shape.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+        std::vector<int64_t> vi;
+        vi.assign(shape.begin(), shape.end());
+        auto pat = op::Constant::create<int64_t>(element::i64, Shape{vi.size()}, vi);
+
+        if (as_type_ptr<opset3::Reshape>(input) || as_type_ptr<opset3::Squeeze>(input) ||
```
diyessi, line 149: Use `is_type<...>` when you just want to know if it is of the type.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+            as_type_ptr<opset3::Unsqueeze>(input))
+        {
+            input = input->input_value(0).get_node_shared_ptr();
+            reshape = make_shared<opset3::Reshape>(input->output(0), pat, false);
```
diyessi, line 153: These are all single outputs so `output(0)` works, but it's be safer to use `node->input_value(0)` as a guard against admittedly unlikely future changes that might see a multi-output node show up there.

src/ngraph/pass/nop_elimination.cpp
diyessi: `data_shape.rank()` could be non-static.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    }
+
     auto unsqueeze = as_type_ptr<opset3::Unsqueeze>(node);
     auto input = unsqueeze->input_value(0).get_node_shared_ptr();
+    auto data_shape = input->input_value(0).get_partial_shape();
     auto squeeze = as_type_ptr<opset3::Squeeze>(input);
+    auto replace_unsqueeze_only = [&](const vector<int64_t>& axes) {
+        auto axes_const = op::Constant::create<int64_t>(element::i64, Shape{axes.size()}, axes);
+        auto new_unsq = make_shared<opset3::Unsqueeze>(
+            input->input_value(0).get_node_shared_ptr()->output(0), axes_const);
```
diyessi, line 229: Just `input->input_value(0)`

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+        if (out_shape.rank().get_length() < data_shape.rank().get_length())
+        {
+            // check if single squeeze can handle this
+            auto axes = get_squeeze_axes(data_shape, out_shape);
+            if (data_shape.rank().get_length() - axes.size() == out_shape.rank().get_length())
+            {
+                auto axes_const =
+                    op::Constant::create<int64_t>(element::i64, Shape{axes.size()}, axes);
+                auto new_sq = make_shared<opset3::Squeeze>(
+                    input->input_value(0).get_node_shared_ptr()->output(0), axes_const);
```
diyessi, line 267: Just `input->input_value(0)`

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    }
+
     auto squeeze = as_type_ptr<opset3::Squeeze>(node);
     auto input = squeeze->input_value(0).get_node_shared_ptr();
+    auto data_shape = input->input_value(0).get_partial_shape();
+    auto unsqueeze = as_type_ptr<opset3::Unsqueeze>(input);
+    auto replace_squeeze_only = [&](const vector<int64_t>& axes) {
+        auto axes_const = op::Constant::create<int64_t>(element::i64, Shape{axes.size()}, axes);
+        auto new_sq = make_shared<opset3::Squeeze>(
+            input->input_value(0).get_node_shared_ptr()->output(0), axes_const);
```
diyessi, line 309: `input->input_value(0)`

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+        if (out_shape.rank().get_length() > data_shape.rank().get_length())
+        {
+            // check if single unsqueeze can handle this
+            auto axes = get_unsqueeze_axes(data_shape, out_shape);
+            if (data_shape.rank().get_length() + axes.size() == out_shape.rank().get_length())
+            {
+                auto axes_const =
+                    op::Constant::create<int64_t>(element::i64, Shape{axes.size()}, axes);
+                auto new_unsq = make_shared<opset3::Unsqueeze>(
+                    input->input_value(0).get_node_shared_ptr()->output(0), axes_const);
```
diyessi, line 346: `input->input_value(0)`

src/ngraph/pass/nop_elimination.cpp
ashokei: shape is out_shape, which we check for dynamic shape before the if. will use that here; will remove get_shape.

src/ngraph/pass/nop_elimination.cpp
ashokei: these functions are local utils, only called for static rank.

ashokei:
"If there are any dynamic dimensions they could potentially be 1"

@diyessi these optimizations are similar to (& based-on) partial shape inference, where dynamic dimensions could be 1.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,230 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
 
+static bool replace_squeeze_unsqueeze(const std::shared_ptr<Node>& node)
+{
+    auto shape_ps = node->get_output_partial_shape(0);
+    if (!shape_ps.is_dynamic() && shape_size(shape_ps.get_shape()) != 0)
+    {
+        shared_ptr<Node> reshape;
+        auto input = node->input_value(0).get_node_shared_ptr();
+        auto shape = shape_ps.get_shape();
+        std::vector<int64_t> vi;
```
lazarevevgeny, line 145: You can merge lines 145 and 146 into one:
`std::vector<int64_t> vi(shape.begin(), shape.end());`

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,230 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
     return false;
 }
 
+static bool replace_squeeze_unsqueeze(const std::shared_ptr<Node>& node)
+{
+    auto shape_ps = node->get_output_partial_shape(0);
+    if (!shape_ps.is_dynamic() && shape_size(shape_ps.get_shape()) != 0)
```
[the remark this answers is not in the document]
jane-intel: IDK if we really need to

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,230 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+        }
+    }
+    return axes;
+}
+
+static std::vector<int64_t> get_squeeze_axes(const PartialShape& data_shape,
+                                             const PartialShape& out_shape)
+{
+    std::vector<int64_t> axes;
+    size_t o = 0;
```
jane-intel, line 196: nit: naming

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,235 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
     return false;
 }
 
+static bool replace_squeeze_unsqueeze(const std::shared_ptr<Node>& node)
+{
+    auto out_shape = node->get_output_partial_shape(0);
+    if (!out_shape.is_dynamic() && shape_size(out_shape.get_shape()) != 0)
+    {
+        shared_ptr<Node> reshape;
+        auto input = node->input_value(0).get_node_shared_ptr();
```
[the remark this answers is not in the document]
ashokei: for node, as this is only used for checking input node type (reshape/squeeze/unsqueeze). It's input_value is safe to use as these are all single out nodes.

src/ngraph/pass/nop_elimination.cpp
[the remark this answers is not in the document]
ashokei: added support for unknown dim using -1 placeholder

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,264 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+}
+
+static bool replace_squeeze_unsqueeze(const std::shared_ptr<Node>& node)
+{
+    auto shape_ps = node->get_output_partial_shape(0);
+    if (shape_ps.rank().get_length() == 0)
+    {
+        return false;
+    }
+    size_t num_unknown = 0;
```
lazarevevgeny, line 161: No need to introduce the variable  `num_unknown`. Just do the following:
`if (count_unknown_dims(shape_ps) > 1)
    return false;
`

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,264 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    std::vector<int64_t> vi;
```
lazarevevgeny, line 162: Lets rename `vi` to something more meaningful like `target_shape`

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,263 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    {
+        reshape = make_shared<opset3::Reshape>(input->input_value(0), pat, false);
+    }
+    else
+    {
+        reshape = make_shared<opset3::Reshape>(node->input_value(0), pat, false);
+    }
+
+    // skip if reshape is nop
+    if (reshape->get_input_partial_shape(0).same_scheme(shape_ps))
```
lazarevevgeny, line 194: I don't like the idea that we have a nop Reshape elimination here. This transformation should not care about this since we should have a separate transformation which will remove all redundant reshapes. 

If there are no more other significant comments then please ignore this one and proceed with the merge.

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,263 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    {
+        return replace_squeeze_unsqueeze(node);
+    }
+
     auto unsqueeze = as_type_ptr<opset3::Unsqueeze>(node);
     auto input = unsqueeze->input_value(0).get_node_shared_ptr();
+    auto data_shape = input->input_value(0).get_partial_shape();
     auto squeeze = as_type_ptr<opset3::Squeeze>(input);
+    auto replace_unsqueeze_only = [&](const vector<int64_t>& axes) {
+        auto axes_const = op::Constant::create<int64_t>(element::i64, Shape{axes.size()}, axes);
```
lazarevevgeny, line 259: I am not sure how important this but should we use v3::Constant everywhere instead of op::Constant?

@diyessi @GlebKazantaev could you comment?

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,263 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    auto shape_ps = node->get_output_partial_shape(0);
+    if (shape_ps.rank().get_length() == 0)
+    {
+        return false;
+    }
+    if (count_unknown_dims(shape_ps) > 1)
+    {
+        return false;
+    }
+    std::vector<int64_t> target_shape;
```
GlebKazantaev, line 165: Why not to use ngraph::Shape here?

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,263 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+        else
+        {
+            target_shape.emplace_back(shape_ps[i]);
+        }
+    }
+
+    shared_ptr<Node> reshape;
+    auto input = node->input_value(0).get_node_shared_ptr();
+    auto pat =
+        op::Constant::create<int64_t>(element::i64, Shape{target_shape.size()}, target_shape);
```
GlebKazantaev, line 181: opset3

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,263 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    if (shape_ps.rank().get_length() == 0)
+    {
+        return false;
+    }
+    if (count_unknown_dims(shape_ps) > 1)
+    {
+        return false;
+    }
+    std::vector<int64_t> target_shape;
+    for (auto i = 0; i < shape_ps.rank().get_length(); i++)
```
GlebKazantaev, line 166: Why don't we have iterators for PartialShape? may be there is a reason? It brings so many burden that could avid with iterators...
`std::transform(s.begin(), s.end(), std::back_inserter(shape), [](const ngraph::Dimension & dim) -> ngraph::Dimension { return dim.is_dynamic() ? -1 : dim; });`

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,263 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+    }
+    else
+    {
+        reshape = make_shared<opset3::Reshape>(node->input_value(0), pat, false);
+    }
+
+    // skip if reshape is nop
+    if (reshape->get_input_partial_shape(0).same_scheme(shape_ps))
+    {
+        return replace_output_update_name(node->output(0), reshape->input_value(0));
```
GlebKazantaev, line 196: The same comment about `replace_output_update_name` usage

src/ngraph/pass/nop_elimination.cpp
```diff
@@ -134,62 +134,263 @@ static bool eliminate_reshape_v1(const std::shared_ptr<Node>& node)
...
+            rc += 1;
+        }
+    }
+    return rc;
+}
+
+static bool replace_squeeze_unsqueeze(const std::shared_ptr<Node>& node)
+{
+    auto shape_ps = node->get_output_partial_shape(0);
+    if (shape_ps.rank().get_length() == 0)
```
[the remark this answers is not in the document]
ashokei: this function is local utility and only called for non-dynamic rank.

src/ngraph/pass/nop_elimination.cpp
ashokei: this is 2nd argument to reshape, it expects a node (target_shape is used to create it)

src/ngraph/pass/nop_elimination.cpp
ashokei: perhaps this feature can be added as another PR.

src/ngraph/graph_util.cpp
```diff
@@ -912,3 +912,20 @@ bool ngraph::replace_output_update_name(Output<Node> output, const Output<Node>&
...
     return false;
 }
+
+bool ngraph::replace_node_update_name(std::shared_ptr<Node> target,
+                                      std::shared_ptr<Node> replacement)
+{
+    for (auto& output : target->output(0).get_target_inputs())
+    {
+        if (as_type<ngraph::op::Parameter>(replacement->input_value(0).get_node()) &&
+            as_type<op::Result>(output.get_node()))
```
GlebKazantaev, line 922: No, we can replace node with another node even if output is Result. 
Just keep this:
```suggestion
    replace_node(target, replacement);
    replacement->set_friendly_name(target->get_friendly_name());
    copy_runtime_info(target, replacement);
    return true;
```

GlebKazantaev:
@ashokei I've left you one last comment but approved in advance.

src/ngraph/graph_util.cpp
ashokei: that is only a check to avoid replacement for single layer graph (to help OV unittests).

GlebKazantaev: If you replace smth with node you won't have empty graph.

ashokei: yes, but we do not want this for single layer graph as this will fail OV unittests. (eg: param->squeeze->result becomes param->reshape->result, which fails OV unittests).

PR #4665, opened 2020-04-30
Constant Folding using Evaluate method for reshape::v0/v1

pruthviIntel:


src/ngraph/op/reshape.cpp
```diff
@@ -298,3 +362,9 @@ void op::v1::Reshape::generate_adjoints(autodiff::Adjoints& /* adjoints */,
 {
     throw ngraph_error("generate_adjoints not implemented for Reshape");
 }
+
+bool op::v1::Reshape::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
```
diyessi, line 367: `v1::Reshape` is just a memcpy of all the bytes from the input to the output.
However, with `v1::Reshape` the output shape is an input which may have a -1 in it. If there is a -1, you need to figure out the actual length of that axis. In all cases, you need to set the actual output shape/element type before using it.

src/ngraph/op/reshape.cpp
```diff
@@ -298,3 +360,135 @@ void op::v1::Reshape::generate_adjoints(autodiff::Adjoints& /* adjoints */,
...
+    // infer and set output shape if the output shape contain -1
+    // and zero value dimension
+    size_t output_rank = inputs[1]->get_shape()[0];
+    Shape out_shape_val;
+
+#define COMPUTE_OUPUT_SHAPE(T)                                                                     \
```
diyessi, line 371: Use a template and base it on `element_type_traits` to ensure that the the C++ type is the correct one for the element type.

PR #4666, opened 2020-04-30
Attribute visitor doc, simplification

diyessi:
The virtual `AttributeVisitor::on_attribute` for string and bool are deprecated. Use the equivalent `on_adapter` methods.
Related to #4670


src/ngraph/attribute_visitor.hpp
```diff
@@ -27,136 +28,125 @@ namespace ngraph
...
+        /// \brief Start visiting a nested structure
+        virtual void start_structure(const std::string& name);
+        /// \brief Finish visiting a nested structure
+        virtual void finish_structure();
+        /// \brief Get the id for a pointer
+        virtual int64_t get_id(void*);
+        /// \brief Retrieve a pointer from the id table
+        virtual void* get_ptr(int64_t id);
+        /// \brief Add a pointer to the id table
+        virtual int64_t insert_ptr(void*);
```
arogowie-intel, line 145: What are those methods for? Some usage scenario example would be helpful.

src/ngraph/attribute_adapter.hpp
```diff
@@ -23,278 +23,387 @@
...
+        void set(const VAT& value) override
+        {
+            m_ref = value;
+            m_buffer_valid = false;
+        }
+
+    protected:
+        AT& m_ref;
+        VAT m_buffer;
+        bool m_buffer_valid{false};
```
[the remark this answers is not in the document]
diyessi: I want to get rid of the base class ones; they are only used by some adapters.

src/ngraph/attribute_visitor.hpp
diyessi: Since nodes are deserialized a node at a time from formats that have their own way to interconnect nodes, this (slightly different now) is a way for the serialization format to associate its ids with previously processed nodes.

python/src/pyngraph/node_factory.cpp
```diff
@@ -41,26 +41,19 @@ class DictAttributeDeserializer : public ngraph::AttributeVisitor
     {
     }
 
-    virtual void on_attribute(const std::string& name, std::string& value) override
```
ilyachur, line 44: Why did you remove the adapter for string?

src/ngraph/op/tensor_iterator.cpp
```diff
@@ -118,41 +161,184 @@ op::TensorIterator::ConcatOutputDescription::ConcatOutputDescription(uint64_t bo
...
+{
+    // We assume the ops in body lambda are alread registered
+    ParameterVector body_parameters;
+    ResultVector body_results;
+    if (m_body)
+    {
+        body_parameters = m_body->get_parameters();
+        body_results = m_body->get_results();
+    }
+    visitor.on_attribute("body_parameters", body_parameters);
```
ilyachur, line 316: `TensorIterator` doesn't contain "body_parameters" attribute (https://docs.openvinotoolkit.org/latest/_docs_ops_infrastructure_TensorIterator_1.html). I don't fully understand how this method read the body attribute.

src/ngraph/attribute_visitor.hpp
```diff
@@ -27,137 +28,125 @@ namespace ngraph
...
+        /// The generic visitor. There must be a definition of AttributeAdapter<T> that can convert
+        /// to a ValueAccessor<U> for one of the on_adpater methods.
+        template <typename AT>
+        void on_attribute(const std::string& name, AT& value)
+        {
+            AttributeAdapter<AT> adapter(value);
+            on_adapter(get_name_with_context(name), adapter);
+        }
+        /// \returns The nested context of visits
+        const std::vector<std::string>& get_context() const { return m_context; }
```
ilyachur, line 128: Context, what is it?

src/ngraph/attribute_visitor.hpp
```diff
@@ -27,137 +28,125 @@ namespace ngraph
...
+        virtual void start_structure(const std::string& name);
+        /// \brief Finish visiting a nested structure
+        virtual void finish_structure();
+        using node_id_t = std::string;
+        static const node_id_t invalid_node_id;
+        /// \brief Associate a node with an id.
+        ///
+        /// No node may be used as an attribute unless it has already been registered with an ID.
+        /// References to nodes are visited with a ValueAccessor of their ID.
+        virtual void register_node(const std::shared_ptr<Node>& node, node_id_t id);
```
ilyachur, line 141: Do we need to register all nodes for the TensorIterator?

src/ngraph/attribute_adapter.cpp
```diff
@@ -31,388 +31,32 @@ using namespace ngraph;
 
 namespace ngraph
 {
-    // float
     constexpr DiscreteTypeInfo AttributeAdapter<float>::type_info;
-    const double& AttributeAdapter<float>::get()
```
ilyachur, line 36: Why did you remove all `AttibuteAdapter`s?

python/src/pyngraph/node_factory.cpp
diyessi: The `on_adapter` for string remains. Anything that turns into a string, like enums, goes through `on_adapter`.

src/ngraph/op/tensor_iterator.cpp
diyessi: How do you get the results/parameters associated with the body? I need to make sure there's a way for you to do that.

src/ngraph/attribute_visitor.hpp
diyessi: All nodes used in the body. If you deserialize them before TensorIterator and register them as you deserialize with whatever ID you use to refer to them in the body, everything will work.

src/ngraph/attribute_visitor.hpp
diyessi: Suppose an X op has a `before_pad_spec` and an `after_pad_spec` and both are some struct `XPadSpec`. There'd be `visitor.on_attribute("before_pad_spec", m_before_pad_spec")` and that would get to the pad spec adapter which would do `visitor.on_fill("fill", m_fill")` and `visitor.on_size("size", m_size)`. The context would be `["before_pad_spec"]`. When we get to `after_pad_spec` the context would be `["after_pad_spec"]`.  If there were a vector of pad specs, you could have contexts like `["before_pad_spec", "0"]`, `["before_pad_spec", "1"]`, etc. We need to make sure however contexts work matches up with however your XML reading works since I think the XML format is not something that can be changed.

src/ngraph/attribute_adapter.cpp
diyessi: They are still there; same API, simpler template definition in the header.

src/ngraph/op/tensor_iterator.cpp
ilyachur: In order to get results/parameters for body, in IR reader we just parse all body and operations with Result type are body results, the operations with Parameter type is body parameters.

src/ngraph/attribute_visitor.hpp
ilyachur: Is it means that I should detect operations in the TI body for each topology and register them? If yes it is not flexible.
How should I detect the set of operations that TI for some topology contains? AlsoTI can contain some custom operations.

diyessi: In the spec for TensorIterator, each layer in the body has an ID, so those are the IDs. How do you plan to deserialize those nodes in the body and connect their inputs? The same way you deserialize the other nodes in the graph, or does the visitor deserialize them? That is the question. It doesn't matter if the nodes are custom.

ilyachur: Could you please provide a test or more detailed explanation which will illustrate the way how we can create the TI layer?

In order to be on the same page... At the current moment I really don't fully understand the way, how we can create TI from Visitors API.

src/ngraph/axis_vector.hpp
```diff
@@ -51,6 +52,25 @@ namespace ngraph
         NGRAPH_API AxisVector& operator=(AxisVector&& v) noexcept;
     };
 
+    template <>
+    class NGRAPH_API AttributeAdapter<AxisVector> : public ValueAccessor<std::vector<int64_t>>
```
arogowie-intel, line 56: Why not using `IndirectVectorValueAccessor` ?

diyessi: Good idea.

src/ngraph/attribute_adapter.hpp
```diff
@@ -23,278 +23,387 @@
...
         {
         }
 
         static constexpr DiscreteTypeInfo type_info{"AttributeAdapter<double>", 0};
         const DiscreteTypeInfo& get_type_info() const override { return type_info; }
-        const double& get() override;
-        void set(const double& value) override;
     };
 
     /// \brief Access a bool as a bool
```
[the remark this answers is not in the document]
diyessi: Got it and a few others.

src/ngraph/axis_vector.hpp
diyessi: Done.

PR #4668, opened 2020-04-30
EmbeddingBagOffsetsSums op shell

nbpatel:
The Op spec 
https://gitlab-icv.inn.intel.com/inference-engine/dldt/blob/1c91942962d48c7c87c393a63fd1555fe4a8a1bf/docs/ops/sparse/EmbeddingBagOffsetsSum_3.md


src/ngraph/op/embeddingbag_offsets_sum.cpp
```diff
@@ -0,0 +1,167 @@
...
+                              get_input_partial_shape(DEFAULT_INDEX).compatible(PartialShape{}),
+                              "DEFAULT_INDEX must be a scalar");
+    }
+
+    element::Type result_et = get_input_element_type(EMB_TABLE);
+
+    const PartialShape& emb_table_shape = get_input_partial_shape(EMB_TABLE);
+    const PartialShape& offsets_shape = get_input_partial_shape(OFFSETS);
+
+    PartialShape result_shape;
```
diyessi, line 141: Why don't you add a `PartialShape(const Rank&)` constructor and then you (and others) can skip doing things like `result_dims`

src/ngraph/op/embeddingbag_offsets_sum.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    set_output_type(0, result_et, result_shape);
+}
+
+shared_ptr<Node>
+    op::v3::EmbeddingBagOffsetsSum::clone_with_new_inputs(const OutputVector& new_args) const
+{
+    check_new_args_count(this, new_args);
+    return make_shared<EmbeddingBagOffsetsSum>(
```
diyessi, line 165: You need one constructor here for each of the number of args.

src/ngraph/op/embeddingbag_offsets_sum.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include "ngraph/op/util/index_reduction.hpp"
+
+namespace ngraph
+{
+    namespace op
+    {
+        namespace v3
+        {
+            /// \brief Returns embeddings for given indices
+            class NGRAPH_API EmbeddingBagOffsetsSum : public Op
```
lazarevevgeny, line 29: Again, we need a base class for all EmbeddingBagOffsetsXXX ops.

src/ngraph/op/embeddingbag_offsets_sum.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+op::v3::EmbeddingBagOffsetsSum::EmbeddingBagOffsetsSum(const Output<Node>& emb_table,
+                                                       const Output<Node>& indices,
+                                                       const Output<Node>& offsets)
+    : Op({emb_table, indices, offsets})
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::v3::EmbeddingBagOffsetsSum::validate_and_infer_types()
```
lazarevevgeny, line 52: Missing visitor

nbpatel: visitor is only for attributes, isn't it?

lazarevevgeny: Correct. But even if the op does not have attributes we need to implement the method and return "true" so the whole system knows that the visitor is implemented and there are no attributes in the op.

@diyessi, please, correct me if I am wrong.

src/ngraph/op/util/embeddingbag_offsets_base.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    element::Type result_et = get_input_element_type(EMB_TABLE);
+
+    const PartialShape& emb_table_shape = get_input_partial_shape(EMB_TABLE);
+    const PartialShape& offsets_shape = get_input_partial_shape(OFFSETS);
+
+    PartialShape result_shape;
+    if (emb_table_shape.rank().is_static())
+    {
+        std::vector<Dimension> result_dims(emb_table_shape.rank().get_length());
```
lazarevevgeny, line 135: The same comment as in another PR with EmbeddingBagPackedSum

src/ngraph/op/embeddingbag_offsets_sum.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+}
+
+shared_ptr<Node>
+    op::v3::EmbeddingBagOffsetsSum::clone_with_new_inputs(const OutputVector& new_args) const
+{
+    check_new_args_count(this, new_args);
+    if (new_args.size() == 3)
+    {
+        return make_shared<EmbeddingBagOffsetsSum>(new_args.at(0), new_args.at(1), new_args.at(2));
```
lazarevevgeny, line 55: Use `op::v3::EmbeddingBagOffsetsSum` instead of just `EmbeddingBagOffsetsSum `

PR #4670, opened 2020-04-30
[Py] API for ops with struct attributes + visitor.

arogowie-intel:
Ad `visit_attributes` as well as Python API for following operators:

- [x] PriorBox
- [x] PriorBoxClustered
- [x] DetectionOutput
- [x] Interpolate
- [x] Proposal
- [ ] TensorIterator

This PR requires changes from #4666. That branch is already merged into this, thus there is a lot off changed stuff not directly related to this PR.


python/src/ngraph/utils/input_validation.py
```diff
@@ -33,3 +34,104 @@ def assert_list_of_ints(value_list, message):  # type: (Iterable[int], str) -> N
...
+    :param      op_name:        The operator name which attributes are checked.
+    :param      attr_key:       The attribute name.
+    :param      value:          The value to check.
+    :param      val_type:       Required value type.
+    :param      cond:           The optional function running additional checks.
+
+    :raises     UserInputError:
+    :return:    True if attribute satisfies all criterias. Otherwise False.
+    """
+    if not np.issubdtype(type(value), val_type):
```
postrational, line 52: I'm not sure the type checking here will work as you expect.
```
>>> a = np.array([123], dtype=np.int32)
>>> np.issubdtype(type(a), np.int32)
False
```

arogowie-intel: The `_check_value` function is intended to be used only with scalar values and _internally_ it is so. Moreover this function is _hidden_ since it's name starts with an underscore. Do you want to enforce scalar input in this function?

PR #4675, opened 2020-05-01
nGraph v3 op ExtractImagePatches

noim210:
This is shell of ExtractImagePatches which is close the following tensorflow op:
https://www.tensorflow.org/api_docs/python/tf/image/extract_patches

Actual op spec for OpenVino can be found from here:
https://gitlab-icv.inn.intel.com/inference-engine/dldt/blob/6a6a6ffd59d3188abeb2e1de8476132b3e52e20c/docs/ops/movement/ExtractImagePatches_3.md

In this page, the output spec is not fully cleared. I tried to mimic the behavior of extract_patches of tensorflow. As OpenVion spec requirement said that input is a 4D tensor. I created this op for 4D input.


src/ngraph/op/extractimagepatches.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include "ngraph/op/op.hpp"
+
+namespace ngraph
+{
+    namespace op
+    {
+        namespace v3
+        {
+            typedef struct
```
diyessi, line 27: Why is this a struct instead of the four attributes?

src/ngraph/op/extractimagepatches.hpp
```diff
@@ -0,0 +1,117 @@
...
+                bool visit_attributes(AttributeVisitor& visitor) override;
+
+                virtual std::shared_ptr<Node>
+                    clone_with_new_inputs(const OutputVector& new_args) const override;
+
+                const ExtractImagePatchesAttrs& get_attrs() const { return m_attrs; }
+                void set_attrs(const Shape& sizes,
+                               const Strides& strides,
+                               const Shape& rates,
+                               const std::string& str_padding)
```
diyessi, line 58: Use the enum for padding, not a string. 
Unless there's some reason the attributes need to be packaged up as a struct, give each a set/get method.

src/ngraph/op/extractimagepatches.cpp
```diff
@@ -0,0 +1,106 @@
...
+    : Op({image})
+    , m_attrs(attrs)
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::v3::ExtractImagePatches::validate_and_infer_types()
+{
+    Shape input_shape = get_input_shape(0);
+
```
diyessi, line 38: The input shape could be dynamic, in which case you make the output shape dynamic and skip the shape-based checking.

test/type_prop/extractimagepatches.cpp
```diff
@@ -0,0 +1,86 @@
...
+// limitations under the License.
+//*****************************************************************************
+
+#include "gtest/gtest.h"
+#include "ngraph/ngraph.hpp"
+#include "util/type_prop.hpp"
+
+using namespace std;
+using namespace ngraph;
+
```
diyessi, line 23: Add tests for dynamic shape input.

test/type_prop/extractimagepatches.cpp
```diff
@@ -0,0 +1,86 @@
...
+    auto sizes = Shape{3, 3};
+    auto strides = Strides{5, 5};
+    auto rates = Shape{1, 1};
+    auto padding = string("valid");
+    auto padtype_padding = op::PadType::VALID;
+    auto eip_attributes =
+        op::v3::ExtractImagePatches::CreateExtractImagePatchesAttrs(sizes, strides, rates, padding);
+    auto extractimagepatches = make_shared<op::v3::ExtractImagePatches>(data, eip_attributes);
+
+    EXPECT_EQ(extractimagepatches->get_element_type(), element::i32);
```
diyessi, line 36: Use `get_output_element_type(0)` here and below.

src/ngraph/op/extractimagepatches.cpp
yimeisun123: get_partial_shape, then, check if it is static or not to determine the next action.

src/ngraph/op/extractimagepatches.hpp
noim210: i was following interpolate and after reading the team's wiki I misunderstood and thought It is desired to make struct. 
Do you recommend to change it to remove struct? In that case I will update the code.

src/ngraph/op/extractimagepatches.hpp
noim210: My first implementation was using enum padding, however later I saw in the following link that  the attribute "padding" is string. 
https://gitlab-icv.inn.intel.com/inference-engine/dldt/blob/6a6a6ffd59d3188abeb2e1de8476132b3e52e20c/docs/ops/movement/ExtractImagePatches_3.md

That's why i changed it to string.

src/ngraph/op/extractimagepatches.cpp
noim210: thanks I will update this

src/ngraph/op/extractimagepatches.hpp
noim210: Thanks i got rid of "struct" type of attribute.

src/ngraph/op/extractimagepatches.hpp
noim210: Thanks,  I used enum only.

test/type_prop/extractimagepatches.cpp
noim210: added dynamic shape test

src/ngraph/op/extractimagepatches.cpp
```diff
@@ -0,0 +1,124 @@
...
+    if (input_Pshape.is_dynamic())
+    {
+        set_input_is_relevant_to_shape(0);
+        auto output_Pshape = PartialShape::dynamic(4);
+        set_output_type(0, get_input_element_type(0), output_Pshape);
+    }
+    else
+    {
+        Shape input_shape = get_input_shape(0); // as input shape is static
+        size_t out_rows((input_shape[2] - 1) / m_patch_movement_strides[0]);
```
diyessi, line 77: What padding type is this for? You overwrite padding below for VALID. The spec says for same upper/lower rows/cols match output.

If input_shape[2] were unsigned 0, subtracting 1 will give a very large number. I also think this gives the wrong values. If stride is 2, shape of 0->0, 1, 2->1, 3, 4->2, so (shape+stride-1)/stride

Also need to make sure strides are not 0.

src/ngraph/op/extractimagepatches.cpp
```diff
@@ -0,0 +1,124 @@
...
+        size_t out_cols((input_shape[3] - 1) / m_patch_movement_strides[1]);
+        if (m_padding == PadType::VALID)
+        {
+            out_rows =
```
diyessi, line 81: Why compute the valid paddings and then throw them away? Switch on padding type instead.

noim210: thanks  i will do that.

src/ngraph/op/extractimagepatches.cpp
noim210: Thank you for discussing these thing with me, 
i will check following things:
so i will these 4 things:
   1. zero stride check
   2. stride > patch size check
   3. in_rows and in_cols zero check?
   4. Double check the dimension result with convolution and pooling validation_utils

test/attributes.cpp
```diff
@@ -1377,3 +1377,25 @@ TEST(attributes, logical_xor_op)
...
+
+TEST(attributes, extractimagepatches_op)
+{
+    FactoryRegistry<Node>::get().register_factory<opset3::ExtractImagePatches>();
+    auto data = make_shared<op::Parameter>(element::i32, Shape{64, 3, 10, 10});
+
+    auto sizes = Shape{3, 3};
+    auto strides = Strides{5, 5};
+    auto rates = Shape{1, 1};
+    // auto padding = string("valid");
```
lazarevevgeny, line 1389: Please, remove

src/ngraph/op/extractimagepatches.cpp
noim210: I updated the code.  none of the code doesn't have rates, so i used my tensorflow code to check the result of the dimensions

src/ngraph/op/extractimagepatches.cpp
noim210: I updated the code according to your suggestions

test/attributes.cpp
noim210: I removed the unnecessary comment portions. 
About finalizing the operations, you can use the tensor flow code I used to check the behavior, I am sharing below:

import numpy as np;                                                                                                                                      
import tensorflow as tf;
import sys;
def tprint(a):
    with tf.Session() as sess:  print(a.eval())
n=8;
images = [[[[x * n + y + 1] for y in range(n)] for x in range(n)]]
output = tf.extract_image_patches(images=images,
                       sizes=[1, 3, 3, 1],
                       strides=[1, 5, 5, 1],
                       rates=[1, 2, 2, 1],
                       padding='VALID')
                       #padding='SAME')

print("printing output: ");
print(output)
print("printing tensor: ");
tprint(output);

src/ngraph/op/extractimagepatches.hpp
```diff
@@ -0,0 +1,70 @@
...
+                /// \param data 4-D Input data to extract image patches
+                /// \param sizes Patch size in the format of [size_rows, size_cols]
+                /// \param strides Patch movement stride in the format of [stride_rows, stride_cols]
+                /// \param rates Element seleciton rate for creating a patch. in the format of
+                /// [rate_rows, rate_cols] \param padding Padding type. it can be any value from
+                /// valid, same_lower, same_upper
+                ExtractImagePatches(const Output<Node>& data,
+                                    const Shape sizes,
+                                    const Strides strides,
+                                    const Shape rates,
```
yimeisun123, line 44: Why not use const Shape&, const Strides&

src/ngraph/op/extractimagepatches.hpp
```diff
@@ -0,0 +1,70 @@
...
+                                    const PadType padding);
+
+                void validate_and_infer_types() override;
+                bool visit_attributes(AttributeVisitor& visitor) override;
+
+                virtual std::shared_ptr<Node>
+                    clone_with_new_inputs(const OutputVector& new_args) const override;
+
+                const Shape& get_sizes() const { return m_patch_sizes; }
+                void set_sizes(const Shape _sizes) { m_patch_sizes = _sizes; }
```
yimeisun123, line 54: const Shape& ? 
also, suggest to use "sizes" instead of "_sizes" just to be consistant with the style. 
Same suggestions for the other setters.

noim210: sure, I am updating these.

noim210: about  "const Shape& " as return value, this will be rvalue while returning the result to a function. This will make the code run faster.
I followed this from "max_pool" , "avg_pool",  "roi_pool" & "broadcast". 

Thank you very much for the feedback. 
I updated the issues of the naming according to your suggestion.

src/ngraph/op/extractimagepatches.hpp
noim210: oh, I misunderstood your comment, I will fix it. Thank you.

src/ngraph/op/extractimagepatches.hpp
noim210: I misunderstood your comment, I fixed it now according to your suggestion. Thank you.

yimeisun123:
Need to run "make style" to get rid of the style check error.

test/type_prop/extractimagepatches.cpp
```diff
@@ -0,0 +1,93 @@
...
+    auto rates = Shape{2, 2};
+    auto padtype_padding = op::PadType::VALID;
+    auto extractimagepatches =
+        make_shared<op::v3::ExtractImagePatches>(data, sizes, strides, rates, padtype_padding);
+
+    EXPECT_EQ(extractimagepatches->get_output_element_type(0), element::i32);
+    EXPECT_EQ(extractimagepatches->get_output_shape(0), (Shape{64, 27, 1, 1}));
+}
+
+TEST(type_prop, extractimagepatches_dynamic_shape)
```
lazarevevgeny, line 80: The overall review is good. Probably we may change the specification soon in terms of attribute "padding". We will decide soon on should we support value "same" or "same_lower" + "same_upper". Once we do this, please, add additional tests for these case(s).

noim210: thanks for the feed back. and sure let me know the final spec, i will add new tests.

lazarevevgeny: The final update of the specification https://gitlab-icv.inn.intel.com/inference-engine/dldt/merge_requests/7105/diffs will be done within a couple of hours.

noim210: thanks, I am going to add two more tests and merge

noim210: I have added actually 4 new tests : 2 for same_upper and 2 for same_lower.

test/attributes.cpp
noim210: I think the specs are finalized now. Can you accept this "review request",  I can't merge because of this requested review.

src/ngraph/op/extractimagepatches.cpp
```diff
@@ -0,0 +1,157 @@
...
+        set_output_type(0, get_input_element_type(0), output_shape);
+    }
+}
+
+bool op::v3::ExtractImagePatches::visit_attributes(AttributeVisitor& visitor)
+{
+    visitor.on_attribute("sizes", m_patch_sizes);
+    visitor.on_attribute("strides", m_patch_movement_strides);
+    visitor.on_attribute("rates", m_patch_selection_rates);
+    visitor.on_attribute("padding", m_padding);
```
lazarevevgeny, line 144: The attribute has been renamed to "auto_pad"

src/ngraph/op/extractimagepatches.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+    NODE_VALIDATION_CHECK(
+        this,
+        m_padding == PadType::VALID || m_padding == PadType::SAME_LOWER ||
+            m_padding == PadType::SAME_UPPER,
+        "Attribute padding should be in either valid or same_lower or same_upper.");
+
+    if (input_Pshape.is_dynamic())
+    {
+        set_input_is_relevant_to_shape(0);
```
lazarevevgeny, line 78: If only the batch dimension is dynamic we still can calculate the 3 dimensions out of 4. So please, change the condition in line 76 so the "else" branch will work for case when the only batch dimension is dynamic. Also add a unit test for this case.

src/ngraph/op/extractimagepatches.cpp
```diff
@@ -0,0 +1,157 @@
...
+        auto output_Pshape = PartialShape::dynamic(4);
+        set_output_type(0, get_input_element_type(0), output_Pshape);
+    }
+    else
+    {
+        Shape input_shape = get_input_shape(0); // as input shape is static
+        size_t out_rows(0);
+        size_t out_cols(0);
+
+        if (m_patch_movement_strides[0] > input_shape[2] ||
```
lazarevevgeny, line 88: Is this case really possible? The same comment for line 102. If they are possible then we need to have a unit test for each of them.

src/ngraph/op/extractimagepatches.cpp
```diff
@@ -0,0 +1,157 @@
...
+            m_patch_movement_strides[1] > input_shape[3])
+        {
+            if (m_padding == PadType::VALID)
+            {
+                out_rows = 0; // no padding allowed for VALID
```
lazarevevgeny, line 93: I don't quite understand why the out_rows and out_cols are 0 in this case. If the rates value is bigger than the corresponding input dimension then the first element will still have index 0 in input so it will be in the output patch. The rest of the input elements will be in position m_patch_movement_strides[0] * i > input_shape[2] so they will not be in the output patch. So we will have out_rows = out_cols = 1.

src/ngraph/op/extractimagepatches.cpp
noim210: Thanks. updated the name.

src/ngraph/op/extractimagepatches.cpp
noim210: fixed according to your suggestion. Thanks.

src/ngraph/op/extractimagepatches.cpp
noim210: These two "corner" cases are suggested by Scott. Actual questions were:
1. what happen if stride length is larger than data dimensions
2. what will happen if data is blank.

However, you are right, it looks like I don't need to have special case for these. (explain in the reply of your next comment.

I fixed the code calculations for large stride but smaller sizes.
According to your suggestion, I have added more test cases.

src/ngraph/op/extractimagepatches.cpp
noim210: Thanks for showing me the bug. I solve the issues. 
You are right. In case of larger stride, the actual equation will handle this scenario. However, there might a corner case when rates are very larger than the input dimension, in that case there is a possibility my out_rows and out_cols become negative in padding::valid case (the Padding::valid should give (0,0) and same_lower and same_upper should result (1,1) in this situation)
I fixed this by using signed int for out_rows and out_cols and later check it's negativity. Before adding this to output_shape, it was type_cast-ed to size_t.

PR #4676, opened 2020-05-01
EmbeddingBagPackedSum op shell

nbpatel:


src/ngraph/op/embeddingbag_packedsum.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+shared_ptr<Node>
+    op::v3::EmbeddingBagPackedSum::clone_with_new_inputs(const OutputVector& new_args) const
+{
+    if (new_args.size() == 2)
+    {
+        return make_shared<EmbeddingBagPackedSum>(new_args.at(0), new_args.at(1));
+    }
+    else
```
diyessi, line 114: If someone passes in 0 arguments this will segfault. Check that new_args has the same size as the inputs and explicitly check for 2 and 3 args and error if neither.

src/ngraph/op/embeddingbag_packedsum.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include "ngraph/op/util/index_reduction.hpp"
+
+namespace ngraph
+{
+    namespace op
+    {
+        namespace v3
+        {
+            /// \brief Returns embeddings for given indices
+            class NGRAPH_API EmbeddingBagPackedSum : public Op
```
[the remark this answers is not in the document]
diyessi: Interface (not based on Node) or implementation base (like unary_elementwise)?

src/ngraph/op/util/embeddingbag_packed_base.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    element::Type result_et = get_input_element_type(EMB_TABLE);
+
+    const PartialShape& emb_table_shape = get_input_partial_shape(EMB_TABLE);
+    const PartialShape& indices_shape = get_input_partial_shape(INDICES);
+
+    PartialShape result_shape;
+    if (emb_table_shape.rank().is_static())
+    {
+        std::vector<Dimension> result_dims(emb_table_shape.rank().get_length());
```
lazarevevgeny, line 83: I think that the easier solution to calculate the `result_dims` is to take the `emb_table_shape` and change the zero element to `indices_shape[0]` or `Dimension::dynamic()` like you do in line 84.

src/ngraph/op/embeddingbag_packedsum.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+}
+
+shared_ptr<Node>
+    op::v3::EmbeddingBagPackedSum::clone_with_new_inputs(const OutputVector& new_args) const
+{
+    check_new_args_count(this, new_args);
+    if (new_args.size() == 2)
+    {
+        return make_shared<EmbeddingBagPackedSum>(new_args.at(0), new_args.at(1));
```
lazarevevgeny, line 44: Use `op::v3::EmbeddingBagPackedSum` instead of `EmbeddingBagPackedSum`

postrational:
Please add the operator to the Python API. In order to do so, add a function similar to this:
https://github.com/NervanaSystems/ngraph/blob/dynamic/python/src/ngraph/ops.py#L2482-L2497

PR #4677, opened 2020-05-01
EmbeddingSegmentsSum op shell

nbpatel:


src/ngraph/op/embedding_segments_sum.cpp
```diff
@@ -0,0 +1,210 @@
...
+                                                   const Output<Node>& segment_ids,
+                                                   const Output<Node>& num_segments)
+    : Op({emb_table, indices, segment_ids, num_segments})
+{
+    constructor_validate_and_infer_types();
+}
+
+void op::v3::EmbeddingSegmentsSum::validate_and_infer_types()
+{
+    enum
```
diyessi, line 57: Let's make this an enum class and stick it in the class definition. Validation isn't the only thing that needs to look at inputs.

src/ngraph/op/embedding_segments_sum.cpp
```diff
@@ -0,0 +1,210 @@
...
+            new_args.at(0), new_args.at(1), new_args.at(2), new_args.at(3));
+    }
+    else if (new_args.size() == 5)
+    {
+        return make_shared<EmbeddingSegmentsSum>(
+            new_args.at(0), new_args.at(1), new_args.at(2), new_args.at(3), new_args.at(4));
+    }
+    else
+    {
+        return make_shared<EmbeddingSegmentsSum>(new_args.at(0),
```
diyessi, line 203: This will crash if too few new_args. Check all the sizes explicitly.

src/ngraph/op/embedding_segments_sum.cpp
nbpatel: that enum is just to make the validation routine more readable..dont see a reason why we need a enum class for it.

diyessi: Passes can use them too.

src/ngraph/op/embedding_segments_sum.cpp
```diff
@@ -0,0 +1,205 @@
...
+    set_output_type(0, result_et, result_shape);
+}
+
+shared_ptr<Node>
+    op::v3::EmbeddingSegmentsSum::clone_with_new_inputs(const OutputVector& new_args) const
+{
+    check_new_args_count(this, new_args);
+    if (new_args.size() == 4)
+    {
+        return make_shared<EmbeddingSegmentsSum>(
```
lazarevevgeny, line 184: `EmbeddingSegmentsSum` -> `op::v3::EmbeddingSegmentsSum`

src/ngraph/op/embedding_segments_sum.cpp
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    element::Type result_et = get_input_element_type(EMB_TABLE);
+
+    const PartialShape& emb_table_shape = get_input_partial_shape(EMB_TABLE);
+
+    PartialShape result_shape;
+    if (emb_table_shape.rank().is_static())
+    {
+        std::vector<Dimension> result_dims(emb_table_shape.rank().get_length());
```
lazarevevgeny, line 160: The same comment as in other EmbeddingBag op reviews.

src/ngraph/op/embedding_segments_sum.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+                virtual std::shared_ptr<Node>
+                    clone_with_new_inputs(const OutputVector& new_args) const override;
+
+            private:
+                static constexpr int EMB_TABLE = 0;
+                static constexpr int INDICES = 1;
+                static constexpr int SEGMENT_IDS = 2;
+                static constexpr int NUM_SEGMENTS = 3;
+                static constexpr int PER_SAMPLE_WEIGHTS = 4;
```
lazarevevgeny, line 86: Please, pay attention to a change in an order of inputs described here https://gitlab-icv.inn.intel.com/inference-engine/dldt/merge_requests/7490

src/ngraph/op/embedding_segments_sum.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    element::Type result_et = get_input_element_type(EMB_TABLE);
+
+    const PartialShape& emb_table_shape = get_input_partial_shape(EMB_TABLE);
+
+    PartialShape result_shape;
+    if (emb_table_shape.rank().is_static())
+    {
+        result_shape = emb_table_shape;
+        result_shape[0] = Dimension::dynamic();
```
lazarevevgeny, line 161: Please, check that the input with `num_segments` is Constant and if so use to set the `result_shape[0]`

PR #4678, opened 2020-05-02
Broadcast v3 evaluate 

gaurides:


src/ngraph/runtime/interpreter/int_executable.hpp
```diff
@@ -468,6 +468,25 @@ class INTERPRETER_BACKEND_API ngraph::runtime::interpreter::INTExecutable : publ
                                     broadcast_axes);
             break;
         }
+        case OP_TYPEID::Broadcast_v3:
```
diyessi, line 471: You don't want this; the interpreter tries evaluate first and only goes to the switch if evaluate returns false.

gaurides: ok, thanks. I will remove it before I do final push. I wanted to test if Broadcast v3, BIDIRECTIONAL is functional with current kernel. And my test works fine.

gaurides:
This PR had some other dependencies. So closing this and there's a newer PR 4703 (https://github.com/NervanaSystems/ngraph/pull/4703)

PR #4681, opened 2020-05-04
add evaluate for unary ops

gaurides:


src/ngraph/op/abs.cpp
```diff
@@ -43,3 +46,55 @@ void op::Abs::generate_adjoints(autodiff::Adjoints& adjoints, const OutputVector
...
+    inline bool evaluate(const HostTensorPtr& arg0, const HostTensorPtr& out, const size_t count)
+    {
+        using T = typename element_type_traits<ET>::value_type;
+        runtime::reference::abs<T>((arg0->get_data_ptr<ET>()), (out->get_data_ptr<ET>()), count);
+        return true;
+    }
+
+    bool evaluate_abs(const HostTensorPtr& arg0, const HostTensorPtr& out, const size_t count)
+    {
+        if (arg0->get_element_type() != out->get_element_type())
```
diyessi, line 62: Use `out->set_unary(arg0)`. It takes care of the shape and element type. Your check here is actually wrong because the element type on out is likely to be dynamic at this point.

gaurides:
TESTNOW

src/ngraph/op/sqrt.cpp
```diff
@@ -48,3 +51,61 @@ void op::Sqrt::generate_adjoints(autodiff::Adjoints& adjoints, const OutputVecto
...
+    inline bool evaluate(const HostTensorPtr& arg0, const HostTensorPtr& out, const size_t count)
+    {
+        using T = typename element_type_traits<ET>::value_type;
+        /*std::vector<T> in_vec(count);
+        arg0->read(in_vec.data(), count * sizeof(T));
+        if (std::any_of(in_vec.begin(), in_vec.end(), [](T i) { return i < T(0); }))
+        {
+            std::cout << " ----------- returning false "
+            return false;
+        }*/
```
rkimballn1, line 67: Remove this.
Also only use line comment '//' in the code

test/backend/sqrt.in.cpp
```diff
@@ -61,3 +61,25 @@ NGRAPH_TEST(${BACKEND_NAME}, sqrt)
...
+{
+    Shape shape{4};
+    auto A = make_shared<op::Parameter>(element::f32, shape);
+    auto f = make_shared<Function>(make_shared<op::Sqrt>(A), ParameterVector{A});
+
+    auto backend = runtime::Backend::create("${BACKEND_NAME}");
+
+    // Create some tensors for input/output
+    auto a = backend->create_tensor(element::f32, shape);
+    copy_data(a, vector<float>{-1, -4, -81, -100});
```
rkimballn1, line 75: I would add both positive and negative numbers to check that it does not just fill the array with NaN

PR #4682, opened 2020-05-04
[Py] Support for different opsets ver from ops.py

arogowie-intel:


python/src/pyngraph/node_factory.cpp
```diff
@@ -139,27 +142,29 @@ class NodeFactory
...
-        else if (opset_name_ == "opset1")
-        {
-            return ngraph::get_opset1();
-        }
-        else if (opset_name_ == "opset2")
-        {
-            return ngraph::get_opset2();
-        }
-        else
+        std::string opset_ver_ = opset_ver;
```
tomdol, line 147: You can just grab `opset_ver` by value and avoid this copy.

PR #4684, opened 2020-05-04
Strided Slice op - evaluate and constant_fold

wenzhe-nrv:


src/ngraph/op/strided_slice.cpp
```diff
@@ -227,3 +232,154 @@ void op::v1::StridedSlice::generate_adjoints(autodiff::Adjoints& /* adjoints */,
...
+            TYPE_CASE(f32)(in, slice_plan, out);
+            break;
+            TYPE_CASE(f64)(in, slice_plan, out);
+            break;
+        default: rc = false; break;
+        }
+        return rc;
+    }
+
+    bool constant_fold_strided_slice(Node* strided_slice_node,
```
diyessi, line 303: Won't evaluate handle this?

wenzhe-nrv: no. I referred to the `op::v0::ShapeOf::constant_fold`. 
Are you suggesting to put the folding into evaluate? Could you point me to the latest example?

PR #4695, opened 2020-05-06
Cyphers/opsetcheck

diyessi:


src/ngraph/runtime/ie/ie_executable.cpp
```diff
@@ -74,10 +74,15 @@ namespace
     }
 }
 
+namespace
+{
+    set <
+}
+
```
rkimballn1, line 81:
```suggestion
namespace
{
    set <
}

```

src/ngraph/opsets/opset.cpp
```diff
@@ -112,3 +112,24 @@ const ngraph::OpSet& ngraph::get_opset3()
...
+    static OpSet opset;
+    if (!opset_is_initialized)
+    {
+        std::lock_guard<std::mutex> guard(init_mutex);
+        if (!opset_is_initialized)
+        {
+#define NGRAPH_OP(NAME, NAMESPACE) opset.insert<NAMESPACE::NAME>();
+#include "ngraph/opsets/opset1_tbl.hpp"
+#include "ngraph/opsets/opset2_tbl.hpp"
+#include "ngraph/opsets/opset3_tbl.hpp"
```
arogowie-intel, line 129: I'm not sure whether this is correct. IE engine opset is always the highest. Opset `X` contains ops from opset `X-1` and **replaces** some ops with their new versions. With the above code the Opset's member `m_op_types` would have more elements than `m_name_type_info_map`. Since the higher opset versions would overwrite operators with same names.  This way one can call ie.:  `contains_type(const NodeTypeInfo& type_info)` for operator from opset1 and would get `true`, while only `opset3` could be created (for example `Topk`).

diyessi: I guess it's not really an opset. What we really want is the union of the set of ops. Adding an accessor for `m_op_types` will let us form that. I'll make those changes.

diyessi:
> Won't this create an opset which includes all ops, including the ones removed in `opset3`, such as `Reverse`?
> 
> ```
> // Moved out of opset2, it was added to opset1 by mistake
> // NGRAPH_OP(Reverse, ngraph::op::v1)
> ```
> 
> Why would we want this?

What we really want is the set of ops IE will execute; ones in opsets 1-3. I refactored to just get the set of those type infos and not call it an opset.

src/ngraph/runtime/ie/ie_executable.cpp
```diff
@@ -74,17 +74,30 @@ namespace
     }
 }
 
+namespace
+{
+    Î std::set<NodeTypeInfo>& get_ie_ops()
```
arogowie-intel, line 79: Some artifact.
```suggestion
    std::set<NodeTypeInfo>& get_ie_ops()
```

PR #4701, opened 2020-05-07
Update non_zero output shape to follow the OpenVino spec

yimeisun123:
Revert the output shape to follow the op spec for non-scalar input, i.e., output shape is {input_shape_rank, num_non_zero}. For scalar input, the output shape is {0, 0} if value is 0 or {1, 1} otherwise.


src/ngraph/op/non_zero.cpp
```diff
@@ -63,9 +64,16 @@ void op::v3::NonZero::validate_and_infer_types()
                           m_output_type == element::i64 || m_output_type == element::i32,
                           "Output type must be i32 or i64");
 
-    // Rank and dimension values of output shape depend on both input shape
-    // and input data
-    set_output_type(0, m_output_type, PartialShape::dynamic());
+    // For scalar non-zero value case, onnx test case expects output shape {1, 1}
+    if (input_shape.rank() == 0)
+    {
+        set_output_type(0, m_output_type, PartialShape{Dimension::dynamic(), Dimension::dynamic()});
```
diyessi, line 70: Why dynamic instead of 1,1?

yimeisun123: At this point, we don't know if the value of the scalar yet. If it is non-zero, then, it is {1, 1}, if it is 0, then, the out shape will be {0, 0}.

rkimballn1:
PR has zero files changed

PR #4703, opened 2020-05-08
Add evaluate for broadcast v3, v1, v0

gaurides:


src/ngraph/op/broadcast.cpp
```diff
@@ -310,6 +322,73 @@ void op::v0::Broadcast::generate_adjoints(autodiff::Adjoints& adjoints, const Ou
...
+
+    bool evaluate_broadcast(const HostTensorPtr& arg0,
+                            const HostTensorPtr& out,
+                            const AxisSet broadcast_axes,
+                            const Shape output_shape)
+    {
+        if (arg0->get_element_type() != out->get_element_type())
+        {
+            return false;
+        }
```
rkimballn1, line 348: Is this a thing? I would think that would be an invalid broadcast since it is doing a Convert as well.
If this is a valid check then it should throw an exception, just returning false seems wrong.

test/eval.cpp
```diff
@@ -218,6 +219,203 @@ TEST(eval, interpret_dynamic_range_sum)
...
+
+    auto result = make_shared<HostTensor>();
+    ASSERT_TRUE(fun->evaluate(
+        {result}, {make_host_tensor<element::Type_t::f32>(Shape{2, 4}, {1, 2, 3, 4, 1, 2, 3, 4})}));
+    EXPECT_EQ(result->get_element_type(), element::f32);
+    EXPECT_EQ(result->get_partial_shape(), (PartialShape{2, 3, 4}));
+    auto result_val = read_vector<float>(result);
+    vector<float> expec{1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4};
+    ASSERT_EQ(result_val, expec);
+}
```
rkimballn1, line 418:
```suggestion
}

```

src/ngraph/op/broadcast.cpp
diyessi: Just set the output element type to the arg0 element type and it will make sure they are compatible. The output element type from validation could be "dynamic" which would not be the same as the arg0 type, but is compatible so setting the element type on output would work.

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -267,3 +269,75 @@ void op::util::BroadcastBase::generate_adjoints(autodiff::Adjoints& adjoints,
...
+                                         broadcast_axes);
+        return true;
+    }
+
+    bool evaluate_broadcast(const HostTensorPtr& arg0,
+                            const HostTensorPtr& out,
+                            const std::pair<bool, AxisSet> pair_broadcast_axes,
+                            const Shape output_shape)
+    {
+        if (arg0->get_element_type() != out->get_element_type())
```
diyessi, line 293: Just set the output element type to the  arg0 element type.

src/ngraph/op/broadcast.cpp
gaurides: No, i think this should never happen, because the constructor sets node output type and we need to set the HostTensor type.
I will remove this check.

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -267,3 +269,75 @@ void op::util::BroadcastBase::generate_adjoints(autodiff::Adjoints& adjoints,
...
+        default: rc = false; break;
+        }
+        return rc;
+    }
+}
+
+bool op::util::BroadcastBase::evaluate(const HostTensorVector& outputs,
+                                       const HostTensorVector& inputs)
+{
+    return evaluate_broadcast(inputs[0], outputs[0], get_broadcast_axes(), get_output_shape(0));
```
diyessi, line 342: `get_output_shape(0)` only works for static shapes. you need to look at arg1 and then do the shape merge with the arg0 shape arg2 axes value and the broadcast mode to compute the output shape and use that,

diyessi: An additional complication is that there is a 2-input and a 3-input form of broadcast

ashokei:
can we remove from constant_folding pass

construct_constant_broadcast();    
construct_constant_dyn_broadcast();

src/ngraph/op/broadcast.cpp
```diff
@@ -163,6 +187,23 @@ bool op::v3::Broadcast::visit_attributes(AttributeVisitor& visitor)
     return true;
 }
 
+bool op::v3::Broadcast::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
+    std::cout << "Bcast v3 evaluate \n";
```
diyessi, line 192: Remove debugging

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -170,33 +206,56 @@ void op::util::BroadcastBase::validate_and_infer_types()
...
+    NGRAPH_CHECK(start_axis >= 0);
+    for (size_t i = 0; i < result_shape.size(); i++)
+    {
+        if (i < start_axis || result_shape[i] != arg_shape[i - start_axis])
+        {
+            broadcast_axes.insert(i);
+        }
+    }
+    axes_known = true;
+    std::cout << "get_broadcast_axes_numpy_pdpd, axes_known = true, broadcast_axes = "
```
diyessi, line 233: Debuggging

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -170,33 +206,56 @@ void op::util::BroadcastBase::validate_and_infer_types()
...
+    std::vector<size_t> axes(target_shape_size);
+    std::iota(axes.begin(), axes.end(), 0);
+    for (auto i = axes_mapping_val.rbegin(); i != axes_mapping_val.rend(); ++i)
+    {
+        axes.erase(axes.begin() + *i);
+    }
+    broadcast_axes.insert(axes.begin(), axes.end());
+
+    axes_known = true;
+    std::cout << "get_broadcast_axes_none, axes_known = true, broadcast_axes = " << broadcast_axes
```
diyessi, line 254: Debugging

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -267,3 +307,219 @@ void op::util::BroadcastBase::generate_adjoints(autodiff::Adjoints& adjoints,
...
+}
+
+Shape op::util::BroadcastBase::get_target_shape(const HostTensorPtr& input1)
+{
+    Shape target_shape;
+    const auto shape_constant = as_type_ptr<op::v0::Constant>(input_value(1).get_node_shared_ptr());
+    if (shape_constant)
+    {
+        target_shape = shape_constant->get_shape_val();
+        std::cout << "*** from node Target shape = " << target_shape << "\n";
```
diyessi, line 473: debugging

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -267,3 +307,219 @@ void op::util::BroadcastBase::generate_adjoints(autodiff::Adjoints& adjoints,
...
+    }
+    else
+    {
+        target_shape = get_target_shape_from_ht(input1);
+        std::cout << "*** Calculated Target shape = " << target_shape << "\n";
```
diyessi, line 478: debugging

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -267,3 +307,219 @@ void op::util::BroadcastBase::generate_adjoints(autodiff::Adjoints& adjoints,
...
+    }
+    return target_shape;
+}
+
+bool op::util::BroadcastBase::evaluate(const HostTensorVector& outputs,
+                                       const HostTensorVector& inputs)
+{
+    std::cout << "BcastBase evaluate \n";
```
diyessi, line 486: debugging

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -267,3 +307,219 @@ void op::util::BroadcastBase::generate_adjoints(autodiff::Adjoints& adjoints,
...
+
+    if (m_mode.m_type == BroadcastType::NONE)
+    {
+        AxisVector axes_mapping_val;
+        const auto axes_mapping_constant =
+            as_type_ptr<op::v0::Constant>(input_value(2).get_node_shared_ptr());
+        if (axes_mapping_constant)
+        {
+            axes_mapping_val = axes_mapping_constant->get_axis_vector_val();
+            std::cout << "axes_mapping_val from node input = " << axes_mapping_val << "\n";
```
diyessi, line 501: debugging

src/ngraph/op/util/broadcast_base.cpp
```diff
@@ -267,3 +307,219 @@ void op::util::BroadcastBase::generate_adjoints(autodiff::Adjoints& adjoints,
...
+        }
+        else
+        {
+            // read from HT and save as AxisVector
+            get_axis_vector_from_ht(inputs[2], axes_mapping_val, arg_shape);
+            std::cout << "axes_mapping_val from HT = " << axes_mapping_val << "\n";
```
diyessi, line 507: debugging

PR #4704, opened 2020-05-08
Cyphers/moremax

diyessi:
Make maximum value tracking handle more ops and element types.


src/ngraph/validation_util.cpp
```diff
@@ -922,84 +929,138 @@ void ngraph::opset1::infer_conv_backprop_auto_padding(const Shape& input_data_sh
...
+        MaxValue(uint64_t value)
+            : m_value(value)
         {
         }
-        /// \brief int64_t assoiated with the output
-        MaxValue(int64_t value)
-            : m_element_type(element::Type_t::i64)
-            , m_int64_t(value)
+        MaxValue(vector<uint64_t> slices, int64_t slice_axis)
+            : m_slices(slices)
```
rkimballn1, line 945: Either pass vector by const ref or use `m_slices(std::move(slices))` to save a copy. This is doing two copies.

diyessi: Got it.

PR #4705, opened 2020-05-08
add evaluate for Concat

ashokei:


src/ngraph/op/concat.hpp
```diff
@@ -69,7 +72,7 @@ namespace ngraph
                 /// \brief m_concat_axis stores m_axis plus the number of rank for each iteration
                 int64_t m_concat_axis = -1;
             };
-        }
+        } // namespace v0
         using v0::Concat;
-    }
-}
+    } // namespace op
+} // namespace ngraph
```
rkimballn1, line 78: Don't add these

src/ngraph/op/concat.cpp
```diff
@@ -142,3 +143,61 @@ void op::Concat::generate_adjoints(autodiff::Adjoints& adjoints, const OutputVec
         pos = next_pos;
     }
 }
+
+namespace
+{
+    template <element::Type_t ET>
+    inline bool
+        evaluate(const HostTensorVector& args, const HostTensorPtr& out, int64_t concatenation_axis)
+    {
```
diyessi, line 152: You need to figure the output shape from the input host tensors and set it on the output, in case some of the inputs are dynamic.

ashokei: shape inference should have handled for static, wasn't sure we are supporting dynamic inputs? will update.

diyessi: We are supporting dynamic.

ashokei: https://github.com/NervanaSystems/ngraph/pull/4705/commits

PR #4707, opened 2020-05-08
Add evaluate for gather

nbpatel:


src/ngraph/op/gather.cpp
```diff
@@ -208,3 +211,80 @@ shared_ptr<Node> op::v1::Gather::clone_with_new_inputs(const OutputVector& new_a
...
+
+namespace
+{
+    template <element::Type_t ET>
+    inline bool evaluate(const HostTensorPtr& arg0,
+                         const HostTensorPtr& arg1,
+                         const HostTensorPtr& out,
+                         size_t axis)
+    {
+        using T = typename element_type_traits<ET>::value_type;
```
diyessi, line 223: Somewhere around here you need to set the output shape/element type based on the input argument so that the dynamic input case works.

nbpatel: added, also a unit test

src/ngraph/op/gather.cpp
```diff
@@ -208,3 +210,113 @@ shared_ptr<Node> op::v1::Gather::clone_with_new_inputs(const OutputVector& new_a
...
+}
+
+bool op::v0::Gather::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
+    return evaluate_gather(inputs[0], inputs[1], outputs[0], get_axis());
+}
+
+bool op::v1::Gather::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
+    int64_t axis = inputs[2]->get_data_ptr<int64_t>()[0];
```
diyessi, line 312: The spec doesn't say that the axis input is int64_t and this will do something bad if it's not. Unfortunately, HostTensor doesn't have a `cast_vector` sort of thing. You can do a switch on the element type of inputs[2] and do `*get_data_ptr<element::Type_t::i32>` etc. for each of the integral element types.

src/ngraph/op/gather.cpp
```diff
@@ -208,3 +210,126 @@ shared_ptr<Node> op::v1::Gather::clone_with_new_inputs(const OutputVector& new_a
     check_new_args_count(this, new_args);
     return make_shared<v1::Gather>(new_args.at(PARAMS), new_args.at(INDICES), new_args.at(AXIS));
 }
+
+namespace
+{
+    template <element::Type_t ET>
+    inline bool evaluate(const HostTensorPtr& arg0,
```
rkimballn1, line 217:
```suggestion
    bool evaluate(const HostTensorPtr& arg0,
```

rkimballn1: Don't inline this big function.

src/ngraph/op/gather.cpp
```diff
@@ -208,3 +210,126 @@ shared_ptr<Node> op::v1::Gather::clone_with_new_inputs(const OutputVector& new_a
...
+    if (axis < 0)
+    {
+        const auto& input_rank = get_input_partial_shape(PARAMS).rank();
+        if (input_rank.is_static())
+        {
+            axis += input_rank.get_length();
+        }
+    }
+    return evaluate_gather(inputs[0], inputs[1], outputs[0], axis);
+}
```
rkimballn1, line 335:
```suggestion
}

```

src/ngraph/op/gather.cpp
```diff
@@ -208,3 +210,126 @@ shared_ptr<Node> op::v1::Gather::clone_with_new_inputs(const OutputVector& new_a
...
+
+bool op::v0::Gather::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
+    return evaluate_gather(inputs[0], inputs[1], outputs[0], get_axis());
+}
+
+bool op::v1::Gather::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
+    auto axis = 0;
```
diyessi, line 312: Make it int64_t to be safe.

PR #4708, opened 2020-05-08
Reimplemented Transpose->Reshape optimization

GlebKazantaev:


jane-intel:
@rkimballn1 @diyessi Current PR is recommended to be merged straight to master branch -- OV CI passed.

test/algebraic_simplification.cpp
```diff
@@ -887,7 +887,6 @@ TEST(algebraic_simplification, replace_transpose_with_reshape)
     check_usecase(Shape{10, 20, 1, 1, 1}, vector<int64_t>{0, 2, 3, 4, 1});
     check_usecase(Shape{10, 1, 1, 1, 1}, vector<int64_t>{1, 4, 2, 3, 0});
     check_usecase(Shape{10, 1, 1, 1, 1}, vector<int64_t>{4, 2, 0, 1, 3});
-    check_usecase(Shape{1, 1, 1, 1, 1}, vector<int64_t>{4, 2, 0, 1, 3});
```
rkimballn1, line 890: This looks like a valid test, why was it removed? I realized it does not pass if added but it seems valid.

rkimballn1: ok, so I checked and the test fails because the transpose/reshape is optimized away which is a good thing.

PR #4714, opened 2020-05-11
ONNX Softmax adjustments

tomdol:
This is how ONNX defines Softmax:
https://github.com/onnx/onnx/blob/master/docs/Operators.md#Softmax
Scroll down to examples under the operator definition and expand `softmax_axis`. The 4 first lines define a function that shows how tensors bigger than 1D should be handled.
https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/softmax.py#L30-L33


src/ngraph/frontend/onnx_import/op/softmax.cpp
```diff
@@ -17,27 +17,92 @@
...
+
+                const auto data_minus_max =
+                    std::make_shared<default_opset::Subtract>(coerced_data, reshaped_max);
+
+                const auto exp = std::make_shared<default_opset::Exp>(data_minus_max);
+                const auto sum_exp = std::make_shared<default_opset::ReduceSum>(exp, axis_1);
+                const auto reshaped_sum_exp =
+                    std::make_shared<default_opset::Reshape>(sum_exp, reshape_pattern, true);
+
+                const auto result = std::make_shared<default_opset::Divide>(exp, reshaped_sum_exp);
```
arogowie-intel, line 52: You don't need to create subgraph for Softmax here on your own. You may leverage exisiting operator:
```suggestion
const auto result = std::make_shared<default_opset::Softmax>(data_minus_max, 1);
```
I wonder whether this would enable to release the requirement of static input data rank and simplify op logic.

test/onnx/onnx_import.in.cpp
```diff
@@ -706,61 +706,144 @@ NGRAPH_TEST(${BACKEND_NAME}, onnx_model_matmul)
...
+    {
+        onnx_import::import_onnx_model(
+            file_util::path_join(SERIALIZED_ZOO, "onnx/softmax_invalid_axis_1D.prototxt"));
+        FAIL() << "Softmax model with invalid axis was successfully imported while it should have "
+                  "thrown.";
+    }
+    catch (ngraph::ngraph_error const& err)
+    {
+        SUCCEED();
+    }
```
arogowie-intel, line 831: Wouldn't `ASSERT_THROW` work here?

test/onnx/onnx_import.in.cpp
```diff
@@ -706,61 +706,144 @@ NGRAPH_TEST(${BACKEND_NAME}, onnx_model_matmul)
...
+
+         0.01648154, 0.00583583, 0.00182802, 0.00515374, 0.02491679, 0.00537859, 0.00195794,
+         0.00802367, 0.00254737, 0.0223216,  0.02893419, 0.0142204,  0.02053893, 0.00778581,
+         0.00251907, 0.00111174, 0.00800149, 0.0030324,  0.06658917, 0.0179084,
+
+         0.00181811, 0.01407243, 0.01072611, 0.0069699,  0.01158077, 0.01305647, 0.00053677,
+         0.0017687,  0.08719896, 0.02028982, 0.00236265, 0.01027717, 0.0060709,  0.01216173,
+         0.00388087, 0.00385541, 0.00758048, 0.00909469, 0.04775123, 0.03836337});
+
+    test_case.run(6);
```
arogowie-intel, line 768: Check if using `Softmax` operator for cases >2d gives better accuracy.

test/onnx/onnx_import.in.cpp
```diff
@@ -706,61 +706,144 @@ NGRAPH_TEST(${BACKEND_NAME}, onnx_model_matmul)
...
-            .get_vector());
+    auto test_case = ngraph::test::NgraphTestCase(function, "${BACKEND_NAME}");
+    test_case.add_input<float>({-1.0, 0.0, 1.0});
+    test_case.add_expected_output<float>({0.09003058, 0.24472848, 0.66524094});
+    test_case.run();
+}
+namespace
+{
+    // common input for all Softmax 3D test cases (Shape = {3,4,5})
+    const std::vector<float> SOFTMAX_INPUT = {
```
arogowie-intel, line 733: Just an idea: You may create a `static constexpr std::array<float, <array_size>> sfmax = { ... }` and then later use it like this: `test_case.add_input<float>({sfmax.begin(), sfmax.end()});`

src/ngraph/frontend/onnx_import/op/softmax.cpp
tomdol: Known rank is required to correctly handle 0D and 1D (single op instead of a subgraph). Additionally the static rank requirement also exists in `ngraph::normalize_axis` internally.

test/onnx/onnx_import.in.cpp
tomdol: Yep, that's much cleaner, thanks.

test/onnx/onnx_import.in.cpp
tomdol: Unfortunately that didn't help. I have to leave those custom tolerance values for those 3 tests.

test/onnx/onnx_import.in.cpp
tomdol: While the `std::array` is constexpr, this suggested approach causes 2 copies in runtime array->vector & vector->vector. I'd rather leave it as is.

PR #4718, opened 2020-05-11
Create inverse hyperbolic ops

rkimballn1:


src/ngraph/op/acosh.cpp
```diff
@@ -0,0 +1,86 @@
...
+    bool evaluate(const HostTensorPtr& arg0, const HostTensorPtr& out)
+    {
+        runtime::reference::acosh(
+            arg0->get_data_ptr<ET>(), out->get_data_ptr<ET>(), shape_size(arg0->get_shape()));
+        return true;
+    }
+
+    bool evaluate_acosh(const HostTensorPtr& arg0, const HostTensorPtr& out)
+    {
+        bool rc = true;
```
diyessi, line 54: Need to set the output element type and shape. `out->set_unary(arg0)`
Same for the other functions.

PR #4723, opened 2020-05-12
[Py] Update Python API - RNN ops

mitruska:
Update Python API with RNN ops:
* GRUCell
* LSTMCell
* LSTMSequence
* RNNCell


postrational:
Please run flake8.

python/test/ngraph/test_create_op.py
```diff
@@ -161,6 +161,306 @@ def test_gather_tree(dtype):
...
+                                  activation_alpha,
+                                  activation_beta,
+                                  clip)
+
+    assert node_param.get_type_name() == 'LSTMSequence'
+    assert node_param.get_output_size() == 3
+
+
+@pytest.mark.parametrize('dtype', [np.float32, np.float64])
+def test_lstm_sequence_operator_forward(dtype):
```
arogowie-intel, line 346: Just a note: I think that this test case is redundant. It doesn't check anything more than the above `bidirectional` plus `reverese` ones in terms of Python API functionality. The input validation should be tested by core op UTs in general.

mitruska: I can agree, but I saw in the nGraph LSTMSequence some conditions based on the `direction` so I wanted to be sure that it works for all of them.

src/ngraph/op/fused/rnn_cell.cpp
```diff
@@ -73,7 +73,9 @@ op::RNNCell::RNNCell(const Output<Node>& X,
 
 bool op::RNNCell::visit_attributes(AttributeVisitor& visitor)
 {
-    return op::util::RNNCellBase::visit_attributes(visitor);
+    op::util::RNNCellBase::visit_attributes(visitor);
+    m_activation_f = get_activation_function(0);
```
diyessi, line 77: What is this for?

mitruska: The [python RNNCell test](https://aipg-jenkins-algo.intel.com/blue/organizations/jenkins/ngraph-python-merge/detail/ngraph-python-merge/15170/pipeline/69#step-72-log-435) failed on the CPU backend (the same test for the INTERPRETER passed).
I discovered that for RNNCell created by node factory, the activation function on CPU backend is always 'tanh' despite we passed 'sigmoid' in the `activations` vector attribute.

I just wanted to run the CI with this hot fix and I see that it works now.
But probably this problem required deeper investigation, not in this PR scope.
So I can revert this change and skip the python RNNCell test on the CPU backend.
Like I said for the INTERPRETER backend the CI passed anyway.

PR #4731, opened 2020-05-13
[Py] More ops exposed to python API

tomdol:


python/src/ngraph/ops.py
```diff
@@ -2588,3 +2588,83 @@ def reverse_sequence(input, seq_lengths, batch_axis, seq_axis, name=None):
...
+    """
+    return _get_node_factory().create('Range', [start, as_node(stop), as_node(step)])
+
+
+@nameable_op
+def region_yolo(input, # type: Node
+                num_coords, # type: int
+                num_classes, # type: int
+                num_regions, # type: int
+                do_softmax,  # type: bool
```
postrational, line 2629: According to the spec this is not a required attribute.

python/src/ngraph/ops.py
```diff
@@ -2588,3 +2588,83 @@ def reverse_sequence(input, seq_lengths, batch_axis, seq_axis, name=None):
...
+                                    })
+
+
+@nameable_op
+def reorg_yolo(input, strides): # type: (Node, NodeInput, NodeInput) -> Node
+    """Return a node which produces the ReorgYolo operation.
+
+
+    :param input:  Input data
+    :param step:   The step value for the generated range
```
mitruska, line 2666: `step` is not ReorgYolo attribute, add :param `stride` instead
```suggestion
    :param stride:   The distance between cut throws in output blobs.
```

python/src/ngraph/ops.py
```diff
@@ -2588,3 +2588,83 @@ def reverse_sequence(input, seq_lengths, batch_axis, seq_axis, name=None):
...
+    # type: (Node, NodeInput, NodeInput) -> Node
+    """Return a node which produces the Range operation.
+
+
+    :param start:  The start value of the generated range
+    :param stop:   The stop value of the generated range
+    :param step:   The step value for the generated range
+    :return: Range node
+    """
+    return _get_node_factory().create('Range', [start, as_node(stop), as_node(step)])
```
mitruska, line 2622: We should let users to pass NodeInput type also for `start` input. You can call `as_nodes` which return Node list.

python/src/ngraph/ops.py
```diff
@@ -2588,3 +2588,83 @@ def reverse_sequence(input, seq_lengths, batch_axis, seq_axis, name=None):
     """
     return _get_node_factory().create('ReverseSequence', [input, as_node(seq_lengths)],
                                       {'batch_axis': batch_axis, 'seq_axis': seq_axis})
+
+
+@nameable_op
+def bucketize(data, buckets, output_type='i64', with_right_bound=True):
```
mitruska, line 2594: To be consistent with the rest of pyAPI ops and `@namable_op` decorator, 
please  add  optional arg `name=None,  # type: str`.

python/src/ngraph/ops.py
```diff
@@ -2588,3 +2588,83 @@ def reverse_sequence(input, seq_lengths, batch_axis, seq_axis, name=None):
...
+                                      "do_softmax": do_softmax,
+                                      "mask": mask,
+                                      "axis": axis,
+                                      "end_axis": end_axis,
+                                      "anchors": anchors
+                                    })
+
+
+@nameable_op
+def reorg_yolo(input, strides): # type: (Node, NodeInput, NodeInput) -> Node
```
mitruska: To be consistent with the rest of pyAPI ops and `@namable_op` decorator, 
please  add  optional arg `name=None,  # type: str`.

python/src/ngraph/ops.py
```diff
@@ -2588,3 +2588,82 @@ def reverse_sequence(input, seq_lengths, batch_axis, seq_axis, name=None):
...
+@nameable_op
+def region_yolo(input,  # type: Node
+                coords,  # type: int
+                classes,  # type: int
+                num,  # type: int
+                mask,  # type: List[int]
+                axis,  # type: int
+                end_axis,  # type: int
+                do_softmax=True,  # type: bool
+                anchors=[],  # type: List[float]
```
mitruska, line 2634: AFAIK,  In Python we should't set a list `[]` (mutable) as default argument here but init such an arg in the function body.
https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

python/src/ngraph/ops.py
```diff
@@ -2588,3 +2588,82 @@ def reverse_sequence(input, seq_lengths, batch_axis, seq_axis, name=None):
...
+    """
+    return _get_node_factory().create(
+        'Bucketize',
+        [data, as_node(buckets)],
+        {'output_type': output_type, 'with_right_bound': with_right_bound},
+    )
+
+
+@nameable_op
+def range(start, stop, step):
```
mitruska, line 2613: To be consistent with the rest of pyAPI ops and `@namable_op` decorator, 
please  add  optional arg `name=None,  # type: str`.

PR #4734, opened 2020-05-13
Add an identity kernel for use with integer floor, ceiling, and round

rkimballn1:
Added unit test passing int64 values too large to fit in a double.


src/ngraph/op/util/eval_identity.hpp
```diff
@@ -0,0 +1,20 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//*****************************************************************************
+
+#pragma once
+
+#define IDENTITY(a)                                                                                \
```
gaurides, line 19: IDENTITY makes think of all 1's at first glance. Can we please rename to COPY or IDENTICAL or TYPE_COPY  or something else to convey the copy part?

src/ngraph/runtime/reference/identity.hpp
```diff
@@ -0,0 +1,37 @@
+//*****************************************************************************
```
silee2, line 1: We already have a kernel that does the same.
https://github.com/NervanaSystems/ngraph/blob/master/src/ngraph/runtime/reference/copy.hpp
I don't think we need a new kernel.

rkimballn1: removed

src/ngraph/op/util/eval_identity.hpp
rkimballn1: I changed the name to COPY_TENSOR

PR #4741, opened 2020-05-14
merge master to dynamic

ashokei:


rkimballn1:
I did the command line version of this so dynamic should be up to date

PR #4742, opened 2020-05-14
Add Evaluate to  scatter_elements_update

noim210:
Add Evaluate to scatter_elements_update

original spec: 
https://gitlab-icv.inn.intel.com/inference-engine/dldt/blob/master/docs/ops/movement/ScatterElementsUpdate_3.md


test/eval.cpp
```diff
@@ -881,3 +882,62 @@ TEST(eval, evaluate_dynamic_concat)
...
+    ASSERT_EQ(cval, out);
+
+}
+
+TEST( eval, evaluate_dynamic_scatter_elements_update_basic)
+{
+    const Shape data_shape{3, 3};
+    const Shape indices_shape{2, 3};
+
+    //auto arg1 = make_shared<op::Parameter>(element::f32, PartialShape::dynamic()); //error
```
postrational, line 916: Please remove debug code.

noim210:
> You need to set the output shape based on the shapes of the argument tensors; otherwise it won't work if the inputs are dynamic. To test, make an evaluate test with a parameter that has dynamic shape.

test/eval.cpp
noim210: removed and added more tests. 
Thanks.

PR #4746, opened 2020-05-14
[ONNX] gru rnn v3

arogowie-intel:
- [ ] unit tests for GRU
- [ ] unit tests for RNN


src/ngraph/op/fused/lstm_sequence.hpp
```diff
@@ -66,7 +60,7 @@ namespace ngraph
                                       const Output<Node>& B,
                                       const Output<Node>& P,
                                       const std::int64_t hidden_size,
-                                      const direction lstm_direction,
+                                      const RecurrentSequenceDirection lstm_direction,
```
silee2, line 63: Why do you need an API breaking change?

diyessi: Add a `using direction=RecurrentSequenceDirection` to the class so it won't be API-breaking.

arogowie-intel: I need some enum to use for ONNX's RNN and GRU direction mode. Added using statement.

src/ngraph/frontend/onnx_import/utils/recurrent.cpp
```diff
@@ -0,0 +1,285 @@
...
+                        "RecurrentSequence: unhandled direction mode during decomposition.");
+                }
+                return results;
+            }
+
+            NodeVector
+                RecurrentSequence::recurrent_sequence_pass(const RecurrentCellFunction& kernel,
+                                                           bool is_reverse)
+            {
+                NodeVector h_list;
```
rkimballn1, line 163:
```suggestion
                OutputVector h_list;
```

arogowie-intel:
Moved to OpenVINO repository.

PR #4749, opened 2020-05-15
evaluate and dynamic evaluate for maxpool

ashokei:


src/ngraph/op/max_pool.cpp
```diff
@@ -492,3 +504,113 @@ void op::v1::MaxPool::generate_adjoints(autodiff::Adjoints& adjoints, const Outp
...
+        default: rc = false; break;
+        }
+        return rc;
+    }
+} // namespace
+
+bool op::v0::MaxPool::evaluate(const HostTensorVector& outputs, const HostTensorVector& inputs)
+{
+    auto arg_shape = inputs[0]->get_partial_shape();
+    update_auto_padding(arg_shape);
```
diyessi, line 574: Can't do this. Evaluating should not modify the node.

ashokei: will make it stateless.

PR #4752, opened 2020-05-15
Add eval and test for transpose

swapna-intel:


src/ngraph/op/transpose.cpp
```diff
@@ -90,3 +91,138 @@ void op::v1::Transpose::generate_adjoints(autodiff::Adjoints& /* adjoints */,
 {
     throw ngraph_error("generate_adjoints not implemented for Transpose");
 }
+
+namespace
+{
+    template <typename T>
+    std::vector<T> get_vector(const HostTensorPtr& arg)
+    {
+        std::vector<T> rc;
```
diyessi, line 100: Make this an int64_t vector and you won't need to copy it again.
Use `template<element::Type_t T>` instead of `typename T` so we know that we're getting the right compile-time C++ type for the element type.

swapna-intel: done

PR #4753, opened 2020-05-15
Add evaluate support for MatMul op

yimeisun123:


src/ngraph/op/fused/matmul.cpp
```diff
@@ -102,3 +103,149 @@ shared_ptr<Node> op::MatMul::clone_with_new_inputs(const OutputVector& new_args)
...
+                  const HostTensorPtr& arg1,
+                  const HostTensorPtr& output,
+                  bool transpose_a,
+                  bool transpose_b)
+    {
+        using T = typename element_type_traits<ET>::value_type;
+
+        Shape arg0_shape = arg0->get_shape();
+        Shape arg1_shape = arg1->get_shape();
+        PartialShape output_pshape = output->get_partial_shape();
```
diyessi, line 184: Don't worry about `output_pshape`. Just go directly to the `output_shape` computation in the `else` below.

yimeisun123: The output shape is checked to see if it is already static, if so, no need to do the computation for output shape.

diyessi: Frameworks don't enforce the static shapes, so you might see inputs that don't correspond to the shape computed during shape inference.

PR #4761, opened 2020-05-15
add clamp reference kernel with f16 and bf16 support

adstraw:


src/ngraph/op/fused/clamp.cpp
```diff
@@ -18,12 +18,110 @@
...
+             count);
+            break;
+            TYPE_CASE(u64)
+            (arg,
+             out,
+             double_to_int<uint64_t>(min, ceil_func),
+             double_to_int<uint64_t>(max, floor_func),
+             count);
+            break;
+            TYPE_CASE(f16)(arg, out, static_cast<float16>(min), static_cast<float16>(max), count);
```
diyessi, line 104: Here the casts of large values should go to the fp16/bfloat16 infinities, but you need to test that it works.

test/util.cpp
```diff
@@ -712,3 +712,56 @@ TEST(util, topological_sort_replace)
 
     EXPECT_TRUE(custom_sorter_used);
 }
```
diyessi, line 714: Need some bf16/fp16 cases, including with range out of bf16/fp16 non-infinity range.

src/ngraph/op/fused/clamp.cpp
adstraw: Done.

test/util.cpp
adstraw: Done.

PR #4767, opened 2020-05-16
Enable ngraph passes in IE backend

adk9:
* Also add support for dumping serialized IE network and TF function post optimization.


diyessi:
A lot of those passes don't do anything for IE ops.

adk9:
> A lot of those passes don't do anything for IE ops.

That's good to know. Note that these are applied before the auto-upgrade to opset1. Which ones would you recommend dropping?

diyessi:
Before the opset1 pass you're okay. But if you later start doing opset1 directly from TF, reshape sink/swim won't be doing anything which will be a problem.

src/ngraph/runtime/ie/ie_executable.cpp
```diff
@@ -113,6 +129,12 @@ runtime::ie::IE_Executable::IE_Executable(shared_ptr<Function> func, string devi
     m_network = InferenceEngine::CNNNetwork(func);
     set_parameters_and_results(*func);
 
+#if defined(NGRAPH_IE_DUMP_GRAPHS)
```
rkimballn1, line 132: I wonder if using an environment variable would be better for this so you don't need to rebuild if you want a graph.

diyessi: Using `env_util.hpp`

PR #4776, opened 2020-05-18
Fix a reshape builder when data is reshaped to scalar.

tsocha:


src/ngraph/partial_shape.hpp
```diff
@@ -305,4 +305,6 @@ namespace ngraph
         static constexpr DiscreteTypeInfo type_info{"AttributeAdapter<PartialShape>", 0};
         const DiscreteTypeInfo& get_type_info() const override { return type_info; }
     };
+
+    inline bool is_scalar(const PartialShape& shape) { return shape.rank() == 0; }
```
tomdol, line 309: Can you use [this](https://github.com/NervanaSystems/ngraph/blob/7bb94ca049e720c35adb5844e7deb6c93ddf1b49/src/ngraph/shape.hpp#L98) instead? Or even remove this method since I don't see any references to it.

src/ngraph/builder/reshape.cpp
```diff
@@ -177,10 +178,28 @@ shared_ptr<Node> builder::expand_dims(const Output<Node>& value, size_t axis)
         ->add_provenance_group_members_above({value});
 }
 
+NGRAPH_API
 shared_ptr<Node> builder::opset1::reshape(const Output<Node>& value, const Shape& shape)
 {
-    const auto out_pattern = op::Constant::create(
-        element::i64, Shape{shape.size()}, vector<int64_t>(shape.begin(), shape.end()));
+    shared_ptr<op::Constant> out_pattern;
+    if (value.get_shape() == shape)
```
tomdol, line 185: I suggest using `value.get_output_partial_shape(0).same_scheme(shape)` instead to avoid exceptions when `value`'s shape is not static.

src/ngraph/partial_shape.hpp
tsocha: I can't because `PartialShape` class doesn't have a `size()` method.

src/ngraph/builder/reshape.cpp
```diff
@@ -177,10 +178,28 @@ shared_ptr<Node> builder::expand_dims(const Output<Node>& value, size_t axis)
         ->add_provenance_group_members_above({value});
 }
 
+NGRAPH_API
 shared_ptr<Node> builder::opset1::reshape(const Output<Node>& value, const Shape& shape)
 {
-    const auto out_pattern = op::Constant::create(
-        element::i64, Shape{shape.size()}, vector<int64_t>(shape.begin(), shape.end()));
+    shared_ptr<op::Constant> out_pattern;
```
diyessi, line 184: Could you move `out_pattern` down into the `else` and move `special_zero` and the `return` into the `else` body? Not sure why `special_zero` needs a variable unless it's documentation; could be a constexpr then.

tsocha: Done.

PR #4777, opened 2020-05-18
Use op::v0::Interpolate in opset3

mbencer:
As currently agreed `op::v0::Interpolate` should be used in opset3.
It means that `Resize-10` should produce `op::v0::Interpolate`.


mbencer:
@rkimballn1 @silee2 removed from review after change base to dynamic

jane-intel:
Will be merged to dldt closed source repository in MR#7594

PR #4781, opened 2020-05-18
ShuffleChannels - rename attribute groups->group

postrational:
Rename attribute groups->group to match specification.


src/ngraph/op/fused/shuffle_channels.cpp
```diff
@@ -23,18 +23,18 @@ using namespace ngraph;
 
 constexpr NodeTypeInfo op::ShuffleChannels::type_info;
 
-op::ShuffleChannels::ShuffleChannels(const Output<Node>& data, const int axis, const size_t groups)
+op::ShuffleChannels::ShuffleChannels(const Output<Node>& data, const int axis, const size_t group)
```
diyessi, line 26: Noted in passing:
Would it be a problem to make `groups` a `uint64_t`? The size of `size_t` varies from platform to platform.
Same with `axis` -- it should be `int64_t`

rkimballn1: There is nothing wrong with size_t but there is a problem when people assume that size_t is 64 bits. You should never mix size_t with other types like int64_t, if you use size_t use size_t everywhere it touches.

diyessi: People assume that a lot of places, and the compiler isn't warning about it when it's not true.

lazarevevgeny:
The change in the DLDT repo will be done as part of MR #7528

PR #4783, opened 2020-05-18
[Py] Python API for ops with struct unpacked manually

arogowie-intel:
This PR does not depend on #4666  in order to be able to merge faster.


arogowie-intel:
Most of the code was already reviewed in #4670

PR #4806, opened 2020-05-27
Eigen moved to gitlab

silee2:


cmake/external_eigen.cmake
```diff
@@ -17,24 +17,19 @@
 # Enable ExternalProject CMake module
 include(ExternalProject)
 
-set(EIGEN_GIT_TAG ded1e7b4960f0074fa147a8ed1c9926174958092)
-set(EIGEN_GIT_URL https://github.com/eigenteam/eigen-git-mirror)
+set(EIGEN_GIT_TAG 4e7046063babd95e100ab202d52ca5776997c80c)
```
gaurides, line 20: Where should this tag come from? from one of the release tags in here https://gitlab.com/libeigen/eigen/-/tags? i don't seem to find a match

silee2: The latest commit as of today. nGraph never used an Eigen release tag.

silee2:
Build with Clang works only if C++ standard is changed from C++11 to C++14. Closing PR for now and will reopen once nGraph moves to C++14 or later standard.

PR #4807, opened 2020-05-27
Normalize comments for namespaces

rkimballn1:
I don't know if anyone else will like it, but I like consistent.
If you hate it I'm fine with that too.

All I did was remove the `// namespace blah` comment on the closing bracket of a namespace since it is not consistently applied everywhere.
I also removed the extra newlines on some classes for the closing brackets, again because it is not consistently applied everywhere.


fboemer:
Just happened to see this; would it work to add `FixNamespaceComments: false` to the `.clang-format`?

PR #4808, opened 2020-05-27
A little more node method cleanup

rkimballn1:


src/ngraph/node.hpp
```diff
@@ -330,6 +330,9 @@ namespace ngraph
         const std::deque<descriptor::Output>& get_output_descriptors() const
             NGRAPH_DEPRECATED("use outputs() instead");
 
+        size_t get_input_count() const;
```
diyessi, line 333: There are already `get_input_size()`and `get_output_size()`

rkimballn1: Changed

PR #4810, opened 2020-05-27
[MLIR] fix GDBRegistrationListener assert when running MLIR unit tests

adstraw:


cmake/external_mlir.cmake
```diff
@@ -68,6 +68,11 @@ message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
 message(STATUS "Using modules in: ${LLVM_DIR}")
 message(STATUS "LLVM RTTI is ${LLVM_ENABLE_RTTI}")
 
+set(MLIR_DIR "${MLIR_LLVM_BUILD_DIR}/lib/cmake/mlir"
+    CACHE PATH "Path to MLIR cmake modules")
```
dcaballe, line 72: I would probably rename this to MLIR_CMAKE_PATH since there are quite a few var with MLIR directories already. Or if it's only used in the next line, maybe you can inline the path there?

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1459,9 +1448,9 @@ namespace
         StdIndexedValue iParams(params);
 
         // Construct outer loop for params dims. Exclude the axis dim.
-        SmallVector<ValueHandle, 4> paramsLbs, paramsUbs, paramsIVs;
+        SmallVector<Value, 4> paramsLbs, paramsUbs;
         SmallVector<int64_t, 4> paramsSteps;
-        SmallVector<ValueHandle*, 4> paramsIVPtrs;
+        SmallVector<Value*, 4> paramsIVPtrs;
```
dcaballe, line 1453: This is not used?

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1512,12 +1499,12 @@ namespace
         //                   params[P_0, P_1, .. P_(A-1), indices[I_0, .., I_(M-1)],
         //                          P_(A+1), ... P_(N-1)];
 
-        AffineLoopNestBuilder(indicesIVPtrs, indicesLbs, indicesUbs, indicesSteps)([&] {
+        AffineLoopNestBuilder(indicesIVs, indicesLbs, indicesUbs, indicesSteps)([&] {
             // Load axis value from indices array and cast it to Index Type
-            ValueHandle axisIdx = ValueHandle::create<IndexCastOp>(
-                (ValueHandle)iIndices(indicesIVs), rewriter.getIndexType());
+            Value axisIdx =
+                ValueBuilder<IndexCastOp>((Value)iIndices(indicesIVs), rewriter.getIndexType());
```
dcaballe, line 1505: is the Value casting still necessary? There are plenty of them in the code. Maybe they can be removed.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1722,7 +1706,6 @@ namespace
         ArrayRef<Attribute> padAbove = pooling.padAbove().getValue();
 
         Value result = pass.buildOutputDefs(op, rewriter)[0];
-        NGRAPH_CHECK(src && delta && result, "Unexpected null values in MaxPoolBackprop Op");
```
nmostafa, line 1725: Why removing the asserts ?

adstraw: @nmostafa with the change to Value (instead of ValueHandle) the && operator here was invoking a logical and operation in the Affine dialect via overloaded operator&&.  My theory (could be wrong) is that since we are passing mlir::Value around by value it cannot be null which was the check here.  Please correct me if I am wrong.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
adstraw: Missed one.  Removed.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
adstraw: Looks like it can be removed.  Good catch.

cmake/external_mlir.cmake
adstraw: Done.

adstraw:
TESTNOW

src/contrib/mlir/backend/pass/affine_lowerer.cpp
nmostafa: The Value is a  wrapper around a pointer. I do see a bool cast operator. 
https://github.com/llvm/llvm-project/blob/master/mlir/include/mlir/IR/Value.h#L95

You could split this into multiple asserts.

adstraw: I ended up splitting into multiple asserts.

PR #4812, opened 2020-05-28
Get GPU backend to compile

rkimballn1:
It compiles but I could not run unit tests


src/runtime/gpu/gpu_cuda_kernel_ops.hpp
```diff
@@ -16,54 +16,48 @@
 
 #pragma once
 
+#include "ngraph/op/abs.hpp"
```
diyessi, line 19: `ngraph/ops.hpp` was made for just this sort of occasion.

rkimballn1: I looked for that header in the ops directory. I guess I picked the wrong place to look.
fixed

PR #4827, opened 2020-05-31
Fix mlir build

rkimballn1:
I use build command
`cmake .. -DCMAKE_BUILD_TYPE=Release -DNGRAPH_ONNX_IMPORT_ENABLE=ON -DNGRAPH_ONNXIFI_ENABLE=ON -DNGRAPH_CPU_ENABLE=TRUE -DNGRAPH_USE_PREBUILT_LLVM=TRUE -DNGRAPH_MLIR_ENABLE=TRUE -DNGRAPH_DEX_ONLY=ON -DNGRAPH_USE_PREBUILT_MLIR=0 -DCMAKE_INSTALL_PREFIX=~/ngraph/BUILD-GCC/ngraph_dist`
using clang 6.0
I get the error
```
In file included from /home/rkimball/dev/ngraph/src/contrib/mlir/runtime/cpu/cpu_callbacks.cpp:22:
/home/rkimball/dev/ngraph/src/contrib/mlir/runtime/cpu/cpu_runtime.hpp:41:25: error: flexible array members are a C99
      feature [-Werror,-Wc99-extensions]
                int64_t shapeAndStrides[];
                        ^
In file included from /home/rkimball/dev/ngraph/src/contrib/mlir/runtime/cpu/cpu_runtime.cpp:20:
/home/rkimball/dev/ngraph/src/contrib/mlir/runtime/cpu/cpu_runtime.hpp:41:25: error: flexible array members are a C99
      feature [-Werror,-Wc99-extensions]
                int64_t shapeAndStrides[];
                        ^
1 error generated.
```


silee2:
That error is coming from usage of https://en.wikipedia.org/wiki/Flexible_array_member
And I remember it is need for nGraph MLIR callback function. Clang on Mac had the same error and it is currently suppressed by https://github.com/NervanaSystems/ngraph/blob/master/cmake/clang_4_0_flags.cmake#L63 Looks like the warning suppression needs to expanded to Clang on any platform.

rkimballn1:
Sure enough, the check for Apple|Clang was borked. Not sure how that got changed but it did not work. I make the flags a little nicer and renamed the file to be more true to intent.

PR #4829, opened 2020-05-31
Better style tools

rkimballn1:
* There is a single list of directories to format in the root CMakeLists.txt. This makes the format tools easier for other projects to use and also keeps UNIX and APPLE directories in sync.
* Simplified the scripts, you now need to pass in the absolute path of the directories to fomat, something cmake already knows and the scripts needed to discover.
* Minor rewording of progress messages


rkimballn1:
@diyessi Can you please verify that the macos scripts work as expected?

PR #4834, opened 2020-06-01
[MLIR] add support for unsigned integer comparison

adstraw:


src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1271,10 +1271,22 @@ namespace
...
 
+        auto origOperand = op->getOperands()[0];
+        auto origIntType =
+            origOperand.getType().cast<NGTensorType>().getElementType().dyn_cast<IntegerType>();
+
         AffineLoopNestBuilder(ivs, lbs, ubs, steps)([&] {
             Value val = iLHS(ivs);
             Value zero = createZeroConstant(elemTy);
-            iRes(ivs) = std_select(val > zero, val, zero);
+            auto cmp = val > zero;
```
dcaballe, line 1281: This overloaded '>' operator is actually generating code so you may want to put it in an 'else' branch after 1287 so that we don't generate two compare ops for unsigned.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1271,10 +1271,22 @@ namespace
...
+            auto builder = ScopedContext::getBuilderRef();
```
dcaballe, line 1282: `rewriter` is a special builder that is available in this function. We can use it here instead of retrieving a new builder.

test/util.cpp
```diff
@@ -740,6 +740,58 @@ TEST(util, double_to_int_limits)
...
+    EXPECT_TRUE(std::numeric_limits<int8_t>::max() == double_to_int<int8_t>(x, floor_func));
+    EXPECT_TRUE(std::numeric_limits<int16_t>::max() == double_to_int<int16_t>(x, floor_func));
+    EXPECT_TRUE(std::numeric_limits<int32_t>::max() == double_to_int<int32_t>(x, floor_func));
+    EXPECT_TRUE(std::numeric_limits<int64_t>::max() == double_to_int<int64_t>(x, floor_func));
+
+    EXPECT_TRUE(std::numeric_limits<uint8_t>::max() == double_to_int<uint8_t>(x, floor_func));
+    EXPECT_TRUE(std::numeric_limits<uint16_t>::max() == double_to_int<uint16_t>(x, floor_func));
+    EXPECT_TRUE(std::numeric_limits<uint32_t>::max() == double_to_int<uint32_t>(x, floor_func));
+    EXPECT_TRUE(std::numeric_limits<uint64_t>::max() == double_to_int<uint64_t>(x, floor_func));
 }
```
dcaballe, line 796: We should add a LIT test for unsigned comparison to make sure we have it covered as we have other ops and types.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
adstraw: Done.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
adstraw: Done.

test/util.cpp
adstraw: Done.

adstraw:
> LGTM, Thanks! I see some formatting issues. Other than that, it looks great!

@dcaballe note that formatting was recently changed to LLVM style.  Where are you seeing format issues?

dcaballe:
Sorry, my review went out before getting those commits.

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1135,10 +1135,23 @@ REWRITER(NGReluOp) {
...
+  auto origIntType = origOperand.getType()
+                         .cast<NGTensorType>()
+                         .getElementType()
+                         .dyn_cast<IntegerType>();
+
   AffineLoopNestBuilder(ivs, lbs, ubs, steps)([&] {
     Value val = iLHS(ivs);
     Value zero = createZeroConstant(elemTy);
-    iRes(ivs) = std_select(val > zero, val, zero);
+    if (origIntType && origIntType.isUnsigned()) {
```
diyessi, line 1147: What does `val > zero` turn into? It would be better if this remained as `val > zero` and all of this signed/unsigned distinction were moved there. Otherwise it needs to be repeated everywhere there is a signed/unsigned distinction, as you have had to do. It's too easy to miss a comparison, or for one to be added again in the future.

adstraw: See this code:

https://github.com/llvm/llvm-project/blob/e25f01be0c4c4f0355ae645599b3f54acb712ddd/mlir/lib/Dialect/Affine/EDSC/Builders.cpp#L268

The intrinsic for operator> creates a signed comparison.  And, the Affine dialect is "signless" meaning that all integers are neither signed nor unsigned.  The signedess, in this case, is carried in the predicate for comparison:  signed greater than (sgt) vs. unsigned greater than (ugt).  It is the responsibility of the dialect lowering to Affine to pick the correct predicate.  Which mean, I think, that the if/else in the nGraph Affine lowerer here are necessary.  The unsigned lowering is a bit uglier as there is no intrinsic for it.  We could add it.  But, the Affine operator> will have to pick one way or the other (signed or unsigned) as it has no way to distinguish between signed and unsigned once lowered to Affine.

Your point about "too easy to miss a comparison" is a good one... I missed ArgMin and ArgMax with this commit.  Will follow up with a separate commit for those two ops.

diyessi: I think we should submit an MLIR PR to fix those operators.

adstraw: The fix would involve a redesign of the Affine dialect to use signed integers.  Are we going to hold up this PR waiting on such a change?

adstraw: Note that we have a correctness issue in our MLIR implementation at the moment.  Unsigned integers with the most significant bit set are treated as signed (negative) and therefore we are generating that incorrect answer.

PR #4835, opened 2020-06-02
changing default bcast type from none to numpy

gaurides:


rkimballn1:
Should the autobroadcast default be changed to NUMPY or should the TF frontend set that as the broadcast spec when constructing ops? A default in ngraph will not likely work with all frameworks so some frameworks will need to change it when they construct ops.

gaurides:
@rkimballn1 : so i read what other frameworks do for the Greater op (as example) and default is Numpy autobroadcast . So ngraph should default to Numpy and user can change to None or Pdpd if they need.
OpenVino : https://gitlab-icv.inn.intel.com/inference-engine/dldt/blob/master/docs/ops/comparison/Greater_1.md
ONNX: https://github.com/onnx/onnx/blob/master/docs/Operators.md#Greater
Tensorflow: https://www.tensorflow.org/api_docs/python/tf/math/greater

gaurides:
Bridge needs to use the correct op version, we don't need these changes.

PR #4852, opened 2020-06-08
[MLIR] add MLIR patch + signed compare ops + ArgMin and ArgMax

adstraw:


src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -1107,6 +1107,17 @@ REWRITER(NGArgMinRedOp) {
...
 
+bool is_signed(NGTensorType ngTensorType) {
+  NGRAPH_CHECK(ngTensorType);
+  auto ngIntType = ngTensorType.getElementType().dyn_cast<NGIntegerType>();
+
+  if (ngIntType && ngIntType.isUnsigned()) {
+    return false;
+  } else {
+    return true;
+  }
```
rkimballn1, line 1118:
```suggestion
  return !(ngIntType && ngIntType.isUnsigned());
```

nmostafa:
Perhaps I am missing context, why are we patching MLIR ? This can be a headache in the future as MLIR upstream diverges. Why not try to land that patch upstream instead ? If not possible, and you rather have your change, you could add wrappers on nGraph side. 

Also, I would break this PR into two. The CPU Argmin/max non-MLIR change doesn't really belong here and can make triaging for errors confusing.

adstraw:
The patch is temporary.  My next step is to upstream the gt, lt, ge, le operators to MLIR.

adstraw:
Marking as fully reviewed.  @nmostafa and @dcaballe disagree with the patching approach.  Making a call to press forward with this approach in spite of their feedback.

PR #4855, opened 2020-06-09
Remove deprecated get input descriptor and get output descriptor methods from Node

rkimballn1:
* Remove all deprecated descriptor access methods from node
* Remove the const versions of get_input_descriptor(n) and get_output_descriptor(n)
* Make all remaining descriptor methods in Node private since they are only used within Node


silee2:
@rkimballn1 Mac Ut has six failed cases with "C++ exception with description "node output index is out of range" thrown in the test body."

PR #4859, opened 2020-06-12
Switch to clang-format-6 and apply style

rkimballn1:
If there is anything you don't like let me know and we can perhaps tweak it
MLIR has a separate .clang-format file so it's format is different than ngraph


silee2:
general-ci MLIR prebuilt lacks clang-format-6.0

PR #4863, opened 2020-06-14
Change placement from enum to an integer

rkimballn1:


src/ngraph/graph_util.hpp
```diff
@@ -420,7 +419,7 @@ namespace ngraph
 
     // Assert that nodes in the function is colocated and return that placement
```
diyessi, line 420: Might as well fix the comment while in the neighborhood.

src/ngraph/node.hpp
```diff
@@ -428,10 +427,10 @@ namespace ngraph
         bool has_same_type(std::shared_ptr<const Node> node) const;
 
         /// Get device placement
```
diyessi, line 429: Need a comment explaining now placement is used.

src/ngraph/node.hpp
```diff
@@ -545,7 +544,7 @@ namespace ngraph
         std::deque<descriptor::Input> m_inputs;
         std::deque<descriptor::Output> m_outputs;
         std::unordered_map<Node*, autodiff::Adjoints> m_adjoint_map;
-        Placement m_placement = Placement::DEFAULT;
+        int32_t m_placement = -1;
```
diyessi, line 547: Make a const for this magic -1 value.

src/ngraph/graph_util.hpp
rkimballn1: done

PR #4866, opened 2020-06-15
Remove GetOutputElement helper functions from Node

rkimballn1:
GetOutputElement is deprectated so ngraph should not have methods to make it easy to use. To that end I removed these helpers from `Node`, `Output<Node>`, and `GetOutputElement`. There are two places in the code, the CPU backend, and the GPU backend, where these helpers were used so I made copies of the helper code in those backends so at least ngraph core is clean.


src/ngraph/builder/autobroadcast.cpp
```diff
@@ -139,11 +139,11 @@ namespace ngraph
                                                      const Shape& output_shape,
                                                      const Shape& source_shape)
         {
-            shared_ptr<Node> broadcasted_node = value.as_single_output_node();
+            shared_ptr<Node> broadcasted_node = value.get_node_shared_ptr();
             // If node already has the required shape, return original node
             if (output_shape == value.get_shape())
             {
-                return broadcasted_node;
+                return value.get_node_shared_ptr();
```
diyessi, line 146: Just return `broadcasted_node`

src/ngraph/builder/autobroadcast.cpp
```diff
@@ -249,8 +249,8 @@ namespace ngraph
             // Handle the trivial case...
             if (arg1_in_shape == arg2_in_shape)
             {
-                return make_pair(args.first.as_single_output_node(),
-                                 args.second.as_single_output_node());
+                return make_pair(args.first.get_node_shared_ptr(),
```
diyessi, line 252: I suspect this should be returning `make_pair(args.first, args.second)` since those args aren't necessarily using 1-output nodes.

src/ngraph/node.cpp
```diff
@@ -450,7 +431,7 @@ std::shared_ptr<Node> Node::get_argument(size_t index) const
 {
     NGRAPH_CHECK(
         index < m_inputs.size(), "index '", index, "' out of range in get_argument(size_t index)");
-    return input_value(index).as_single_output_node();
+    return input_value(index).get_node_shared_ptr();
```
diyessi, line 434: I suspect most remaining callers assume the argument is a one-output node and this should really be returning an `Output<Node>`.

src/ngraph/node.cpp
```diff
@@ -787,7 +751,7 @@ NodeVector ngraph::as_node_vector(const OutputVector& values)
     NodeVector node_vector;
     for (auto& value : values)
     {
-        node_vector.push_back(value.as_single_output_node());
+        node_vector.push_back(value.get_node_shared_ptr());
```
diyessi, line 754: This is incorrect. Instead, `as_node_vector` callers need to be changed to work with `OutputVector`. There probably aren't many left.

src/ngraph/op/util/rnn_cell_base.cpp
```diff
@@ -97,7 +97,7 @@ shared_ptr<Node> op::util::RNNCellBase::clip(const Output<Node>& data) const
 {
     if (m_clip == 0.f)
     {
-        return data.as_single_output_node();
+        return data.get_node_shared_ptr();
```
diyessi, line 100: This is incorrect. Need to change it to return `Output<Node>`.

src/ngraph/pattern/matcher.hpp
```diff
@@ -155,7 +155,7 @@ class NGRAPH_API ngraph::pattern::Matcher
     OutputVector& get_matched_values() { return m_matched_list; }
     void reset() {}
     const std::string& get_name() { return m_name; }
-    std::shared_ptr<Node> get_pattern() { return m_pattern_node.as_single_output_node(); }
+    std::shared_ptr<Node> get_pattern() { return m_pattern_node.get_node_shared_ptr(); }
```
diyessi, line 158: This will break patterns that work with multiple output nodes and assume they get back a GOE.

src/ngraph/pattern/op/any_of.hpp
```diff
@@ -62,7 +62,7 @@ class NGRAPH_API ngraph::pattern::op::AnyOf : public Pattern
           const NodeVector& wrapped_values)
         : AnyOf(type,
                 s,
-                [pred](const Output<Node>& value) { return pred(value.as_single_output_node()); },
+                [pred](const Output<Node>& value) { return pred(value.get_node_shared_ptr()); },
```
diyessi, line 65: This will break patterns involving multi-output nodes.

src/ngraph/builder/autobroadcast.cpp
rkimballn1: I changed the method to return pair<Output<Node>, Output<Node>> instead of a pair of nodes.

src/ngraph/node.cpp
rkimballn1: Most uses actually want the node, not the Output<Node>. A very common use is the cast the input argument to a particular op type. There are some misuses of get_arguments, but certainly not the majority.

diyessi: Many of the incorrect uses have already been changed; the replacement approach was replacing getArgument users with something else. It used to be that most uses were incorrect.

rkimballn1:
I blocked this because many of the fusion unit tests are not running against the MLIR backend. I am making a PR to enable those tests for all CPU backend variants and want to wait until those tests are enabled before merging this PR.

rkimballn1:
I was able to run tests forcing MLIR when CPU was asked for an everything passes so I removed the block

PR #4875, opened 2020-06-18
Update: oneDNN v1.5.0

silee2:
Update oneDNN to v1.5.0
Move to new API name. (use `dnnl` instead of `mkldnn`, use `DNNL` instead of `MKLDNN`)
Fix bug in vanilla_rnn scratchpad memory usage.


CMakeLists.txt
```diff
@@ -130,7 +130,7 @@ option(NGRAPH_GENERIC_CPU_ENABLE "Enable build nGraph for generic CPU backend" T
 option(NGRAPH_DEBUG_ENABLE "Enable output for NGRAPH_DEBUG statements" FALSE)
 option(NGRAPH_DEPRECATED_ENABLE "Enable compiler deprecation pragmas for deprecated APIs (recommended only for development use)" FALSE)
 option(NGRAPH_ONNX_IMPORT_ENABLE "Enable ONNX importer" FALSE)
-option(NGRAPH_ENABLE_CPU_CONV_AUTO "Enable mkldnn convolution_auto for CPU" TRUE)
+option(NGRAPH_ENABLE_CPU_CONV_AUTO "Enable dnnl convolution_auto for CPU" TRUE)
```
rkimballn1, line 133:
```suggestion
option(NGRAPH_CPU_CONV_AUTO_ENABLE "Enable dnnl convolution_auto for CPU" TRUE)
```

silee2: Renamed as suggested.

PR #4878, opened 2020-06-22
Pass element::Type by value

rkimballn1:


rkimballn1:
There is no point in passing an enum by reference. The justification is to normalize the use of element::Type so that all calls/uses are consistent.

diyessi:
> There is no point in passing an enum by reference. The justification is to normalize the use of element::Type so that all calls/uses are consistent.

`element::Type` isn't the enum, it's the struct. `element::Type_t` is the enum. It would be nice if the struct users were switched to the enum, but that's a separate thing.

PR #4883, opened 2020-06-23
Merge LLVM build file for Codegen and MLIR

silee2:
And use CMake > 3.12 to support LLVM_ROOT for pointing to prebuilt LLVM and LLVMConfig.cmake
Still keeping a minimum external_mlir.cmake with just MLIR_LLVM_COMMIT_ID to minimize impact on CI. 
Will be remove in a separate PR later.


CMakeLists.txt
```diff
@@ -634,15 +637,12 @@ endif()
 if(NGRAPH_CPU_ENABLE)
     include(cmake/external_onednn.cmake)
 endif()
-if (NGRAPH_MLIR_ENABLE)
-    include(cmake/external_mlir.cmake)
-endif()
 
-if(NGRAPH_CODEGEN_ENABLE)
+if(NGRAPH_CODEGEN_ENABLE OR NGRAPH_MLIR_ENABLE)
```
rkimballn1, line 641: If MLIR is enabled it should set NGRAPH_CODEGEN_ENABLE true. This extra OR condition is not needed.

silee2: NGRAPH_CODEGEN_ENABLE and NGRAPH_MLIR_ENABLE are two different usage case of LLVM and cannot be merged yet. That needs to be done in a separate PR where codegen needs to be updated to work on current LLVM master branch. I cleared up the confusion by introducing NGRAPH_LLVM_ENABLE that is turned on if NGRAPH_CODEGEN_ENABLE or NGRAPH_MLIR_ENABLE is turned on. That makes the intention clear. If LLVM usage is enabled, build LLVM.

PR #4887, opened 2020-06-26
Matcher supporting multi-output nodes

rkimballn1:
* Remove the use and requirement of GetOutputElement in the CPU RNN fusion passes
* Update pattern matcher to match multi-output nodes


src/ngraph/node_input.cpp
```diff
@@ -152,15 +153,35 @@ namespace ngraph
...
+                   << input.get_partial_shape();
+        }
     }
 
     std::ostream& operator<<(std::ostream& out, const Input<const Node>& input)
     {
-        return input.get_node()->write_description(out, 0)
-               << ".input(" << input.get_index() << "):" << input.get_element_type()
-               << input.get_partial_shape();
+        if (is_type<op::GetOutputElement>(input.get_node()))
```
diyessi, line 173: Why not define `write_description` for `GetOutputElement` ?

src/ngraph/node_output.cpp
```diff
@@ -218,17 +233,53 @@ namespace ngraph
...
+    }
+
     std::ostream& operator<<(std::ostream& out, const Output<Node>& output)
     {
-        return output.get_node()->write_description(out, 0)
-               << "[" << output.get_index() << "]:" << output.get_element_type()
-               << output.get_partial_shape();
+        if (is_type<op::GetOutputElement>(output.get_node()))
+        {
+            auto captured_output = output.get_node()->input(0).get_source_output();
```
diyessi, line 254: Use a virtual function on `Node` -- then all the GOE stuff disappears when GOE goes away, and other nodes can provide enhanced printing (I think there are one or two that already do). This would also reduce duplication in the const node case below.

src/ngraph/pattern/matcher.cpp
```diff
@@ -163,7 +185,15 @@ bool pattern::Matcher::match_value(const Output<Node>& pattern_value,
             return false;
         }
     }
-    return pattern_node->match_value(this, pattern_value, graph_value);
+    bool rc = pattern_node->match_value(this, pattern_value, graph_value);
+    if (!rc && is_type<ngraph::op::GetOutputElement>(graph_node))
+    {
```
diyessi, line 190: You can do this `GetOutputElement::match_value`

src/ngraph/node_input.cpp
rkimballn1: Rewritten

src/ngraph/node_output.cpp
rkimballn1: Rewritten

src/ngraph/pattern/matcher.cpp
rkimballn1: I don't understand what you are suggesting. Can you give more detail or a code suggestion?

rkimballn1: This code is no longer needed since the GOE uses that required it have been removed

PR #4893, opened 2020-07-03
Remove as single output

rkimballn1:
Fix MLIR usage and tracked down a few others.


src/ngraph/node.hpp
```diff
@@ -633,11 +632,11 @@ namespace ngraph
     void check_new_args_count(const Node* node, T new_args)
     {
         NODE_VALIDATION_CHECK(node,
-                              new_args.size() == node->get_arguments().size(),
+                              new_args.size() == node->inputs().size(),
                               "copy_with_new_args() expected ",
-                              node->get_arguments().size(),
+                              node->inputs().size(),
```
diyessi, line 637: There's an `input_size` sort of method.

rkimballn1:
```suggestion
                              node->input_size(),
```

PR #4899, opened 2020-07-08
Use oneDNN 1.5.1

silee2:


python/CMakeLists.txt
```diff
@@ -46,5 +46,5 @@ endif()
 
 add_custom_target(python_wheel DEPENDS ngraph ${CMAKE_CURRENT_BINARY_DIR}/dist/)
 if (NGRAPH_CPU_ENABLE)
-    add_dependencies(python_wheel libdnnl)
```
rkimballn1, line 49: You made the alias libdnnl but then never use it. Either remove alias or use it.

cmake/external_onednn.cmake
```diff
@@ -260,42 +263,6 @@ else()
         )
 endif()
 
-if(WIN32)
```
rkimballn1, line 263: Why remove support for win32?

python/CMakeLists.txt
silee2: libdnnl is used inside external_onednn.cmake. DNNL::dnnl is the alias created out of it. I'm using the name DNNL::dnnl to match the target oneDNN exposes. That would make using prebuilt oneDNN easier in the future.

cmake/external_onednn.cmake
silee2: The PR does not remove WIN32 support. The code that is removed is for copying libraries to a new location. The PR passes CMAKE_LIBRARY_OUTPUT_DIRECTORY and CMAKE_ARCHIVE_OUTPUT_DIRECTORY to the external project and the code for copying is not needed anymore.

PR #4911, opened 2020-07-20
Standalone MLIR backend

rkimballn1:
Create an MLIR backend as an independent backend, no longer part of the CPU backend.
* The big change was to make the contrib/mlir code work with ngraph::Function instead of requiring an ngraph::op::CompiledKernel. The MLIR backend is not a fallback backend, it is a first-class backend.
* I added support for Abs just to see if I could
* There are now two independent MLIR backends, the CPU MLIR and the standalone MLIR backend.
    - NGRAPH_CPU_MLIR_ENABLE enables the CPU MLIR backend
    - NGRAPH_MLIR_ENABLE enables the standalone MLIR backend


src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -2588,6 +2595,11 @@ namespace
                 Value zero = createZeroConstant(elemTy);
                 iRes(ivs) = zero - val;
             }
+            else if (isa<NGAbsOp>(op))
+            {
+                Value zero = createZeroConstant(elemTy);
+                iRes(ivs) = std_select(lt(val, zero, is_signed(ngTensorType)), zero - val, val);
```
silee2, line 2601: #4909 introduces API change for `lt`.

rkimballn1: Fixed, and I added a unit the value of -0 to the Abs test vector just in case. Is -0 lt zero? Yes it is which is kind of odd.

CMakeLists.txt
```diff
@@ -637,7 +637,7 @@ endif()
 if(NGRAPH_CPU_ENABLE OR NGRAPH_GENERIC_CPU_ENABLE)
     include(cmake/external_eigen.cmake)
 endif()
-if(NGRAPH_CPU_ENABLE)
+if(NGRAPH_CPU_ENABLE OR NGRAPH_MLIR_ENABLE)
```
adstraw, line 640: Is this actually necessary?

src/contrib/mlir/backend/backend.hpp
```diff
@@ -32,13 +32,13 @@ namespace ngraph
             class MLIRBackend
             {
             public:
-                MLIRBackend(mlir::OwningModuleRef& module, mlir::MLIRContext& context)
+                MLIRBackend(::mlir::OwningModuleRef& module, ::mlir::MLIRContext& context)
```
adstraw, line 35: The prepended :: forces a search from global namespace, correct?  I see you have done this in many places.  Was this necessary to overcome some compile issue?

src/contrib/mlir/backend/pass/affine_lowerer.cpp
```diff
@@ -2596,6 +2603,12 @@ namespace
                 Value zero = createZeroConstant(elemTy);
                 iRes(ivs) = zero - val;
             }
+            else if (isa<NGAbsOp>(op))
```
adstraw, line 2606: Nice!  Good to see you were able to add an op.

src/contrib/mlir/core/pass/ng_dialect_builder.cpp
```diff
@@ -209,10 +203,11 @@ mlir::ArrayAttr NgDialectConversionPass::getShapeAsAttr(T ngShape)
 
 ngraph::Node* NgDialectConversionPass::getOriginArg(ngraph::Node* node) const
 {
-    auto inputMap = m_compiledKernel->get_input_map();
-    auto it = inputMap.find(node->shared_from_this());
-    NGRAPH_CHECK(it != inputMap.end(), "Parameter not in CK input map");
-    return m_compiledKernel->input_values().at(it->second).get_node();
+    return node;
+    // auto inputMap = m_compiledKernel->get_input_map();
```
adstraw, line 207: Clean up comments.

src/contrib/mlir/core/pass/ng_dialect_builder.cpp
```diff
@@ -323,96 +318,128 @@ void NgDialectConversionPass::buildNgDialect(mlir::FuncOp function)
 }
 
 template <>
-mlir::Operation* NgDialectConversionPass::COMPILE_OP_DECL(ngraph::op::Add)
+mlir::Operation*
+    NgDialectConversionPass::createOp<ngraph::op::Add>(NgDialectConversionPass& NgDialectObj,
```
adstraw, line 322: Why do you need to pass a reference to an NGDialectConversionPass here?

src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -143,12 +143,12 @@ if (NGRAPH_CPU_CODEGEN_ENABLE)
         )
 endif()
 
-if (NGRAPH_MLIR_ENABLE)
-    set(SRC
-        ${SRC}
-        builder/mlir_cpu_compiled_kernel.cpp
-        )
-endif()
+# if (NGRAPH_MLIR_ENABLE)
```
adstraw, line 146: clean up comments

src/ngraph/runtime/cpu/CMakeLists.txt
```diff
@@ -248,21 +248,21 @@ if (NGRAPH_CPU_ENABLE)
         set_property(TARGET cpu_backend APPEND PROPERTY LINK_FLAGS "-Wl,-Bsymbolic-functions -Wl,--exclude-libs=ALL")
     endif()
 
-    if (NGRAPH_MLIR_ENABLE)
-        # TODO: can we get away without LLVM/MLIR include path.
-        # Currently mlir backend compiler.hpp include LLVM/MLIR files
-        get_directory_property(MLIR_LLVM_INCLUDEPATH
-                               DIRECTORY ${NGRAPH_MLIR_SOURCE_DIR}
-                               DEFINITION MLIR_LLVM_INCLUDEPATH)
+    # if (NGRAPH_MLIR_ENABLE)
```
adstraw, line 251: clean up comments

test/CMakeLists.txt
```diff
@@ -290,13 +290,17 @@ if (NGRAPH_CPU_ENABLE)
         list(APPEND SRC cpu_codegen.cpp)
         set(ACTIVE_CPU_BACKEND_LIST ${ACTIVE_CPU_BACKEND_LIST} "CPU:CODEGEN")
     endif()
-    if (NGRAPH_MLIR_ENABLE)
-        set(ACTIVE_CPU_BACKEND_LIST ${ACTIVE_CPU_BACKEND_LIST} "CPU:MLIR")
-    endif()
+    # if (NGRAPH_MLIR_ENABLE)
```
adstraw, line 293: clean up comments

src/contrib/mlir/backend/backend.hpp
rkimballn1: I put the ngraph mlir standalone backend as namespace ngraph::runtime::mlir so there was compiler confusion. This fixes the compile error.

src/contrib/mlir/core/pass/ng_dialect_builder.cpp
rkimballn1: This is the signature of the function as defined as COMPILE_OP_DECL. I don't like the macro function signature hiding since it makes the code hard to read so I just expanded the macro to the explicit function call. The signature is unchanged.

src/ngraph/runtime/cpu/CMakeLists.txt
rkimballn1: I want to re-enable MLIR in the CPU backend so I left this in. I did not want to make the compiler too complex so I figured a separate PR to fix the CPU backend would work. If you want I can add MLIR support back to the CPU backend in this PR.

test/CMakeLists.txt
rkimballn1: See above comment

CMakeLists.txt
silee2: It is. MLIR is now separate from CPU and MLIR uses oneDNN callbacks.

adstraw: I forgot about oneDNN callbacks.  Thanks.

src/ngraph/runtime/cpu/CMakeLists.txt
adstraw: I think that would work best, that way this change is an addition and not a subtraction.

silee2:
@rkimballn1 I think you would need to change https://github.com/NervanaSystems/ngraph/blob/master/cmake/external_llvm.cmake#L120 to include NGRAPH_CPU_MLIR_ENABLE in the condition.

CMakeLists.txt
```diff
@@ -173,7 +174,7 @@ if (NGRAPH_CPU_ENABLE)
     set(NGRAPH_INTEL_CPU_ONLY_ENABLE ON)
 endif()
 
-if (NGRAPH_MLIR_ENABLE OR NGRAPH_CODEGEN_ENABLE)
+if (NGRAPH_MLIR_ENABLE OR NGRAPH_CODEGEN_ENABLE OR NGRAPH_CPU_MLIR_ENABLE)
```
adstraw, line 177: This may be a latent bug, but should this say "NGRAPH_CPU_CODEGEN_ENABLE" where it now says "NGRAPH_CODEGEN_ENABLE"?

PR #4915, opened 2020-07-24
Add EVAL backend

rkimballn1:
The EVAL backend exclusively uses the evaluate method on ops. For an op to be supported by EVAL it must properly implement its evaluate method.
EVAL is enabled by default and I have disabled GCPU by default but GCPU still exists.
~Reshape v1 evaluate is borked so I comment it out to leave an ugly stain that should be addressed soon. We may already have a fix.~
Reshape v1 evaluate fixed
* EVAL does not downgrade any ops.
* EVAL reports that it supports dynamic tensors so it does not get wrapped with the Dynamic backend wrapper.


src/ngraph/op/reshape.cpp
```diff
@@ -500,6 +500,6 @@ bool op::v1::Reshape::evaluate(const HostTensorVector& outputs, const HostTensor
         }
         outputs[0]->set_shape(output_shape);
     }
-    const AxisVector order = get_default_order(outputs[0]->get_shape());
+    const AxisVector order = get_default_order(inputs[0]->get_shape());
```
silee2, line 503: @rkimball Is this correct? Then it means there was a long standing bug?

rkimballn1: Yes, the input for v0 Reshape is the input axis order, not the output order. When I enabled the EVAL backend it was segfaulting because the output shape was smaller than the input shape so the axis order vector was too small and we read past the end. With this change unit tests pass and 100% less segfaults.

PR #4923, opened 2020-07-28
Ninja: support building python_wheel.

silee2:


python/setup.py
```diff
@@ -94,9 +94,10 @@ def find_pybind_headers_dir():
 if len([fn for fn in os.listdir(NGRAPH_CPP_LIBRARY_DIR) if re.search("ngraphd", fn)]):
     NGRAPH_CPP_LIBRARY_NAME = "ngraphd"
 
-ONNX_IMPORTER_CPP_LIBRARY_NAME = "onnx_importer"
-if len([fn for fn in os.listdir(NGRAPH_CPP_LIBRARY_DIR) if re.search("onnx_importerd", fn)]):
-    ONNX_IMPORTER_CPP_LIBRARY_NAME = "onnx_importerd"
+if NGRAPH_ONNX_IMPORT_ENABLE in ["TRUE", "ON", True]:
```
swapnai, line 97: why give 3 ways to do this?

swapna-intel: This is my personal github account - was logged in for class assignment!

python/setup.py
```diff
@@ -232,11 +233,12 @@ def cpp_flag(compiler):
 
 library_dirs = [NGRAPH_CPP_LIBRARY_DIR]
 
-libraries = [NGRAPH_CPP_LIBRARY_NAME, ONNX_IMPORTER_CPP_LIBRARY_NAME]
+libraries = [NGRAPH_CPP_LIBRARY_NAME]
 
 extra_compile_args = []
 if NGRAPH_ONNX_IMPORT_ENABLE in ["TRUE", "ON", True]:
```
silee2, line 239: It is copied from this line. Will check if we still need a 3-way check. Seems like the code is from the days when bool type was not normalized.

silee2, line 239: It is copied from this line. Will check if we still need a 3-way check. Seems like the code is from the days when bool type was not normalized.

python/setup.py
silee2: Copied from line 239. Will check if we still need a 3-way check. Seems like the code is from the days when bool type was not normalized.

PR #4932, opened 2020-08-13
Build MLIR with add_subdirectory.

silee2:


test/CMakeLists.txt
```diff
@@ -629,7 +629,7 @@ else()
     )
 endif()
 
-add_custom_target(check
+add_custom_target(ngraph-check
```
silee2, line 632: LLVM has a target called "check", so had to rename this target.

src/ngraph/codegen/compiler.cpp
```diff
@@ -185,6 +185,7 @@ void codegen::CompilerCore::initialize()
     // This is for both Eigen strong and weak inlines
     args.push_back("-mllvm");
     args.push_back("-inline-threshold=1000000");
+    // args.push_back("-fpch-instantiate-templates");
```
silee2, line 188: This is a new Clang option available from LLVM 11 to speed up compilation when using PCH but commented out now. Will enable in a later PR.

PR #4946, opened 2020-08-26
Fetch archive instead of cloning repo for external dependencies.

silee2:


silee2:
PR is getting too big. Splitting into smaller ones.

